PR #5, opened 2023-02-09
Add an AVX2 version of the BLEND_PREMULTIPLIED blend mode

MyreMylar:
It is about 2 to 6 times faster on average than the old SSE2 version of `BLEND_PREMULTIPLIED` depending on the size of the blit with larger blits gaining more from the AVX2 code and very small blits having little difference.

AVX2 is the current 'top level' of our SIMD code in the pygame-ce codebase. It can do about twice the simultaneous operations as SSE2 which is out current 'base level' of SIMD. These levels are based on hardware adoption gathered via the Steam survey and the ability to transfer SIMD code between processor architectures (SSE2 is quite well supported for Arm ports, later stuff not so well).

In this PR I also get rid of the old MMX SIMD version of the `BLEND_PREMULTIPLIED` blitter - it simply wasn't being used on any real hardware anymore, MMX is the oldest SIMD technology and the slightly more modern SSE2 adoption is at 100% in hardware surveys these days, and has been for a long time.

For blitter/blends more operations usually translates to more pixels being processed at once, but this works a lot better if there are a lot of pixels together to work on, otherwise we have to drop down to slower techniques to finish off/entirely do a blit. This is why you see a growth in performance with larger blits that are more composed of big chunks of pixels we can feed into the SIMD blitter.

You can find the original versions of this PR [here](https://github.com/pygame/pygame/pull/3268) and [here](https://github.com/pygame/pygame/pull/3540).

The [Intel guide on intrinisics](https://www.intel.com/content/www/us/en/docs/intrinsics-guide/index.html) is very useful for reviewing and understanding all kinds of SIMD code.

### DIFF
--- a/src_c/alphablit.c
+++ b/src_c/alphablit.c
@@ -581,6 +581,7 @@ SoftBlitPyGame(SDL_Surface *src, SDL_Rect *srcrect, SDL_Surface *dst,
                         src->format->Rmask == dst->format->Rmask &&
                         src->format->Gmask == dst->format->Gmask &&
                         src->format->Bmask == dst->format->Bmask &&
+                        src->format->Amask == 0xFF000000 &&
                         info.src_blend != SDL_BLENDMODE_NONE &&
                         pg_HasSSE_NEON() && (src != dst)) {
                         blit_blend_premultiplied_sse2(&info);
--- a/src_c/alphablit.c
+++ b/src_c/alphablit.c
@@ -109,10 +109,6 @@ blit_blend_rgba_max(SDL_BlitInfo *info);
 
 static void
 blit_blend_premultiplied(SDL_BlitInfo *info);
-#ifdef __MMX__
-static void
-blit_blend_premultiplied_mmx(SDL_BlitInfo *info);
-#endif /*  __MMX__ */
 
 static int
 SoftBlitPyGame(SDL_Surface *src, SDL_Rect *srcrect, SDL_Surface *dst,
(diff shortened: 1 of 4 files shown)


Starbuck5 (approved):
Approved again.

src_c/alphablit.c
```diff
@@ -567,27 +563,33 @@ SoftBlitPyGame(SDL_Surface *src, SDL_Rect *srcrect, SDL_Surface *dst,
...
+                    if (src->format->BytesPerPixel == 4 &&
+                        dst->format->BytesPerPixel == 4 &&
+                        src->format->Rmask == dst->format->Rmask &&
+                        src->format->Gmask == dst->format->Gmask &&
+                        src->format->Bmask == dst->format->Bmask &&
+                        src->format->Amask == 0xFF000000 &&
+                        info.src_blend != SDL_BLENDMODE_NONE &&
+                        pg_HasSSE_NEON() && (src != dst)) {
+                        blit_blend_premultiplied_sse2(&info);
+                        break;
```
robertpfeiffer, line 588: This is both ARM and X86, right?

MyreMylar: Yep, NEON on Arm and SSE2 on X86. the intrinsics are written in SSE2 style but these are then converted to neon intrinsics by the vendored `sse2neon` header.

We should probably update that vendored header at some point as well.

ankith26 (approved):
Can't review in detail but approving nonetheless. I like how this PR removes old mmx code, and the new tests are looking good and make me confident about approving this.

Thanks for the PR :partying_face:

PR #8, opened 2023-02-10
Added `pygame.mixer.music.get_metadata`

Matiiss:
If obj gets freed somewhere before the 2nd if (obj) I have a fix to immediately push to this branch. What about the doc versionchanged? Will this be added to 2.1.3 by chance? (MyreMylar approved in the old repo and this has changes Starbuck requested)


buildconfig/stubs/pygame/mixer_music.pyi
```diff
@@ -18,3 +18,4 @@ def get_pos() -> int: ...
 def queue(filename: FileArg, namehint: str = "", loops: int = 0) -> None: ...
 def set_endevent(event_type: int) -> None: ...
 def get_endevent() -> int: ...
+def get_metadata(filename: Optional[FileArg] = None, namehint: Optional[str] = "") -> Dict[str, str]: ...
```
Starbuck5, line 21: Optional doesn't mean optional, it means can be this type or nonetype. The namehint argument therefore is not "optional," because you can't pass None into it.

MyreMylar (approved):
👍

Starbuck5 (approved):
Looks good, thanks!

PR #12, opened 2023-02-10
Formatted `pixelarray_test.py` using `python setup.py format`

Matiiss:
Because it wasn't formatted for some reason? Not sure if this passes tests but apparently this could potentially be an issue? Either way it would get formatted whenever someone would run this command, so this would fix that.

Although not going lie, the formatting looks a bit atrocious.

This must wait to pass tests obviously.


Matiiss:
Reopen?

PR #1829, opened 2023-02-13
Port `SDL_RenderGeometry`; rendering triangles and quads with `_sdl2.Renderer` and drawing from `_sdl2.Texture`

avaxar:
**PR pygame/pygame#3006 ported from the Illume-maintained pygame library**

This PR implements the relatively new `SDL_RenderGeometry` function from SDL 2.0.18. It lets the programmer draw primitives with `SDL_Renderer`--in turn, `pygame._sdl2.Renderer`. This PR implements the SDL function as: `pygame._sdl2.Texture.draw_triangle`, `pygame._sdl2.Texture.draw_quad`, `pygame._sdl2.Renderer.draw_triangle`, `pygame._sdl2.Renderer.fill_triangle`, `pygame._sdl2.Renderer.draw_quad`, `pygame._sdl2.Renderer.fill_quad`.

I also have added typehints as requested by @MyreMylar.

There are also some minor additions I did:
- Implement the `pygame._sdl2.Window.brightness` property, which corresponds with `SDL_SetWindowBrightness` and `SDL_GetWindowBrightness`.
- Make `res` variables in `video.pyx` be `cdef`'d with an `int` type, to squeeze an extra bit of performance.

# Example
image
```py
import pygame
import pygame._sdl2


pygame.init()

window = pygame._sdl2.Window("amogus", (1066, 600))
renderer = pygame._sdl2.Renderer(window, vsync=True)
surface = pygame.image.load("avaxar.png")
texture = pygame._sdl2.Texture.from_surface(renderer, surface)

running = True
while running:
        if event.type == pygame.QUIT:
            running = False
            break

    renderer.draw_color = (0, 0, 0, 255)
    renderer.clear()

    renderer.draw_color = (255, 255, 255, 255)
    renderer.fill_triangle((50, 50), (200, 100), (50, 100))
    renderer.fill_quad((250, 50), (500, 50), (600, 400), (200, 400))

    renderer.draw_color = (255, 0, 0, 255)
    renderer.draw_triangle((50, 150), (200, 200), (50, 200))
    renderer.draw_quad((250, 150), (500, 150), (600, 500), (200, 500))

    texture.draw_triangle((700, 400), (900, 400), (700, 550))
    texture.draw_quad((600, 100), (900, 100), (1000, 300), (700, 300))

    renderer.present()
```

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -181,8 +181,8 @@ See docs/licenses for licenses of dependencies.
 
 .. |Python3| image:: https://img.shields.io/badge/python-3-blue.svg?v=1
 
-.. |GithubCommits| image:: https://img.shields.io/github/commits-since/pygame/pygame/2.1.2.svg
-   :target: https://github.com/pygame/pygame/compare/2.1.2...main
+.. |GithubCommits| image:: https://img.shields.io/github/commits-since/pygame-community/pygame-ce/2.1.3.svg
+   :target: https://github.com/pygame-community/pygame-ce/compare/2.1.3...main
 
 .. |BlackFormatBadge| image:: https://img.shields.io/badge/code%20style-black-000000.svg
     :target: https://github.com/psf/black


MyreMylar:
> * Implement the `pygame._sdl2.Window.brightness` property, which corresponds with `SDL_SetWindowBrightness` and `SDL_GetWindowBrightness`.

What was the reasoning behind adding this (other than it exists in SDL2)? I tried it and the functionality did not match what I expected. [See](https://wiki.libsdl.org/SDL2/SDL_SetWindowBrightness):

> Despite the name and signature, this method sets the brightness of the entire display, not an individual window. A window is considered to be owned by the display that contains the window's center pixel. (The index of this display can be retrieved using [SDL_GetWindowDisplayIndex](https://wiki.libsdl.org/SDL2/SDL_GetWindowDisplayIndex)().) The brightness set will not follow the window if it is moved to another display.
> 
> Many platforms will refuse to set the display brightness in modern times. You are better off using a shader to adjust gamma during rendering, or something similar.

This property feels confusing to me (it doesn't set the _window_ brightness on windows OS) and not particularly useful.

I also looked [over here](https://github.com/libsdl-org/SDL/blob/873ec097ea9d09d6fb8e50602e6d5d27fd5ff5b3/docs/README-migration.md) and:

> SDL_SetWindowBrightness and SDL_SetWindowGammaRamp have been removed from the API, because they interact poorly with modern operating systems and aren't able to limit their effects to the SDL window.
> 
> Programs which have access to shaders can implement more robust versions of those functions using custom shader code rendered as a post-process effect.

Otherwise the triangle and quad rendering seem to work.

avaxar:
Ah well, then. I'm going to remove `pygame._sdl2.Window.brightness`.

avaxar:
I may have done something I shouldn't have done. I was trying to solve the merge conflict, but accidentally made GH close this PR, and not let me able to open it back. PR has been reopened in #1876.

PR #1830, opened 2023-02-13
Remove table in favor of codeblock

novialriptide:
I never exactly liked the table...


MyreMylar (commented):
I'm afraid I can't bring myself to care between a table and a codeblock despite switching back and forth a couple of times.

I leave this up to everyone else/

Starbuck5:
I say keep the table.

PR #1832, opened 2023-02-13
Collidepoint performance improvements

itzpr3d4t0r:


Matiiss (approved):
LGTM, ran the same tests and the PR is faster which makes sense

PR #1833, opened 2023-02-13
Surface.rblit() function.

itzpr3d4t0r:
Use this program to benchmark. You can fiddle with the image size by changing the img_size variable. You can also fiddle with number of tests/repeats and blend modes.
I used this image:
img

```Python
from statistics import mean
from timeit import repeat

import pygame

pygame.init()

screen = pygame.display.set_mode((480, 480))

img_scale = 25
n = 1
r = 1000
num_tests = 1000
pos = (10, 10)

img = pygame.image.load("img.png").convert()
img = pygame.transform.scale(img, (img_scale, img_scale))
g = globals()


def test(stmt, name):
    data = [mean(repeat(stmt, globals=g, number=n, repeat=r))
            for _ in range(num_tests)]
    print(f"Finished {name} | mean: {mean(data)}")
    return data


rblit = test("screen.rblit(img, pos)", "rblit")
blit = test("screen.blit(img, pos)", "blit")
rblit_add = test("screen.rblit(img, pos, 1)", "rblit add")
blit_add = test("screen.blit(img, pos, None, 1)", "blit add")

plt.plot(rblit, label="rblit")
plt.plot(blit, label="blit")
plt.plot(rblit_add, label="rblit_add")
plt.plot(blit_add, label="blit_add")
plt.legend()
plt.show()
```


MyreMylar:
What we need is a copy-pasteable test program for `rblit()` so we can test the performance boost. The collidepoint one was very handy for this.

Might also give us an idea on the break point for where it is little different to blit.

MyreMylar (commented):
As above.

itzpr3d4t0r:
> As above.

yeah gimme a sec

MyreMylar:
Here are some graphs I generated with the image sizes used:

512x512:
image

256x256:
image

128x128:
image

64x64:
image

32x32:
image

16x16:
image

The way I interpret this, is that the more time you spend inside each blit (i.e. each blit is larger), the less the difference between rblit & blit matters to their overall execution time. 

This change will matter most for blits 64x64 and below.

MyreMylar (commented):
My main thought is that this needs a a bit of unit testing in `surface_test`. probably verifying that the output is the same as blit in most cases (no reason it shouldn't be, but somebody could break it later if we don't slap in a test now. Also, checking that different types of iterators work in the sequence (list, tuple, set, dictionary).

itzpr3d4t0r:
> Here are some graphs I generated with the image sizes used:
> 
> 512x512: image
> 
> 256x256: image
> 
> 128x128: image
> 
> 64x64: image
> 
> 32x32: image
> 
> 16x16: image
> 
> The way I interpret this, is that the more time you spend inside each blit (i.e. each blit is larger), the less the difference between rblit & blit matters to their overall execution time.
> 
> This change will matter most for blits 64x64 and below.
Can i have your code pls?

MyreMylar:
> Can i have your code pls? @MyreMylar

I was just using your code above and changing the image size:

E.g. for 512x512:
```
from statistics import mean
from timeit import repeat

import pygame

pygame.init()

screen = pygame.display.set_mode((480, 480))

img_scale = 512
n = 1
r = 1000
num_tests = 1000
pos = (10, 10)

img = pygame.image.load("images/glow.png").convert()
img = pygame.transform.scale(img, (img_scale, img_scale))
print("image size:", img.get_size())
g = globals()


def test(stmt, name):
    data = [mean(repeat(stmt, globals=g, number=n, repeat=r))
            for _ in range(num_tests)]
    print(f"Finished {name} | mean: {mean(data)}")
    return data


rblit = test("screen.rblit(img, pos)", "rblit")
blit = test("screen.blit(img, pos)", "blit")
rblit_add = test("screen.rblit(img, pos, 1)", "rblit add")
blit_add = test("screen.blit(img, pos, None, 1)", "blit add")

plt.plot(rblit, label="rblit")
plt.plot(blit, label="blit")
plt.plot(rblit_add, label="rblit_add")
plt.plot(blit_add, label="blit_add")
plt.legend()
plt.show()

```

src_c/surface.c
```diff
@@ -1884,6 +1887,66 @@ surf_blit(pgSurfaceObject *self, PyObject *args, PyObject *keywds)
...
 
+static PyObject *
+surf_rblit(pgSurfaceObject *self, PyObject *const *args, Py_ssize_t nargs)
+{
+    SDL_Surface *src, *dest = pgSurface_AsSurface(self);
+    SDL_Rect *src_rect, dest_rect, temp;
+    PyObject *blit_pos, *src_surf;
+    int flags_numeric = 0;
+
+    if (nargs != 2 || nargs != 3) {
```
MyreMylar, line 1898:
```suggestion
    if (nargs != 2 && nargs != 3) {
```

The current check doesn't work.

MyreMylar (commented):
Current testing results:

512x512:
image

64x64
image

16x16:
image

1,000 16x16:
image

Which, to sum it up, shows that this is still showing a decent speedup for smaller blits (gap between green/blue lines and red/orange lines) but by the time you hit 512x1512 there is no noticeable per blit difference because you are spending way more time SIMD pixel blitting per blit than you are on the per blit overhead.

I think, it would be worth specifying in the 'why use this?' part of the documentation that the improvement in performance here is all about reducing the per-blit overhead and not in improving the per-pixel performance. As such it may be worth using when you have lots of smaller blits but makes no difference on large single blits.

I think the main remaining unanswered question from me is - "Beyond simple benchmarks, do we have a practical gaming use case for this where you might get a noticeable framerate improvement from switching from normal blit to rblit?".  

I think we need to answer that question because it directly impacts on whether it is worth increasing the complexity of the pygame API for our users by adding more for them to learn ("What is the difference between blit and rblit?") and another choice for them to make )"Should I use blit or rblit for this blitting operation?").

My intent is to try creating a couple of little test programs with small blits with a traditional FPS counter to see if this reduction in blit overhead makes a noticeable impact.

MyreMylar (commented):
"Real World" Test program 1:

```
from typing import List
import random
import pygame

GRAVITY = 200.0


class Particle:
    def __init__(self, particle_list: List, surf, pos, initial_velocity):
        self.particle_list = particle_list
        self.particle_list.append(self)

        self.surf = surf
        self.pos = pos.copy()
        self.velocity = initial_velocity.copy()

    def update(self, dt):
        self.velocity.y += (GRAVITY * dt)
        self.pos.x += (self.velocity.x * dt)
        self.pos.y += (self.velocity.y * dt)

        if self.pos.y > 1024:
            self.particle_list.remove(self)


def create_test_surf(dominant_colour):
    test_surf = pygame.Surface((8, 8), flags=pygame.SRCALPHA)
    test_surf.fill((dominant_colour.r, dominant_colour.g, dominant_colour.b, 50))
    test_surf.fill((dominant_colour.r, dominant_colour.g, dominant_colour.b, 100), pygame.Rect(1, 1, 6, 6))
    test_surf.fill((dominant_colour.r, dominant_colour.g, dominant_colour.b, 200), pygame.Rect(2, 2, 4, 4))
    test_surf.fill((dominant_colour.r, dominant_colour.g, dominant_colour.b, 250), pygame.Rect(3, 3, 2, 2))
    return test_surf


def create_test_surf_no_alpha(dominant_colour):
    test_surf = pygame.Surface((2, 2))
    test_surf.fill((dominant_colour.r, dominant_colour.g, dominant_colour.b, 255))
    return test_surf


pygame.init()
screen = pygame.display.set_mode((1024, 768))

background = pygame.Surface(screen.get_size())
background.fill((0, 0, 0))

font = pygame.font.Font(None, 40)

recent_frame_times = []
frame_buffer = 200.0
clock = pygame.time.Clock()
running = True

# blit stuff
test_surfs = [create_test_surf(pygame.Color("red")),
              create_test_surf(pygame.Color("white")),
              create_test_surf(pygame.Color("blue"))]

test_r_surfs = [create_test_surf(pygame.Color("yellow")),
                create_test_surf(pygame.Color("green")),
                create_test_surf(pygame.Color("white"))]

particles = []
rparticles = []

spawn_particles = False
spawn_r_particles = False

particle_density = 80

while running:

    frame_time = clock.tick()/1000.0
    recent_frame_times.append(frame_time)
    while len(recent_frame_times) > frame_buffer:
        recent_frame_times.pop(0)
    average_frame_time = sum(recent_frame_times)/frame_buffer
    fps = round(1.0 / max(0.0000001, average_frame_time), 1)

        if event.type == pygame.QUIT:
            running = False
        if event.type == pygame.KEYDOWN and event.key == pygame.K_ESCAPE:
            running = False

        if event.type == pygame.MOUSEBUTTONDOWN and event.button == pygame.BUTTON_LEFT:
            spawn_particles = True
        if event.type == pygame.MOUSEBUTTONUP and event.button == pygame.BUTTON_LEFT:
            spawn_particles = False

        if event.type == pygame.MOUSEBUTTONDOWN and event.button == pygame.BUTTON_RIGHT:
            spawn_r_particles = True
        if event.type == pygame.MOUSEBUTTONUP and event.button == pygame.BUTTON_RIGHT:
            spawn_r_particles = False

    if spawn_particles:
        for _ in range(particle_density):
            Particle(particles, random.choice(test_surfs),
                     pygame.Vector2(pygame.mouse.get_pos()[0], pygame.mouse.get_pos()[1]),
                     pygame.Vector2((random.random()-0.5)*200.0, random.random()*-200.0))

    if spawn_r_particles:
        for _ in range(particle_density):
            Particle(rparticles, random.choice(test_r_surfs),
                     pygame.Vector2(pygame.mouse.get_pos()[0], pygame.mouse.get_pos()[1]),
                     pygame.Vector2((random.random()-0.5)*200.0, random.random()*-200.0))

    screen.blit(background, (0, 0))

    for particle in rparticles:
        particle.update(frame_time)
        screen.rblit(particle.surf, particle.pos)

    for particle in particles:
        particle.update(frame_time)
        screen.blit(particle.surf, particle.pos)

    fps_text = font.render("FPS: " + str(fps), True, "white")
    screen.blit(fps_text, (10, 10))
    pygame.display.update()

```

HOW TO USE:

1. Run the program.
2. hold left mouse button to spawn a fountain of regular blit drawn particles.
3. hold right mouse button to spawn a fountain of new style rblit spawned particles.
4. hold each button for about 30 seconds to get a good sampling of the average frame rate for each type.

On my PC this test program showed that the program ran about 10% faster overall with the rblit particles (FPS went from about 55 frames per second holding down left mouse to about 60 FPS holding down right mouse).

This seems like a decent & practical speedup to me in a plausible scenario as I know particles for explosions are popularly used. Welcome others trying it and giving their opinions.

I'm going to now have a look at the so-called "p-blit" PR with a similar set of tests to those I've just conducted on this one as we should probably only have 1 of these faster blit PRs.

Starbuck5:
Hey @itzpr3d4t0r,

I've been long been skeptical of additional functions that micro-optimize argument parsing runtime, which is why I haven't really been reviewing them. However I think I'm coming around on the idea of a "fast blit" function. Fblits proved that people will eagerly adopt APIs like this. With this in mind, what if this PR moved from "rblit" to become "fblit," a single version of fblits? The functions would have a "consistent" story: that they are faster but less flexible. I talked to Ankith about this and he's also onboard.

I think this builds a bit of consensus on how to move this forward in a way that lots of people will be happy with, hopefully.

What do you think?

Temmie3754:
I think the fblit naming convention is also likely a better choice, but overall this addition will serve as a good way to squeeze more performance out of pygame in scenarios where there are numerous small objects (such as particles, etc) on screen.

itzpr3d4t0r:
A couple of thoughts I've accumulated over time (in no particular order):
- The use of a single "fast" blit function would be quite limited, and mainly focused for when you need multiple blits of a single surface onto another with no area or anything specific, or just with varied blend flags. 
- To be honest, we could still add the area thing and it wouldn't really change performance much, since that would just incour in an extra if and run only if necessary.
- fblits is already faster than blit for a single surface. Marginally faster but still meaningful, this function should be compared to flibts with a single surface.
- In #2185, I've made a version of blit (that's faster than starndard blit and slightly slower than rblit) that retains most of blit functionality (in that it returns a rect and accepts blit area) while also adding the possibility to use a single kwarg to position the surface (so like center=(10, 10)) while still being about 30% faster than using get_rect with the same args. What i believe is that a function that is only slightly slower than rblit with that much more functionality would be more broadly useful.
-  I still have to make better docs for this.

So basically over time i got more attached to pblit because of its broader usecases.

PR #1840, opened 2023-02-13
Remove video.Renderer.blit()

yunline:
image

### DIFF
--- a/buildconfig/stubs/pygame/_sdl2/video.pyi
+++ b/buildconfig/stubs/pygame/_sdl2/video.pyi
@@ -138,13 +138,6 @@ class Renderer:
     logical_size: Iterable[int]
     scale: Iterable[float]
     target: Optional[Texture]
-    def blit(
-        self,
-        source: Union[Texture, Image],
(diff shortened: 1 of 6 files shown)


MyreMylar (approved):
I agree that it makes little sense for us to try and pretend that the GPU Renderer/Texture API from SDL is the same as the CPU blit API. 

I think instead we should focus our compatibility efforts on where we can port from one API to the other (e.g. nicely turning a surface into a Texture). Also, moving all this stuff into C.

I guess we could deprecation warning this, but I'm inclined not to while the whole module is marked experimental.

yunline:
Replace 'Renderer.blit()' with 'Renderer.draw_texture()' and remove the 'Texture.draw()' may be better.

Texture should be a class that 'storage texture data'. Drawing is handled by Renderer. 'Renderer.draw_texture()' would be more intuitive.

yunline:
I also recommend to add a 'target' parameter to all drawing functions of rendeter to draw graphics on a texture.

yunline:
This PR is not good enough.

PR #1841, opened 2023-02-13
draw.polygons() function

itzpr3d4t0r:
I used this test program. Note that the base program is testing hollow polygons, to change that and use filled polygons you should modify the `create_poly` function. Performance benefits vary significantly based on size and on hollow/filled_in draws.
```Python
from random import random
from math import cos, sin
from statistics import fmean
from timeit import timeit

import pygame
from pygame.draw import polygon as draw_polygon, polygons as draw_polygons

from plotting import create_plot

pygame.init()
WIDTH, HEIGHT = 400, 400
GLOB = {
    "polygon": draw_polygon,
    "polygons": draw_polygons,
    "win": win,
}

NUMPOLYS = 1000
NUM = 1
TESTS = 500
schedule = range(1, TESTS + 1)
CONST = 1
DIM = 20
PWIDTH = 1
NUMVERTS = 3


def create_poly(pos: tuple):
    verts_pos = []
    angle = 360 / (NUMVERTS - 1)
    rand_base = random() * 360
    for i in range(NUMVERTS):
        verts_pos.append(
            (int(pos[0] + DIM * cos(rand_base + i * angle)),
             int(pos[1] + DIM * sin(rand_base + i * angle)))
        )
    
    return (int(random() * 255), int(random() * 255),
            int(random() * 255)), verts_pos, 2


lst = [create_poly((random() * WIDTH, random() * HEIGHT)) for _ in
       range(NUMPOLYS)]

GLOB["lst"] = lst

data = (
    [timeit(f"polygons(win, lst)",
            number=NUM if CONST else i, globals=GLOB) for i in schedule],
    
    [timeit("for item in lst:"
            "  polygon(win, *item)",
            number=NUM if CONST else i, globals=GLOB) for i in schedule],
)

labels = "polygons", "for loop"

print(f"CONST = {bool(CONST)},"
      f" CALLS_PER_TEST = {NUM if CONST else 'increases linearly'},"
      f" TESTS = {TESTS},"
      f" POLYS TO DRAW = {NUMPOLYS} \n")

p1 = round((fmean(data[1]) / fmean(data[0]) - 1) * 100, 4)

print(
    f"polygons is: {p1 if p1 > 0 else -p1}%"
    f" {'faster' if p1 > 0 else 'slower'} than for loop")

create_plot("Original", "id-num", "time(s)", schedule, labels, data,
            do_show_info=True)
create_plot("Sorted", "id-num", "time(s)", schedule, labels, data,
            sort_data=True)

# # VISUAL TEST ------------------------------------------------------------------
clock = pygame.time.Clock()
use_draw_polygons = True
keep = True
while keep:
    win.fill((0, 0, 0))
    delta_time = clock.tick_busy_loop(1000)
    if use_draw_polygons:
        draw_polygons(win, lst)
    else:
        for poly in lst:
            draw_polygon(win, *poly)
    
        if event.type == pygame.QUIT:
            keep = False
            if event.key == pygame.K_r:
                use_draw_polygons = not use_draw_polygons
    
    pygame.display.set_caption(str(clock.get_fps()))
    pygame.display.update()
pygame.quit()

```
Plotting module here:
[plotting.txt](https://github.com/pygame-community/pygame-ce/files/10771406/plotting.txt)


MyreMylar:
I'm pretty sure this one will be valuable for people wanting to draw polygons for 3D rendering (without using a GPU), but it could use a neat copy paste-able test program that shows how it is faster than a for loop of draw polygon.

Probably someone has already written a basic 3d model loading and drawing program we could adapt here.

MyreMylar (commented):
As above. I want a convenient test program to compare old vs new.

itzpr3d4t0r:
@MyreMylar Could you re-review this PR? It changed a lot recently.

Temmie3754 (approved):
Code looks good to me and functions as expected on my system, thanks to the test I can confirm a consistent ~30% speedup for hollow polygons and ~20% for filled with no signs of memory leakage. Only change that is needed would be a versionadded comment for the docs for when its merged.

docs/reST/ref/draw.rst
```diff
@@ -148,6 +148,42 @@ object around the draw calls (see :func:`pygame.Surface.lock` and
...
 
+.. function:: polygons
+
+   | :sl:`draw a sequence of polygons on a surface`
+   | :sg:`polygons(surface, draw_sequence) -> None`
+
+   Draws a sequence of polygons on the given surface.
+
+   :param surface: the surface to draw the polygons on
+   :param draw_sequence: a sequence of (points, color, width) where:
```
MyreMylar, line 159:
```suggestion
   :param draw_sequence: a sequence of (color, points, width) where:
```

The original text appears to be wrong. The sequence is actually in the order:
`(color, points, width)`

docs/reST/ref/draw.rst
```diff
@@ -148,6 +148,42 @@ object around the draw calls (see :func:`pygame.Surface.lock` and
...
+
+        - points: a sequence of 3 or more (x, y) coordinates that make up the vertices
+                  of the polygon, each *coordinate* in the sequence must be a
+                  tuple/list/:class:`pygame.math.Vector2` of 2 ints/floats,
+                  e.g. ``[(x1, y1), (x2, y2), (x3, y3)]``
+
+        - color: the color to draw the polygon with, the alpha value is optional if
+                 using a tuple ``(RGB[A])``
```
MyreMylar, line 167:
```suggestion
        - color: the color to draw the polygon with, the alpha value is optional if
                 using a tuple ``(RGB[A])``
                 
        - points: a sequence of 3 or more (x, y) coordinates that make up the vertices
                  of the polygon, each *coordinate* in the sequence must be a
                  tuple/list/:class:`pygame.math.Vector2` of 2 ints/floats,
                  e.g. ``[(x1, y1), (x2, y2), (x3, y3)]``
```

Swap the order of these explanations to match the correct sequence order.

MyreMylar (changes requested):
The speed up still works for me, looks like it got slightly faster 👍 

Apart from the little ordering mix-up pointed out in the line-by-line comments the main thing I also noticed that we haven't added this function to the draw tests.

For an example you can see what has been done for `draw.lines` by searching `draw.lines` and `draw_lines` in `draw_test.py`. Getting a bit of unit testing of this would be helpful in case somebody else decides to come and tweak it for performance later on and breaks something.

Great work!

PR #1844, opened 2023-02-13
Add mixer.Channel.set_source_location()

yunline:
```python
def set_source_location(self, angle:float, distance:float) -> None: ...
```

Example:
```python
import pygame
import time

pygame.init()
pygame.mixer.init()

snd=pygame.mixer.Sound("你吼那么大声干什么嘛.mp3")

ch=snd.play(loops=10)
angle=0
distance=50 # distance from 0 to 255
while ch.get_busy():
    ch.set_source_location(angle,distance)
    angle+=1
    angle%=360
    time.sleep(0.005)
```
You should hear the sound running around you.


docs/reST/ref/mixer.rst
```diff
@@ -477,6 +477,16 @@ change the default buffer by calling :func:`pygame.mixer.pre_init` before
 
       .. ## Channel.fadeout ##
 
+   .. method:: set_position
+
+      | :sl:`set the position of a playing channel`
+      | :sg:`set_volume(angle,distance) -> None`
+
+      Set the position (angle, distance) of a playing chennal.
```
MyreMylar, line 485: 'chennal' should be 'channel'

yunline: Sorry for that

MyreMylar:
We could perhaps have an interactive test here

src_c/mixer.c
```diff
@@ -1160,6 +1160,27 @@ chan_unpause(PyObject *self, PyObject *_null)
...
+    int channelnum = pgChannel_AsInt(self);
+    Sint16 angle;
+    Uint8 distance;
+    PyThreadState *_save;
+
+    if (!PyArg_ParseTuple(args, "hb", &angle, &distance))
+        return NULL;
+
+    MIXER_INIT_CHECK();
+    _save = PyEval_SaveThread();
```
Starbuck5, line 1175: Use `Py_BEGIN_ALLOW_THREADS` and `Py_END_ALLOW_THREADS` instead of this please.

Starbuck5:
I assumed from the title of this PR that this was a way to set the playback time of a channel. I.e. rewind or seek forward to a time.

I'm quite excited to see it's actually about setting the direction of a sound effect. That could be a super cool effect!

However I'm wondering if we can get a more clear name than `set_position` to illustrate the functionality. Maybe `set_direction` or something like `set_playback_angle`? Just throwing out ideas.

src_c/mixer.c
Starbuck5: Oh, I see why you did it like that. So you could return in the if statement.

I'd still prefer it to use the macros, which could be done by saving Mix_SetPosition's return value to a variable and testing it after END_ALLOW_THREADS.

OR, if Mix_SetPosition isn't an expensive function maybe it shouldn't be releasing the GIL at all.

docs/reST/ref/mixer.rst
```diff
@@ -480,6 +480,21 @@ change the default buffer by calling :func:`pygame.mixer.pre_init` before
 
       .. ## Channel.fadeout ##
 
+   .. method:: set_position
+
+      | :sl:`set the position of a playing channel`
+      | :sg:`set_volume(angle,distance) -> None`
```
yunline, line 486: oops, it shouldn't be `set_volume`

test/mixer_test.py
Starbuck5, line 1: It's super nice to give it an interactive test, but it should probably have a basic unit test as well.

Like assert it fails if you pass in wrongly typed arguments.

MyreMylar:
> I assumed from the title of this PR that this was a way to set the playback time of a channel. I.e. rewind or seek forward to a time.
> 
> I'm quite excited to see it's actually about setting the direction of a sound effect. That could be a super cool effect!
> 
> However I'm wondering if we can get a more clear name than `set_position` to illustrate the functionality. Maybe `set_direction` or something like `set_playback_angle`? Just throwing out ideas, what do you think?

I believe this is usually called positional audio. Looking elsewhere OpenAL uses a concept of audio sources for its positional audio implementation - perhaps we could go with something like `set_source_location()` which may be both somewhat familiar to users of OpenAL, cover both direction and angle, and also unique/intriguing enough that other users read the docs to figure out exactly what it does. @Starbuck5 is correct that we are already using the function name `set_pos()` in `mixer.music` to refer to setting the time position in the music file.

src_c/mixer.c
yunline: This is copied from `chan_set_volume`
https://github.com/pygame-community/pygame-ce/blob/e5a27e5d0079cccc098358142942e6d247075f6e/src_c/mixer.c#L1186-L1192
I don't know if it's necessary to release the GIL

yunline:
This function should support float input

src_c/mixer.c
Starbuck5: It'll work, this is just a stylistic concern.

Not really important, so I'll resolve this review thread about it.

Starbuck5 (approved):
Thanks!

PR #1851, opened 2023-02-13
Run python3 setup.py format

yunline:


MyreMylar:
These changes actually already got made a little earlier on main when I was scrambling about trying to get the CI to run.

If you check the actual file in the repo it already looks like the changes so I assume the merge view just hasn't caught up to that.

Closing as completed.

PR #1852, opened 2023-02-13
Add concurrecy check to CI runs.

MyreMylar:
Should cancel old in progress jobs when new one is started in same workflow.

Idea borrowed from Pillow [here](https://github.com/python-pillow/Pillow/blob/main/.github/workflows/test-windows.yml).
GitHub docs [here](https://docs.github.com/en/actions/using-jobs/using-concurrency).

### DIFF
--- a/.github/workflows/build-windows.yml
+++ b/.github/workflows/build-windows.yml
@@ -30,6 +30,10 @@ on:
       # re-include current file to not be excluded
       - '!.github/workflows/build-windows.yml'
 
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
+
 jobs:
   build:
     name: ${{ matrix.name }}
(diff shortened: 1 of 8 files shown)


.github/workflows/build-windows.yml
```diff
@@ -38,13 +42,6 @@ jobs:
       fail-fast: false  # if a particular matrix build fails, don't skip the rest
       matrix:
         include:
-          - {
-            name: "x86_64 (CPython 3.6 3.10 and above)",
-            winarch: AMD64,
-            msvc-dev-arch: x86_amd64,
-            # pattern matches 6 or any 2 digit number
-            pyversions: "cp3{6,[1-9][0-9]}-*"
-          }
```
MyreMylar, line 47: We had these builds twice so I removed them to save some CI slots.

Issue #1672, opened 2023-02-12
pygame.font.Font | empty string (3394)


GalacticEmperor1:
#  **Issue &numero;3394 opened by *[Vazno](https://github.com/Vazno)* at 2022-08-14 11:37:16**

**Description**
Currently:
``pygame.font.Font()`` takes as first argument only path to font, or None
Example:
``pygame.font.Font("path/to/font.ttf", size)`` > Works
And if we give it an empty string ``pygame.font.Font("", size)`` it will raise an error saying: ``FileNotFoundError: No file '' found in working directory 'path/to/working/directory/'.``
What I suggest is to add possibility to call Font class with an empty string as a first param:
``pygame.font.Font("", size)`` which will use system's default font too, like when we insert None.
It looks like [PR](https://github.com/pygame/pygame/pull/1488) close to this theme was closed a few years ago, but it was about pygame.font.Font.render() method.<hr>

#  Comments

# # #  **[Starbuck5](https://github.com/Starbuck5)* commented at 2022-08-15 06:52:31*

We already alias `None` and `"freesansbold.ttf"` to the default font, I don't personally see any need to add another alias.
<hr>

# # #  **[MyreMylar](https://github.com/MyreMylar)* commented at 2022-08-15 08:40:42*

A better option anyway would surely be:

```
font = pygame.font.Font()  #  default font at size 12
font = pygame.font.Font(size=32)  #  default font at size 32
```

with no supplied path parameter at all producing fonts from the default font.
<hr>

# # #  **[PurityLake](https://github.com/PurityLake)* commented at 2022-10-09 23:51:04*

Made a draft PR of a suggested implementation based on MyreMylar's suggestion. # 3489 
<hr>

# # #  **[ankith26](https://github.com/ankith26)* commented at 2022-10-11 15:34:27*

+1 for MyreMylar's suggestion, definitely makes things simpler than both `None` and `""`

PR #1853, opened 2023-02-13
Add defaults to enable passing no arguments when creating a `pygame.Font()`

PurityLake:
#1672 

This is a suggested solution to the Font class. It simply makes the font name and size arguments options with default values of "freesansbold.ttf" and 12 respectively.

This wouldn't be a breaking change as the default usual behavior still works, it just made the args optional instead of mandatory.


src_c/font.c
```diff
@@ -72,6 +72,8 @@ static const char resourcefunc_name[] = "getResource";
 #endif
 static const char font_defaultname[] = "freesansbold.ttf";
 
+static const int font_defaultsize = 12;
```
MyreMylar, line 75: Should we make this default size a bit higher? The pygame default font is tiny at 12 (it comes out only about 6 pixels high for some reason):

image

This is it at 24 instead:
image

On the other hand, other fonts will be large at 24

MyreMylar (commented):
It seems like the first argument needs to nail down what it is called for this to work with keyword arguments

In the code it is called "font", in the typehints it is called "name", in the docs it is called "file_path".

PurityLake:
I will sort that out this evening

src_c/font.c
PurityLake: That is a good question, maybe 16 splitting the different between 12 and 24?

docs/reST/ref/font.rst
```diff
@@ -170,7 +171,8 @@ solves no longer exists, it will likely be removed in the future.
    the height of the font in pixels. If the filename is ``None`` the pygame
    default font will be loaded. If a font cannot be loaded from the arguments
    given an exception will be raised. Once the font is created the size cannot
-   be changed.
+   be changed. If no arguments are given then the default font will be used and
```
oddbookworm, line 174: Maybe add a note in the docs that the parameter name changed in 2.1.4?

oddbookworm (changes requested):
As Myre commented, the default font is super tiny at the default size
```
import pygame

pygame.init()

default_font = pygame.font.Font()
text1 = default_font.render("Hello World!", True, "red")

freesans_font = pygame.font.Font("freesansbold.ttf")
text2 = freesans_font.render("Hello World!", True, "red")

running = True
while running:
        if event.type == pygame.QUIT:
            running = False
            
    screen.fill("black")
    screen.blit(text1, (0, 0))
    screen.blit(text2, (0, 250))
    pygame.display.flip()
```
image

docs/reST/ref/font.rst
PurityLake: I have added that detail now.

PurityLake:
I have updated the font size to 20, I feel 20 is quite big enough for it's purposes

buildconfig/stubs/pygame/font.pyi
```diff
@@ -32,7 +32,7 @@ class Font:
     italic: bool
     underline: bool
     strikethrough: bool
-    def __init__(self, name: Optional[FileArg], size: int) -> None: ...
+    def __init__(self, filename: Optional[FileArg], size: int) -> None: ...
```
MyreMylar, line 35:
```suggestion
    def __init__(self, filename: Optional[FileArg] = None, size: int = None) -> None: ...
```

MyreMylar: At the minute the type hints don't work because they are missing the default values.

MyreMylar (changes requested):
Other than the type hints, this seems good.

Fix that and I will approve 👍

buildconfig/stubs/pygame/font.pyi
```diff
@@ -32,7 +32,7 @@ class Font:
     italic: bool
     underline: bool
     strikethrough: bool
-    def __init__(self, name: Optional[FileArg], size: int) -> None: ...
+    def __init__(self, filename: Optional[FileArg] = None, size: int = None) -> None: ...
```
MyreMylar, line 35:
```suggestion
    def __init__(self, filename: Optional[FileArg] = None, size: int = 20) -> None: ...
```

src_c/doc/font_doc.h
```diff
@@ -8,7 +8,7 @@
 #define DOC_PYGAMEFONTGETFONTS "get_fonts() -> list of strings\nget all available fonts"
 #define DOC_PYGAMEFONTMATCHFONT "match_font(name, bold=False, italic=False) -> path\nfind a specific font on the system"
 #define DOC_PYGAMEFONTSYSFONT "SysFont(name, size, bold=False, italic=False) -> Font\ncreate a Font object from the system fonts"
-#define DOC_PYGAMEFONTFONT "Font(filename, size) -> Font\nFont(pathlib.Path, size) -> Font\nFont(object, size) -> Font\ncreate a new Font object from a file"
+#define DOC_PYGAMEFONTFONT "Font(filename=None, size=12) -> Font\nFont(filename, size) -> Font\nFont(pathlib.Path, size) -> Font\nFont(object, size) -> Font\ncreate a new Font object from a file"
```
MyreMylar, line 11:
```suggestion
#define DOC_PYGAMEFONTFONT "Font(filename=None, size=20) -> Font\nFont(filename, size) -> Font\nFont(pathlib.Path, size) -> Font\nFont(object, size) -> Font\ncreate a new Font object from a file"
```

MyreMylar (commented):
Whoops looks like my last suggestion had the wrong default value for size, but upon reviewing I noticed it was still set to 12 all over the shop as well.

MyreMylar (approved):

src_c/doc/font_doc.h
```diff
@@ -73,6 +73,7 @@ pygame.font.SysFont
 create a Font object from the system fonts
 
 pygame.font.Font
+ Font(filename=None, size=12) -> Font
```
MyreMylar, line 76:
```suggestion
 Font(filename=None, size=20) -> Font
```

src_py/ftfont.py
```diff
@@ -34,7 +34,7 @@ class Font(_Font):
     __unull = "\x00"
     __bnull = b"\x00"
 
-    def __init__(self, file, size=-1):
+    def __init__(self, file=None, size=-1):
```
ankith26, line 37: shouldn't this be `size=20` too?

PurityLake: Yes it should, sorry, didn't think to change that

MyreMylar, line 37:
```suggestion
    def __init__(self, file=None, size=20):
```

oddbookworm (approved):
LGTM!

PR #1860, opened 2023-02-14
Modified `_sdl2.RendererDriverInfo`

yunline:
- Renamed `RendererDriverInfo` as `RendererDriver`.
- Changed the API of setting the renderer driver ( `driver` parameter of Renderer ).
- Add an API to get current driver of a renderer.
- Docs added ( which haven't been done before ).

Example:
```python
import pygame

print("Available renderer drivers:")
for i in video.RendererDriver.get_drivers():
    print(i)
print("=======")

win=video.Window("sb")
r=video.Renderer(win,driver=video.RendererDriver(1))
print("Using renderer driver:")
print(r.driver)

r.draw_color=(255,255,255,255)
r.fill_rect((10,10,100,100))
r.present()

while 1:
    pygame.event.get()
```
image

### DIFF
--- a/buildconfig/stubs/pygame/_sdl2/video.pyi
+++ b/buildconfig/stubs/pygame/_sdl2/video.pyi
@@ -1,3 +1,4 @@
+from __future__ import annotations
 from typing import Any, Generator, Iterable, Optional, Tuple, Union
 
 from pygame.color import Color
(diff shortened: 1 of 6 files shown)


yunline:
reference: https://github.com/pygame/pygame/issues/3627

PR #1873, opened 2023-02-16
Use `SDL_GetTicks64` instead of `SDL_GetTicks`

avaxar:
Supposed to close pygame/pygame#3658. This replaces the every of `SDL_GetTicks` with `SDL_GetTicks64`. `SDL_GetTicks` would overflow once every 49.71 days, as it returns an `Uint32`. Of course, most programs wouldn't run that long, but it's still nice to fix such case for the few running it that long. Using `SDL_GetTicks64` returns an `Uint64`, which extends the overflowing cycle to 584,542,046 years. If anyone's planning to make a graphical time-capsule using pygame to the end of Earth, hit me up to replace it with `SDL_GetTicks128`.

### DIFF
--- a/src_c/event.c
+++ b/src_c/event.c
@@ -1610,10 +1610,10 @@ _pg_event_wait(SDL_Event *event, int timeout)
     /* Custom re-implementation of SDL_WaitEventTimeout, doing this has
      * many advantages. This is copied from SDL source code, with a few
      * minor modifications */
-    Uint32 finish = 0;
+    Uint64 finish = 0;
 
     if (timeout > 0)
-        finish = SDL_GetTicks() + timeout;
+        finish = SDL_GetTicks64() + timeout;
 
     while (1) {
         _pg_event_pump(1); /* Use our custom pump here */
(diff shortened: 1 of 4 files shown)


Starbuck5:
https://wiki.libsdl.org/SDL2/SDL_GetTicks64

> This function is available since SDL 2.0.18.

avaxar:
Ah, thought it was from 2.0.8.

Starbuck5:
You could add a PG_GetTicks64 that either calls SDL_GetTicks or SDL_GetTicks64 and returns a 64 bit int either way.

pgcompat.c or one of the pygame.h's would be a good location.

avaxar:
Alright. No wonder the CIs failed. Their SDL versions were less than 2.0.18.

Matiiss (approved):
LGTM

MyreMylar (approved):
I think at some point we'll just want to upgrade to using `SDL_GetPerformanceCounter()` on modern systems as in Ankith's old timing PR. 

AFAIK it is just superior in every way.

Still this should help people who run pygame-ce apps for months at a time. Which is probably someone!

robertpfeiffer:
> Still this should help people who run pygame-ce apps for months at a time. Which is probably someone!

:wave:

PR #1874, opened 2023-02-16
Better VSync

robertpfeiffer:
This is a work in progress. Right now we force a SDL_Renderer/GPU code path when vsync without other flags indicating GPU is passed.

Resizing and fullscreen work, but toggling fullscreen does not. The behaviour is meant to be indistinguishable from regular un-SCALED pygame software rendering. Re-sizing the window changes the size of the display surface. This feels like a horrible hack, but it runs smooth.


MyreMylar (commented):
Gave this a bit of a test on my windows machine. 

1. Vsync itself works great in the unscaled window. My 'moving a block around' test program runs much smoother with vsync turned on. Hooray!
2. Tried a 'traditional pygame toggle' rather than using `toggle_fullscreen()` and it also doesn't work. At the third call to set_mode() it just breaks (though you can toggle back out again).

This was the test code I tried:

```
import pygame


pygame.init()


pygame.display.set_caption('Vsync toggle test')
vsync = 1
window_surface = pygame.display.set_mode((1920, 1080), flags=pygame.FULLSCREEN, vsync=vsync)

background = pygame.Surface((1920, 1080))
background.fill(pygame.Color(0, 0, 0))


font = pygame.font.Font(None, 48)
fps_render = font.render('0 FPS', True, pygame.Color('#FFFFFF'))


clock = pygame.time.Clock()
is_running = True

while is_running:
    time_delta = clock.tick()/1000.0
        if event.type == pygame.QUIT:
            is_running = False

        if event.type == pygame.KEYDOWN and event.key == pygame.K_RETURN:
            if vsync:
                vsync = 0
            else:
                vsync = 1
            window_surface = pygame.display.set_mode((1920, 1080), flags=pygame.FULLSCREEN, vsync=vsync)

    fps_render = font.render(f"{clock.get_fps():.2f}" + " FPS", True, pygame.Color('#FFFFFF'))

    window_surface.blit(background, (0, 0))
    window_surface.blit(fps_render, (1650, 50))

    pygame.display.update()
```

Slapping the return key to toggle out of and then, failing to toggle back into vsync mode.

robertpfeiffer:
It works on my machine...

robertpfeiffer:
how about this one on main and on this PR?

```py
import pygame
import os

os.environ["SDL_HINT_RENDER_SCALE_QUALITY"]="best"
pygame.init()


pygame.display.set_caption('Vsync toggle test')
vsync = 1
scaled= 1
window_surface = pygame.display.set_mode((1280,720), flags=pygame.FULLSCREEN|pygame.SCALED, vsync=vsync)

font = pygame.font.Font(None, 48)
fps_render = font.render('0 FPS', True, pygame.Color('#FFFFFF'))


clock = pygame.time.Clock()
is_running = True

x = 0

flags=pygame.FULLSCREEN|pygame.SCALED

while is_running:
    time_delta = clock.tick()/1000.0
        if event.type == pygame.QUIT:
            is_running = False

        if event.type == pygame.KEYDOWN and event.key == pygame.K_s:
            scaled = int(not scaled)
            if scaled:
                flags=pygame.FULLSCREEN|pygame.SCALED
            else:
                flags=pygame.FULLSCREEN
            window_surface = pygame.display.set_mode((1280,720), flags=flags, vsync=vsync)

        if event.type == pygame.KEYDOWN and event.key == pygame.K_RETURN:
            vsync= int(not vsync)
            window_surface = pygame.display.set_mode((1280,720), flags=flags, vsync=vsync)

    fps_render = font.render(f"{clock.get_fps():.2f}" + " FPS", True, pygame.Color('#FFFFFF'))
    
    w, h = window_surface.get_size()
    x= (x+3)%w

    window_surface.fill(0)

    for i in range(10):
        x1=(x+10*i)%w
        pygame.draw.line(window_surface, (255,255,255), (x1,0), (x1,h), 3)
    
    for i in range(10):
        x1=(x+100+3*i)%w
        pygame.draw.line(window_surface, (255,255,255), (x1,0), (x1,h))


    window_surface.blit(fps_render, (50, 50))

    pygame.display.update()

```

MyreMylar:
The behaviour (on my windows machine) with the above test seems the same on main as it is with his PR - basically you can only start a vsync'd display once.

You can start in vsync and then leave it successfully - fine. You can't then call `set_mode()` to re-enter a vsync'd display.

You can also start with a non-vsync'd display and then enter a vsync'd one successfully. You can then leave it, but can't get back to it.

When it breaks the behaviour is just a frozen screen as if the display has stopped updating/flipping.

I guess it's been this way for a while on windows?

MyreMylar (approved):
Going to approve this as it doesn't actually change the annoying behaviour on windows and improves lots of other things regarding vsync.

robertpfeiffer:
I think this needs more supporting functionality to be useful: A function to determine if we are currently running with vsync, and a function to determine the current refresh rate. That last item might need to be two different functions for SDL API reasons.

novialriptide (approved):
Running the code that blubber posted works on my high refresh rate monitor (`Lenovo G27-20 27" FHD Gaming Monitor 165Hz`) at around 165fps.

novialriptide (approved):
I see nothing out of the ordinary with the new changes, LGTM.

docs/reST/ref/display.rst
```diff
@@ -173,21 +173,42 @@ required).
    .. versionadded:: 2.0.0 ``SCALED``, ``SHOWN`` and ``HIDDEN``
 
    By setting the ``vsync`` parameter to ``1``, it is possible to get a display
-   with vertical sync, but you are not guaranteed to get one. The request only
-   works at all for calls to ``set_mode()`` with the ``pygame.OPENGL`` or
-   ``pygame.SCALED`` flags set, and is still not guaranteed even with one of
-   those set. What you get depends on the hardware and driver configuration
-   of the system pygame is running on. Here is an example usage of a call
+   with vertical sync at a constant frame rate. Subsequent calls to
+   :func:`pygame.display.flip()` will block (i.e. *wait*) until the screen has
```
MyreMylar, line 177:
```suggestion
   :func:`pygame.display.flip()` or  :func:`pygame.display.update()` will block (i.e. *wait*) until the screen has
```

docs/reST/ref/display.rst
```diff
@@ -723,7 +744,43 @@ required).
 
    .. versionadded:: 2.0.0
 
+.. function:: is_vsync
+
+   | :sl:`Returns True if vertical synchronisation for pygame.display.flip() is enabled`
```
MyreMylar, line 749:
```suggestion
   | :sl:`Returns True if vertical synchronisation for pygame.display.flip() and pygame.display.update() is enabled`
```

MyreMylar (commented):
Some small doc suggestions

robertpfeiffer:
Merging this now, more doc changes will come with a display docs PR anyway...

PR #1863, opened 2023-02-14
Implement `pitch` argument for `pygame.image.frombuffer`, `pygame.image.fromstring`, and `pygame.image.frombytes`

avaxar:
Supposed to close pygame/pygame#3630 if I had the access. This PR implements a ~~`stride`~~ `pitch` argument for the functions `pygame.image.frombuffer`, `pygame.image.fromstring`, and `pygame.image.frombytes`. The argument is placed at the very back, so that it wouldn't ruin the behavior of existing code.

# Examples

```py
>>> pygame.image.fromstring(b"\xFF\xFF..\xFF\xFF..", (2, 2), "P", False, 4)
<Surface(2x2x8 SW)>
>>> pygame.image.frombytes(b"\xFF\xFF..\xFF\xFF..", (2, 2), "P", False, 4)
<Surface(2x2x8 SW)>
>>> pygame.image.frombuffer(b"\xFF\xFF..\xFF\xFF..", (2, 2), "P", 4)
<Surface(2x2x8 SW)>
```

```py
>>> pygame.image.fromstring(b"\xFF\xFF\xFF...\xFF\xFF\xFF...", (1, 2), "RGB", False, 6)
<Surface(1x2x24 SW)>
>>> pygame.image.frombytes(b"\xFF\xFF\xFF...\xFF\xFF\xFF...", (1, 2), "RGB", False, 6)
<Surface(1x2x24 SW)>
>>> pygame.image.frombuffer(b"\xFF\xFF\xFF...\xFF\xFF\xFF...", (1, 2), "RGB", 6)
<Surface(1x2x24 SW)>
```

```py
>>> pygame.image.fromstring(b"\xFF\xFF\xFF\xFF...\xFF\xFF\xFF\xFF...", (1, 2), "RGBA", False, 7)
<Surface(1x2x32 SW)>
>>> pygame.image.frombytes(b"\xFF\xFF\xFF\xFF...\xFF\xFF\xFF\xFF...", (1, 2), "RGBA", False, 7)
<Surface(1x2x32 SW)>
>>> pygame.image.frombuffer(b"\xFF\xFF\xFF\xFF...\xFF\xFF\xFF\xFF...", (1, 2), "RGBA", 7)
<Surface(1x2x32 SW)>
```

PR #1875, opened 2023-02-16
Add "ABGR" format to pygame.image.fromstring

novialriptide:
Ported from: https://github.com/pygame/pygame/pull/3674
Fixes: https://github.com/pygame/pygame/issues/3629

Looking at the SDL_PIXELFORMAT page (https://wiki.libsdl.org/SDL2/SDL_PixelFormatEnum), they have
```
SDL_PIXELFORMAT_RGBA32
SDL_PIXELFORMAT_ARGB32
SDL_PIXELFORMAT_BGRA32
SDL_PIXELFORMAT_ABGR32
```
Similarly, in pygame.image.fromstring, we support
```
RGBA
ARGB
BGRA
```
But with no `ABGR`

Somebody might need that support someday, and it seems like an obvious addition to what we support already.

### DIFF
--- a/src_c/image.c
+++ b/src_c/image.c
@@ -1159,6 +1159,28 @@ image_frombytes(PyObject *self, PyObject *arg)
         }
         SDL_UnlockSurface(surf);
     }
+    else if (!strcmp(format, "ABGR")) {
+        if (len != (Py_ssize_t)w * h * 4)
+            return RAISE(
+                PyExc_ValueError,
+                "Bytes length does not equal format and resolution size");
+        surf = SDL_CreateRGBSurface(SDL_SRCALPHA, w, h, 32,
+#if SDL_BYTEORDER == SDL_LIL_ENDIAN
+                                    0xFF << 24, 0xFF << 16, 0xFF << 8, 0xFF);
+#else
+                                    0xFF << 16, 0xFF << 24, 0xFF, 0xFF << 8);
+#endif
+        if (!surf)
+            return RAISE(pgExc_SDLError, SDL_GetError());
+        SDL_LockSurface(surf);
+        for (looph = 0; looph < h; ++looph) {
+            Uint32 *pix = (Uint32 *)DATAROW(surf->pixels, looph, surf->pitch,
+                                            h, flipped);
+            memcpy(pix, data, w * sizeof(Uint32));
+            data += w * sizeof(Uint32);
+        }
+        SDL_UnlockSurface(surf);
+    }
     else
         return RAISE(PyExc_ValueError, "Unrecognized type of format");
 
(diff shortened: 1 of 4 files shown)


MyreMylar:
Debian fail is:

```
====================================================================
  ERROR: test_frombuffer_ABGR (pygame.tests.image_test.ImageModuleTest)
    File "/usr/local/lib/python3.9/dist-packages/pygame/tests/image_test.py", line 1176, in test_frombuffer_ABGR
      argb_surf = pygame.image.frombuffer(abgr_buffer, (4, 4), "ABGR")
  pygame.error: Unknown pixel format
  
  Ran 39 tests in 8.986s
  
  FAILED (errors=1, skipped=1)
```

You might need to do something like this:

```
        try:
            # test if hardware supports ABGR surfaces
            pygame.Surface((64, 64), flags=pygame.SRCALPHA, masks=(0, 0xFF00, 0xFF0000, 0xFF000000))
        except pygame.error:
            pass
        else:
            argb_surf = pygame.image.frombuffer(abgr_buffer, (4, 4), "ABGR")
            self.assertEqual(argb_surf.get_at((0, 0)), pygame.Color(20, 10, 255, 200))
            self.assertEqual(argb_surf.get_at((1, 1)), pygame.Color(255, 255, 255, 127))
            self.assertEqual(argb_surf.get_at((2, 2)), pygame.Color(0, 0, 0, 79))
            self.assertEqual(argb_surf.get_at((3, 3)), pygame.Color(20, 200, 50, 255))
```

If I have those masks correct.

novialriptide:
> You might need to do something like this:

The Debian tests still fail when I insert that code into the tests. Also, I feel like that's a hacky sort of solution. There should be a better way.

src_c/image.c
```diff
@@ -1147,6 +1147,28 @@ image_frombytes(PyObject *self, PyObject *arg)
...
+        for (looph = 0; looph < h; ++looph) {
+            Uint32 *pix = (Uint32 *)DATAROW(surf->pixels, looph, surf->pitch,
+                                            h, flipped);
+            memcpy(pix, data, w * sizeof(Uint32));
+            data += w * sizeof(Uint32);
+        }
+        SDL_UnlockSurface(surf);
+    }
+    else if (!strcmp(format, "ABGR")) {
+        if (len != (Py_ssize_t)w * h * 4)
```
avaxar, line 1163: You haven't implemented the changes with the merging of #1863, implementing a `pitch` argument for every format.

avaxar (changes requested):

src_c/image.c
novialriptide: Unfortunate

Issue #1862, opened 2023-02-14
Shall .c files generated by cython be included in the .gitignore?


yunline:
Shall .c files generated by cython be included in the .gitignore?
Auto-merged .c files may be unuseable and needs to be regenerated.

Starbuck5:
Generating the cython at build time is not a simple thing to do well.

I tried to do this, https://github.com/pygame/pygame/pull/2916, but I gave up after encountering lots of errors.

We're hoping to migrate away from the cython code, so maybe that makes it easier to live with this less than ideal set up for now.

But yes, for now we need to be careful to merge cython PRs serially.

zoldalma999:
Done by #2831

PR #1877, opened 2023-02-16
Generate Cython code at build time

yunline:
- [x] Generate Cython code at build time.
- [x] Remove `cython` option from setup.py. (but keep the `cython_only`)
- [x] Add cython-generated files into .gitignore .
- [x] #2395

Reference: 
[https://github.com/pygame/pygame/pull/2916](https://github.com/pygame/pygame/pull/2916)
#1862


.github/workflows/build-windows.yml
```diff
@@ -156,7 +158,7 @@ jobs:
           set DISTUTILS_USE_SDK=1
           set MSSdk=1
           python -m pip install setuptools wheel requests numpy Sphinx
-          python setup.py docs
+          python3 setup.py docs
```
Starbuck5, line 161: What, why?

.github/workflows/build-windows.yml
```diff
@@ -131,6 +131,8 @@ jobs:
       CIBW_BUILD: ${{ matrix.pyversions }}
       CIBW_ARCHS: ${{ matrix.winarch }}
 
+      CIBW_BEFORE_BUILD: python -m pip install setuptools wheel Cython requests
+
```
Starbuck5, line 135: Why not just move the cython install into the line 158/160 pip install?

.github/workflows/build-windows.yml
yunline: ???
I will change it back

.github/workflows/build-windows.yml
yunline: If moved, the CIBW would not run correctly ( LNK1104 error )

Starbuck5: That's crazy.

MyreMylar (approved):
I'm generally approving of this as a stop gap before steadily turning cython files to C files.

pmp-p:
Though the wasm CI is already doing that, this is probably a bad idea for testing against python git where cython files **may** need to be edited.

This is what i do locally after 3.11 and 3.12(git) are built:

- i build with python3.11 : cython files + pygame
- Then if i know cython files will fail on 3.12, i patch the 3.11 generated one
- then continue build without running cython_only.

yunline:
> Though the wasm CI is already doing that, this is probably a bad idea for testing against python git where cython files **may** need to be edited.

Maybe we can solve this problem by adding a 'non-cython' option.

MyreMylar:
slightly longer term (but getting closer) the plan is to clear out all the cython code entirely.

Is this PR useful enough to current cython development to get over the downsides? Or should we close it. I tend to avoid cython entirely.

setup.py
```diff
@@ -288,11 +279,12 @@ def consume_arg(name):
...
         cythonize_one(**kwargs)
-    
-    if cython_only:
-        sys.exit(0)
 
-no_compilation = any(x in ['lint', 'format', 'docs'] for x in sys.argv)
+no_compilation = any(x in ['lint', 'format', 'docs','cython_only'] for x in sys.argv)
+
+if not no_compilation: # Compile cython before compilation
+    compile_cython()
```
dr0id, line 286: Why are those lines not in the build command? Or are these files needed elsewhere too?

setup.py
```diff
@@ -947,6 +939,16 @@ class LintCommand(LintFormatCommand):
 class FormatCommand(LintFormatCommand):
     format = True
 
+@add_command("cython_only")
+class CythonOnlyCommand(Command):
+
+    user_options = []
+    def initialize_options(self):pass
```
dr0id, line 946: please add a description:
`    description = "short doc what this command does"`

can be seen with: `python setup.py --help-commands`

there is also `python setup.py cython_only --help`
not sure what that would produce at the moment.

yunline: image
It seems other commands don't have descriptions too. Maybe we need another pr to fix that.

dr0id: yes, see https://github.com/pygame-community/pygame-ce/pull/2360

ankith26:
This PR has been superseded by #2831 because we recently introduced the new buildconfig, so I'm closing this 

But thanks for starting the work on this!

PR #1881, opened 2023-02-16
Installing & updating grub to cppcheck as it seems to fail

MyreMylar:
At the moment.


yunline:
Also in [build-ubuntu-sdist.yml](https://github.com/pygame-community/pygame-ce/actions/runs/4197175086/jobs/7279149216#step:3:415)

MyreMylar:
This didn't work.

PR #1882, opened 2023-02-16
Bump version of actions/checkout

MyreMylar:
To avoid deprecation warning when running actions.


ankith26 (approved):
LGTM thanks for the PR :star:

PR #1884, opened 2023-02-17
Faster simple scaling algorithm using SDL functions

Starbuck5:
Formerly known as https://github.com/pygame/pygame/pull/3610

For reduced code footprint and increased performance.

My testing shows a 40% speedup for normal scales as well as scales with destination surfaces.

The scale output does change, but I can only notice the difference when I switch between them in at the same location onscreen in an image editor.

Benchmark script:
```py
import pygame
import os
import time

pygame.init()

os.chdir("../../official_pygame/examples/data")
im1 = pygame.image.load("fist.png").convert_alpha()

start = time.time()
for _ in range(100000): #100000
    im2 = pygame.transform.scale_by(im1, 0.5)
print(time.time() - start)

pygame.quit()
raise SystemExit
```
I also fixed a bug where the destination surface could be the right bit depth but wrong pixel format. Before, it would stretch the image onto the dest surface even though the color information would be all messed up. Now, I use the error handling of SDL_SoftStretch to catch the format errors. This does change the error from a ValueError to a pygame.error, but I think this behavior is so rare that it doesn't need to be emulated.

Here is a compilation of files scaled in certain ways by the existing algorithm (normal) and the new algorithm (stretch):
[scale_stretch.zip](https://github.com/pygame-community/pygame-ce/files/10764201/scale_stretch.zip)
[scale_normal.zip](https://github.com/pygame-community/pygame-ce/files/10764202/scale_normal.zip)


MyreMylar (approved):
Already spent some time testing and approving this three weeks ago.

Still looks good to me 👍

docs/reST/ref/transform.rst
```diff
@@ -47,6 +47,10 @@ Instead, always begin with the original image and scale to the desired size.)
    the destination must be the same size as the size (width, height) passed in. Also
    the destination surface must be the same format.
 
+   .. versionchanged:: 2.1.4 internal scaling algorithm was replaced with a nearly
```
MyreMylar, line 50:
```suggestion
   .. versionchanged:: 2.2.0 internal scaling algorithm was replaced with a nearly
```

src_c/transform.c
```diff
@@ -544,27 +437,21 @@ scale_to(pgSurfaceObject *srcobj, pgSurfaceObject *dstobj, int width,
...
         Py_BEGIN_ALLOW_THREADS;
-        if (width == 2 * src->w && height == 2 * src->h) {
-            scale2xraw(src, retsurf);
-        }
-        else {
-            stretch(src, retsurf);
-        }
-        Py_END_ALLOW_THREADS;
 
+        stretch_result_num = SDL_SoftStretch(src, NULL, retsurf, NULL);
```
ankith26, line 447: Hmmm... The SDL docs for [`SDL_SoftStretch`](https://wiki.libsdl.org/SDL2/SDL_SoftStretch) says

> Please use [`SDL_BlitScaled()`](https://wiki.libsdl.org/SDL2/SDL_BlitScaled) instead.

I wonder whether that applies for our usecase?

oddbookworm: I'm not too well-versed in this part of SDL, but I tried to test this implementation and I got something weird. It would throw an error about "Surfaces must not be locked during blit" with a straight replacement, and if I removed the surface lock, it did work but took 4-5x as long as current scaling

Starbuck5: @ankith26 Myremylar had the same question when reviewing this PR when it existed upstream.

We do not want a blit, we want a copy. A blit might be an alpha blit, it might take into account the background, so we need to have a plain background, and getting a new surface with a plain background is hard. Because SDL defaults to (0,0,0,0) when we would need (0,0,0,255). 

So it's much simpler to just "stretch" and have it overwrite whatever the destination pixels are.

ankith26 (approved):
Okay I'm convinced now. I like how this PR removes code on our side, thanks for the PR now :tada: 

PS to any other reviewers/members: Don't merge this ATM, gotta bump dev version first

PR #1888, opened 2023-02-17
Add brew install pkg-config to mac builds

ankith26:
Somehow `pkg-config` is missing in the mac runners now, which has led to new fails.
A workaround is implemented in this PR: we install `pkg-config` with brew.


MyreMylar (approved):

MyreMylar (approved):
hooray it works! 👍

oddbookworm (approved):
LGTM!

PR #1891, opened 2023-02-18
Fix cppcheck crash.

yunline:
Fix error in step `Install deps`.

### DIFF
--- a/.github/workflows/build-ubuntu-sdist.yml
+++ b/.github/workflows/build-ubuntu-sdist.yml
@@ -56,6 +56,7 @@ jobs:
       # install numpy from pip and not apt because the one from pip is newer,
       # and has typestubs
       run: |
+        sudo apt-mark hold grub-efi-amd64-signed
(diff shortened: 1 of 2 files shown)


oddbookworm (approved):
I'm not sure about this being a permanent solution because I'm not entirely sure if updating this is ultimately needed for the CI or if the docker images are just having a stroke, but the Ubuntu 20.04 check passed and I'm fairly confident that it resolves the issue with the cppcheck workflow as well because they complain about the same thing. So I'm approving it on the basis that this will at least get us back on track for getting PR's to pass the checks.

ankith26:
Hmmm. This fix works for now, but I suspect this fix would not be needed once the issue is fixed on the github actions side

MyreMylar:
I reported the underlying issue to GitHub here: https://github.com/orgs/community/discussions/47863

We could just merge this for now and monitor that issue to see if they take action to fix it?

Starbuck5 (approved):
I say we merge this for now.

Issue #1850, opened 2023-02-13
Drop CPython 3.6 support


itzpr3d4t0r:
Python 3.6 is officially dead:
![image](https://user-images.githubusercontent.com/103119829/218437104-a8978162-ea42-44f7-8313-344cd0369f6a.png)
 
I propose we drop compatibility for 3.6. This would allow us to lower CI runtime and let us use stuff that 3.7+ only supports, such as FASTCALL (speaking of C API).

MightyJosip:
Looking at some of the most used python libraries on PyPi (or documentation page if PyPi doesn't have Requires section)
| library | version supported |
|---------|-------------------|
| pandas  | 3.8               |
| numpy   | 3.8               | 
| keras   | 3.7               |
| tensorflow   | 3.7               |
| scipy  | 3.8              |
| pytorch  | 3.7             |
| flask  | 3.7             |
| opencv  | 3.6             |

pmp-p:
i think opencv is abi3 3.7+

PR #1896, opened 2023-02-19
Drop CPython 3.6 support

yunline:
Issue: #1850


Starbuck5:
@ankith26 What do you want to do about the ubuntu-18.04 build here?

ankith26:
If we are dropping 3.6 we also have to drop the 18.04 build (gh actions have deprecated it anyways)

MyreMylar (approved):
LGTM 👍 

I can't find any more CPython 3.6 references outside of a single comment (which is still relevant).

MyreMylar:
Looking at > 3.7 stuff we could get rid of the blocks like this:

```
#if PY_VERSION_HEX < 0x03070000
#ifdef WITH_THREAD
    PyEval_InitThreads();
#endif /* WITH_THREAD */
#endif
```

itzpr3d4t0r:
yeah the `_pygame.h` one should be related to the `FASTCALL` macros we introduced a while back for back compat with 3.6.

MyreMylar (approved):
Changes look good 👍

oddbookworm (approved):
LGTM! But seeing as how major a change this is, I'm not going to merge it because I think we might want at least one more approval?

Starbuck5 (approved):

src_c/_pygame.h
```diff
@@ -281,41 +281,13 @@ supported Python version. #endif */
...
-
-#else /* PY_VERSION_HEX >= 0x03070000 */
 /* compat macros are no-op on python versions that support fastcall */
 #define PG_FASTCALL_NAME(func) func
 #define PG_DECLARE_FASTCALL_FUNC(func, self_type)
 #define PG_WRAP_FASTCALL_FUNC(func, self_type)
 
 #define PG_FASTCALL METH_FASTCALL
 
-#endif /* PY_VERSION_HEX >= 0x03070000 */
```
ankith26, line 317: not sure whether it should be a part of this PR, or block this PR. But with this change many of these macro references can be removed in the codebase

.github/workflows/build-macos.yml
```diff
@@ -83,10 +83,10 @@ jobs:
         # of these builds take roughly the same time
         include:
           - { 
-            name: "x86_64 (CPython 3.6 3.10 and above)",
+            name: "x86_64 (CPython 3.10 and above)",
             macarch: x86_64,
-            # pattern matches 6 or any 2 digit number
-            pyversions: "cp3{6,[1-9][0-9]}-*"
+            # pattern matches any 2 digit number
+            pyversions: "cp3{10,[1-9][0-9]}-*"
```
ankith26, line 89:
```suggestion
            pyversions: "cp3[1-9][0-9]-*"
```
The explicit 10 here is not needed, the pattern covers for this

ankith26 (approved):
Overall the PR looks good, left two minor reviews for your consideration. You may choose to fix these in this PR, or (you or anyone else can) do a different PR shortly after merging this.

Thanks for the PR! :tada:

src_c/_pygame.h
ankith26: Cool!

PR #1898, opened 2023-02-19
Multiline text

Starbuck5:
Originally opened as [https://github.com/pygame/pygame/pull/3538](https://github.com/pygame/pygame/pull/3538)

Adds multiline text, through a wraplength parameter on font.render.
Also adds a font.align property to control left/right/center aligned text.


oddbookworm:
Works very well with this code that I was using to test dynamic width rescaling
```
import pygame

pygame.init()

f = pygame.font.Font(None, 50)
f.align = pygame.FONT_RIGHT

class TextBox:
    def __init__(self, width: int, pos: pygame.Vector2, text: str = ''):
        self.font = f
        self.width = width
        self.pos = pos
        self.text = text
        
        self.img = self.font.render(self.text, True, "black", None, self.width)
        self.rect = self.img.get_rect(topleft=self.pos)
        
    def change_width(self, amount):
        self.width += amount
        self.img = self.font.render(self.text, True, "black", None, self.width)
        self.rect = self.img.get_rect(topleft=self.pos)
        
    def draw(self, screen):
        screen.blit(self.img, self.rect)
        pygame.draw.rect(screen, "red", self.rect, 1)
        
buffer_size = 10
text = TextBox(200, pygame.Vector2(10, 10), "Hello\nWorld!")
mouse_rect = pygame.Rect(0, 0, buffer_size, buffer_size)

running = True
while running:
        if event.type == pygame.QUIT:
            running = False
    
    mouse_pressed = pygame.mouse.get_pressed()
    if any(mouse_pressed):
        mousepos = pygame.mouse.get_pos()
        mouse_rect.center = mousepos
        
        if text.rect.colliderect(mouse_rect) and mousepos[0]-buffer_size/2<text.rect.right<mousepos[0]+buffer_size/2:
            new_width = mousepos[0] - text.rect.right
            text.change_width(new_width)
            
    screen.fill("white")
    text.draw(screen)
    pygame.display.flip()
```

oddbookworm (approved):

src_c/font.c
```diff
@@ -425,9 +483,10 @@ font_render(PyObject *self, PyObject *args)
     Uint8 rgba[] = {0, 0, 0, 0};
     SDL_Surface *surf;
     const char *astring = "";
+    int wraplength = 0;
```
novialriptide, line 486: We should add an error handler for negatives. No segmentation fault happens if we do, but we get this error: `pygame.error: Invalid parameter 'wrapLength'` which I think is ugly.

docs/reST/ref/font.rst
```diff
@@ -296,6 +312,9 @@ solves no longer exists, it will likely be removed in the future.
         pygame supports rendering UCS4 unicode including more languages and
         emoji.
 
+      .. versionchanged:: 2.1.4 newline characters now will break text into
```
novialriptide, line 315: Should specify `wraplength` keyword

novialriptide (changes requested):
## Review has been ported from the original pygame project

Testcases for `wraplength` seem to be missing. Although I don't know if that was intentional or not since we're just implementing SDL's wraplength.

Align seems to work just fine for me, just one slight problem..
```python
a = font.render("hi starbuck", True, (255, 255, 255), wraplength=10)
```

This doesn't actually work, returning this:
```
PS C:\Users\novia\Documents\GitHub\pygame> python tester.py
pygame 2.1.4.dev1 (SDL 2.0.22, Python 3.11.0)
  File "C:\Users\novia\Documents\GitHub\pygame\tester.py", line 14, in <module>
    a = font.render("hi starbuck", True, (255, 255, 255), wraplength=10)
        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
TypeError: Font.render() takes no keyword arguments
```

We need to do this instead.
```python
a = font.render("hi starbuck", True, (255, 255, 255), (0, 0, 0, 0), 10)
```

The documentation is misleading, as it suggests that it would accept keyword arguments by showing its default values.

Instead of using `PyArg_ParseTuple()`, should we use `PyArg_ParseTupleAndKeywords()`? It'll be a performance loss, but the method that works is a messy implementation in my opinion.

## Example Code
```python
import pygame
import sys
pygame.init()

font = pygame.font.Font(None, 25)

while True:
    for e in pygame.event.get():
        if e.type == pygame.QUIT:
            sys.exit()
    
    screen.fill((0, 0, 0))
    a = font.render("hi starbuck", True, (255, 255, 255), (0, 0, 0, 0), 10)
    screen.blit(a, (0, 0))
    pygame.display.flip()
```

docs/reST/ref/font.rst
Starbuck5: Wraplength would be a versionadded. This specifically is a versionchanged. In 2.1.3 `\n` would render as a box, in this version it will actually break the line, regardless of your wraplength setting.

Starbuck5:
> The documentation is misleading, as it suggests that it would accept keyword arguments by showing its default values.

> Instead of using PyArg_ParseTuple(), should we use PyArg_ParseTupleAndKeywords()? It'll be a performance loss, but the method that works is a messy implementation in my opinion.

Plenty of pygame is like this. It still works.

src_c/font.c
Starbuck5: Sure, I suppose. More explicit error message can't hurt.

Starbuck5:
I'm not sure whether I will or should add support for keyword arguments to this function in this PR.

MyreMylar (approved):
LGTM 👍 

Tested with my own program and also the code posted above and it all seems to behave as it should with different widths and text alignments.

novialriptide (approved):
LGTM!

Starbuck5:
* Force push just to squash the commits

Issue #1845, opened 2023-02-13
A way to detect whether Pygame-CE is being used.


Sjmarf:
This would be useful in cases where a CE-exclusive feature is used, but you want to fallback to an alternative if the user doesn't have Pygame CE installed. Or, it could be used to warn the user upon program startup that they aren't using the correct version of Pygame.

MyreMylar:
Good point.

We will likely have to add a testable flag/method in the pygame module if the fork and the upstream continue with parallel development (i.e. there is a 2.1.4 of pygame and a 2.1.4 of pygame-ce with different featuresets). 

For now there is no real difference between the last development release of upstream pygame and the full 2.1.3 release of pygame-ce.

Starbuck5:
I propose a `pygame.IS_CE` attribute.

One would test it with `getattr(pygame, "IS_CE", False)`

This could be implemented on `__init__.py` in src_py.

PR #1900, opened 2023-02-19
Add `IS_CE` to constants

yunline:
Issue: #1845


buildconfig/stubs/pygame/__init__.pyi
```diff
@@ -52,12 +51,6 @@ from .math import Vector2 as Vector2, Vector3 as Vector3
 from .cursors import Cursor as Cursor
 from .bufferproxy import BufferProxy as BufferProxy
 from .mask import Mask as Mask
-from ._debug import print_debug_info as print_debug_info
```
yunline, line 55: Why are these changed???

oddbookworm (changes requested):
In my testing, this worked great. `getattr(pygame, "IS_CE", False)` returned `True`. I think docs should be added for this feature however, which is the reason for the change request. Probably best to document it in https://github.com/pygame-community/pygame-ce/blob/main/docs/reST/ref/pygame.rst IMO

MyreMylar:
This also seems to be crashing the CI in the sphinx docs.

Perhaps we could just do a:

```
DEC_CONSTS(IS_CE, 1);
```

instead?

yunline:
> This also seems to be crashing the CI in the sphinx docs.
> 
> Perhaps we could just do a:
> 
> ```
> DEC_CONSTS(IS_CE, 1);
> ```
> 
> instead?

I doubt if the crash was caused by IS_CE.
And IS_CE should be a bool but DEC_CONSTS genetates an integer constant.

MyreMylar (approved):
Tested with:

```
import pygame


if getattr(pygame, "IS_CE", False):
    print("Running pygame CE")
else:
    print("Not running pygame CE")
```

And it works as it should.

oddbookworm (approved):
LGTM!

PR #1903, opened 2023-02-19
Move cython based controller module to controller_old

zoldalma999:
This is the first step to porting the module to C. I don't want to have terrible diff with the next pr (as that is somewhat big as-is), so this would be a requirement to opening my next pr. I think keeping it with an _old prefix is fine for now if anyone wants to use the old version for some reason (but it should be removed sometime in the future).


MyreMylar (approved):
LGTM 👍 

Excited to see someone start to tackle this beast!

Starbuck5:
I know that people do actually use _sdl2.controller, and this seems to move the module at the python-top-level as well.

Wouldn't it be preferable to keep the top level module intact and just move the internal files around to prepare for the ones?

ankith26 (approved):
LGTM, thanks for contributing :apple:

PR #1904, opened 2023-02-19
Delete old unused scrap module code

MyreMylar:
Resurrected from closed PR upstream.

Just cleans out a bunch of old unused code from the scrap module, some of which is now deprecated on mac OS.
(Take 2 on this, rebasing did not work out).

Test program for old scrap (this should still work on linux and mac after this code is removed):

```
import pygame
from io import BytesIO
import pygame.scrap

display_surf = pygame.display.set_mode((800, 600))

test_surface = pygame.Surface((50, 50))
test_surface.fill((255, 0, 0))

pygame.init()

pygame.scrap.init()

print("Is scrap init?: ", pygame.scrap.get_init())

clock = pygame.time.Clock()

running = True

while running:
        if event.type == pygame.QUIT:
            running = False

        if event.type == pygame.KEYDOWN and event.key == pygame.K_F1:
            text = pygame.scrap.get(pygame.SCRAP_TEXT)
            if text:
                print("There is text in the clipboard.")
                print("This is it:", str(text))
            else:
                print("There does not seem to be text in the clipboard.")

        if event.type == pygame.KEYDOWN and event.key == pygame.K_F2:
            image = pygame.scrap.get(pygame.SCRAP_BMP)
            if image:
                print("found bmp image in buffer")
                test_surface = pygame.image.load(BytesIO(image), ".bmp")
            else:
                print("no image found")

        if event.type == pygame.KEYDOWN and event.key == pygame.K_F3:
            print("Available types in clipboard: ", pygame.scrap.get_types())

        if event.type == pygame.KEYDOWN and event.key == pygame.K_F4:
            pygame.scrap.put(pygame.SCRAP_TEXT, b"A text to copy")

    display_surf.fill((0, 0, 0))
    display_surf.blit(test_surface, (100, 100))

    pygame.display.flip()

    clock.tick(60)

```

Expected output after pressing F4 then F1:

```
Is scrap init?:  True
There is text in the clipboard.
This is it: b'A text to copy'
```


Starbuck5 (approved):

oddbookworm (approved):
LGTM. From my limited testing, current bytes-based scrap is broken on linux either way (exact same error message for pre- and post-PR)

PR #1905, opened 2023-02-19
Rect `collidelist()`/`collidelistall()` speedup

itzpr3d4t0r:
Changed:
- The functions now use `METH_O` instead of `METH_VARARGS` with a `PyArg_ParseTuple(args, "O", &list)`. This makes them 100% compatible as before.
- There are two paths for different types of sequences: a fast path for sequences that allow faster iteration and a normal path for general sequences.

Added:
- A new macro called `pgSequenceFast_Check` that checks if a sequence object is compatible with PySequence_Fast. Currently, the macro is defined as `pgSequenceFast_Check(o) (PyList_Check(o) || PyTuple_Check(o))`. This will make it easier to add another check if Python introduces another fast sequence type in the future.
- More tests for both functions

Perf. tests:
TLDR:
collidelist: 66% faster on my pc.
collidelistall: 53% faster on my pc.
Of course the longer the sequence, the better the benefits.
```
OLD
===| collidelist |===
1): 216.64ms
2): 217.03ms
3): 216.89ms
4): 215.63ms
5): 215.41ms
Mean: 216.32ms
===| collidelistall |===
1): 252.01ms
2): 248.19ms
3): 256.87ms
4): 255.00ms
5): 256.72ms
Mean: 253.76ms

NEW
===| collidelist |===
1): 130.00ms
2): 129.85ms
3): 129.98ms
4): 131.40ms
5): 130.17ms
Mean: 130.28ms
===| collidelistall |===
1): 163.49ms
2): 166.11ms
3): 165.83ms
4): 165.82ms
5): 166.47ms
Mean: 165.54ms
```
I used this program:
```Python
from pygame import Rect

from statistics import fmean
from timeit import repeat


rects = [
    Rect(0, 0, 100, 100),
    Rect(50, 50, 100, 100),
    Rect(100, 100, 100, 100),
    Rect(150, 150, 100, 100),
    Rect(200, 200, 100, 100),
    Rect(250, 250, 100, 100),
    Rect(300, 300, 100, 100),
    Rect(350, 350, 100, 100),
    Rect(400, 400, 100, 100),
    Rect(450, 450, 100, 100),
    Rect(500, 500, 100, 100),
    Rect(550, 550, 100, 100),
    Rect(600, 600, 100, 100),
    Rect(650, 650, 100, 100),
    Rect(700, 700, 100, 100),
]

r = Rect(799, 799, 33, 33)
rep = 10
N = 5
tot = 0

print("===| collidelist |===")
for i in range(N):
    data = repeat("r.collidelist(rects)", globals=globals(), repeat=rep)
    tot += fmean(data)
    print(f"{i+1}): {fmean(data)*1000:.2f}ms")
print(f"Mean: {tot/N*1000:.2f}ms")

print("===| collidelistall |===")
tot = 0
for i in range(N):
    data = repeat("r.collidelistall(rects)", globals=globals(), repeat=rep)
    tot += fmean(data)
    print(f"{i+1}): {fmean(data)*1000:.2f}ms")
print(f"Mean: {tot/N*1000:.2f}ms")
```

### DIFF
--- a/src_c/_pygame.h
+++ b/src_c/_pygame.h
@@ -316,6 +316,12 @@ supported Python version. #endif */
 
 #endif /* PY_VERSION_HEX >= 0x03070000 */
 
+/* Update this function if new sequences are added to the fast sequence
+ * type. */
+#ifndef pgSequenceFast_Check
+#define pgSequenceFast_Check(o) (PyList_Check(o) || PyTuple_Check(o))
+#endif /* ~pgSequenceFast_Check */
+
 /*
  * event module internals
  */
--- a/src_c/rect.c
+++ b/src_c/rect.c
@@ -808,11 +808,11 @@ pg_rect_collidelistall(pgRectObject *self, PyObject *arg)
             }
         }
     }
-    /* If the sequence is not a fast sequence, we have to use the slower
-     * PySequence_GetItem() function to get the items. */
+    /* Slower path for general sequences. */
     else {
         for (loop = 0; loop < PySequence_Length(arg); loop++) {
-            PyObject *obj = PySequence_GetItem(arg, loop);
+            PyObject *obj = PySequence_ITEM(arg, loop);
+
             if (!obj || !(argrect = pgRect_FromObject(obj, &temp))) {
                 Py_XDECREF(obj);
                 Py_DECREF(ret);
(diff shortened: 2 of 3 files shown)


novialriptide (approved):
LGTM. I couldn't find any issues with the code here, but it's probably worth a second review from either Starbuck5, ankith, or Myre.

Matiiss (approved):
LGTM, got about a 20% - 30% decrease in time taken using the given test code 👍

MyreMylar (approved):
LGTM 👍 

Ran substantially faster on my machine and the code looks good.

PR #1909, opened 2023-02-20
Rewrite stars.py

pycoinfu:
Refactored the `stars.py` example: made it use OOP and less math; simplified the code


pycoinfu:
Just realized I forgot to bring the quote at the end back, not sure if it needs to be there though 😅

yunline (commented):
Maybe define a class and pack `create_particle()`, `update_particles()` etc into it is better.
Like:
```python
class Particles:
    def __init__(self):
        self._particles=[]
        ...

    def create_particle(self, ... ):
        ...

    ...

    def mainloop(self):
        ...
```

examples/stars.py
```diff
@@ -1,103 +1,135 @@
 #!/usr/bin/env python
-""" pg.examples.stars
+"""
+pg.examples.stars
```
novialriptide, line 3: To stay consistent with the other examples, keep it like `""" pg.examples.stars`

examples/stars.py
```diff
@@ -1,103 +1,135 @@
...
-        pos[0] = pos[0] + vel[0]
-        pos[1] = pos[1] + vel[1]
-        if not 0 <= pos[0] <= WINSIZE[0] or not 0 <= pos[1] <= WINSIZE[1]:
-            vel[:], pos[:] = init_star()
-        else:
-            vel[0] = vel[0] * 1.05
-            vel[1] = vel[1] * 1.05
+
+class Particle:
+    """
```
novialriptide, line 21: Nitpick: I don't think this comment here is needed

novialriptide (changes requested):

examples/stars.py
pycoinfu: Thanks for the review 👍🏻

novialriptide (approved):
LGTM

MyreMylar (approved):
LGTM 👍 

I do have a couple of fun suggestions:

See below; adding some random colours to draw, and regaining the 'single pixel' look of the original by using draw line:
```
        self.colour_list = [(255, 210, 125),
                            (255, 163, 113),
                            (166, 168, 255),
                            (255, 250, 134),
                            (168, 123, 255)]

    def draw(self, display: pg.Surface):
        pg.draw.line(display, random.choice(self.colour_list),
                     self.pos, self.pos)
```

Also, add a density variable to make more stars:

```
        for _ in range(particle_densisty):
            create_particle(particles, spawn_point)
```

I liked a density of about 10.

pycoinfu:
> LGTM 👍
> 
> I do have a couple of fun suggestions:
> 
> See below; adding some random colours to draw, and regaining the 'single pixel' look of the original by using draw line:
> 
> ```
>         self.colour_list = [(255, 210, 125),
>                             (255, 163, 113),
>                             (166, 168, 255),
>                             (255, 250, 134),
>                             (168, 123, 255)]
> 
>     def draw(self, display: pg.Surface):
>         pg.draw.line(display, random.choice(self.colour_list),
>                      self.pos, self.pos)
> ```
> 
> Also, add a density variable to make more stars:
> 
> ```
>         for _ in range(particle_densisty):
>             create_particle(particles, spawn_point)
> ```
> 
> I liked a density of about 10.

Thanks for reviewing my PR 😄
I've implemented the suggestions, and it really does look better now!

oddbookworm (approved):
LGTM!

PR #1910, opened 2023-02-20
Don't merge: Try building and then installing wheel on macOS 11 and running tests

MyreMylar:
OK, this is working now, building then installing a failing wheel on macOS 11.

If anyone wants to have a go at the macOS 11 text bug then this might help you test out any fixes.


ankith26:
Thanks to this PR we can confirm #1922 works on mac 11. The mac 10.15 fail here is unrelated to the issue, but starbuck has locally tested that PR on an older 10.x mac so I'm confident the PR works there too

MyreMylar:
Finished with this now.

PR #1912, opened 2023-02-20
get_just_pressed and get_just_released

ScriptLineStudios:
duplicate of my old pr - https://github.com/pygame/pygame/pull/3237


docs/reST/ref/key.rst
```diff
@@ -269,6 +269,45 @@ for ``KMOD_NONE``, which should be compared using equals ``==``). For example:
...
+   and is therefore similar to checking KEYUPS in the event loop.
+
+   .. note::
+      If you require getting the key releases in order use the event queue KEYUP events.
+
+   ::
+      if pygame.key.get_just_released()[pygame.K_b]:
+         print("B key just released")
+
+   .. versionadded:: 2.1.3
```
oddbookworm, line 307: Should be 2.1.4

docs/reST/ref/key.rst
```diff
@@ -269,6 +269,45 @@ for ``KMOD_NONE``, which should be compared using equals ``==``). For example:
...
+   and is therefore similar to checking KEYDOWNS in the event loop.
+
+   .. note::
+      If you require getting the key presses in order use the event queue KEYDOWN events
+
+   ::
+      if pygame.key.get_just_pressed()[pygame.K_b]:
+         print("B key just pressed")
+
+   .. versionadded:: 2.1.3
```
oddbookworm, line 288: Should be 2.1.4

oddbookworm (changes requested):
I'd be happy to approve this once you update the versionadded tags to 2.1.4

robertpfeiffer:
I really want this functionality merged, because it's something that exists in HaxeFlixel and Unity and everywhere. My main concern is what API/behaviour/guarantees this should give. I think as is, it wouldn't work in any of my games that get() different event types. I think the reset should only be triggered when `dopump` is true, or better yet, the reset code should live inside _pg_event_pump. You have my approval conditional on these changes.

I know this change will make the docs slightly more unwieldy, and I don't know how your code will interact with keyboard events introduced by pygame.event.post, but I think this is worth including now.

docs/reST/ref/key.rst
MyreMylar, line 288:
```suggestion
   .. versionadded:: 2.2.0
```

docs/reST/ref/key.rst
MyreMylar, line 307:
```suggestion
   .. versionadded:: 2.2.0
```

MyreMylar (commented):
Updating the version strings

oddbookworm (approved):
LGTM!

src_c/event.c
```diff
@@ -1907,6 +1923,14 @@ _pg_get_all_events(void)
...
             if (!_pg_event_append_to_list(list, &eventbuf[loop]))
                 goto error;
+
+            if (eventbuf[loop].type == SDL_KEYDOWN) {
+                pressed_keys[(int)eventbuf[loop].key.keysym.scancode] = 1;
+            }
+
+            if (eventbuf[loop].type == SDL_KEYUP) {
+                released_keys[(int)eventbuf[loop].key.keysym.scancode] = 1;
+            }
```
robertpfeiffer, line 1933: I think this should probably live in an event filter.

src_c/event.c
```diff
@@ -1988,6 +2012,9 @@ pg_event_get(PyObject *self, PyObject *args, PyObject *kwargs)
 
     _pg_event_pump(dopump);
 
+    memset(pressed_keys, 0, sizeof(pressed_keys));
+    memset(released_keys, 0, sizeof(released_keys));
+
```
robertpfeiffer, line 2017: And this should really live inside _pg_event_pump

src_c/key.c
```diff
@@ -194,6 +194,49 @@ key_get_pressed(PyObject *self, PyObject *_null)
...
+        PyObject *key_elem;
+        key_elem = PyBool_FromLong(released_keys[i]);
+        PyTuple_SET_ITEM(key_tuple, i, key_elem);
+    }
+    ret_obj = PyObject_CallFunctionObjArgs((PyObject *)&pgScancodeWrapper_Type,
+                                           key_tuple, NULL);
+    Py_DECREF(key_tuple);
+    return ret_obj;
+}
+
```
robertpfeiffer, line 239: The external API looks good and sensible and I have no problem here

robertpfeiffer (changes requested):
I don't think resetting on pygame.event.get() only is a good idea, and resetting on pygame.event.get(pump=False) is a very bad idea. I think it should reset on calls to pump.

I like the idea in general, I like the API, but I think we need to talk about this a little more. I'd also be open to implementing the change myself.

Starbuck5 (approved):
Directly pushed fixes for couple small issues I saw with it.

I like your changes @robertpfeiffer, this seems like a good use for an event filter.

This has got to be a record for the most people who've committed to a single PR branch?

src_c/event.c
```diff
@@ -1596,7 +1612,17 @@ _pg_event_pump(int dopump)
...
+        /* This needs to be reset on pump, e.g. on calls to pygame.event.get(),
+         * but not on pygame.event.get(pump=False) */
+        memset(pressed_keys, 0, sizeof(pressed_keys));
+        memset(released_keys, 0, sizeof(released_keys));
     }
+
+    /* this could also be in an event filter like pg_event_filter
+     * but here it more predictable. */
+    SDL_FilterEvents(_pg_register_keys_just_pressed, NULL);
```
ankith26, line 1624: As a middle term goal, I'd like to move away from custom pump hacks (yes I know, my fault for putting some of it in the first place, but that was years ago at this point). I also had a WIP PR at some point to remove this function entirely (which I plan to revive at around SDL3 release time)

If this bit of code could be moved to the global event filter and still work as expected, that would be great. However not a PR blocker, I think.

As for why this should be a big deal, we are blocked on using stuff like `SDL_WaitEventTimeout` that implicitly call SDLs "vanilla pump". And `SDL_WaitEventTimeout` is a lot better than our current custom implementation (that is actually old SDL code - the newest actually waits for events, as opposed to repeated polling)

src_c/key.c
```diff
@@ -194,6 +194,54 @@ key_get_pressed(PyObject *self, PyObject *_null)
...
+get_just_pressed()
+{
+    char *pressed_keys = pgEvent_GetKeyDownInfo();
+    PyObject *key_tuple = NULL;
+    PyObject *ret_obj = NULL;
+
+    if (!(key_tuple = PyTuple_New(SDL_NUM_SCANCODES)))
+        return NULL;
+
+    VIDEO_INIT_CHECK();
```
ankith26, line 207: minor nitpick: This should be at the top of the function. If it's here, it leaks the tuple allocated above in the exit condition

Starbuck5:
@ankith26 Thanks for taking a look at this.

About your reviews on this PR, you should push those changes directly. Scriptline seems busy right now, and more focused on their recent PRs. Plus, we've got a record to set with the committers on this branch!

MyreMylar:
This has developed a small conflict. Looks like it could be ready to go for 2.4.0 with a tiny bit of love as it already has three approvals.

src_c/event.c
```diff
@@ -92,6 +92,10 @@ static SDL_TimerID _pg_repeat_timer = 0;
 static SDL_Event _pg_repeat_event;
 static SDL_Event _pg_last_keydown_event = {0};
 
+/* Not used as text, acts as an array of bools */
+char pressed_keys[SDL_NUM_SCANCODES];
+char released_keys[SDL_NUM_SCANCODES];
```
ankith26, line 97:
```suggestion
static char pressed_keys[SDL_NUM_SCANCODES] = { 0 };
static char released_keys[SDL_NUM_SCANCODES] = { 0 };
```

src_c/key.c
```diff
@@ -194,6 +194,54 @@ key_get_pressed(PyObject *self, PyObject *_null)
     return ret_obj;
 }
 
+static PyObject *
+get_just_pressed()
```
ankith26, line 198:
```suggestion
get_just_pressed(PyObject *self, PyObject *_null)
```

src_c/key.c
```diff
@@ -194,6 +194,54 @@ key_get_pressed(PyObject *self, PyObject *_null)
...
+        PyTuple_SET_ITEM(key_tuple, i, key_elem);
+    }
+    ret_obj = PyObject_CallFunctionObjArgs((PyObject *)&pgScancodeWrapper_Type,
+                                           key_tuple, NULL);
+    Py_DECREF(key_tuple);
+    return ret_obj;
+}
+
+static PyObject *
+get_just_released()
```
ankith26, line 222:
```suggestion
get_just_released(PyObject *self, PyObject *_null)
```

src_c/key.c
```diff
@@ -194,6 +194,54 @@ key_get_pressed(PyObject *self, PyObject *_null)
...
+    PyObject *key_tuple = NULL;
+    PyObject *ret_obj = NULL;
+
+    if (!(key_tuple = PyTuple_New(SDL_NUM_SCANCODES)))
+        return NULL;
+
+    int i;
+    for (i = 0; i < SDL_NUM_SCANCODES; i++) {
+        PyObject *key_elem;
+        key_elem = PyBool_FromLong(pressed_keys[i]);
```
ankith26, line 212:
```suggestion
        key_elem = PyBool_FromLong(pressed_keys[i]);
        if (!key_elem) {
            Py_DECREF(key_tuple);
            return NULL;
        }
```

src_c/key.c
```diff
@@ -194,6 +194,54 @@ key_get_pressed(PyObject *self, PyObject *_null)
...
+    PyObject *key_tuple = NULL;
+    PyObject *ret_obj = NULL;
+
+    if (!(key_tuple = PyTuple_New(SDL_NUM_SCANCODES)))
+        return NULL;
+
+    int i;
+    for (i = 0; i < SDL_NUM_SCANCODES; i++) {
+        PyObject *key_elem;
+        key_elem = PyBool_FromLong(released_keys[i]);
```
ankith26, line 236:
```suggestion
        key_elem = PyBool_FromLong(released_keys[i]);
        if (!key_elem) {
            Py_DECREF(key_tuple);
            return NULL;
        }
```

ankith26 (changes requested):
Some reviews (mostly notes to self) that I will address when I get back to this PR later today

ankith26 (approved):
I have pushed some changes, and now this PR looks good to go for me.

I also wrote a little test script to validate that my changes don't break anything, here is it for anyone else interested in testing things out
```py
import pygame

win = pygame.display.set_mode((500, 500))

letters = [pygame.K_a + i for i in range(26)]

while True:
    cur_pressed = set()
    cur_released = set()
        if event.type == pygame.QUIT:
            raise SystemExit

            cur_pressed.add(event.key)

        if event.type == pygame.KEYUP:
            cur_released.add(event.key)
    
    for i in letters:
        if (i in cur_pressed) != pygame.key.get_just_pressed()[i]:
            raise RuntimeError(f"{i}")

        if (i in cur_released) != pygame.key.get_just_released()[i]:
            raise RuntimeError(f"{i}")

    pygame.time.wait(10)
```
Just run this script and spam letters randomly. There should be no error raised if the implementation is alright.

I'll wait for someone else to merge this PR though, because the last change (as of now) is mine

Starbuck5 (approved):

Issue #1915, opened 2023-02-21
Remove experimental tag in the documentation


novialriptide:
Related pull request that introduced it: https://github.com/pygame/pygame/pull/3511

PR #1920, opened 2023-02-21
Remove experimental tags for certain functions

PurityLake:
Refers to #1915 

Simply removes the experimental tags from docs of [this PR](https://github.com/pygame/pygame/pull/3511)


Matiiss (approved):
LGTM
I don't really see a need for these, if something here appears broken at some point, people will ask questions either way and a bug report may get submitted, it just may not be as quick, but unlikely that people would have read and followed this warning in the docs either.

MyreMylar (approved):
LGTM 👍 

I think if these functions don't work out we would have to deprecate them anyway, which is just the same as any non-experimental marked function.

PR #1917, opened 2023-02-21
Fix freetype dylib with install_name_tool

Starbuck5:
Fixes the issue originally reported in https://github.com/pygame/pygame/issues/3596

```py
Python 3.10.9 (v3.10.9:1dd9be6584, Dec  6 2022, 14:37:36) [Clang 13.0.0 (clang-1300.0.29.30)] on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>> import pygame
pygame-ce 2.1.4.dev1 (SDL 2.0.22, Python 3.10.9)
>>> pygame.font
<module 'pygame.font' from '/Library/Frameworks/Python.framework/Versions/3.10/lib/python3.10/site-packages/pygame/font.cpython-310-darwin.so'>
>>> pygame.font.init()
>>> 
```

Freetype works as well.
Tested on my old macOS 10.13 laptop.

Wrote an explanation of how this works in the fix dylib python script.

PR #1922, opened 2023-02-22
Fix mac freetype dylib circular import issue, make sure CI generated wheels are self-contained

ankith26:
The freetype self-link was happening in the dependency build step itself, the fix was to remove the old freetype that harfbuzz needs, and re-install freetype (with harfbuzz link) while lying to freetype that harfbuzz doesn't need freetype. But after the new freetype is built linked to harfbuzz, this harfbuzz can also dynamically load the new freetype dylib in theory.

Also in this PR I introduce a little change, the installed dependencies are essentially deleted while a wheel is being tested: this is to ensure the self-contained-ness of the wheel.

An alternative solution to what #1917 fixes

### DIFF
--- a/.github/workflows/build-macos.yml
+++ b/.github/workflows/build-macos.yml
@@ -131,11 +131,16 @@ jobs:
         export MAC_ARCH="${{ matrix.macarch }}"
         brew install pkg-config
         cd buildconfig/macdependencies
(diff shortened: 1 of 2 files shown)


Starbuck5 (approved):
I tested this locally. Font and freetype work, test suites pass, text shaping demo works.

Nice job!

MyreMylar (approved):
LGTM.

Well... it looks like a hack! But it works and it seems like hacks were the only way to deal with this one. I guess this saga helped us all discover a range of new mac dev hacks to take with us into the future.

PR #1926, opened 2023-02-22
Color swizzling

### LINKED ISSUE #1919: Implementing swizzling RGBA values for `pygame.Color`
Taking pygame/pygame#3705 here...

# Description

 There should be a way to construct a new `pygame.Color` from joining multiple color channels of a `pygame.Color` together in respect to RGB(A), simply per a single property access, otherwise known as swizzling. This is alike to GLSL's [swizzling](https://www.khronos.org/opengl/wiki/Data_Type_(GLSL)#Swizzling), where one could construct a vector from another by joining the values of multiple axes in a single identifier (i.e `my_vec.xyz` instead of `vec3(my_vec.x, my_vec.y, my_vec.z)`).

# Example Concept

```py
>>> color = pygame.Color(1, 2, 3, 4)
>>> color.rgb  # Alpha would still be specified optionally; `255` by default.
(1, 2, 3, 255)
>>> color.bgr
(3, 2, 1, 255)
>>> color.rrr
(1, 1, 1, 255)
>>> color.arar
(4, 1, 4, 1)
>>> color.brag
(3, 1, 4, 2)
>>> color.grab
(2, 1, 4, 3)
```

# Notes

 Though, yes, I am aware how many attributes would need to be implemented in C for every single case. Assuming that we'd be supporting combining 3 and 4 axes, that means 4^3 + 4^4 = 320 cases that would need to be implemented. It could be shrunk down to 4^4 = 256 cases if we use default args for the alpha channel for the other 4^3 cases. And we could shrink it even more into one packed function if we hijack `__getattr__`.

ScriptLineStudios:
Closes #1919


gresm:
Should color swizzling also support different color types like ``hsla`` or ``hsva``

src_c/color.c
```diff
@@ -2025,6 +2025,131 @@ _color_getbuffer(pgColorObject *color, Py_buffer *view, int flags)
...
+            if (used[j] == current) {
+                PyErr_SetString(
+                    PyExc_AttributeError,
+                    "Attribute assignment conflicts with swizzling.");
+                return -1;
+            }
+        }
+        switch (current) {
+            case 'r':
+                self->data[RED] = PyLong_AsLong(PySequence_GetItem(val, RED));
```
oddbookworm, line 2065: Compiler complains about implicit typecasting here, should cast to `Uint8`

src_c/color.c
```diff
@@ -2025,6 +2025,131 @@ _color_getbuffer(pgColorObject *color, Py_buffer *view, int flags)
...
+                break;
+            case 'g':
+                self->data[GREEN] =
+                    PyLong_AsLong(PySequence_GetItem(val, GREEN));
```
oddbookworm, line 2069: see above about casting

src_c/color.c
```diff
@@ -2025,6 +2025,131 @@ _color_getbuffer(pgColorObject *color, Py_buffer *view, int flags)
...
+                break;
+            case 'b':
+                self->data[BLUE] =
+                    PyLong_AsLong(PySequence_GetItem(val, BLUE));
```
oddbookworm, line 2073: see above about casting

src_c/color.c
```diff
@@ -2025,6 +2025,131 @@ _color_getbuffer(pgColorObject *color, Py_buffer *view, int flags)
...
+                break;
+            case 'a':
+                self->data[ALPHA] =
+                    PyLong_AsLong(PySequence_GetItem(val, ALPHA));
```
oddbookworm, line 2077: see above about casting

src_c/color.c
```diff
@@ -2025,6 +2025,131 @@ _color_getbuffer(pgColorObject *color, Py_buffer *view, int flags)
...
+{
+    Py_ssize_t len = PySequence_Length(attr_name);
+    PyObject *res;
+
+    if (len == 1) {
+        return PyObject_GenericGetAttr((PyObject *)self, attr_name);
+    }
+
+    if (len == 3 || len == 4) {
+        Uint8 data[len];
```
oddbookworm, line 2102: This doesn't really work. Compiler tries to figure out what size this is at compile time, but fails and errors out

oddbookworm (changes requested):
This overall looks good, but there are some problems that keep it from compiling

src_c/color.c
oddbookworm: I fixed it in my local, so here's what I did to get it to compile (you don't have to do it this way, but you can try it if you want):
```
    if (len == 3 || len == 4) {
        Uint8* data = malloc(sizeof(Uint8) * len);
        res = pgColor_New(data);
        ((pgColorObject *)res)->data[ALPHA] = 255;
        free(data);
    }
```

gresm:
It would be nice if there were some sort of macro helper functions in ``pgcompact.h`` (or more appropriate header file for that matter) for values swizzling? Why rewrite everything from scratch that was already written for vectors and colors when new class (like Rect) get value swizzling, when the idea is usually very similar?

ScriptLineStudios:
closing as discussed here: https://discord.com/channels/772505616680878080/772940667231928360/1078546221708349530

ScriptLineStudios:
closing as discussed here: https://discord.com/channels/772505616680878080/772940667231928360/1078546221708349530

Issue #1927, opened 2023-02-23
Accessing Cameras with the Same Name


mahyarmirrashed:
**Environment:**

- **Operating system** (e.g. Windows, Linux(Debian), Linux(Ubuntu), Mac): Windows 11
- **Python version** (e.g. 3.7.9, 3.8.5) : 3.8.16
- **SDL version** (e.g. SDL 2.0.12): 2.26.3
- **PyGame version** (e.g. 2.0.0.dev10, 1.9.6): 2.1.3
- **Relevant hardware** (e.g. if reporting a bug about a controller, tell us the brand & name of it): Not necessary.

**Current behavior:**

I have three web cameras (from the same manufacturer) connected to my USB C dock.

![image](https://user-images.githubusercontent.com/59240843/220219616-37d4a7be-4f36-4ced-a520-943b648b789c.png)

I create a sample program:

```python
>>> import pygame
pygame 2.1.3 (SDL 2.26.3, Python 3.8.16)
Hello from the pygame community. https://www.pygame.org/contribute.html
>>> import pygame.camera
>>> pygame.init()
(5, 0)
>>> pygame.camera.init()
>>> pygame.camera.list_cameras()
['PC Camera', 'PC Camera', 'HP Wide Vision HD Camera', 'PC Camera']
```

This is annoying. They are all listed as "PC Camera". Therefore, there is no way to access any individual camera, afaik.

**Expected behavior:**

It would be different if these were named tuples with an option that distinguishes them by their ID.

e.g.

```python
>>> pygame.camera.list_cameras()
[CameraSource(name="PC Camera", uuid=""), etc.]
```

mahyarmirrashed:
The best solution here is adding an ability to access the camera objects via index or some other UUID. We could have a dataclass with hidden fields that is then managed by whatever backend. As long as the interface with the user remains consistent, that should be good.

Starbuck5:
Hi, thank you for opening an issue.

I see this as an API oversight in the camera API.

Here's my solution: make all backends support passing indices.
In type hints format:
`Camera(device: int | str, desired_size: Tuple[int, int], format: str)`

Starbuck5:
I opened a super initial PR to try this out (on windows).

If you'd like to test it out, here is the correct wheel for your system I believe:
[...]

PR #1928, opened 2023-02-23
Camera by index

Starbuck5:
Early work, at least. (Windows only)

Ideally before merge we could get this API consistent across the backends.

Not sure what will getting indexing to work on Linux will entail. Maybe @oddbookworm wants to be the Linux lead on camera stuff? 😉 

For #1927

### DIFF
--- a/src_c/_camera.c
+++ b/src_c/_camera.c
@@ -1845,27 +1845,43 @@ camera_init(pgCameraObject *self, PyObject *arg, PyObject *kwargs)
 
     return 0;
 #elif defined(PYGAME_WINDOWS_CAMERA)
-    PyObject *name_obj = NULL;
-    WCHAR *dev_name = NULL;
+    PyObject *identifier_obj = NULL;
     int w, h;
     char *color = NULL;
-    IMFActivate *p = NULL;
+    IMFActivate *act = NULL;
 
     w = DEFAULT_WIDTH;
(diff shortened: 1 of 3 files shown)


oddbookworm:
> Not sure what will getting indexing to work on Linux will entail. Maybe @oddbookworm wants to be the Linux lead on camera stuff? 😉

I can take a crack at it, although I'll need to get some cameras... I've never actually used the camera module

pmp-p:
i've left some thoughts about camera(s) here https://github.com/pygame-community/pygame-ce/issues/1927#issuecomment-1445201829

Coming from broadcast, i'm used to external video/audio feeds with virtual patch bays and i think the camera default should just be ONE camera picked up from a list,

And working correctly in a very consistent manner across all platforms : by that i mean fixing query_image() erratic behaviour, probably with an async warmup and grabbing a first frame. **It's very bad to crash a program because camera failed to start correctly or does not produces frames in the expected way**


Advanced use of camera(S) means codecs/clocks/threads or queue/SMTPE controls/media offline feed  and eventually a bit of async or polling with query_image().

 I really think it's a bit too much for core pygame module. Most likely robotics or surveillance projects may implement that sooner or later in a more complete way.

Starbuck5:
This still needs to be done, but maybe will just be on top of SDL3's camera API when that comes in. And having a super old draft PR open doesn't help at all.

PR #1930, opened 2023-02-23
Vector2 complex

gresm:
Migrating [#2997](https://github.com/pygame/pygame/pull/2997).

Original description:
> Allows to convert python "complex" object into pygame Vector2 and adds vector2.to_complex() to convert Vector2 into "complex" object.


docs/reST/ref/math.rst
```diff
@@ -455,11 +456,23 @@ Multiple coordinates can be set using slices or swizzling
...
-   
+
+   .. method:: to_complex
+
+      | :sl:`Converts vector to complex number.`
+      | :sg:`to_complex() -> complex`
+
+      Returns complex number where x is real and y is imaginary.
+
+      .. versionadded:: 2.1.3
```
oddbookworm, line 467: this should be changed to 2.1.4

docs/reST/ref/math.rst
```diff
@@ -67,6 +67,7 @@ Multiple coordinates can be set using slices or swizzling
    | :sg:`Vector2() -> Vector2(0, 0)`
    | :sg:`Vector2(int) -> Vector2`
    | :sg:`Vector2(float) -> Vector2`
+   | :sg:`Vector2(complex) -> Vector2`
```
oddbookworm, line 70: Maybe add a version changed note to specify that this particular construction was added in 2.1.4

oddbookworm (changes requested):
Overall, looks really good! I just have a couple of small edits to the docs that I'd like to see

docs/reST/ref/math.rst
gresm: Good point, it was an old pr and I forgot to change it.

gresm: Done!

docs/reST/ref/math.rst
gresm: Done!

oddbookworm (approved):
LGTM!

MyreMylar:
Just copying over my original example program that demonstrates complex number vector rotation by multiplication:

```
import math

import pygame
from pygame import Vector2
from math import sin, cos, radians

pygame.init()

window_surface = pygame.display.set_mode((800, 600))

angle = 90
start_vec = Vector2(0, -100)

# Two rotation options here: standard Vector2 rotate(), or complex conversion rotation by multiplication
# rotated_vec = start_vec.rotate(angle)

# first we create the complex number to perform the rotation and convert our start vector to complex
complex_rotation = complex(cos(radians(angle)),
                           sin(radians(angle)))
complex_start_vec = complex(start_vec)

# then to do the actual rotation is only a multiply, this could allow for a lot of cheap rotations
# at run time where the angle to rotate at is known ahead of time
rotated_vec = Vector2(complex_start_vec * complex_rotation)

centre_vec = Vector2(400, 300)

translated_start_vec = start_vec + centre_vec
translated_rotated_vec = rotated_vec + centre_vec

running = True
while running:

        if event.type == pygame.QUIT:
            running = False

    window_surface.fill((0, 0, 0))

    pygame.draw.line(window_surface, (255, 255, 255), centre_vec, translated_start_vec)
    pygame.draw.line(window_surface, (255, 0, 255), centre_vec, translated_rotated_vec)

    pygame.display.flip()
```

I think this would be the main potential use case - performing lots of rotations.

robertpfeiffer:
I still think the API should be something like `Vector2.from_complex(complex)` or `Vector2.complex_as_2d(complex)`, for two reasons: One is consistency with maths. The other is consistency with existing Vector2 behaviour. `Vector2(1)` is interpreted as $ \begin{pmatrix} 1 \\ 1 \end{pmatrix} $ and `1 == 1+0j`. Wanting `Vector2(1+0j)` to mean $ \begin{pmatrix} 1 \\ 0 \end{pmatrix} $ feels like a footgun.
At least complex numbers in Python are not automatically simplified to fractions or floats, but they are automatically promoted.

gresm:
@robertpfeiffer Indeed, this edge-case could be a confusing edge case,  but I'm not sure if it will be problematic: if a person decides to use complex number for vector initialization, they probably would already expect the current behavior. The real issue would be *i*f  Python automatically simplify** complex numbers with zero as imaginary. I also think adding ``Vector2.complex_as_2d(complex)`` could lead to class methods pollution, where it can be done in an already existing one (without overfilling ``Vector2.__init__()`` with features - too much is bad, but too little is also bad).

oddbookworm:
@gresm you should run `python setup.py docs --fullgeneration` to fix the lint check

docs/reST/ref/math.rst
```diff
@@ -77,6 +78,9 @@ Multiple coordinates can be set using slices or swizzling
       Inherited methods of vector subclasses now correctly return an instance of the 
       subclass instead of the superclass
 
+   .. versionchanged:: 2.1.4
```
MyreMylar, line 81:
```suggestion
   .. versionchanged:: 2.2.0
```

docs/reST/ref/math.rst
```diff
@@ -455,11 +459,23 @@ Multiple coordinates can be set using slices or swizzling
...
-   
+
+   .. method:: to_complex
+
+      | :sl:`Converts vector to complex number.`
+      | :sg:`to_complex() -> complex`
+
+      Returns complex number where x is real and y is imaginary.
+
+      .. versionadded:: 2.1.4
```
MyreMylar, line 470:
```suggestion
      .. versionadded:: 2.2.0
```

MyreMylar (commented):
Retargeting 2.1.4 -> 2.20

gresm:
@oddbookworm Should be fixed now...

Starbuck5:
I agree with robertpfieffer that Vector2 complex functionality should be through a classmethod rather than adding more work into the Vector2 constructor.

Starbuck5:
Thanks for spending the time and effort to put this forward gresm, it really is not easy to come in and make a PR on our C code.

It seems like there is no consensus on what the right API should be, or if this functionality should be added at all, and this has been stale for a long time, so I'm going to go ahead and close.

PR #1933, opened 2023-02-23
Rewrite docs of display.set_mode()

yunline:
### Easier to read
image


robertpfeiffer (changes requested):
To me, some of the changed language sounds a bit more awkward than before. Also I am currently working on vsync, so the docs will have to be changed then anyway.

The re-ordering of topics makes sense though.

yunline:
This PR is not good enough

PR #1940, opened 2023-02-25
Surface, Draw, and Transform segfault fixes

oddbookworm:
fixes a whole bunch o' segfaults related to surface, draw, and transform


oddbookworm:
This is a reopening of https://github.com/pygame/pygame/pull/3680 with additions from https://github.com/pygame/pygame/pull/3525

src_c/include/_pygame.h
```diff
@@ -480,6 +480,13 @@ PYGAMEAPI_EXTERN_SLOTS(math);
  *  functions in Python 3.
  */
 
+#define SURF_INIT_CHECK(SDL_SurfObj)                                    \
+    {                                                                   \
+        if (!SDL_SurfObj) {                                             \
```
ankith26, line 485: super minor nitpick: it's a bit confusing to see a macro variable to be prefixed with `SDL_`

Conventionally single-variable names are used in such contexts

oddbookworm: Adjusted

src_c/surface.c
```diff
@@ -1827,6 +1820,9 @@ surf_fill(pgSurfaceObject *self, PyObject *args, PyObject *keywds)
 static PyObject *
 surf_blit(pgSurfaceObject *self, PyObject *args, PyObject *keywds)
 {
+    SDL_Surface *surf = pgSurface_AsSurface(self);
+    SURF_INIT_CHECK(surf)
+
     SDL_Surface *src, *dest = pgSurface_AsSurface(self);
```
MyreMylar, line 1826: why are we getting `self` as an SDL_Surface twice here? Think we could just add the check on `src` or `dest` after the existing call to `pgSurface_AsSurface()` (since they are pointing to the same thing here).

src_c/surface.c
```diff
@@ -1843,8 +1839,9 @@ surf_blit(pgSurfaceObject *self, PyObject *args, PyObject *keywds)
         return NULL;
 
     src = pgSurface_AsSurface(srcobject);
-    if (!dest || !src)
-        return RAISE(pgExc_SDLError, "display Surface quit");
+    if (!dest) {
+        SURF_INIT_CHECK(src)
+    }
```
MyreMylar, line 1844: this check is different now. Previously it would error if either `src` **or** `dest` were null, now it will only error if both `src` **and** `dest` are null.

src_c/surface.c
```diff
@@ -1898,6 +1895,9 @@ surf_blit(pgSurfaceObject *self, PyObject *args, PyObject *keywds)
 static PyObject *
 surf_blits(pgSurfaceObject *self, PyObject *args, PyObject *keywds)
 {
+    SDL_Surface *surf = pgSurface_AsSurface(self);
+    SURF_INIT_CHECK(surf)
+
     SDL_Surface *src, *dest = pgSurface_AsSurface(self);
```
MyreMylar, line 1901: Similar issue as above. The check here could just be moved below the existing call to  `pgSurface_AsSurface()`

src_c/transform.c
```diff
@@ -951,6 +960,9 @@ surf_rotozoom(PyObject *self, PyObject *args, PyObject *kwargs)
                                      &scale))
         return NULL;
     surf = pgSurface_AsSurface(surfobj);
+    surf = pgSurface_AsSurface(surfobj);
```
MyreMylar, line 963:
```suggestion
```

MyreMylar (approved):
LGTM 👍

ankith26 (approved):

Issue #1941, opened 2023-02-25
Sprite module refactoring


MyreMylar:
This is a topic that has come up several times in the past, the sprite module has some parts that are outdated, some parts that are  confusing and a few areas that could be tweaked to improve performance and likely help prepare it for some future refactorings.

I opened a draft PR on this topic back on the upstream repository, but I concluded there that it would be better to break the changes down into a series of smaller and easier to review changes.

**GOALS**

- Boost sprite group performance in most common looping operations (principally draw).
- Add `special_flags` support to group draw operations.
- Deprecate & remove from docs redundant Group types (OrderedUpdates is no longer relevant from 3.7 onwards as all python dictionaries are insertion ordered, RenderPlain & RenderClear should also be deprecated as they are just crowding up the docs and source).
- Generally prepare for 'future' upgrades like FRect using sprites and Renderer Texture sharing Sprite Groups.

**Refactoring To Do list**

- [ ] Add `.image` and `.rect` to sprite as properties. I'm not sure why these were never attached to the basic sprite class despite being used in every sprite group. They will default to None. Adding them as properties allows us to keep the underlying data types hidden (potentially useful for FRect type usage later), and more importantly lets us attach sprite group operations to them when a new value is assigned to either `.rect` or `.image`, which will be important for performance refactors.
- [ ] switch out `.blits()` usage for `.fblits()` usage in sprite groups to enable easy support of `special_flags` parameter in draw.
- [ ] Use a 'drawlist' as the basic data type for drawing sprites in all sprite groups. right now only some of the groups store sprites in a list with others making use of a dictionary to do this. Lists are faster to iterate over.
- [ ] Assemble the drawlist only when changes are made to sprite group's membership or to sprite's `.image` or `.rect` properties rather than on every call to draw. This will just improve performance a bit in large groups.
- [ ] Remove redundant code from OrderedUpdates, add deprecation warnings to it and RenderPlain/RenderClear on creation. Remove all three from docs (? - this last part could wait until the deprecation expires, probably should have a deprecation policy so it is clear).

I think that is about it. If anyone has any further ideas along the lines of improving what is already here, or removing parts of it for improved clarity (rather than adding new features to the sprite module) then please add a comment.

itzpr3d4t0r:
> switch out .blits() usage for .fblits() usage in sprite groups to enable easy support of special_flags parameter in draw.

[...]

PR #1942, opened 2023-02-25
Add image & rect properties to Sprite

MyreMylar:
Hopefully, I didn't miss anything. Related to #1941 

I can't see any good reason not to have these as properties. Sprite has never been an actual abstract base class, and python has always allowed you to just add a `.image` and `.rect` onto instances of `Sprite` without having to make a derived class.

By adding these as properties we gain hooks we can use on assignments for other refactoring later to boost performance. 

I expect 90% of all sprite group using sprites derive from Sprite.


oddbookworm (approved):
LGTM!

itzpr3d4t0r (approved):
LGTM.

Starbuck5 (approved):
Seems like a good idea, thanks.

Issue #1932, opened 2023-02-23
SDL2 video ideas, revisited


Mega-JC:
This GH issue aims to both port and superseed https://github.com/pygame/pygame/issues/2691 , to help bring the `_sdl2.video` API to a stage that allows for straightforward porting to C, as proposed in #1929 , and a proper adoption in pygame.

None of these ideas are final, and discussion around them is encouraged.

- **Give `Texture` and `Image` `get_width()`, `get_height()`, and `get_size()` methods like surfaces**
    - _Reasoning_: Hard to document, potentially confusing
- **Remove readonly attributes from `Window`, `Texture` and `Renderer`**
- **Rename `Renderer.present()` to `.flip()` / `.update()`**
- **Create API for supporting multiple windows in `pygame.display`**
- **Change `Renderer.fill_rect()` to impersonate `Surface.fill()`**
- **Change `Renderer.draw_*` methods to take in a color directly**
    - _Reasoning_: The current “C like” set color flag then draw system is uncommon in pygame.
- **`pygame.draw` compatibility for `Texture` objects**
    - _Reasoning_: A unified system for drawing that works for both `Surface` and `Texture`. `SDL_RenderGeometry()` (SDL 2.0.18) might come in handy.
    - _Other ideas_:
        - A dedicated drawing module for `Textures`, using either `SDL_RenderGeometry()` or  `SDL2_gfx`, called e.g. `gpudraw` , given the complexity of `pygame.draw` and plans for merging `gfxdraw` into it
- **Eliminate `Image` objects, add `Texture.subtexture`**
- **Add compatibility for drawing `Surface` unto `Renderer`**
    - Reasoning: By e.g. making a streaming `Texture` the size of the `Renderer`, and putting surface data into there before drawing it onto the Renderer, we could save people the trouble of adding a conversion step for dynamic surfaces like Text rendered every frame, as some things are always going to be Surfaces.
- **`pygame.transform` compatibility for `Texture` objects**

robertpfeiffer:
We might be able to lock and memory-map a texture to a write-only PyGame Surface: https://wiki.libsdl.org/SDL2/SDL_LockTexture, but please understand the performance implications

yunline:
> * Remove readonly attributes from `Window`, `Texture` and `Renderer`

There are currently 2 readonly attributes in `Window`.
One is `display_index` (which i think is useful and doesn't need to be removed).

Another one is `id`.
[...]

PR #1945, opened 2023-02-26
Remove attributes in `_sdl2.video`, turn them into setters/getters

yunline:


ankith26:
Thanks for the PR!

Not sure whether this is the right way to proceed, sure it makes the API more consistent with the older `Surface` API, but personally I think attributes are a more "modern" way of doing things, and any future API we add should follow the newer trends.

Ideally this sort of PRs generally need an issue discussing this first with other fellow contributors (or was it done already elsewhere and I missed it? Please do correct me if I missed it)

yunline:
> Thanks for the PR!
> 
> Not sure whether this is the right way to proceed, sure it makes the API more consistent with the older `Surface` API, but personally I think attributes are a more "modern" way of doing things, and any future API we add should follow the newer trends.
> 
> Ideally this sort of PRs generally need an issue discussing this first with other fellow contributors (or was it done already elsewhere and I missed it? Please do correct me if I missed it)

Issue: https://github.com/pygame-community/pygame-ce/issues/1932
Discord: https://discord.com/channels/772505616680878080/772940667231928360/1079359592926040146

yunline:
> Thanks for the PR!
> 
> Not sure whether this is the right way to proceed, sure it makes the API more consistent with the older `Surface` API, but personally I think attributes are a more "modern" way of doing things, and any future API we add should follow the newer trends.
> 
> Ideally this sort of PRs generally need an issue discussing this first with other fellow contributors (or was it done already elsewhere and I missed it? Please do correct me if I missed it)

Issue: https://github.com/pygame-community/pygame-ce/issues/1932
Discord: https://discord.com/channels/772505616680878080/772940667231928360/1079359592926040146

buildconfig/stubs/pygame/_sdl2/video.pyi
```diff
@@ -51,27 +51,35 @@ class Window:
     ) -> None: ...
     @classmethod
     def from_display_module(cls) -> Window: ...
-    grab: bool
-    relative_mouse: bool
+    def get_geab(self) -> bool: ...
```
Mega-JC, line 54:
```suggestion
    def get_grab(self) -> bool: ...
```

Mega-JC:
This PR covers converting `@property` attributes into separate getters and setters, but ignores other attributes. This leads to some odd API differences. Ideally, `Image.alpha` should be `Image.get_alpha()` and `Image.set_alpha()`, the same also goes for `Image.blend_mode`.

PR #1948, opened 2023-02-27
Wasm static multiphase init for dynamic loading

pmp-p:
Support dynamic loading for cpython-wasm or pyodide :
`import pygame_static` from the wasm library stored in the wheel will trigger complete import of pygame-ce. 

Such a wheel has been tested with pygbag for months now.

NB: there's nothing provided in that PR to automate building of that wheel.

Vendor patching is required for camera module, audio capture, font listing and possibly various sensors potentially available on mobile with SDL2, but it is left up to implementation import system.

eg https://github.com/pygame-web/platform_wasm for pygbag (WIP).

NB: regarding pyodide it's bound to be used with PR https://github.com/pyodide/pyodide/pull/3508 from @ryanking13. Vendor patching may be implemented in pyodide build system though i think it would de-sync with pygame-ce releases.

 I think a centralized place for runtime patching is the best : since not much people know to do that either for python-wasm or pyodide yet, it's better they gather at the same place.

cc @rth


MyreMylar (approved):
LGTM 👍

ankith26 (approved):
Not an expert around these bits of code, but it looks okay on a glance.

Thanks for the PR :pinata:

PR #1956, opened 2023-02-28
Font.set direction

### LINKED ISSUE #1952: Support setting text direction
Port https://wiki.libsdl.org/SDL2_ttf/TTF_SetFontDirection

Languages like Arabic (right to left) need this to render properly. (Along with a call to Font.set_script as well)

Would be cool to have in the 2.1.4 update as part of a "text enhancements" story: multiline text, Font.set_script for internationalization... Font.set_direction for internalization as well.

See https://github.com/pygame/pygame/pull/3330 for inspiration.

oddbookworm:
Will close #1952. I used this script to test the functionality
```
import pygame

pygame.init()

screen_rect = screen.get_rect()

font = pygame.font.Font(None, 50)

font.set_direction(pygame.DIRECTION_LTR)
ltr = font.render("LTR TEXT", True, "red")
ltr_rect = ltr.get_rect(topleft=screen_rect.topleft)

font.set_direction(pygame.DIRECTION_RTL)
rtl = font.render("RTL TEXT", True, "red")
rtl_rect = rtl.get_rect(topright=screen_rect.topright)

font.set_direction(pygame.DIRECTION_TTB)
ttb = font.render("TTB TEXT", True, "red")
ttb_rect = ttb.get_rect(bottomleft=screen_rect.bottomleft)

font.set_direction(pygame.DIRECTION_BTT)
btt = font.render("BTT TEXT", True, "red")
btt_rect = btt.get_rect(bottomright=screen_rect.bottomright)

running = True
while running:
        if event.type == pygame.QUIT:
            running = False
            
    screen.fill("black")
    screen.blit(ltr, ltr_rect)
    screen.blit(rtl, rtl_rect)
    screen.blit(ttb, ttb_rect)
    screen.blit(btt, btt_rect)
    pygame.display.flip()
```
but the vertical font renders seem to be the wrong way?


docs/reST/ref/font.rst
```diff
@@ -492,6 +492,30 @@ solves no longer exists, it will likely be removed in the future.
 
       .. ## Font.set_script ## 
 
+   .. method:: set_direction
+
+      | :sl:`set the script direction for text shaping`
+      | :sg:`set_direction(direction=int) -> None`
+
+      **Experimental:** feature still in development available for testing and feedback. It may change.
+      `Please leave feedback with pygame-ce authors <https://github.com/pygame-community/pygame-ce/pull/1956>`_
```
Starbuck5, line 501: I believe we've decided to move away from this pattern, and we had a PR strip all the existing ones out.

src_c/font.c
```diff
@@ -667,6 +667,69 @@ font_set_script(PyObject *self, PyObject *arg)
...
+/* Can't use SDL_TTF_VERSION_ATLEAST until SDL_ttf 2.0.15 is minimum supported
+ */
+#if SDL_VERSIONNUM(SDL_TTF_MAJOR_VERSION, SDL_TTF_MINOR_VERSION, \
+                   SDL_TTF_PATCHLEVEL) >= SDL_VERSIONNUM(2, 20, 0)
+    TTF_Font *font = PyFont_AsFont(self);
+    int direction;
+    char *kwds[] = {"direction", NULL};
+
+    if (!PyArg_ParseTupleAndKeywords(arg, kwarg, "i", kwds, &direction)) {
+        return RAISE(PyExc_TypeError, "direction argument must be an int");
```
Starbuck5, line 682: This comes with the exception already set, so you don't need to specify your own error here.

Starbuck5:
I wonder if the vertical direction mismatch is because of SDL's coordinate system being +y as down-- maybe harfbuzz has -y as down and this was not adjusted in SDL_ttf.

pmp-p (approved):

docs/reST/ref/font.rst
oddbookworm: I was actually copying this structure from set_script above. If it needs to be removed, I'm fine removing it and then opening an issue to remove it from set_script

MyreMylar: Yes do this.

 I think the two PRs (adding set_script and removing experimental) missed each other like ships in the night.

MyreMylar:
I think this one should go in 2.1.4 if we are going for a text rendering improvements bonanza. it goes hand in hand with the `set_script()` changes to enable arabic text rendering.

MyreMylar (approved):
Other than the small issues raised above this looks good.

It is a bit of a shame about the bug in SDL with the up and down direction but I think the left to right/right to left support is the important thing to get in.

MyreMylar:
As discussed on discord: Probably best to patch the up & down text direction issue on our end for the current versions of SDL (and below) as it will always be like that and is now confirmed as an SDL bug. 

Please don't merge until that's sorted/

pmp-p:
i think we should break the build on SDL ttf <= 2.0.15 , 2.0.15 included to be safe.  TTF_SetFontScriptName did not work correctly on wasm with it so there's probably an internal bug only visible at runtime which will only rise with set_script().

Starbuck5:
> i think we should break the build on SDL ttf <= 2.0.15 , 2.0.15 included to be safe. TTF_SetFontScriptName did not work correctly on wasm with it so there's probably an internal bug to SDL2_ttf only visible at runtime which will only rise with set_script().

Not sure about this, but pmp-p asked me to merge this on discord so I believe he's okay with the changes right now.

PR #1837, opened 2023-02-13
Add "from_existing_window" function to the _sdl2.Window

yunline:
# Add "from_existing_window" function to the _sdl2.Window

*This PR is cloned from [https://github.com/pygame/pygame/pull/3685](https://github.com/pygame/pygame/pull/3685).*

**2023-02-06T21:59:19Z** [yunline](https://github.com/yunline) wrote:
> ```python
  > from_existing_window(hwnd: int) -> Window
  > ```
  > With this function, pygame window could be inserted into another window (Tk/Qt/MFC/...) .
  > Here is an example based on Tkinter：
  > 
  > ```python
  > import tkinter
  > import threading
  > 
  > root=tkinter.Tk()
  > root.geometry("640x480")
  > btn=tkinter.Button(root,text="HelloWorld\n你好世界")
  > btn.pack()
  > cv=tkinter.Canvas(root)
  > cv.pack()
  > 
  > off=0
  > def pg_loop(hwnd):
  >     import pygame
  >     from pygame._sdl2 import video
  > 
  >     pygame.display.init()
  > 
  >     win=video.Window.from_existing_window(hwnd)
  >     
  >     win.focus()         # focus/restore is available
  >     #win.grab=1         # grab is available
  >     #win.opacity=0.5    # opacity is available
  >     #win.hide()         # hide/show is availabe
  >     #win.resizable=1    # available but works weirdly
  > 
  >     ren=video.Renderer(win)
  > 
  >     pos=(20,20)
  >     while not off:
  >         ren.draw_color=(0,0,0,0)
  >         ren.clear()
  >         ren.draw_color=(255,0,0,0)
  >         ren.fill_rect((*pos,10,10))
  >         ren.present()
  > 
  >         for event in pygame.event.get():
  >             if event.type==1024:#MOUSEMOTION
  >                 pos=event.pos
  >             else:
  >                 print(event)
  > 
  > 
  > if __name__ == "__main__":
  >     #hwnd=int(root.frame(),16)  # get the handle of the root window
  >     hwnd=cv.winfo_id()          # get the handle of the canvas widget
  > 
  >     p=threading.Thread(target=pg_loop,args=(hwnd,))
  > 
  >     p.start()
  >     root.mainloop()
  >     off=1
  > ```
  > image
  >   

**2023-01-27T19:18:18Z** [MyreMylar](https://github.com/MyreMylar) wrote:
> Related Stale PR: https://github.com/pygame/pygame/pull/2981
  > 
  > To reference when reviewing  

[...]

Issue #1943, opened 2023-02-26
Support SDL_WINDOWID


Starbuck5:
Original issue: https://github.com/pygame/pygame/issues/1574

yunline:
Relevant PR: #1837

Starbuck5:
That is relevant, but just to be clear, I don't see that PR closing this issue.

Since upstream pygame merged a back compat fix for the old SDL_WINDOWID environment variable, we should as well, to minimize our compatibility differences.

PR #1957, opened 2023-02-28
Add `hwnd` parameter to display.set_mode()

yunline:
```python
display.set_mode( ... , hwnd :int = 0 ) # hwnd: The handle of the foreign window.
```
Embed a pygame window to a foreign window by setting the `hwnd` parameter to the handle of the target window. If this parameter is not set, the programe will check the environment variable `SDL_WINDOWID`. If `SDL_WINDOWID` is not set, a new window will be created.

**Test code:**
```python
import tkinter
import threading

root=tkinter.Tk()
root.geometry("640x480")
btn=tkinter.Button(root,text="HelloWorld")
btn.pack()
cv=tkinter.Canvas(root)
cv.pack()

off=0
def pg_loop(hwnd):
    import pygame

    sf=pygame.display.set_mode(hwnd=hwnd) 

    pos=(20,20)
    while not off:
        sf.fill((0,0,0))
        pygame.draw.rect(sf,(255,0,0),(*pos,10,10))
        pygame.display.update()

            if event.type==pygame.MOUSEMOTION:
                pos=event.pos

if __name__ == "__main__":
    hwnd=cv.winfo_id()
    threading.Thread(target=pg_loop,args=(hwnd,)).start()
    root.mainloop()
    off=1
```

**This feature is compatible with OpenGL**
Reference about OpenGL compatibility:
- https://github.com/pygame/pygame/pull/3685#issuecomment-1408199295
- https://wiki.libsdl.org/SDL2/SDL_CreateWindowFrom
- https://gamedev.stackexchange.com/a/119903
- https://github.com/pygame-community/pygame-ce/pull/1837#issuecomment-1434783345


oddbookworm:
This PR has the same problem as #1943, where keyboard events are not being detected properly and persistently

MyreMylar (commented):
First this PR has conflicts that need resolving.

I personally think that a keyword argument like this is an easier thing for users to work with than the old pygame hint style - however I'm also not sure about this entire approach to embedding.

It seems to me, especially now that we are having our own proper window class that the generally proper way to embed _aspects_ of pygame into a foreign window is by drawing/blitting everything to a pygame.Surface/Texture/Whatever SDL 3 uses and then drawing that surface into the foreign window via whatever methods the window manager has for drawing image data to it's windows. It would be nice to carry this neat separation between windows and drawing to a surface/texture onwards into embedding pygame drawing elsewhere.

Event loop code is always going to be messy when mixing and matching window management from different APIs as almost all window management APIs also include some form of input event handling.

I feel like a tutorial/conversion code effort on how to embed a pygame Surface/Texture into three popular window manager/GUI libraries (e.g. tkinter, PyQt5, Kivy) would be the optimal path possibly with some discussion of what event management should be left to the owning window manager (mouse/keyboard) and what could be done in pygame (probably controllers?). 

Anyway, that is my overall feeling - this is better than the old pygame way, but that way is backwards compatible so, ugh, we should keep it around but not advertise it. However the overall approach to embedding a window inside a window is weird and we would probably do better with directly converting a pygame Surface/texture to whatever the external window manager of choice uses.

I've only really seen this approach used with tkinter where it doesn't work very well, so its possible my opinion could be changed by smooth running in another window manager.

damusss:
I'm taking the liberty to close this pull request. It's been stagnating for three years, and we are now discouraging and moving away from using set_mode as a window creation tool entirely. I'm not sure whether the hwnd parameter should be allowed in the constructor at all, but even if it should, it should be something to worry about for the pygame.Window class.

PR #1961, opened 2023-03-01
Add `.point_size` attribute modifier for fonts

novialriptide:
Original PR: https://github.com/pygame/pygame/pull/3688

This pull request has been significantly improved since its original state.

## Video
vid

## Test code
```python
import pygame
import sys

pygame.init()
screen = pygame.display.set_mode((690, 690))

font = pygame.font.Font(None, 10)

while True:
    for x in pygame.event.get():
        if x.type == pygame.QUIT:
            sys.exit()
        if x.type == pygame.KEYUP:
            if x.key == pygame.K_UP:
                # font.set_pointsize(font.get_pointsize() + 10)
                font.pointsize += 10
            if x.key == pygame.K_DOWN:
                # font.set_pointsize(font.get_pointsize() - 10)
                font.pointsize -= 10

    screen.fill((0, 0, 0))
    screen.blit(font.render("among us is real", True, (255, 255, 255)), (250, 250))
    
    pygame.display.flip()
```


MyreMylar:
I'd like to put this in 2.1.4, because it is font related - if you can get it working.

buildconfig/stubs/pygame/font.pyi
```diff
@@ -60,5 +61,7 @@ class Font:
     def get_descent(self) -> int: ...
     def set_script(self, script_code: str) -> None: ...
     def set_direction(self, direction: int) -> None: ...
+    def get_pointsize(self) -> int: ...
+    def set_pointsize(self, val: int) -> None: ...
```
MyreMylar, line 65: My inclination here would be to follow the [zen of python maxim](https://peps.python.org/pep-0020/):

> There should be one-- and preferably only one --obvious way to do it.

and just have `pointsize` as a property.

However, I'd also rename that property `point_size` with an underscore to fit with [PEP8](https://peps.python.org/pep-0008/#method-names-and-instance-variables)

Pygame has historically been a little inconsistent with the naming of attributes and properties in line with PEP8, so there are examples of both ways in the code base.

docs/reST/ref/font.rst
```diff
@@ -260,6 +260,18 @@ solves no longer exists, it will likely be removed in the future.
 
       .. ## Font.align ##
 
+   .. attribute:: pointsize
+
+      | :sl:`Gets or sets the font's pointsize`
+      | :sg:`pointsize -> int`
+
+      The font's pointsize will not entirely be accurate upon
+      initialization.
```
MyreMylar, line 269: As far as I can tell this is only true if you are talking about the default font which does that weird:

```
fontsize = (int)(fontsize * .6875);
```
thing. In all other cases the initial font size will be reported correctly by the `.point_size` attribute. This should probably be made clear, as right now it sounds like the point size is never accurate unless you set it first with the attribute.

docs/reST/ref/font.rst
```diff
@@ -260,6 +260,18 @@ solves no longer exists, it will likely be removed in the future.
...
+
+      .. versionadded:: 2.2
+
+      .. ## Font.underline ##
```
MyreMylar, line 273:
```suggestion
      .. ## Font.point_size##
```

Or `Font.pointsize` if we go with that.

src_c/font.c
```diff
@@ -631,6 +631,82 @@ font_size(PyObject *self, PyObject *text)
...
+static int
+font_setter_pointsize(PyFontObject *self, PyObject *value, void *closure)
+{
+#if SDL_TTF_VERSION_ATLEAST(2, 0, 18)
+    TTF_Font *font = PyFont_AsFont(self);
+    int val = PyLong_AsLong(value);
+
+    if (val <= 0) {
+        PyErr_SetString(PyExc_ValueError,
+                        "pointsize cannot be equal or less than 0");
```
MyreMylar, line 655:
```suggestion
                        "point_size cannot be equal to, or less than 0");
```

src_c/font.c
```diff
@@ -631,6 +631,82 @@ font_size(PyObject *self, PyObject *text)
...
+static PyObject *
+font_set_ptsize(PyObject *self, PyObject *arg)
+{
+#if SDL_TTF_VERSION_ATLEAST(2, 0, 18)
+    TTF_Font *font = PyFont_AsFont(self);
+    int val = PyLong_AsLong(arg);
+
+    if (val <= 0) {
+        PyErr_SetString(PyExc_ValueError,
+                        "pointsize cannot be equal or less than 0");
```
MyreMylar, line 693:
```suggestion
                        "point_size cannot be equal to, or less than 0");
```

src_c/font.c
```diff
@@ -631,6 +631,82 @@ font_size(PyObject *self, PyObject *text)
...
+        return NULL;
+    }
+
+    if (TTF_SetFontSize(font, val) == -1) {
+        PyErr_SetString(pgExc_SDLError, SDL_GetError());
+    }
+    ((PyFontObject *)self)->ptsize = val;
```
MyreMylar, line 700: Should we still be setting the `ptsize` here if we just errored in the check above?

Perhaps encase this in an else statement block?

test/font_test.py
```diff
@@ -545,6 +545,42 @@ def test_size(self):
...
+    def test_pointsize_property(self):
+        if pygame_font.__name__ == "pygame.ftfont":
+            return  # not a pygame.ftfont feature
+
+        ttf_version = pygame_font.get_sdl_ttf_version()
+        if ttf_version < (2, 0, 18):
+            return
+
+        pygame_font.init()
+        f = pygame_font.Font(None, 25)
```
MyreMylar, line 557: You could load a font here to get test results from a non-default font instead of the silly 68% sized default font.

MyreMylar (commented):
This works well in my testing 👍 Nice work and thanks for fixing it up.

I have some style and philosophical(?) quibbles that you may, or may not, agree with so I'm leaving this review on comment for now.

Starbuck5:
I believe we should call this “size” to match the existing freetype attribute.

novialriptide:
> I believe we should call this “size” to match the existing freetype attribute.

@Starbuck5 and I were in call and realized that `font.size` is already taken.

docs/reST/ref/font.rst
```diff
@@ -473,6 +485,25 @@ solves no longer exists, it will likely be removed in the future.
...
+
+      .. ## Font.set_point_size ##
+
+   .. method:: get_point_size
+
+      | :sl:`get the point_size of the font`
+      | :sg:`get_point_size() -> int`
+
+      Returns the point_size of the font. Will not be accurate upon
+      initializing the font object.
```
Starbuck5, line 503: Why will it not be accurate on init?

MyreMylar: this is only true for the default font and because of that stupid resize by 0.6 thing.

src_py/ftfont.py
```diff
@@ -155,6 +155,14 @@ def size(self, text):
 
         return self.get_rect(text).size
 
+    def get_point_size(self):
+        """"""
+        return self.get_point_size()
```
dr0id, line 160: Maybe I don't understand this , but isn't that method calling itself?

dr0id: Maybe you wanted to call the method of the base class, e.g. _Font.get_point_size(self) or super().get_point_size() ?

docs/reST/ref/font.rst
dr0id: I would rephrase to point out the special case of using the default font (passing `None`).

src_py/ftfont.py
novialriptide: I'm not entirely sure what this means, but the errors went away when I inserted that code :shrug:. If you look at the file there are similar stuff as well.

src_c/font.c
novialriptide: I don't think we should set the ptsize, since `TTF_SetFontSize` returns `-1` when unsuccessful.

dr0id: why isn't there a `return NULL;` or something after the pyErr_SetString(...) call on line 699?

docs/reST/ref/font.rst
Starbuck5: @novialriptide This is not resolved, there's still a location in the new docs that says this.

> Returns the point_size of the font. Will not be accurate upon initializing the font object when the font name is initalized as ``None``

The phrasing seems a bit weird to me. Also initalized = typo.

Do we even need to mention this special case explicitly in the docs? Or maybe it's a note in the Font constructor docs instead?

docs/reST/ref/font.rst
```diff
@@ -275,6 +275,18 @@ solves no longer exists, it will likely be removed in the future.
...
 
+   .. attribute:: point_size
+
+      | :sl:`Gets or sets the font's point_size`
+      | :sg:`point_size -> int`
+
+      Returns the point_size of the font. Will not be accurate upon initializing
+      the font object when the font name is initalized as None
+
+      .. versionadded:: 2.2
```
zoldalma999, line 286: Version number needs update

docs/reST/ref/font.rst
```diff
@@ -490,6 +502,26 @@ solves no longer exists, it will likely be removed in the future.
...
+
+   .. method:: get_point_size
+
+      | :sl:`get the point_size of the font`
+      | :sg:`get_point_size() -> int`
+
+      Returns the point_size of the font. Will not be accurate upon
+      initializing the font object when the font name is initalized
+      as ``None``
+
```
zoldalma999, line 522: Should probably have a versionadded tag

docs/reST/ref/font.rst
```diff
@@ -490,6 +502,26 @@ solves no longer exists, it will likely be removed in the future.
...
       .. ## Font.get_height ##
 
+   .. method:: set_point_size
+
+      | :sl:`set the point_size of the font`
+      | :sg:`set_point_size(val) -> int`
+
+      Sets the point_size of the font.
+
+      .. ## Font.set_point_size ##
```
zoldalma999, line 512: Should probably have a versionadded tag

src_c/font.c
```diff
@@ -636,6 +636,83 @@ font_size(PyObject *self, PyObject *text)
...
+    int val = PyLong_AsLong(value);
+
+    if (val <= 0) {
+        PyErr_SetString(PyExc_ValueError,
+                        "point_size cannot be equal to, or less than 0");
+        return -1;
+    }
+
+    if (TTF_SetFontSize(font, val) == -1) {
+        PyErr_SetString(pgExc_SDLError, SDL_GetError());
```
zoldalma999, line 665: This should return -1 here

dr0id (approved):
LGTM after you have solved @zoldalma999  comments

src_c/include/pygame_font.h
```diff
@@ -28,6 +28,7 @@ struct TTF_Font;
 typedef struct {
     PyObject_HEAD TTF_Font *font;
     PyObject *weakreflist;
+    unsigned int ptsize;
```
Starbuck5, line 31: Why is this unsigned, when `ptsize` is always handled as signed in the methods?

src_py/ftfont.py
```diff
@@ -155,6 +155,14 @@ def size(self, text):
...
+
+    def set_point_size(self, val):
+        """"""
+        return self.set_point_size(val)
+
```
Starbuck5, line 165: I agree with this Dr0id, I don't think this is right.

Your test cases specifically exit instead of testing ftfont compatibility. If you took that out you'd get some sort of error. Recursion I think.

Ftfont is the Font API on top of Freetype. So this should be calling into the freetype equivalent of this API: https://pyga.me/docs/ref/freetype.html#pygame.freetype.Font.size

docs/reST/ref/font.rst
```diff
@@ -490,6 +502,30 @@ solves no longer exists, it will likely be removed in the future.
 
       .. ## Font.get_height ##
 
+   .. method:: set_point_size
+
+      | :sl:`set the point_size of the font`
+      | :sg:`set_point_size(val) -> int`
+
+      Sets the point_size of the font.
```
Starbuck5, line 510: What is a point size in this context?

docs/reST/ref/font.rst
```diff
@@ -490,6 +502,30 @@ solves no longer exists, it will likely be removed in the future.
...
+      .. ## Font.set_point_size ##
+
+   .. method:: get_point_size
+
+      | :sl:`get the point_size of the font`
+      | :sg:`get_point_size() -> int`
+
+      Returns the point_size of the font. Will not be accurate upon
+      initializing the font object when the font name is initalized
+      as ``None``
```
Starbuck5, line 523: I still do not like this `font None` caveat line.

buildconfig/stubs/pygame/font.pyi
```diff
@@ -34,6 +34,7 @@ class Font:
     underline: bool
     strikethrough: bool
     align: int
+    point_size: int
```
Starbuck5, line 37: So this is late for this feedback, but why do an attribute and a getter/setter?

It seems more like a computed value than an "intrinsic" value, so I'd say just have a getter/setter. The implementation and the docs of TTF_SetFontSize mention clearing the cache of glyphs, so it's more of an operation than should behind an attribute imo.

MyreMylar: A property seems like the right thing here to me. Since the SDL C functions it is wrapping are getters and setters a single property called `.point_size` makes the most sense to me. Replacing getter and setter functions is the main reason the property concept was introduced.

I would drop the `,get_point_size()` and `.set_point_size()` methods before removing the property.

docs/reST/ref/font.rst
MyreMylar: Could we not make this value accurate (i.e. do the same multiplication on it in the constuctor where font name is `None`)? Leave only the initial value entered with font as `None` as 'inaccurate', and instead document that historical weirdness in the same place we mention the default font being used when passing in `None`?

buildconfig/stubs/pygame/font.pyi
Starbuck5: The caching is in freetype, and every time this function is called the cache is flushed.

On second thought I'm okay with it as just a property. But all the other ones are getsets and properties, so it seems best to just go with that. (Which is what this PR is already doing).

docs/reST/ref/font.rst
novialriptide: Sorry for getting on this pull request 2 weeks later...

Shouldn't the `point_size` be figured out by the user? As in like googling the answer.

novialriptide:
Whoops, I've been in hiatus for a while. I've pushed some commits resolving all reviews.

src_c/font.c
```diff
@@ -636,6 +636,84 @@ font_size(PyObject *self, PyObject *text)
     return Py_BuildValue("(ii)", w, h);
 }
 
+static PyObject *
+font_getter_point_size(PyFontObject *self, void *closure)
+{
+#if SDL_TTF_VERSION_ATLEAST(2, 0, 18)
+    return PyLong_FromLong(self->ptsize);
+#else
+    PyErr_SetString(PyExc_SystemError,
```
Starbuck5, line 645: SystemError?

src_c/font.c
```diff
@@ -636,6 +636,84 @@ font_size(PyObject *self, PyObject *text)
...
+    return NULL;
+#endif
+}
+
+static int
+font_setter_point_size(PyFontObject *self, PyObject *value, void *closure)
+{
+#if SDL_TTF_VERSION_ATLEAST(2, 0, 18)
+    TTF_Font *font = PyFont_AsFont(self);
+    int val = PyLong_AsLong(value);
```
Starbuck5, line 656: Wait what if value passed isn't an int?

src_c/font.c
```diff
@@ -636,6 +636,84 @@ font_size(PyObject *self, PyObject *text)
...
+
+    if (TTF_SetFontSize(font, val) == -1) {
+        PyErr_SetString(pgExc_SDLError, SDL_GetError());
+        return -1;
+    }
+    self->ptsize = val;
+
+    return 0;
+#else
+    PyErr_SetString(PyExc_SystemError,
```
Starbuck5, line 672: SystemError?

src_c/font.c
```diff
@@ -636,6 +636,84 @@ font_size(PyObject *self, PyObject *text)
...
+#endif
+}
+
+static PyObject *
+font_get_ptsize(PyObject *self, PyObject *args)
+{
+#if SDL_TTF_VERSION_ATLEAST(2, 0, 18)
+    return PyLong_FromLong(((PyFontObject *)self)->ptsize);
+#else
+    PyErr_SetString(PyExc_SystemError,
```
Starbuck5, line 684: SystemError again...
How did I not see these before.

https://docs.python.org/3/library/exceptions.html#SystemError
^ Raised on interpreter exceptions

We should raise an SDLError here in all these spots.

src_c/font.c
```diff
@@ -636,6 +636,84 @@ font_size(PyObject *self, PyObject *text)
...
+
+    if (TTF_SetFontSize(font, val) == -1) {
+        PyErr_SetString(pgExc_SDLError, SDL_GetError());
+        return NULL;
+    }
+    ((PyFontObject *)self)->ptsize = val;
+
+    Py_RETURN_NONE;
+#else
+    PyErr_SetString(PyExc_SystemError,
```
Starbuck5, line 711: SystemError

docs/reST/ref/font.rst
Starbuck5: Then you could end up with a situation where you could get_point_size() and then set_point_size() with that same size and the font would suddenly be bigger.

Starbuck5: I still don't like the line, it's just informative enough to start to tell people what's up without actually giving them enough information.

I think it would be better with less information (don't mention it because people won't notice the edge case) or more information (Here's what's up, yada yada yada)

Starbuck5 (approved):
Thanks!

I've noticed this raises a deprecation warning if you give it a float, and the SDL_TTF_VERSION_ATLEAST could be relaxed around getters. But that can be fixed in the future.

PR #1967, opened 2023-03-01
wasm freetype 2 init and some cleanup

pmp-p:
seems ok enough to run pygame_gui on browser.

### DIFF
--- a/src_c/_freetype.c
+++ b/src_c/_freetype.c
@@ -30,9 +30,17 @@
 #define MODULE_NAME "_freetype"
 #define FONT_TYPE_NAME "Font"
 
+#define DEFAULT_FONT_NAME "freesans"
+#define DEFAULT_FONT_FILE "freesansbold.ttf"
+#define PKGDATA_MODULE_NAME "pygame.pkgdata"
+#define RESOURCE_FUNC_NAME "getResource"
+
 /*
  * FreeType module declarations
  */
+
+PyObject *_freetypemodule = NULL;
+
 static const Scale_t FACE_SIZE_NONE = {0, 0};
 
 static int
@@ -207,10 +215,6 @@ free_string(PGFT_String *);
                                                                \
         _var = PyObject_IsTrue(_pyobj);                        \
     }
-
-#define DEFAULT_FONT_NAME "freesansbold.ttf"
-#define PKGDATA_MODULE_NAME "pygame.pkgdata"
-#define RESOURCE_FUNC_NAME "getResource"
 
 static PyObject *
 load_font_res(const char *filename)
@@ -687,7 +691,11 @@ _ftfont_init(pgFontObject *self, PyObject *args, PyObject *kwds)
     SDL_RWops *source;
 
     FreeTypeInstance *ft;
+#if defined(__EMSCRIPTEN__)
+    ft = FREETYPE_MOD_STATE(_freetypemodule)->freetype;
+#else
     ASSERT_GRAB_FREETYPE(ft, -1);
+#endif
 
     if (!PyArg_ParseTupleAndKeywords(args, kwds, "O|O&lIi", kwlist, &file,
                                      obj_to_scale, (void *)&face_size,
@@ -721,7 +729,7 @@ _ftfont_init(pgFontObject *self, PyObject *args, PyObject *kwds)
         self->resolution = FREETYPE_STATE->resolution;
     }
     if (file == Py_None) {
-        file = load_font_res(DEFAULT_FONT_NAME);
+        file = load_font_res(DEFAULT_FONT_FILE);
 
         if (!file) {
             PyErr_SetString(PyExc_RuntimeError, "Failed to find default font");
@@ -2035,6 +2043,14 @@ _ft_get_error(PyObject *self, PyObject *_null)
     Py_RETURN_NONE;
 }
 
+#if defined(__EMSCRIPTEN__)
+static PyObject *
+_ft_get_version(PyObject *self, PyObject *args, PyObject *kwargs)
+{
+    return Py_BuildValue("iii", FREETYPE_MAJOR, FREETYPE_MINOR,
+                         FREETYPE_PATCH);
+}
+#else
 static PyObject *
 _ft_get_version(PyObject *self, PyObject *args, PyObject *kwargs)
 {
@@ -2081,6 +2097,7 @@ _ft_get_version(PyObject *self, PyObject *args, PyObject *kwargs)
                              FREETYPE_PATCH);
     }
 }
+#endif  // defined(__EMSCRIPTEN__)
 
 static PyObject *
 _ft_get_cache_size(PyObject *self, PyObject *_null)
@@ -2120,7 +2137,7 @@ _ft_get_init(PyObject *self, PyObject *_null)
 static PyObject *
 _ft_get_default_font(PyObject *self, PyObject *_null)
 {
-    return PyUnicode_FromString(DEFAULT_FONT_NAME);
+    return PyUnicode_FromString(DEFAULT_FONT_FILE);
 }
 
 static int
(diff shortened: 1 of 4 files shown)


src_py/sysfont.py
```diff
@@ -201,14 +201,42 @@ def initsysfonts_darwin():
...
 
-# read the fonts on unix
+# read the fonts on posix/unix
 def initsysfonts_unix(path="fc-list"):
-    """use the fc-list from fontconfig to get a list of fonts"""
+    """if not embedded, use the fc-list from fontconfig to get a list of fonts"""
     fonts = {}
 
-    if sys.platform == "emscripten":
+    print("208: initsysfonts_unix {path=}")
```
MyreMylar, line 209: should this print still be in here?

MyreMylar (approved):
LGTM 👍

src_py/sysfont.py
pmp-p: nope, thx

oddbookworm (approved):
LGTM, but I'm not 100% comfortable enough with my understanding of the freetype module code to merge it, so I'm hoping that at least one other person could take a look

Starbuck5:
Why do you need to initialize the freetype library instance in module initialization rather than in pygame.freetype.init?

pmp-p:
@Starbuck5 the static init is a bit different and now keeps a global reference so that other subfunctions can reach the pointer. If module init was done like before the ref would be unreachable as soon as leaving the scope of module init

Starbuck5:
I went ahead and pushed a fix for the build failures, since it was caused by a PR of mine getting merged.

> @Starbuck5 the static init is a bit different and now keeps a global reference so that other subfunctions can reach the pointer. If module init was done like before the ref would be unreachable as soon as leaving the scope of module init

I don't understand. It looks like you use the freetype module state to store two others things perfectly fine (cache_size and resolution), why can't you store a freetype library pointer there and have it get initialized in `_ft_autoinit`?

My motivation for caring is that the less special case code for emscripten is needed throughout pygame, the easier things will be to maintain in the future.

pmp-p:
iirc _ft_init() was not able to initialize properly on wasm, and i did not investigate much but i'll have another look. Though as said earlier I'm not sure of usefullness of that module as I think using os fonts is doomed.

pmp-p:
i probably used a bad test the init looks ok, tested with pygame_gui

src_c/_freetype.c
```diff
@@ -2201,27 +2218,34 @@ MODINIT_DEFINE(_freetype)
...
-    FREETYPE_MOD_STATE(module)->freetype = 0;
-    FREETYPE_MOD_STATE(module)->cache_size = 0;
-    FREETYPE_MOD_STATE(module)->resolution = PGFT_DEFAULT_RESOLUTION;
+#if defined(__EMSCRIPTEN__)
+//_PGFT_Init(&(FREETYPE_MOD_STATE(_freetypemodule)->freetype), 0);
+#pragma message "NO _PGFT_Init"
+    FREETYPE_MOD_STATE(_freetypemodule)->freetype = 0;
+#else
+    FREETYPE_MOD_STATE(_freetypemodule)->freetype = 0;
+#endif
```
Starbuck5, line 2233: Now that EMSCRIPTEN and non-emscripten have the same code path, why leave the `#if / #else / #endif` ?

src_c/_freetype.c
```diff
@@ -2040,6 +2048,14 @@ _ft_get_error(PyObject *self, PyObject *_null)
...
 }
 
+#if defined(__EMSCRIPTEN__)
+static PyObject *
+_ft_get_version(PyObject *self, PyObject *args, PyObject *kwargs)
+{
+    return Py_BuildValue("iii", FREETYPE_MAJOR, FREETYPE_MINOR,
+                         FREETYPE_PATCH);
+}
+#else
```
Starbuck5, line 2058: Is there a reason the other `_ft_get_version` doesn't work?

src_c/_freetype.c
pmp-p: oops bad edit/commit

src_c/_freetype.c
pmp-p: no it's just for reducing code size

Starbuck5: I'm surprised you're looking that hard for size optimizations that you'd take out part of a method like this. Does pygame-wasm have freetype compiled into the build such that the compile time and runtime will always be the same?

Since this is for size optimization rather than functionality, there should be a comment explaining that in the source code.

src_c/_freetype.c
```diff
@@ -692,7 +696,11 @@ _ftfont_init(pgFontObject *self, PyObject *args, PyObject *kwds)
     SDL_RWops *source;
 
     FreeTypeInstance *ft;
+#if defined(__EMSCRIPTEN__)
+    ft = FREETYPE_MOD_STATE(_freetypemodule)->freetype;
+#else
     ASSERT_GRAB_FREETYPE(ft, -1);
+#endif
```
Starbuck5, line 703: Is this change needed? It seems like ASSERT_GRAB_FREETYPE does the same thing but also NULL checks `->freetype` for making sure the module is initialized.

src_c/_freetype.c
pmp-p: > compile time and runtime will always be the same?

yes sdl2, sdl2_ttf, freetype and harfbuzz are all linked statically in the wheel but i did not mean size as optimization but more smaller code == less complexity. I'll test if both linked False/True work - which i did not - hence the fencing.

src_c/_freetype.c
pmp-p: > Is this change needed?

Unsure, i'll have to test it. It's possible the init failure was only linked to the python code afterall ( or some rwops in pygame_gui which i found meanwhile ).

src_c/_freetype.c
```diff
@@ -30,9 +30,17 @@
 #define MODULE_NAME "_freetype"
 #define FONT_TYPE_NAME "Font"
 
+#define DEFAULT_FONT_NAME "freesans"
```
dr0id, line 33: not sure, where is `src_py/sysfont.py` used? It seems unused.

src_c/_freetype.c
```diff
@@ -2125,7 +2142,7 @@ _ft_get_init(PyObject *self, PyObject *_null)
 static PyObject *
 _ft_get_default_font(PyObject *self, PyObject *_null)
 {
-    return PyUnicode_FromString(DEFAULT_FONT_NAME);
+    return PyUnicode_FromString(DEFAULT_FONT_FILE);
```
dr0id, line 2145: This does change the return value to the filename as until now it was the font name? 
E.g. it was `freesans` and now it is `freesansbold.ttf`.

src_c/_freetype.c
pmp-p: i think maybe it should be used as a constant to generate the default mapping generated in the python module.
{ DEFAULT_FONT_NAME : DEFAULT_FONT_NAME }
which i have hardcoded.

src_c/_freetype.c
dr0id: you are right, found a test:
```
    def test_get_default_font(self):
        self.assertEqual(pygame_font.get_default_font(), "freesansbold.ttf")

```
this is resolved for me

src_c/_freetype.c
dr0id: if you look at how `_ftfont_getname` is implemented: it will get the name from the font object of the library and pass it to the `font.name` attribute. So not sure why to keep `DEFAULT_FONT_NAME` at all.

src_py/sysfont.py
```diff
@@ -201,14 +201,40 @@ def initsysfonts_darwin():
...
-    if sys.platform == "emscripten":
+    # these are embedded and cannot get os to list fonts a simple way.
+    if hasattr(sys, "getandroidapilevel") or sys.platform == "emscripten":
+        from pathlib import Path
+
+        # default font
+        import importlib.resources
+        import pygame
+
+        entry = importlib.resources.files(pygame) / "freesansbold.ttf"
```
dr0id, line 217: Would this be the place to use the constant from the c code?

dr0id (approved):
Ideally the string constants would appear once in all the source. But I know it might not be practical.

Starbuck5:
So @pmp-p I still have reservations about the two places EMSCRIPTEN is used in the freetype C code, I have an unsubstantiated hunch that the code would work fine with the non-emscripten code paths.

If you just want to get this PR merged without testing that, I would be fine if there were comments in the code explaining that. Like "it's not that getting the linked version doesn't work on emscripten, it's just that it hasn't been tested." I just don't want some future maintainer to see those EMSCRIPTEN blocks and assume that the non-emscripten code definitely won't work on web for some reason.

pmp-p:
@Starbuck5 you are right about the blocks, but i kinda hoped to target wasi just by pretending being in emscripten as i have a wasi SDL2 port with sixel/framebuffer output as a side project.

also i need more time for testing i'm quite busy with real life atm.

pmp-p:
it seems that PR alone is not sufficient to init SDL_ttf correctly on wasm ( when dynamic / side module ), static has no problem. But the goal is to make a wheel


it seems TTF_Init() is not called at all. I  had to add it directly in PyInit_pygame_static() of src_c/static.c to get it effective ...

it's a blocker for https://github.com/pyodide/pyodide/issues/289

Issue #1969, opened 2023-03-02
Add get_name() method to the different Font classes 


Martinus2004:
**Description**

ReaMart proposed this in the discord: 

>  I made a little program that lets you browse the list of sysfonts. I have a list of the Font objects, which I loop trough and blit to the screen with some text.I figured it would be a nice feature if the text that was rendered would include the name of the font, so I tried doing something like font.get_name, but there was no such thing so yeah.

Not only for sysfont this would be handy in my opinion. Let's say a user wanted to get the name of a certain font in some other file or deep down in the code. It would be handy to get that name using the method rather than having to look it up yourself. Also if you have a list of Font objects, this would come in handy (like ReaMart had).

MyreMylar:
I agree that this would be nice. The `freetype` submodule implements something similar with the `.name` attribute for it's `Font` class. However, that is a read only attribute which I know @Starbuck5 doesn't like.

PR #1971, opened 2023-03-03
Add font name getter

novialriptide:
Fixes: https://github.com/pygame-community/pygame-ce/issues/1969

I dunno if we want `get_name()`, it seems "unpythonic" in my opinion. Should we discuss deprecating the `get_x()` in a future pr? We could just remove them from the docs to act like they never existed.

This article/post could be a good reference for deprecating getter/setter methods.
https://discuss.python.org/t/use-instance-attributes-or-class-properties-inside-the-class-definition/13290/3


Matiiss:
Sure, getters/setters that are a `get_` and a complementary `set_` method, could be replaced w/ property getters/setters, however, just getters might as well stay as `get_` methods, since they are not to be changed as in this case with this name, it's read-only, but read-only attributes can cause confusion.

src_c/doc/font_doc.h
```diff
@@ -10,6 +10,7 @@
 #define DOC_PYGAMEFONTSYSFONT "SysFont(name, size, bold=False, italic=False) -> Font\ncreate a Font object from the system fonts"
 #define DOC_PYGAMEFONTFONT "Font(filename, size) -> Font\nFont(pathlib.Path, size) -> Font\nFont(object, size) -> Font\ncreate a new Font object from a file"
 #define DOC_FONTBOLD "bold -> bool\nGets or sets whether the font should be rendered in (faked) bold."
+#define DOC_FONTNAME "name -> str\nGets the font's name."
```
MyreMylar, line 13:
```suggestion
#define DOC_PYGAMEFONTNAME "name -> str\nGets the font's name."
```
Annoyingly the freetype module already stole the `DOC_FONTNAME` macro name

src_c/font.c
```diff
@@ -671,6 +680,7 @@ font_set_script(PyObject *self, PyObject *arg)
  * Getters and setters for the pgFontObject.
  */
 static PyGetSetDef font_getsets[] = {
+    {"name", (getter)font_getter_name, NULL, DOC_FONTNAME, NULL},
```
MyreMylar, line 683:
```suggestion
    {"name", (getter)font_getter_name, NULL, DOC_PYGAMEFONTNAME, NULL},
```

MyreMylar (approved):
LGTM 👍

src_c/doc/font_doc.h
Starbuck5: This file is autogenerated, so I'm not sure how we'd get around that.

Starbuck5: Figured it out: you need to add it to the `#undef` lines in static.c

Starbuck5 (approved):
This looks cool, thanks!

docs/reST/ref/font.rst
```diff
@@ -198,6 +198,17 @@ solves no longer exists, it will likely be removed in the future.
 
       .. ## Font.bold ##
 
+   .. attribute:: name
+
+      | :sl:`Gets the font's name.`
+      | :sg:`name -> str`
+
+      Returns the font's name.
```
Starbuck5, line 206: I already approved, but this is a small fix if you'd like.

Looking at the pygame.freetype docs https://pyga.me/docs/ref/freetype.html#pygame.freetype.Font.name, they mark their version "Read only" in the content. That would be a good thing to add here.

They also call it the "proper font name," which imo has a nice ring to it.

test/font_test.py
```diff
@@ -548,6 +548,10 @@ def test_size(self):
 
         self.assertNotEqual(size, bsize)
 
+    def test_font_name(self):
+        f = pygame_font.Font(None, 20)
+        self.assertEqual(f.name, "FreeSans")
+
```
ankith26, line 554: Another suggestion, for improving tests. We should test that trying to set this attr fails with the expected error

novialriptide:
Once all tests have passed, I'll merge this.

PR #1972, opened 2023-03-03
Fix rect drawing issues around surface edges

### LINKED ISSUE #1968: Only 4 pixels show when drawing a rect on the edge of the screen
**Environment:**
- Windows
- Python 3.11.2
- SDL 2.0.22
- pygame-ce 2.1.3

**Current behavior:**
When drawing a (rounded) rect outside of the screen on the left side that sticks out by one pixel to the right, only the upper and bottom pixels are drawn.
**Expected behavior:**
It should draw a 1 pixel wide vertical black line on the left edge of the screen, instead of only the upper and bottom parts of the expected line.
**Screenshots**
image

**Steps to reproduce:**
1. run the example
2. press the left arrow key until you see "this is where it goes wrong" on screen
3. observe the black pixels on the left side of the screen

**Test code**
```
import pygame
import sys
mainClock = pygame.time.Clock()
pygame.init()
pygame.display.set_caption('testing, use arrow keys to move')
screen_width = 1280
screen_height = 720
WINDOW_SIZE = (900,600)
screen = pygame.display.set_mode(WINDOW_SIZE,0,32)
running = True
display = pygame.Surface((300,200)) # defines the small surf, that will be scaled to fit the screen
def check_for_quit():
        if event.type == pygame.QUIT:
            pygame.quit()
            sys.exit()
            if event.key == pygame.K_ESCAPE:
                pygame.quit()
                sys.exit()

inventory_background_rect = pygame.rect.Rect(0,0,50,WINDOW_SIZE[1]/3) # defines the rect
inventory_background_rect_x = 0
font = pygame.font.SysFont("arialblack",10)
while running:
    check_for_quit()
    display.fill("white")
    #movement
    if pygame.key.get_pressed()[pygame.K_LEFT]:
        inventory_background_rect_x -= 0.1
    if pygame.key.get_pressed()[pygame.K_RIGHT]:
        inventory_background_rect_x += 0.1

    if int(inventory_background_rect_x) == -49:
        display.blit(font.render("this is where it goes wrong",0,"black"),(40,20))

    inventory_background_rect.x = int(inventory_background_rect_x) # sets the x of the rec to the x variable
    pygame.draw.rect(display,"black",inventory_background_rect,2,0,0,20,0,20) # draws the rect
    screen.blit(pygame.transform.scale(display,WINDOW_SIZE),(0,0)) # fits the small dispay Surf to the screen
    pygame.display.update()
    mainClock.tick(60)
```

Temmie3754:
This resolves #1968 and a similar issue encountered where the drawn rect would be stretched to the edge of the screen when in close proximity.

### DIFF
--- a/src_c/draw.c
+++ b/src_c/draw.c
@@ -1064,73 +1064,43 @@ static int
 clip_line(SDL_Surface *surf, int *x1, int *y1, int *x2, int *y2, int width,
           int xinc)
 {
-    int p1 = *x1 - *x2;
-    int p2 = -p1;
-    int p3 = *y1 - *y2;
-    int p4 = -p3;
-    int q1 = *x1 - surf->clip_rect.x;
-    int q2 = surf->clip_rect.w + surf->clip_rect.x - *x1;
-    int q3 = *y1 - surf->clip_rect.y;
-    int q4 = surf->clip_rect.h + surf->clip_rect.y - *y1;
-    int old_x1 = *x1;
-    int old_y1 = *y1;
-    double nmax = 0;
-    double pmin = 1;
-    double r1, r2;
+    int left, right, top, bottom;
     if (xinc) {
-        q1 += width;
-        q2 -= width;
+        left = MIN(*x1, *x2) - width;
+        right = MAX(*x1, *x2) + width;
+        top = MIN(*y1, *y2);
+        bottom = MAX(*y1, *y2);
     }
     else {
-        q3 += width;
-        q4 -= width;
+        left = MIN(*x1, *x2);
+        right = MAX(*x1, *x2);
+        top = MIN(*y1, *y2) - width;
+        bottom = MAX(*y1, *y2) + width;
     }
-    if ((p1 == 0 && q1 < 0) || (p2 == 0 && q2 < 0) || (p3 == 0 && q3 < 0) ||
-        (p4 == 0 && q4 < 0))
+    if (surf->clip_rect.x > right || surf->clip_rect.y > bottom ||
+        surf->clip_rect.x + surf->clip_rect.w < left ||
+        surf->clip_rect.y + surf->clip_rect.h < top) {
         return 0;
-    if (p1) {
-        r1 = (double)q1 / p1;
-        r2 = (double)q2 / p2;
-        if (p1 < 0) {
-            if (r1 > nmax)
-                nmax = r1;
-            if (r2 < pmin)
-                pmin = r2;
+    }
+    if (*y1 == *y2) {
+        if (*x2 < *x1) {
+            *x2 = MAX(surf->clip_rect.x, *x2);
+            *x1 = MIN(surf->clip_rect.x + surf->clip_rect.w, *x1);
         }
         else {
-            if (r2 > nmax)
-                nmax = r2;
-            if (r1 < pmin)
-                pmin = r1;
+            *x1 = MAX(surf->clip_rect.x, *x1);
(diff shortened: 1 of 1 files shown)


src_c/draw.c
```diff
@@ -1488,7 +1500,10 @@ draw_line_width(SDL_Surface *surf, Uint32 color, int x1, int y1, int x2,
     dy = abs(y2 - y1);
     sy = y1 < y2 ? 1 : -1;
     err = (dx > dy ? dx : -dy) / 2;
-    if (clip_line(surf, &x1, &y1, &x2, &y2)) {
+    clip = clip_line(surf, &x1, &y1, &x2, &y2, width, xinc) ||
+           clip_line(surf, &x1 + width, &y1, &x2, &y2, -width, xinc) ||
+           clip_line(surf, &x1, &y1, &x2, &y2, 0, 0);
+    if (clip) {
```
MyreMylar, line 1506: I'm not sure I understand why we have three calls to clip line here. The purpose of clip line appears to be to clip a line to a surface, if it is not doing it right I think that should probably be fixed inside `clip_line` not by calling it three times with different parameters.

MyreMylar (changes requested):
Good work on fixing the reported issue on the left hand side of the clip surface, and generally improving the drawn rect's surface edge clipping.

Unfortunately there still remain some issues with clipping wide lines I can still see disappearing line chunks on the right hand side/bottom of surfaces.

I adapted this test program to demonstrate:

```
import pygame


WIDTH, HEIGHT = 800, 600
FPS = 60

pygame.init()
clock = pygame.time.Clock()

test_surf = pygame.Surface((100, 100))
test_surf.fill((255, 255, 255))

rect = pygame.Rect(0, 0, 100, 100)

test_surf_pos = (150, 150)

running = True
while running:
    clock.tick(FPS)
    screen.fill("black")

    events = pygame.event.get()
    for event in events:
        if event.type == pygame.QUIT:
            running = False

    test_surf.fill("white")
    rect.center = pygame.mouse.get_pos()[0] - test_surf_pos[0], pygame.mouse.get_pos()[1] - test_surf_pos[0]
    pygame.draw.rect(test_surf, "red", rect, 30, 0, 0, 20, 0, 20)
    screen.blit(test_surf, test_surf_pos)

    pygame.display.flip()

```

Just try moving the D shaped rect with the mouse towards the bottom right of the white square surface.

see:

image

and:

image

I'm also a little unsure what is happening with the multiple calls to clip line. I understood the purpose of clip line to be 'clipping a line to be drawn to the surface passed in', I can see why the function needs the line width parameter to determine when to clip or not with wide lines, but not why it has to be called three times.

Perhaps we need some diagrams here to help clarify things?

src_c/draw.c
Temmie3754: It's mostly a workaround for the clip_line function to check if the edges of the line intersect with the surface as well to determine whether it is within bounds (which in itself could have some issues if all three lines as it were were not within the surface), I'll see about rewriting the clip_line entirely to remove the excess number of calls.

Temmie3754:
Alright I have made those changes, this method checks if the bounding box of the line intersects with the surface to decide if it draws it. It will also trim the excess off in cases where the line is vertical or horizontal but does not do so otherwise. I could look into implementing a more rigorous check to see where the line intersects along the surface and calculate where to cut it off based on the width for non orthogonal cases but I worry that would introduce too much overhead into the check where it won't be needed since the line won't be drawn outside the surface anyway.

Temmie3754:
Closing since #1998 supersedes this PR.

Issue #1970, opened 2023-03-02
Print a warning if the font is not found.


AndreyVarvar:
Currently, if you type a non-existing font on your machine into the `pygame.font.SysFont()` function, the pygame automatically swaps it to a default font with no warning, and the user has to orient on the font itself to find that he misspelled it. My suggestion is to print a warning, if the font typed in is non-existing. Something like `Warning: your font was not found, so it got replaced with a default font.` should be fine.

Sjmarf:
I think I'll try to implement this. I assume `logging.warning` is the proper way to give warnings to the user?

Sjmarf:
And what should the error message say? `The system font '{name}' couldn't be found. Using the default font instead.` could work?

AndreyVarvar:
Yes, I think this is good.

Sjmarf:
#1973

PR #1973, opened 2023-03-03
`font.SysFont` - Warn user if no font was found

Sjmarf:
See issue #1970 

When a non-existent font is passed to `pygame.font.SysFont`, Pygame uses a default font instead and doesn't raise an error. I've added a warning (not an error) that is shown when the user does this.
`The system font 'non-existent font' couldn't be found. Using the default font instead.`
`None of the specified system fonts (non-existent font, non-existent font2) could be found. Using the default font instead.`

Let me know if you would prefer a different error message, or if there's anything else you'd like me to change :)

Here's some test code:

```py
import pygame

pygame.font.init()

# Shows the following warning:
# The system font 'non-existent font' couldn't be found. Using the default font instead.
pygame.font.SysFont("non-existent font",40)
pygame.font.SysFont(bytes("non-existent font",'utf-8'),40)

# Shows the following warning:
# None of the specified system fonts (non-existent font, non-existent font2) could be found. Using the default font instead.
pygame.font.SysFont(("non-existent font","non-existent font2"),40)
pygame.font.SysFont((bytes("non-existent font",'utf-8'),"non-existent font2"),40)

# No warning is shown for any of the following:
pygame.font.SysFont("arial",40)
pygame.font.SysFont(("arial","comic sans"),40)
pygame.font.SysFont(("non-existent font", "arial"),40)
pygame.font.SysFont(bytes("arial",'utf-8'),40)
pygame.font.SysFont((bytes("non-existent font",'utf-8'), 'arial'),40)
pygame.font.SysFont(None,40)
```

### DIFF
--- a/src_py/sysfont.py
+++ b/src_py/sysfont.py
@@ -18,7 +18,7 @@
 # Pete Shinners
 # pete@shinners.org
 """sysfont, used in the font module to find system fonts"""
-
+import logging
 import os
 import sys
 import warnings
@@ -451,6 +451,23 @@ def SysFont(name, size, bold=False, italic=False, constructor=None):
                     gotitalic = italic
             if fontname:
                 break
+
+        else:
+            if len(name) > 1:
+                name = list(name)
+                for i in range(len(name)):
+                    if isinstance(name[i], bytes):
+                        name[i] = name[i].decode()
+                logging.warning(
+                    f"None of the specified system fonts ({', '.join(name)}) "
+                    f"could be found. Using the default font instead."
+                )
+            else:
+                single_name = name[0].decode() if type(name[0]) is bytes else name[0]
+                logging.warning(
+                    f"The system font '{single_name}' couldn't be "
+                    f"found. Using the default font instead."
+                )
 
     set_bold = set_italic = False
     if bold and not gotbold:
--- a/src_py/sysfont.py
+++ b/src_py/sysfont.py
@@ -22,7 +22,6 @@
 import warnings
 import os
 import sys
-import warnings
 from os.path import basename, dirname, exists, join, splitext
 
 from pygame.font import Font
--- a/src_py/sysfont.py
+++ b/src_py/sysfont.py
@@ -459,16 +459,19 @@ def SysFont(name, size, bold=False, italic=False, constructor=None):
                 for idx, single_name in enumerate(name):
                     if isinstance(single_name, bytes):
                         name[idx] = single_name.decode()
+
+                names = "', '".join(name)
                 warnings.warn(
-                    f"None of the specified system fonts ({', '.join(name)}) "
-                    "could be found. Using the default font instead."
+                    f"None of the specified system fonts "
+                    f"('{names}') could be found. "
+                    f"Using the default font instead."
                 )
             else:
                 single_name = (
                     name[0].decode() if isinstance(name[0], bytes) else name[0]
                 )
                 warnings.warn(
-                    f"The system font {single_name} couldn't be "
+                    f"The system font '{single_name}' couldn't be "
                     "found. Using the default font instead."
                 )
 
--- a/src_py/sysfont.py
+++ b/src_py/sysfont.py
@@ -18,7 +18,8 @@
 # Pete Shinners
 # pete@shinners.org
 """sysfont, used in the font module to find system fonts"""
-import logging
+
+import warnings
 import os
 import sys
 import warnings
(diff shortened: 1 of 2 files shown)


Matiiss:
I feel like this should be implemented in the existing loop, if a font is not found, you just add it to a list and if it never `break`s out of the loop and enters the `else` block, then you just already have a list of fonts that you can add to the warning.

Sjmarf:
I'm working on some unit tests right now. 

Would your suggested change be faster though? We've already got a list of fonts, and if we fail to find _any_ of those fonts, we can just use that same list in the warning. Granted, I have to loop through the list again to convert `bytes` objects to `str`. But that extra loop only runs if none of the available fonts load - which only occurs in a small portion of cases.

Python-ZZY:
font.SysFont is not a good idea.
i think font.Font will be better because you dont have to worry the font wasnt found.
It supports a font file path as an argument and will throw an error if the font is not found

Sjmarf:
In the current implementation, `font.SysFont` uses a default font if the font isn’t found. Are you suggesting that we should make `font.SysFont` raise an error if the font isn’t found, instead of falling back to a default? That would cause back-compat issues, though…

Sjmarf:
@Matiiss I seem to remember you saying that Pygame should still warn the user if they pass a tuple of font-names, but some of the fonts couldn't be loaded. Currently, it only warns the user if _none_ of the fonts could be loaded. But now I can't see that comment anymore? Is that still the plan, or no? Or am I just remembering wrong? 

If not, then I don't see how your suggested change would be an improvement? We've already got a list of fonts, and if we fail to find any of those fonts, we can just use that same list in the warning. Granted, I have to loop through the list again to convert `bytes` objects to `str`. But that extra loop only runs if none of the available fonts load - which only occurs in a small portion of cases.

Matiiss:
> @Matiiss I seem to remember you saying that Pygame should still warn the user if they pass a tuple of font-names, but some of the fonts couldn't be loaded. Currently, it only warns the user if _none_ of the fonts could be loaded. But now I can't see that comment anymore? Is that still the plan, or no? Or am I just remembering wrong?
> 
> If not, then I don't see how your suggested change would be an improvement? We've already got a list of fonts, and if we fail to find any of those fonts, we can just use that same list in the warning. Granted, I have to loop through the list again to convert `bytes` objects to `str`. But that extra loop only runs if none of the available fonts load - which only occurs in a small portion of cases.

Yeah, I'm not sure what I was thinking yesterday... this should be fine I guess

Sjmarf:
Ah, no worries! Happens to us all sometimes 😄

Sjmarf:
I think I'm happy with my code now 👍

src_py/sysfont.py
```diff
@@ -452,6 +452,26 @@ def SysFont(name, size, bold=False, italic=False, constructor=None):
...
 
+        else:
+            if len(name) > 1:
+                name = list(name)
+
+                for idx, single_name in enumerate(name):
+                    if isinstance(single_name, bytes):
+                        name[idx] = single_name.decode()
+                warnings.warn(
+                    f"None of the specified system fonts ({', '.join(name)}) "
```
Matiiss, line 463: Probably would be great if the font names could be also separated from the rest of the text by single quotes like so
```py
f"... fonts ('{'\', \''.join(name)}') "
```
so that it looks a bit neater and is a bit clearer

src_py/sysfont.py
```diff
@@ -452,6 +452,26 @@ def SysFont(name, size, bold=False, italic=False, constructor=None):
...
+                    "could be found. Using the default font instead."
+                )
+            else:
+                single_name = (
+                    name[0].decode() if isinstance(name[0], bytes) else name[0]
+                )
+                warnings.warn(
+                    f"The system font {single_name} couldn't be "
```
Matiiss, line 471: Same here, but simpler I guess
```py
"... font '{single_name}' ..."
```

Matiiss (changes requested):

src_py/sysfont.py
Sjmarf: Done 👍

Matiiss (approved):
LGTM, works as I'd expect 👍

src_py/sysfont.py
```diff
@@ -452,6 +452,29 @@ def SysFont(name, size, bold=False, italic=False, constructor=None):
             if fontname:
                 break
 
+        else:
```
Starbuck5, line 455: So it looks like this block has different name parsing logic than the block above.

Perhaps the whole thing should be refactored so it parses into a list of string font names once, then looks through that list to try to find the font, then raises the warning with the names it has.

Sjmarf: Sure 👍

src_py/sysfont.py
```diff
@@ -452,6 +455,20 @@ def SysFont(name, size, bold=False, italic=False, constructor=None):
...
+            if len(name) > 1:
+                names = "', '".join(name)
+                warnings.warn(
```
dr0id, line 461: I wonder if the warning should be documented in the docs.

dr0id (commented):
Should freetype also emit a warning if a font is not found?

Sjmarf:
> Should freetype also emit a warning if a font is not found?

My understanding was that the `freetype` SysFont just calls the `font` SysFont anyway - so I don't need to change anything over there. Correct me if I'm wrong, though.

MyreMylar (approved):
LGTM 👍

PR #1984, opened 2023-03-05
deprecated existing scrap API and replaced it with simpler API

oddbookworm:
Reopening of #3 because my original fork was borked and I had to create a fresh fork. Rather than fighting to get the old PR relinked to the new fork, I'm just reopening this PR


MyreMylar (commented):
All the 2.1.4s now need changing to 2.2.0 s

MyreMylar (approved):
LGTM still 👍

docs/reST/ref/scrap.rst
```diff
@@ -235,6 +260,57 @@ For an example of how the scrap module works refer to the examples page
...
+   .. deprecated:: 2.2.0
+
    .. ## pygame.scrap.set_mode ##
 
-.. ## pygame.scrap ##
+.. function:: put_text
+
+   | :sl:`Places text into the clipboard.`
+   | :sg:`put_text(text) -> None`
+
```
robertpfeiffer, line 271: Maybe we should drive the point home that this uses unicode strings, not binary data like the legacy API

docs/reST/ref/scrap.rst
```diff
@@ -235,6 +260,57 @@ For an example of how the scrap module works refer to the examples page
...
+   .. ## pygame.scrap.put_text
+
+.. function:: get_text
+   
+   | :sl:`Gets text from the clipboard.`
+   | :sg:`get_text() -> str`
+
+   Gets text from the clipboard and returns it. If the clipboard is empty,
+   returns an empty string. This is a different clipboard than the rest of 
+   the scrap module, except for ``put_text`` and ``has_text``.
```
robertpfeiffer, line 294: I'd call "the rest of the module" the "legacy API", and mention that unlike the legacy API, this is guaranteed to work cross platform.

src_c/scrap.c
```diff
@@ -93,6 +100,11 @@ _scrap_init(PyObject *self, PyObject *args)
 {
     VIDEO_INIT_CHECK();
 
+    if (PyErr_WarnEx(PyExc_DeprecationWarning,
+                     "pygame.scrap.init deprecated since 2.1.4", 1) == -1) {
```
Starbuck5, line 104: 2.1.4s in this file need to be changed.

docs/reST/ref/scrap.rst
Starbuck5: I think it's okay like this. People expect a string to mean a "real unicode string" in Python 3.

docs/reST/ref/scrap.rst
Starbuck5: It will work on the platforms that SDL supports it on, at least.

Starbuck5 (approved):
Besides that note about the version numbers, this looks great to me. Thanks!

src_c/scrap.c
oddbookworm: fixed

docs/reST/ref/scrap.rst
oddbookworm: This is also wrong. I have pushed a commit fixing this

PR #1985, opened 2023-03-05
add *venv* to gitignore

oddbookworm:
I use the `venv` pattern for my virtual environments, and I often name them a bit differently to distinguish between python versions. So, I would like to add this wildcard match to the gitignore so that it matches anything like `venv`, `3.10venv`, `venv3.10`, etc

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -35,6 +35,7 @@
 # Other
 envdev*
 .virtualenv*
+*venv*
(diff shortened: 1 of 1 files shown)


Starbuck5 (approved):
Sure

ankith26 (approved):
:tada:

Issue #1987, opened 2023-03-06
Incorrect version check in font.c


Starbuck5:
In `font_set_direction`

```c
/*  There is a bug in SDL2 up to 2.26.3 (the current release version as of
   writing this) This bug flips the top-to-bottom and bottom-to-top rendering.
   So, this is a compat patch for that behavior
 */
#if SDL_VERSIONNUM(SDL_TTF_MAJOR_VERSION, SDL_TTF_MINOR_VERSION, \
                   SDL_TTF_PATCHLEVEL) <= SDL_VERSIONNUM(2, 26, 3)
```

This is not a bug in SDL, it's a bug in SDL_ttf, and therefore the check should be based on SDL_ttf versions. Probably just check if it's below 2.22.0 (the next minor SDL_ttf version).

This is an oversight I didn't spot in https://github.com/pygame-community/pygame-ce/pull/1956, a recent PR.

oddbookworm:
I think [simplyrohan](https://github.com/simplyrohan) wants to handle this one

PR #1991, opened 2023-03-06
Change font.c to compare with SDL_ttf instead of SDL itself

simplyrohan:
Fixes: #1987
Compares SDL_ttf version to 2.22.0, as requested by @Starbuck5.

### DIFF
--- a/src_c/font.c
+++ b/src_c/font.c
@@ -779,7 +779,7 @@ font_set_direction(PyObject *self, PyObject *arg, PyObject *kwarg)
             break;
         }
 
-/*  There is a bug in SDL2 up to 2.26.3 (the current release version as of
+/*  There is a bug in SDL_ttf up to 2.22.0 (the current release version as of
    writing this) This bug flips the top-to-bottom and bottom-to-top rendering.
    So, this is a compat patch for that behavior
  */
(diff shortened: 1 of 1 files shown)


src_c/font.c
```diff
@@ -784,7 +784,7 @@ font_set_direction(PyObject *self, PyObject *arg, PyObject *kwarg)
    So, this is a compat patch for that behavior
```
oddbookworm, line 784: You should probably edit the comment just above to reflect the correction as well

oddbookworm (changes requested):
The comment above the code should also be edited to be consistent with the change

src_c/font.c
simplyrohan: Done

src_c/font.c
```diff
@@ -779,12 +779,12 @@ font_set_direction(PyObject *self, PyObject *arg, PyObject *kwarg)
             break;
         }
 
-/*  There is a bug in SDL2 up to 2.26.3 (the current release version as of
+/*  There is a bug in SDL_ttf up to 2.22.0 (the current release version as of
```
Starbuck5, line 782: 2.22.0 is the next release version as of writing.

simplyrohan: I think I fixed this with my latest commit...

oddbookworm: No, the comment says that 2.22.0 is the current release, not the next minor release

simplyrohan: Alright, hope this wraps it up

Starbuck5 (approved):
Thanks!

oddbookworm (approved):
LGTM! Thanks for fixing my mistake 😅

PR #1993, opened 2023-03-06
Rework doc_ file generation

Starbuck5:
In #1971, it came up that the doc_ header writing system sometimes generates duplicate names, which is a problem in the static build for webassembly. That can be handled by a few strategic undefines in places, but it got me into the header build system.

Changes
- delete the "docs in a comment" lines, I don't think they're necessary
- change the doc_ generation scheme to take a the fully qualified object name, ditch the first bit "pygame" (everything's in pygame), and separate everything with underscores. I think this is a better scheme than before-- simpler to generate, I'm not sure if it's possible to get a conflict, and it reads like an object name.
- adds module docstrings to scrap and pixelarray, fixes freetype.Font.descender having ascender docstring.

For example:
DOC_CLOCKGETRAWTIME -> DOC_TIME_CLOCK_GETRAWTIME
DOC_PYGAMEIMAGEFROMBYTES -> DOC_IMAGE_FROMBYTES 

Was this worth the effort it took me? No. But it's done now, so I might as well put a PR up.


MyreMylar (approved):
LGTM 👍 

Probably some PRS in progress will need to adjust to these new scheme so I'd like to get this merged earlier.

zoldalma999 (approved):
Looks good to me!

PR #1994, opened 2023-03-06
Fix color swizzling as per request in #1989

### LINKED ISSUE #1989: Color swizzling behavior
Color swizzling was just merged in #1935. My apologies for not getting to reviewing it and expressing my concern while the PR was open.

I think the swizzling behavior as currently implemented in a confusing way. 

In my opinion 
`pygame.Color(10, 100, 200).bgr` should yield `(200, 100, 10)`,  not `<Color(200, 100, 10, 255)>`
`pygame.Color(10, 100, 200, 25).bag` should yield `(200, 25, 100)`, not `<Color(200, 25, 100, 255)>`

I see that vectors have some runtime behavior like this, but honestly it might be better for Colors if they just returned n-length tuples with their swizzled attributes. Swizzling attributes shouldn't yield a sequence of a different length in my opinion, and the auto-filled alpha when there is explicit alpha is given is strange.

If someone wants to permute a color's values into another color object, they can accomplish that explicitly with a run through the constructor.
```py
color = pygame.Color(10, 100, 200, 25)
pygame.Color(*color.bgra)
```

avaxar:
Closes #1989.

Color swizzling with 3 attributes will now return a `tuple` instead of a 4-element-sized `pygame.Color`, with the alpha channel being inferred as opaque (`255`).

### DIFF
--- a/docs/reST/ref/color.rst
+++ b/docs/reST/ref/color.rst
@@ -24,9 +24,9 @@
 
    Color objects support swizzling for their ``RGBA`` attributes, which allows
    the creation of new color objects with the corresponding swizzled attributes
-   as its ``RGBA`` attributes. For example, ``color.bgr`` provides a shortcut to
-   doing ``Color(color.b, color.g, color.r)``. Swizzling with 2 or more than 4
-   attributes will return a tuple consisting of the corresponding elements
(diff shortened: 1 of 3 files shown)


ankith26 (approved):
thanks for the PR :tada:

PR #1998, opened 2023-03-07
Fix issues and inconsistency with draw_line_width 

Temmie3754:
This change resolves issues #1968 and https://github.com/pygame/pygame/issues/3199 through a rewrite of the draw_line_width function. Drawing lines from a to b should now also be the same as drawing from b to a.

### DIFF
--- a/src_c/draw.c
+++ b/src_c/draw.c
@@ -1481,7 +1481,7 @@ static void
 draw_line_width(SDL_Surface *surf, Uint32 color, int x1, int y1, int x2,
                 int y2, int width, int *drawn_area)
 {
-    int dx, dy, err, e2, sx, sy, y;
+    int dx, dy, err, e2, sx, sy, y, start_numx, start_numy, end_numx, end_numy;
     int start_x = surf->clip_rect.x;
     int start_y = surf->clip_rect.y;
     int end_x = surf->clip_rect.x + surf->clip_rect.w;
@@ -1506,6 +1506,8 @@ draw_line_width(SDL_Surface *surf, Uint32 color, int x1, int y1, int x2,
         xinc = 1;
         start_x -= width;
         end_x += width;
+        start_numy = start_y;
+        end_numy = end_y;
     }
     else {
         start_y -= width;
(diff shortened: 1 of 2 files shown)


src_c/draw.c
```diff
@@ -1469,112 +1458,88 @@ drawhorzlineclipbounding(SDL_Surface *surf, Uint32 color, int x1, int y1,
...
-        if (width == 1)
-            draw_line(surf, x1, y1, x2, y2, color, drawn_area);
-        else {
-            if (xinc) {
-                left_top = x1 - (width - 1) / 2;
-                right_bottom = x1 + width / 2;
+    if (xinc) {
+        while (y1 != (y2 + sy)) {
+            if (surf->clip_rect.y <= y1 && y1 <= end_y) {
+                for (int x =
```
MightyJosip, line 1507: This for loop should be replaced with `drawhorzline` and `add_line_to_drawn_list` for performance reasons.

MightyJosip (changes requested):

MightyJosip (approved):
It is good now. If you would like to do it, can you also add the tests for this

MyreMylar (approved):
LGTM 👍 

Fixes the original issue that was visible in this test program:

```
import pygame


WIDTH, HEIGHT = 800, 600
FPS = 60

pygame.init()
clock = pygame.time.Clock()

test_surf = pygame.Surface((100, 100))
test_surf.fill((255, 255, 255))

rect = pygame.Rect(0, 0, 100, 100)

test_surf_pos = (150, 150)

running = True
while running:
    clock.tick(FPS)
    screen.fill("black")

    events = pygame.event.get()
    for event in events:
        if event.type == pygame.QUIT:
            running = False

    test_surf.fill("white")
    rect.center = pygame.mouse.get_pos()[0] - test_surf_pos[0], pygame.mouse.get_pos()[1] - test_surf_pos[0]
    pygame.draw.rect(test_surf, "red", rect, 30, 0, 0, 20, 0, 20)
    screen.blit(test_surf, test_surf_pos)

    pygame.display.flip()

```

Try moving the D shape around inside the white surface with the mouse. On main you get lots of artefacts when parts of the D are clipped off screen and parts, which should still be visible get culled. After this PR everything looks as it should.

itzpr3d4t0r:
I tested this and confirm to be faster with this PR:

```
OLD:
0.3472604100001263
0.34475235999998405
0.3701231299995925
0.33968151999979457
0.3420160299998315
0.34054873000004593
0.3426739000000453
0.3402624100001049
0.35369226999973763
0.34050359000011665

NEW
0.23466548999986117
0.2332005099999151
0.23741479000018445
0.23511996999986878
0.23276994999960152
0.23450463999997737
0.2321502300004795
0.23337097000003268
0.23315945000049396
0.2355904100000771
```
Code:
```Python
from random import randint

import pygame
from pygame.draw import line as draw_line
from timeit import repeat
from statistics import mean

pygame.init()
WIDTH, HEIGHT = 400, 400


def rand_color():
    return randint(0, 255), randint(0, 255), randint(0, 255)


def create_lines(number, width):
    return [(rand_color(),
             (randint(0, WIDTH), randint(0, HEIGHT)),
             (randint(0, WIDTH), randint(0, HEIGHT)), width)
            for _ in range(number)]


lst = create_lines(100, 3)

for _ in range(10):
    print(mean(repeat("for line in lst: draw_line(win, *line)", globals=globals(), number=1000, repeat=10)))
```

Issue #1990, opened 2023-03-06
Add support for keyword arguments through modules


Starbuck5:
> I think we should be using keyword arguments on any function with more than two parameters that isn't designed to be a 'call me every frame' type of function (e.g. blitting). Probably we discussed this before - but there, I've said it again now.

Posted by Myremylar in https://github.com/pygame-community/pygame-ce/pull/1898#issuecomment-1448589814

That seems like a good policy to me.

Starbuck5:
Places I've identified to start off
- [ ] Color init
- [ ] font.render
- [ ] image.load  / loadext ? (only two parameters, but it would be weird to do this for image.save but not image.load?)
- [ ] image.save / saveext
- [ ] image.to(string|bytes)

PR #1999, opened 2023-03-08
Add kwargs to pygame.Color init

PurityLake:
Refers to #1990 

I have added a way to parse the kwargs `r` `g` `b` and `a` for `pygame.Color`


PurityLake:
Main issue I forsee with my implementation is that this is technically valid:

```python
pygame.Color(r=(10, 20, 30, 40))
```

Is this something that should be avoided?

novialriptide:
I just want to drop this here for now.

## Test Code

```python
from timeit import default_timer as timer
import pygame

start = timer()
calls = 10000000
for x in range(calls):
    color = pygame.Color(69,69,69,69)
elapsed_time = timer() - start
print(f"{calls} calls took {elapsed_time} seconds")
```

## Results
A bit of a performance loss, not significant though.
```
# pygame-ce 2.1.4 (SDL 2.0.22, Python 3.11.0)
10000000 calls took 1.374039599999378 seconds

# this pr
10000000 calls took 1.430627199999435 seconds
```

novialriptide:
> Is this something that should be avoided?

@PurityLake 

I think this an implementation should be added to prevent this to avoid bad code, good catch.
You could use `PyTuple_Check()` for each object that's been parsed to check if the inputted value is a tuple or not.

MyreMylar:
I'm not 100% sure that we should be adding keyword arguments to `Color`.

I know it technically has four arguments down one path (over my two argument baseline for conversion), but:

1. Each of the four (assuming we are doing the r=,g=,b=,a= path to Color creation) is kind of the same argument, in that they accept the same range of values (0 to 255) and only the alpha is optional. There is limited scope for user confusion here as long as you are familiar with the widely used RGB/RGBA colour ordering of parameters.
2. Colour also accepts single arguments as you've pointed out above which works OK in the pass by value paradigm but starts to become confusing when the keyword for a html hex colour string is also `r`.
3. I suspect there is currently a reasonable amount of colour object creation in once per frame game loops. I'm thinking of `surface.fill()` calls mostly but probably also `draw.thing()` calls. I doubt the increase in object creation time will be the thing that significantly slows down someone's game loop but creating core common objects like `Color` (and `Rect` and `Surface`) are going to be frequently called so we should be extra careful with alterations to them that downgrade performance.

However - I'm also prepared to be convinced the other way as well, if anyone has any other thoughts.

PurityLake:
@MyreMylar the idea of using keyword args makes sense to me. It can be handy but with trying to remain backwards compatible while using kwargs, creating `Color` objects with these added checks isn't feasible unless we go down the route of caching values somehow which can have it's own problem.

It might be best to close this PR to preserve performance.

PurityLake:
Closing this PR due to performance issues introduced

PR #2000, opened 2023-03-08
Added keyword arguments to `Font.render`

PurityLake:
Refers to #1990 

Added keyword argument options for the arguments in `Font.render`


PurityLake:
Oddly enough, despite the docs saying `wraplength` is an arg to `Font.render` it was never passed on to the C implementation.

Also the C implementation didn't receive the `antialias` argument.

Starbuck5:
That’s not what ftfont is.
Ftfont is “font” on the freetype API.

PurityLake:
Ah apologies, will rectify that mistake

docs/reST/ref/font.rst
```diff
@@ -321,6 +321,9 @@ solves no longer exists, it will likely be removed in the future.
 
       .. versionadded:: 2.1.4 wraplength parameter
 
+      .. versionchanged:: 2.1.4 all arguments can be pass by keyword arguments 
+         of the respective names.
```
Starbuck5, line 325: version changed = 2.2.0

"all arguments can be passed by keyword arguments  of the respective names" is more grammatically correct.

Viewing kwargs notices in other places, it could be something like "2.2.0 now supports keyword arguments"

src_c/font.c
```diff
@@ -486,8 +486,12 @@ font_render(PyObject *self, PyObject *args)
     const char *astring = "";
     int wraplength = 0;
 
-    if (!PyArg_ParseTuple(args, "OpO|Oi", &text, &antialias, &fg_rgba_obj,
-                          &bg_rgba_obj, &wraplength)) {
+    static char *kwlist[] = {"text",       "antialias",  "color",
+                             "background", "wraplength", NULL};
```
Starbuck5, line 490: If we're ever going to change these, now would be the time.

Fygame.freetype has `fgcolor` and `bgcolor` instead of `color` and `background` (As kwargs already).

Opinions?

PurityLake: fgcolor/bgcolor would make more sense, yes. I can make that change and adjust the docs and other places the parameter names are mentioned

MyreMylar (approved):
LGTM 👍

test/font_test.py
```diff
@@ -245,6 +245,26 @@ def test_render_args(self):
         self.assertEqual(tuple(screen.get_at((0, 0)))[:3], (255, 255, 255))
         self.assertEqual(tuple(screen.get_at(font_rect.topleft))[:3], (255, 255, 255))
 
+        # this test should only be run with FontTest not FtFontTest
```
Starbuck5, line 248: Why? Ftfont should be changed so that you use the keywords with the same names.

PurityLake: Fair point

docs/reST/ref/font.rst
```diff
@@ -278,7 +278,7 @@ solves no longer exists, it will likely be removed in the future.
    .. method:: render
 
       | :sl:`draw text on a new Surface`
-      | :sg:`render(text, antialias, color, background=None, wraplength=0) -> Surface`
+      | :sg:`render(text, antialias, fgcolor, bgcolor=None, wraplength=0) -> Surface`
```
dr0id, line 281: below in the text 
`The color argument is the color of the...`  

Update `color` it is now `fgcolor`.

dr0id, line 281: same for the 'bgcolor' change, in the text below `background` is still used.

dr0id (changes requested):
Thinking about the parameter names I find `color` and `background` more understandable than `fgcolor` and `bgcolor`. Reading this I would wonder what `fgcolor` is. 

Maybe `textcolor` and `backgroundcolor` would be nicer (but longer) and would need a change in Font and in Freetype.

Starbuck5:
> Maybe textcolor and backgroundcolor would be nicer (but longer) and would need a change in Font and in Freetype.

We can't change freetype since it already has those as kwargs. What about `color` and `bgcolor` ?

MyreMylar:
> > Maybe textcolor and backgroundcolor would be nicer (but longer) and would need a change in Font and in Freetype.
> 
> We can't change freetype since it already has those as kwargs. What about `color` and `bgcolor` ?

`color` and `bgcolor` would match the names used in classic HTML tags.

Starbuck5:
@PurityLake This looks to be nearly done, just needs a bit of sprucing up.

It seems we've come to a consensus on the kwarg names being `color` and `bgcolor`, so that needs to be updated in the docs/types/code/test.

The versionchanged needs to be updated to 2.3.0, since that's the next version now.

After those changes I think this is good to go.

test/font_test.py
```diff
@@ -8,7 +8,7 @@
 
 import pygame
 from pygame import font as pygame_font  # So font can be replaced with ftfont
-
+from pygame._freetype import STYLE_DEFAULT
```
Starbuck5, line 11: Looks like an extraneous bit of some local test got included?

docs/reST/ref/font.rst
```diff
@@ -278,7 +278,7 @@ solves no longer exists, it will likely be removed in the future.
    .. method:: render
 
       | :sl:`draw text on a new Surface`
-      | :sg:`render(text, antialias, color, background=None, wraplength=0) -> Surface`
+      | :sg:`render(text, antialias, color, bgcolor=None, wraplength=0) -> Surface`
```
Starbuck5, line 281: You haven't updated `background`  in the description of the method below. It should be explaining the `bgcolor` argument now.

Starbuck5 (approved):

PR #2001, opened 2023-03-08
Change mixer to use FileNotFoundError and fix related tests

Temmie3754:
This resolves https://github.com/pygame/pygame/issues/2792 by changing the error emitted and the test checking the error. This also fixes an issue with a test incorrectly failing because "allowedchanges" was not set to 0.

### DIFF
--- a/src_c/music.c
+++ b/src_c/music.c
@@ -382,12 +382,7 @@ _load_music(PyObject *obj, char *namehint)
     MIXER_INIT_CHECK();
 
     rw = pgRWops_FromObject(obj, &ext);
-    if (rw ==
-        NULL) { /* stop on NULL, error already set is what we SHOULD do */
-        PyErr_Fetch(&_type, &error, &_traceback);
-        PyErr_SetObject(PyExc_FileNotFoundError, error);
-        Py_XDECREF(_type);
-        Py_XDECREF(_traceback);
+    if (rw == NULL) {
         return NULL;
     }
     if (namehint) {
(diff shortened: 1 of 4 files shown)


Starbuck5:
Thanks for this.
The mixer.music documentation should get a `versionchanged` label to reflect this change.

MyreMylar (approved):
LGTM 👍 

As Starbuck5 said, just need to add something like:
```
.. versionchanged:: 2.2.0 - Now raises a FileNotFoundError when no file is found, rather than a pygame.error
```
line to `music.rst` (roughly line 37/38 with spacing around it).

MyreMylar (approved):
LGTM 👍

src_c/music.c
```diff
@@ -385,7 +385,7 @@ _load_music(PyObject *obj, char *namehint)
     if (rw ==
         NULL) { /* stop on NULL, error already set is what we SHOULD do */
         PyErr_Fetch(&_type, &error, &_traceback);
-        PyErr_SetObject(pgExc_SDLError, error);
+        PyErr_SetObject(PyExc_FileNotFoundError, error);
         Py_XDECREF(_type);
         Py_XDECREF(_traceback);
         return NULL;
```
Starbuck5, line 391: I believe this whole block can be removed, since pgRWops_FromObject will set the FileNotFounderror.

Just spotted that :)

Starbuck5 (approved):
Thanks!

PR #2002, opened 2023-03-08
Another minor fblits refactor

itzpr3d4t0r:
Removed some unnecessary inits, added a couple more comments, and reduced nesting level in one place. Renamed `flags_numeric` to `blend_flags`. `flags_numeric` was just inherited from blits and was meant to differentiate from the python flag object pointer, but that's not a thing in fblits hence the rename.


Starbuck5 (approved):
Sure, seems fine 👍

ankith26 (approved):
thankies :cake:

PR #2003, opened 2023-03-09
Circle fix

robertpfeiffer:
I just cherry-picked my own old fix (already approved before all this) plus the bit by @jhilmer that makes the drawing algorithm more correct.

### DIFF
--- a/src_c/draw.c
+++ b/src_c/draw.c
@@ -2009,6 +2009,14 @@ draw_circle_filled(SDL_Surface *surf, int x0, int y0, int radius, Uint32 color,
     int ddF_y = -2 * radius;
     int x = 0;
     int y = radius;
+    int xmin = INT_MAX;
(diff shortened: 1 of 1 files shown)


MyreMylar (approved):
LGTM 👍 

Used this test program, which produces the smeared bar across the screen on current main, and nothing visible (correctly) after this PR:

```
import pygame

pygame.init()

window_surf = pygame.display.set_mode((800, 600))
window_surf.fill((0, 0, 0))

# Draw circle very far off-screen (over int max)
pygame.draw.circle(window_surf, (255, 255, 255), (-1e30, 300), 10)

clock = pygame.time.Clock()
running = True

while running:
    frame_time = clock.tick(60)/1000.0

        if event.type == pygame.QUIT:
            running = False

    pygame.display.update()
```

MightyJosip (approved):
Works great

PR #2004, opened 2023-03-09
Modified docs css

yunline:
## Modified the css to fix the overflowing code
(and made it compatible with mobile browser)

## before:
image
image

## after:
image
image

### DIFF
--- a/docs/reST/themes/classic/static/pygame.css_t
+++ b/docs/reST/themes/classic/static/pygame.css_t
@@ -67,6 +67,7 @@ div.body table.matrix th.stub {
 /* Let a float element be beside a code block */
 div.highlight-python.notranslate {
     display: inline-block;
+    width: 100%;
 }
 
 /* -- page layout ----------------------------------------------------------- */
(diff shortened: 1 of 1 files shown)


Starbuck5 (approved):
Looks good to me. Thanks!

novialriptide:
I don't really like this CSS change, it results in a change where I have a lot of available space and I have to scroll instead.
image

Starbuck5:
> I don't really like this CSS change, it results in a change where I have a lot of available space and I have to scroll instead.

I don't see how that's different from what it does right now: https://pyga.me/docs/ref/transform.html#pygame.transform.threshold

novialriptide (approved):
> > I don't really like this CSS change, it results in a change where I have a lot of available space and I have to scroll instead.
> 
> I don't see how that's different from what it does right now: https://pyga.me/docs/ref/transform.html#pygame.transform.threshold

That's my bad, lol. Maybe in a future pr we should fix that..

PR #2005, opened 2023-03-09
Add test that the surface to be pre-multiplied has pixels

MyreMylar:
This prevents a crash when you try to pre-multiply a surface with zero pixels.


Starbuck5 (approved):

ankith26 (approved):
Thanks! :balloon:

PR #2007, opened 2023-03-10
Fix the segfault of `Surface.convert()`.

yunline:
Fix the segfault on converting an 8-bit surface.
**Test Code:**
```python
import pygame

pygame.display.init()

filename='alien3.png' # 8bit PNG (from /pygame/example/data)
#filename='lena.png' # 32bit PNG

img=pygame.image.load(filename)
img.convert()
# ↑ Should get a 'pygame.error: No video mode has been set'
```

### DIFF
--- a/src_c/surface.c
+++ b/src_c/surface.c
@@ -1562,11 +1562,11 @@ surf_convert(pgSurfaceObject *self, PyObject *args)
     }
     else {
         newsurf = pg_DisplayFormat(surf);
-        if (newsurf == NULL) {
-            SDL_FreeSurface(newsurf);
-            return RAISE(pgExc_SDLError, SDL_GetError());
-        }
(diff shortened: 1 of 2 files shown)


yunline:
https://github.com/pygame-community/pygame-ce/blob/f5a7a9a0458b4452fa5f59aef5d39d5370228b41/src_c/surface.c#L1569
Segfault happens in this line because `newsurf` is NULL.

Starbuck5:
Upon digging, I believe normally this error is raised in `surf_subtype_new`. But sometimes the surface is accessed to do stuff before that, such as in the `if(colorkey)` block.

It may be safer to put the `newsurf` NULL check after the whole newsurf creation block but before it's used. That way it would catch a theoretically case where it goes the argobject route, gets a newsurf as NULL, and then does colorkey stuff.

Starbuck5 (approved):
Thanks for this!

ankith26 (approved):

PR #2009, opened 2023-03-10
Fix text input example

yunline:
Fonts `microsoftjhenghei` ,`microsoftyahei` and `msgothic` were not set correctly and could never be selected. 
**Before:**
```
pygame-ce 2.2.0.dev1 (SDL 2.0.22, Python 3.10.6)
Using font: Arial
```
image

**After:**
```
pygame-ce 2.2.0.dev1 (SDL 2.0.22, Python 3.10.6)
Using font: Microsoft YaHei
```
image

### DIFF
--- a/examples/textinput.py
+++ b/examples/textinput.py
@@ -28,9 +28,9 @@ class TextInput:
         "notosanscjktcregular",
         "notosansmonocjktcregular",
         "notosansregular",
-        "microsoftjhenghei,microsoftjhengheiuilight",
-        "microsoftyahei,microsoftyaheiuilight",
-        "msgothic,msuigothic,mspgothic",
+        "microsoftjhenghei",
+        "microsoftyahei",
+        "msgothic",
         "msmincho",
         "Arial",
     ]
--- a/examples/textinput.py
+++ b/examples/textinput.py
@@ -27,10 +27,10 @@ class TextInput:
     FONT_NAMES = [
         "notosanscjktcregular",
         "notosansmonocjktcregular",
-        "notosansregular,",
-        "microsoftjhengheimicrosoftjhengheiuilight",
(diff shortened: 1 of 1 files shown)


yunline:
Once you installed a Japanese language pack on Windows, the `MS Mincho` will be installed to your system too. This font will be selected instead of `Arial`. 
image
However, most users in other languages do not have a Japanese language pack.

Starbuck5:
I believe this is my fault, in a roundabout way. Sysfont used to have a bug where font names would get globbed together like that when the windows registry defined fonts with multiple names separated by "&". See https://github.com/pygame/pygame/pull/2640.

examples/textinput.py
```diff
@@ -27,10 +27,10 @@ class TextInput:
...
         "notosanscjktcregular",
         "notosansmonocjktcregular",
-        "notosansregular,",
-        "microsoftjhengheimicrosoftjhengheiuilight",
-        "microsoftyaheimicrosoftyaheiuilight",
-        "msgothicmsuigothicmspgothic",
+        "notosansregular",
+        "microsoftjhenghei,microsoftjhengheiuilight",
+        "microsoftyahei,microsoftyaheiuilight",
+        "msgothic,msuigothic,mspgothic",
```
Starbuck5, line 33: Is there any point keeping the secondary names around still? Like `microsoftjhengheiuilight` or `microsoftyaheiuilight` or `msuigothic`?

yunline: Humm, maybe those are not needed because they are subclasses of the former ones.

Starbuck5 (approved):
Thanks! Good catch.

robertpfeiffer:
Can we search for GNU Unifont in here, and use that as a fallback *if and only if* it is available?

robertpfeiffer (approved):

PR #2010, opened 2023-03-11
Add keyword for image

yunline:
Issue: #1990


src_c/image.c
```diff
@@ -91,29 +91,31 @@ image_load_basic(PyObject *self, PyObject *obj)
...
+        return PyObject_Call(extloadobj, arg, kwarg);
 }
 
 static PyObject *
-image_load(PyObject *self, PyObject *arg)
+image_load(PyObject *self, PyObject *arg, PyObject *kwarg)
 {
     PyObject *obj;
     const char *name = NULL;
+    static char *kwds[] = {"filename", "namehint", NULL};
```
Starbuck5, line 108: Thinking about this, `filename` doesn't seem quite right. Since it is really a `filepath`. It also doesn't jive with the case where it's actually a file object. It looks like mixer.Sound has a bunch of special code to handle this "branched keyword" case, where you want two exclusive kwargs for different types of function calls, for real files vs file objects. I'd rather not have all that complex code, so what about calling the kwarg `file` instead?

This would necessitate changing the short signature in the docs.

Starbuck5:
Needs notes in the docs about the kwarg additions.

src_c/image.c
yunline: Thanks. Fixed now.

docs/reST/ref/image.rst
```diff
@@ -68,8 +68,8 @@ following formats.
 .. function:: load
 
    | :sl:`load new image from a file (or file-like object)`
-   | :sg:`load(filename) -> Surface`
-   | :sg:`load(fileobj, namehint="") -> Surface`
+   | :sg:`load(file) -> Surface`
+   | :sg:`load(file, namehint="") -> Surface`
```
oddbookworm, line 72: This seems like a weird change to me, but I'm not 100% opposed to it. I'm just not sure about it. Not enough to prevent me from approving though

docs/reST/ref/image.rst
```diff
@@ -100,13 +100,15 @@ following formats.
...
+
    .. ## pygame.image.load ##
 
 .. function:: save
 
    | :sl:`save an image to file (or file-like object)`
-   | :sg:`save(Surface, filename) -> None`
-   | :sg:`save(Surface, fileobj, namehint="") -> None`
+   | :sg:`save(Surface, file) -> None`
+   | :sg:`save(Surface, file, namehint="") -> None`
```
oddbookworm, line 111: same as my other comment

oddbookworm (approved):
Works perfectly fine in my testing and the docs look fine. There's a couple of things that look weird to me, but I don't think they're enough to stop my approval

Starbuck5 (approved):
Thanks!

docs/reST/ref/image.rst
Starbuck5: Since the args are not the same anymore, it should be collapsed into a single `load(file, namehint="")` as the one signature. But it's a tiny thing and I want to merge the PR anyway, so I will.

PR #2011, opened 2023-03-11
Renamed "`the_args`" to "`blend_flags`" everywhere

itzpr3d4t0r:
I renamed “the_args” to “blend_flags” because they are not regular arguments but flags for blending modes. However, I’m not sure how to handle the discrepancy between the python code and the C code. The python code expects “special_flags” as an argument name in all blit functions, but the C code uses “blend_flags” as a variable name when parsing. How should we resolve this?

### DIFF
--- a/docs/reST/c_api/surface.rst
+++ b/docs/reST/c_api/surface.rst
@@ -42,11 +42,11 @@ Header file: src_c/include/pygame.h
    This is a macro. Argument *x* is assumed to be a Surface, or subclass of
    Surface, instance.
 
-.. c:function:: int pgSurface_Blit(PyObject *dstobj, PyObject *srcobj, SDL_Rect *dstrect, SDL_Rect *srcrect, int the_args)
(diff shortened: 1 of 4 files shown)


itzpr3d4t0r:
Ok turns out i need to work on this a bit more.

itzpr3d4t0r:
Ok the majority of the work seems done but there are a couple files I don't know how to change:
image

oddbookworm:
> Actually I think I'm fine, those files aren't tracked by git.

The `docs/generated` directory is created by running `python setup.py docs`, so the html in there is fine to leave alone (run `python setup.py docs --fullgeneration` in the event that the changes you made to the docs aren't 100% reflected with the normal docs gen)

oddbookworm:
Here's what the modified docs page looks like when I generate locally
image

oddbookworm (approved):
LGTM! Blitting worked just fine after from my testing and the docs page looks fine to me

MyreMylar (approved):
LGTM 👍

Issue #2014, opened 2023-03-12
Additional repr information for surfaces


SSS-Says-Snek:
Upstream issue: https://github.com/pygame/pygame/issues/3380

I'll try to implement this

zoldalma999:
Fixed by #2016

PR #2016, opened 2023-03-13
Add global alpha and colorkey info to Surf repr

SSS-Says-Snek:
Implements #2014

### DIFF
--- a/src_c/surface.c
+++ b/src_c/surface.c
@@ -499,8 +499,37 @@ surface_str(PyObject *self)
         return PyUnicode_FromString("<Surface(Dead Display)>");
     }
 
-    return PyUnicode_FromFormat("<Surface(%dx%dx%d SW)>", surf->w, surf->h,
-                                surf->format->BitsPerPixel);
+    PyObject *colorkey = surf_get_colorkey((pgSurfaceObject *)self, NULL);
+    if (colorkey == NULL) {
+        return NULL;
+    }
+    PyObject *global_alpha = surf_get_alpha((pgSurfaceObject *)self, NULL);
+    if (global_alpha == NULL) {
+        Py_DECREF(colorkey);
+        return NULL;
+    }
+
+    // 49 is max len of format str, plus null terminator
+    char format[50] = "<Surface(%dx%dx%d";
+    if (PyObject_IsTrue(colorkey)) {
+        strcat(format, ", colorkey=%S");
+    }
+    if (PyObject_IsTrue(global_alpha)) {
+        strcat(format, ", global_alpha=%S");
+    }
+    strcat(format, ")>");
+
+    // Because PyUnicode_FromFormat ignores extra args, if we have no colorkey,
+    // but alpha, we can "move up" the global alpha to this spot No need to do
+    // this vice-versa, as it ignores extra args
+    PyObject *formatted_str = PyUnicode_FromFormat(
+        format, surf->w, surf->h, surf->format->BitsPerPixel,
+        PyObject_IsTrue(colorkey) ? colorkey : global_alpha, global_alpha);
+
+    Py_DECREF(colorkey);
+    Py_DECREF(global_alpha);
+
+    return formatted_str;
 }
 
 static intptr_t
--- a/test/surface_test.py
+++ b/test/surface_test.py
@@ -63,7 +63,31 @@ class SurfaceTypeTest(unittest.TestCase):
 
     def test_print(self):
         surf = pygame.Surface((70, 70), 0, 32)
-        self.assertEqual(repr(surf), "<Surface(70x70x32 SW)>")
+        self.assertEqual(repr(surf), "<Surface(70x70x32)>")
+
+        surf_global = pygame.Surface((70, 70), pygame.SRCALPHA, 32)
+        self.assertEqual(repr(surf_global), "<Surface(70x70x32, global_alpha=255)>")
+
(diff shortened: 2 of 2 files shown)


SSS-Says-Snek:
I think I need to cast the self ptr, I'll try when I get back

yunline (approved):
LGTM 👍

Starbuck5:
It seems like it should still have the parentheses, even though it's not a real constructor.

`<Surface 20x20x32>` -> `<Surface(20x20x32)>`

src_c/surface.c
```diff
@@ -499,8 +499,23 @@ surface_str(PyObject *self)
...
-                                surf->format->BitsPerPixel);
+    PyObject *colorkey = surf_get_colorkey((pgSurfaceObject *)self, NULL);
+    PyObject *global_alpha = surf_get_alpha((pgSurfaceObject *)self, NULL);
+
+    // 46 is max len of format str, plus null terminator
+    char format[47] = "<Surface %dx%dx%d";
+    if (PyObject_IsTrue(colorkey))
+        strcat(format, " colorkey=%S");
+    if (PyObject_IsTrue(global_alpha))
+        strcat(format, " global_alpha=%S");
```
itzpr3d4t0r, line 510: I think there should be commas in between the different "attributes" so like:
```suggestion
        strcat(format, ", colorkey=%S");
    if (PyObject_IsTrue(global_alpha))
        strcat(format, ", global_alpha=%S");
```

SSS-Says-Snek: Sure, I was mainly following from the issue

SSS-Says-Snek:
I'll add those things when I get home

src_c/surface.c
```diff
@@ -499,8 +499,23 @@ surface_str(PyObject *self)
...
-                                surf->format->BitsPerPixel);
+    PyObject *colorkey = surf_get_colorkey((pgSurfaceObject *)self, NULL);
+    PyObject *global_alpha = surf_get_alpha((pgSurfaceObject *)self, NULL);
+
+    // 49 is max len of format str, plus null terminator
+    char format[50] = "<Surface(%dx%dx%d";
+    if (PyObject_IsTrue(colorkey))
+        strcat(format, ", colorkey=%S");
+    if (PyObject_IsTrue(global_alpha))
+        strcat(format, ", global_alpha=%S");
```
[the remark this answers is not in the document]
SSS-Says-Snek: To me, the parentheses surrounding each part of the repr as suggested in the upstream issue is redundant now that commas separate them

src_c/surface.c
```diff
@@ -499,8 +499,23 @@ surface_str(PyObject *self)
         return PyUnicode_FromString("<Surface(Dead Display)>");
     }
 
-    return PyUnicode_FromFormat("<Surface(%dx%dx%d SW)>", surf->w, surf->h,
-                                surf->format->BitsPerPixel);
+    PyObject *colorkey = surf_get_colorkey((pgSurfaceObject *)self, NULL);
+    PyObject *global_alpha = surf_get_alpha((pgSurfaceObject *)self, NULL);
```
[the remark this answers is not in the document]
SSS-Says-Snek: Yeah, maybe we can directly find the alpha

src_c/surface.c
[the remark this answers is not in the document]
itzpr3d4t0r: Sry i meant curly braces:
```suggestion
    if (PyObject_IsTrue(colorkey)) {
        strcat(format, ", colorkey=%S");
        }
    if (PyObject_IsTrue(global_alpha)) {
        strcat(format, ", global_alpha=%S");
        }
```

src_c/surface.c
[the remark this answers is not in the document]
SSS-Says-Snek: Now that I think about it, we can tolerate a small perf hit, as people don't really print surfaces that often, and even so, it's probably just for debug

src_c/surface.c
```diff
@@ -499,8 +499,25 @@ surface_str(PyObject *self)
...
```
Starbuck5, line 503: These are both leaking memory.

MyreMylar: you should be able to fix this with:
```
Py_DECREF(colorkey);
Py_DECREF(global_alpha);
```

Once you are done with the objects, perhaps on line 514.

itzpr3d4t0r: No he has to put the `PyUnicode_FromFormat` at line 518 in a separate variable before the return, DECREF them and the nreturn the variable.

SSS-Says-Snek: Oops, IG I thought that PyUnicode_FromFormat eats the references

itzpr3d4t0r (approved):
LGTM

src_c/surface.c
```diff
@@ -499,8 +499,30 @@ surface_str(PyObject *self)
...
```
Starbuck5, line 503: Sorry to not spot this earlier, but I believe these should forward errors

I.e.

```c
PyObject *colorkey = surf_get_colorkey((pgSurfaceObject *)self, NULL);
if (colorkey == NULL) {
    return NULL;
}
PyObject *global_alpha = surf_get_alpha((pgSurfaceObject *)self, NULL);
if (global_alpha == NULL) {
    Py_DECREF(colorkey);
    return NULL;
}
```

Starbuck5 (approved):

Issue #2025, opened 2023-03-16
The current blur algorithm works badly with indexed surfaces.


yunline:
![image](https://user-images.githubusercontent.com/31395137/225644381-a748246f-8cd1-4927-95fe-39f66b401cbc.png)

**Test Code:**
```python
import pygame
import sys

pygame.display.init()
dsf=pygame.display.set_mode((1280,720))

sf=pygame.transform.scale(pygame.image.load("alien3.png"),(512,512))
print(sf)

sf_blur=sf.copy()

dsf.fill((0,150,0))
dsf.blit(sf,(0,0,1,1))

pygame.transform.gaussian_blur(sf,10,dest_surface=sf_blur)
dsf.blit(sf_blur,(640,0,1,1))
pygame.display.update()

while 1:
    for event in pygame.event.get():
        if event.type==256:
            sys.exit()
```

yunline:
Maybe a warning should be raised while handling an 8-bit surface?
Or, just state the side effect in docs?

PR #2026, opened 2023-03-17
Raise `ValueError` when bluring an indexed surface.

yunline:
Fix https://github.com/pygame-community/pygame-ce/issues/2025


MyreMylar (approved):
LGTM 👍

ankith26 (commented):
I wonder if this can be fixed somehow instead of erroring? But then again, indexed/palleted surfaces must be super rare these days so I'm not sure if it'd be worth the effort

PR #2028, opened 2023-03-17
transform.blur: fix #2027

### LINKED ISSUE #2027: Blur algorithm works badly when the pitch is not equal to width*bytes_per_pixel.
```python
import pygame
import sys

pygame.font.init()

font=pygame.font.SysFont('Arial',100)

pygame.display.init()

sf=font.render('HelloWorld',1,(255,255,255))#.convert_alpha()
print(sf)
print(sf.get_pitch())
print(sf.get_width()*sf.get_bitsize()//8)

sf_blur=pygame.transform.box_blur(sf,10)

dsf.fill((0,150,0))
dsf.blit(sf,(0,0,1,1))
dsf.blit(sf_blur,(640,0,1,1))
pygame.display.update()

while 1:
        if event.type==pygame.QUIT:
            sys.exit()
```
```
pygame-ce 2.2.0.dev1 (SDL 2.0.22, Python 3.10.6)
<Surface(487x115x32 SW)>
2016
1948
```
image

And, when I use the `convert_alpha()` to convert the surface, the bug disappeared:
```
pygame-ce 2.2.0.dev1 (SDL 2.0.22, Python 3.10.6)
<Surface(487x115x32 SW)>
1948
1948
```
image

yunline:
Fix #2027
image

### DIFF
--- a/src_c/transform.c
+++ b/src_c/transform.c
@@ -3074,39 +3074,41 @@ box_blur(SDL_Surface *src, SDL_Surface *dst, int radius, SDL_bool repeat)
     Uint8 *srcpx = (Uint8 *)src->pixels;
     Uint8 *dstpx = (Uint8 *)dst->pixels;
     Uint8 nb = src->format->BytesPerPixel;
-    int w = dst->w, h = dst->h, pitch = dst->pitch;
+    int w = dst->w, h = dst->h;
+    int dst_pitch = dst->pitch;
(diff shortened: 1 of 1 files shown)


Starbuck5 (approved):
I found this same problem when trying to blur a subsurface, and this PR fixed it.

So it should be added into 2.2, as it's a bugfix for a new 2.2 feature.

Thanks.

PR #2031, opened 2023-03-17
Fix issue with converting surfaces to 8 bit surfaces with palettes

Temmie3754:
This resolves issue https://github.com/pygame/pygame/issues/2477 by checking if the new surface needs a palette and creating the default one for it if it does.

### DIFF
--- a/src_c/surface.c
+++ b/src_c/surface.c
@@ -1457,6 +1457,7 @@ surf_convert(pgSurfaceObject *self, PyObject *args)
             /* will be updated later, initialize to make static analyzer happy
              */
             int bpp = 0;
+            SDL_Palette *palette = SDL_AllocPalette(default_palette_size);
             SDL_PixelFormat format;
 
             memcpy(&format, surf->format, sizeof(format));
@@ -1566,8 +1567,6 @@ surf_convert(pgSurfaceObject *self, PyObject *args)
                     /* Give the surface something other than an all white
                      * palette.
                      */
-                    SDL_Palette *palette =
-                        SDL_AllocPalette(default_palette_size);
                     SDL_SetPaletteColors(palette, default_palette_colors, 0,
                                          default_palette_size);
                     format.palette = palette;
@@ -1575,6 +1574,7 @@ surf_convert(pgSurfaceObject *self, PyObject *args)
             }
             newsurf = SDL_ConvertSurface(surf, &format, 0);
             SDL_SetSurfaceBlendMode(newsurf, SDL_BLENDMODE_NONE);
+            SDL_FreePalette(palette);
         }
     }
     else {
--- a/src_c/surface.c
+++ b/src_c/surface.c
@@ -1567,9 +1567,9 @@ surf_convert(pgSurfaceObject *self, PyObject *args)
                      * palette.
                      */
                     SDL_Palette *palette =
-                        SDL_AllocPalette(default_palette_size - 1);
+                        SDL_AllocPalette(default_palette_size);
                     SDL_SetPaletteColors(palette, default_palette_colors, 0,
-                                         default_palette_size - 1);
+                                         default_palette_size);
                     format.palette = palette;
                 }
             }
--- a/src_c/surface.c
+++ b/src_c/surface.c
@@ -1556,6 +1556,17 @@ surf_convert(pgSurfaceObject *self, PyObject *args)
                  * that at least one entry is not black.
                  */
                 format.palette = NULL;
+            if (SDL_ISPIXELFORMAT_INDEXED(SDL_MasksToPixelFormatEnum(
+                    format.BitsPerPixel, format.Rmask, format.Gmask,
+                    format.Bmask, format.Amask))) {
+                /* Give the surface something other than an all white palette.
+                 */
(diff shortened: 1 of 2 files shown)


src_c/surface.c
```diff
@@ -1556,6 +1556,23 @@ surf_convert(pgSurfaceObject *self, PyObject *args)
...
+                    format.Bmask, format.Amask))) {
+                if (SDL_ISPIXELFORMAT_INDEXED(surf->format->format)) {
+                    format.palette = surf->format->palette;
+                }
+                else {
+                    /* Give the surface something other than an all white
+                     * palette.
+                     */
+                    SDL_Palette *palette =
+                        SDL_AllocPalette(default_palette_size - 1);
```
yunline, line 1570: `-1` here is not needed.

src_c/surface.c
```diff
@@ -1556,6 +1556,23 @@ surf_convert(pgSurfaceObject *self, PyObject *args)
...
+                    SDL_SetPaletteColors(palette, default_palette_colors, 0,
+                                         default_palette_size - 1);
```
yunline, line 1572: Also here.

yunline (changes requested):

yunline (approved):
LGTM

src_c/surface.c
```diff
@@ -1556,6 +1556,23 @@ surf_convert(pgSurfaceObject *self, PyObject *args)
...
+                }
+                else {
+                    /* Give the surface something other than an all white
+                     * palette.
+                     */
+                    SDL_Palette *palette =
+                        SDL_AllocPalette(default_palette_size);
+                    SDL_SetPaletteColors(palette, default_palette_colors, 0,
+                                         default_palette_size);
+                    format.palette = palette;
```
MyreMylar, line 1573: should we be using:
```
SDL_SetSurfacePalette()
```

here and then freeing our allocation with:

```
SDL_FreePalette()
```

?

MyreMylar (commented):
Is there are chance you are leaking memory here for the all white palette? The SDL functions for setting the surface palette are careful to destroy/free (the functions have recently been renamed) any existing palettes before assigning a new one.

It looks like here you might potentially be just replacing the all white palette with the default palette leaving a dangling allocation for the old white palette.

On a wider scale it might be cool if this conversion function actually generated an appropriate 8 bit palette for the image being converted using something like the [Median cut](https://en.wikipedia.org/wiki/Median_cut) algorithm. See [here ](https://github.com/jkirkwin/gifgen/blob/master/palettize/median_cut.cpp) for one example. This would be beyond the scope of this PR though.

Temmie3754:
Thanks for catching that, it does indeed look as if there is a memory leak, I'll work on fixing that.

src_c/surface.c
```diff
@@ -1556,8 +1557,24 @@ surf_convert(pgSurfaceObject *self, PyObject *args)
                  * that at least one entry is not black.
                  */
                 format.palette = NULL;
+            if (SDL_ISPIXELFORMAT_INDEXED(SDL_MasksToPixelFormatEnum(
+                    format.BitsPerPixel, format.Rmask, format.Gmask,
+                    format.Bmask, format.Amask))) {
+                if (SDL_ISPIXELFORMAT_INDEXED(surf->format->format)) {
+                    format.palette = surf->format->palette;
```
MyreMylar, line 1564:
```suggestion
                    SDL_SetPixelFormatPalette(format, surf->format->palette);
```

src_c/surface.c
```diff
@@ -1556,8 +1557,24 @@ surf_convert(pgSurfaceObject *self, PyObject *args)
...
+                }
+                else {
+                    /* Give the surface something other than an all white
+                     * palette.
+                     */
+                    SDL_SetPaletteColors(palette, default_palette_colors, 0,
+                                         default_palette_size);
+                    format.palette = palette;
```
MyreMylar, line 1572:
```suggestion
                    SDL_SetPixelFormatPalette(format, palette);
```

MyreMylar: Just added these suggestions to minimise risk of memory leaking on the format's _old_ palette.

src_c/surface.c
```diff
@@ -1556,8 +1557,24 @@ surf_convert(pgSurfaceObject *self, PyObject *args)
                  * that at least one entry is not black.
                  */
                 format.palette = NULL;
+            if (SDL_ISPIXELFORMAT_INDEXED(SDL_MasksToPixelFormatEnum(
+                    format.BitsPerPixel, format.Rmask, format.Gmask,
+                    format.Bmask, format.Amask))) {
```
Starbuck5, line 1562: Can't you use format.format here?

Starbuck5:
This should get a unit test to make sure the issue doesn't regress in the future.

src_c/surface.c
Temmie3754: No since format.format doesn't get updated to match bpp and masks so it needs to get the correct enum value from the masks and bpp.

Temmie3754:
Added a unit test

Starbuck5 (approved):
Thanks!

PR #2032, opened 2023-03-18
fixed banding bug with draw.circle

oddbookworm:
When `draw.circle` is fed a circle centered to the left of the screen's left edge, a band appears across the entire screen. I tracked it to [this commit](https://github.com/pygame-community/pygame-ce/commit/046efe181375a9f0078f07c880ac7b2e304e0e01). I fixed the bug and added a test case to prevent regression.

Test code:
```python
import pygame

running = True
while running:
        if event.type == pygame.QUIT:
            running = False
            
    screen.fill("black")
    pygame.draw.circle(screen, "white", (-1, 250), 10)
    pygame.display.flip()
```
Result on current `main`:
broken

Result after this PR:
fixed

### DIFF
--- a/src_c/draw.c
+++ b/src_c/draw.c
@@ -2019,7 +2019,6 @@ draw_circle_filled(SDL_Surface *surf, int x0, int y0, int radius, Uint32 color,
     int ddF_y = -2 * radius;
     int x = 0;
     int y = radius;
-    int xmin = INT_MAX;
     int xmax = INT_MIN;
 
     if (x0 < 0) {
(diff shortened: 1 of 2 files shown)


Matiiss (changes requested):
This needs a test for when center is greater than surface width so it doesn't make a trail to the left.

src_c/draw.c
```diff
@@ -2023,7 +2023,7 @@ draw_circle_filled(SDL_Surface *surf, int x0, int y0, int radius, Uint32 color,
     int xmax = INT_MIN;
```
dr0id, line 2024: `MIN(x, xmin)` on line 2045: xmin is always INT_MAX no? Can x be ever bigger than xmin?

oddbookworm:
> I assume the bug was introduced in [c88a1d8](https://github.com/pygame-community/pygame-ce/commit/c88a1d8f7b31099cec84dc5cb7aeebdada783e83). Or was it introduced by [ef2399e](https://github.com/pygame-community/pygame-ce/commit/ef2399ecfd7e5b1d2b4560458f6ec393b3a634fe)?

It was introduced in [c88a1d8](https://github.com/pygame-community/pygame-ce/commit/c88a1d8f7b31099cec84dc5cb7aeebdada783e83). If I go back one commit to [f5a7a9a](https://github.com/pygame-community/pygame-ce/commit/f5a7a9a0458b4452fa5f59aef5d39d5370228b41), then the problem does not appear

> Does your fix re-introduce the original problem described by @novialriptide in [pygame/pygame#3143](https://github.com/pygame/pygame/issues/3143)?

No. The behavior from novial's test code is exactly as expected. Pure black screen.

src_c/draw.c
oddbookworm: You are correct. I removed the redundant logic associated with the `xmin` variable. Tests still pass on my local, performance isn't significantly affected in my testing, and everything still seems to work as expected.

test/draw_test.py
```diff
@@ -5618,6 +5618,57 @@ def test_circle__diameter(self):
...
+            (-1, height / 2),
+            (width + 1, height / 2),
+            (width / 2, -1),
+            (width / 2, height + 1),
+        ]
+        for center in test_centers:
+            surf.fill("black")
+            self.draw_circle(surf, "white", center, radius)
+
+            # pixel at the center should be black, not white
```
dr0id, line 5669: suggestion: this comment would be helpful as assert message in case the test should fail ( assertion methods have a `msg=` in the kw args). This would remove the comment but still provide the information.

dr0id (approved):
LGTM

dr0id (approved):
LGTM

MyreMylar (approved):
LGTM 👍  fixes the issue and I can't see any other related issues with fully or partially offscreen circles.

Matiiss (approved):
LGTM 👍

Issue #2008, opened 2023-03-10
Investigate font test failures


Starbuck5:
Running the test suite (`python -m pygame.tests`) locally on Windows yields this for me:

```
======================================================================
FAIL: test_match_font_all_exist (pygame.tests.font_test.FontModuleTest)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "C:\Users\charl\AppData\Local\Programs\Python\Python39\lib\site-packages\pygame\tests\font_test.py", line 121, in test_match_font_all_exist
    self.assertTrue(os.path.isabs(path) and os.path.isfile(path))
AssertionError: False is not true

======================================================================
FAIL: test_match_font_all_exist (pygame.tests.ftfont_test.FtFontModuleTest)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "C:\Users\charl\AppData\Local\Programs\Python\Python39\lib\site-packages\pygame\tests\font_test.py", line 121, in test_match_font_all_exist
    self.assertTrue(os.path.isabs(path) and os.path.isfile(path))
AssertionError: False is not true

----------------------------------------------------------------------
Ran 1828 tests in 39.370s

FAILED (failures=2)
```

Why is sysfont returning fonts that don't exist / aren't absolute paths?

yunline:
Can't reperduce this on my computer.
```
PS X:\github\pygame-ce> & C:/Users/yunline/AppData/Local/Programs/Python/Python310/python.exe x:/github/pygame-ce/test/font_test.py
pygame-ce 2.2.0.dev1 (SDL 2.0.22, Python 3.10.6)
............................................................
----------------------------------------------------------------------
Ran 60 tests in 14.313s

OK
PS X:\github\pygame-ce> 
```

yunline:
Oh I reperduced this after deleting the font `MS Mincho`.
```
pygame-ce 2.2.0.dev1 (SDL 2.0.22, Python 3.10.6)
.........F..................................................
======================================================================
FAIL: test_match_font_all_exist (__main__.FontModuleTest)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "x:\github\pygame-ce\test\font_test.py", line 121, in test_match_font_all_exist
    self.assertTrue(os.path.isabs(path) and os.path.isfile(path))
AssertionError: False is not true

----------------------------------------------------------------------
Ran 60 tests in 5.991s

FAILED (failures=1)
```

yunline:
[...]

PR #2034, opened 2023-03-19
Issue 2008: fix initial font loading in windows where there is a regi…

dr0id:
Issue #2008: fix initial font loading in windows where there is a registry entry but the actual file is missing


dr0id:
not sure, would this need a unittest?

MyreMylar (approved):
LGTM 👍 

Don't think it needs a unit test, one would be difficult to construct anyway and I think this is unlikely to go backwards.

Starbuck5:
I ran a test for performance, since we don't want to hang on sysfont init.

Before this PR, sysfont init on Windows takes 0.006 seconds. Now it takes 0.03 seconds. 5x slower, but 0.03 seconds is fine.

Starbuck5 (approved):

PR #2039, opened 2023-03-21
Make Texture and Renderer compatible with float input (floor() based pixel rasterisation/conversion changing to round())

yunline:
**Modified:**
- `Texture.draw`
- `Renderer.draw_line`
- `Renderer.draw_point`
- `Renderer.draw_rect`
- `Renderer.fill_rect`
- `Renderer.draw_triangle`
- `Renderer.draw_quad`

**Removed:**
- `Texture.draw_internal`

**Refactor**
- `Image.draw`

**Notice**
~~`SDL_RenderDrawXxxxF` is only available after SDL 2.0.10. 
If SDL version<2.0.10, `SDL_RenderDrawXxxx` will be used as a fall back.~~

**Test Code:**
```python
import pygame

win=video.Window()

r=video.Renderer(win)

img=pygame.image.load('lena.png')
tx=video.Texture.from_surface(r,img)
im=video.Image(tx,srcrect=(256,256,256,256))

r.draw_color=(0,128,30,255)
r.clear()

im.draw(dstrect=(0.6,0,128,128))
tx.draw(dstrect=(129,0,128,128))

r.draw_color=(255,255,0,255)

r.draw_line((0,192.6),(128,192.6))
r.draw_line((0,193),(128,193))

r.fill_rect((256.6,256,128,128))
r.fill_rect((385,256,128,128))

r.present()

while 1:
    pygame.event.get()
```

**Before:**
image

**After:**
image


src_c/cython/pygame/_sdl2/video.pyx
```diff
@@ -704,6 +729,9 @@ cdef class Texture:
             csrcrect = pgRect_FromObject(srcrect, &src)
             if not csrcrect:
                 raise TypeError("the argument is not a rectangle or None")
+        
+        if SDL_VERSION_ATLEAST(2,0,10):
+            self._frect_draw(csrcrect,dstrect,angle,origin,flip_x,flip_y)
```
dr0id, line 735: I probably don't understand it yet, but on line 734 the '_frect_draw' method is called. The lines that follow, why are they needed? It looks to me as they where the fallback to draw with int input, or not?

src_c/cython/pygame/_sdl2/video.pyx
```diff
@@ -1229,57 +1233,103 @@ cdef class Renderer:
...
         if res < 0:
             raise error()
 
     def draw_rect(self, rect):
         # https://wiki.libsdl.org/SDL_RenderDrawRect
         cdef SDL_Rect _rect
-        cdef SDL_Rect *rectptr = pgRect_FromObject(rect, &_rect)
+        cdef SDL_Rect *rectptr
+        cdef SDL_FRect _frect
+        cdef SDL_FRect *frectptr
```
dr0id, line 1266: still need to declare all variables up front? I mean they are only used in one branch of the if. Resolve at will.

dr0id:
The two images look a bit different. Is that intentional, that the gap between the two yellow rects does not exist in the second one?
And the horizontal line looks like it has less width. Is that intentional too?

src_c/cython/pygame/_sdl2/video.pyx
yunline: It seems cython doesn't allow declaration in if statement

src_c/cython/pygame/_sdl2/video.pyx
yunline: Oops, a return is needed here.
it should be like:
```py
if SDL_VERSION_ATLEAST(2,0,10):
    self._frect_draw(csrcrect,dstrect,angle,origin,flip_x,flip_y)
    return
```
I'll fix this soon.
Thank you.

yunline:
> The two images look a bit different. Is that intentional, that the gap between the two yellow rects does not exist in the second one? And the horizontal line looks like it has less width. Is that intentional too?

Before this PR, the float input is cast to int. This causes the gap and the abnormal line width.

dr0id (approved):
LGTM, not sure why the one ci build fails

MyreMylar (approved):
Seems like this effectively switches float handling behaviour for these functions from a `floor()` to a `round()` (by removing the cast to int) - and simplifies handling of FRects.

I suspect that given the current type hints saying `int` that this behaviour could be argued as not really defined but it is a breaking change.

Internally on SDLs side it looks like there will be no change as the int versions of the drawing functions just call the floating point ones after converting the input ints back to floats. All we are really changing here is the rounding behaviour from `floor()` to `round()`.

We should mention this in the release notes for those who might be affected.

Approved conditional on those merge conflicts being resolved.

buildconfig/stubs/pygame/_sdl2/video.pyi
```diff
@@ -146,18 +146,18 @@ class Renderer:
...
     def draw_triangle(
-        self, p1: Iterable[int], p2: Iterable[int], p3: Iterable[int]
+        self, p1: Iterable[float], p2: Iterable[float], p3: Iterable[float]
     ) -> None: ...
     def fill_triangle(
         self, p1: Iterable[float], p2: Iterable[float], p3: Iterable[float]
     ) -> None: ...
     def draw_quad(
-        self, p1: Iterable[int], p2: Iterable[int], p3: Iterable[int], p4: Iterable[int]
+        self, p1: Iterable[float], p2: Iterable[float], p3: Iterable[float], p4: Iterable[float]
```
ankith26, line 160: Maybe all `Iterable[font]` should be replaced by `Coordinate` as appropriate? `Iterable[int]` is actually broader that what we can actually handle in code.

yunline: Yes, but some of them are out of scope of this PR. I'll open another one to fix them after this PR is merged

Starbuck5:
I'd like to link in this other issue for consideration, #2409.

It's also taking about float->int conversion for drawing, but it only wants to consistently floor, not round.

Temmie3754:
> Although reading this PR's changes I don't see anything about round() either. Is the title out of date?
>
I think they mean that by using floats rather than rounding to the floor as a float to int conversion does, the underlying float drawing methods round to find the correct pixel which is why it doesn't appear in the PR.

MyreMylar (approved):
LGTM 👍

Issue #2047, opened 2023-03-23
SystemError when unplugging a controller.


yunline:
**How to reperduce:**
1. Plug your controller in
2. Run this test script
```python
import pygame

from pygame._sdl2 import controller

controller.init()

for x in range(pygame.joystick.get_count()):
    if controller.is_controller(x):
        c = controller.Controller(x)

pygame.init()

while 1:
    for event in pygame.event.get():
        print(event)
```
3. Unplug the controller when the script is running
4. Then we got this
```
KeyError: 1

The above exception was the direct cause of the following exception:

Traceback (most recent call last):
  File "x:\python\pygame_test\event_joy.py", line 14, in <module>   
    for event in pygame.event.get():
SystemError: <built-in function get> returned a result with an exception set
```

PR #2048, opened 2023-03-23
Remove `event._joy_instance_map`

yunline:
Fix issues:
https://github.com/pygame-community/pygame-ce/issues/2047
https://github.com/pygame/pygame/issues/3634

Removing for 2 reasons.
- ~~`_joy_instance_map` is mapping `instance_id` to `instance_id`, which is pointless.~~
- `_joy_instance_map` also causes `KeyError` (which is not caught and causes the `SystemError`) when a controller is being unplugged but was not in the dict.


zoldalma999 (changes requested):
This will break code as implemented.

joy_instance_map does not map `instance_id` to `instance_id`, it maps `instance_id` to `device_index` (or it tries its best to). Which is needed for backwards compatibility. With your implementation, we will be returning instance_ids instead of device_indices Every code that uses `event.joy` with `joystick.get_id` will break, but not instantly. Only after they unplug and plug a joystick back in (since until this happens, the to ids *should* match up, though it is not guaranteed). And not in an obvious way, like an error, but their event just won't match up with their joystick.

However I do agree with removing the mapping dictionary. A better solution to this would be to add a joystick function that returns a joystick object from an instance id. This should work since 1) joysticks don't send events 'till created and 2) the `pgJoystickObject` keeps track of its id already (or if it cannot find it, it would just set the joy attribute to -1 or something like that. There is no proper way to fix this, since the two ids don't properly match up). It would fix the bug mentioned and would work in a backwards compatible manner. The only place where this is not needed is the `JOYDEVICEADDED` event (which already uses device_indices).

If you need help, or are unsure of something, I'd be happy to help you complete the pr. I'd also be open to implementing the change myself on this branch directly.

yunline:
> This will break code as implemented.

Sorry about this. It's my bad.
I'll try to fix it soon.

yunline:
I added a new c api `pgJoystick_GetDeviceIndexByInstanceID()` to get the device index.
But there's a issue, when a joystick is initialized by controller module instead of joystick module, the "joy" attribute will be always `-1` because the joystick object is not created.
However, I assume that people who use the  deprecated "joy" attribute don't use _sdl2.controller. So I think we don't have to solve this problem.

zoldalma999 (approved):
Looks good to me! Thanks for the pr!🎉

> But there's a issue, when a joystick is initialized by controller module instead of joystick module, the "joy" attribute will be always -1 because the joystick object is not created.

Yeah, that is fine. We should deprecate `event.joy` anyway, since it is not really supposed to be used anymore (but `instance_id` instead)

Also, sorry for the long delay. I should be (hopefully) more active now

Starbuck5 (approved):
Zoldalma walked me through what this does, and it seems legit.

PR #2049, opened 2023-03-23
added consistent dest_surface description in transform.rst

VantaTree:


Starbuck5:
Depth and format are not interchangeable.

You could have 2 32 bit surfaces, one of which is RGBA and one of which is BGRA.

VantaTree:
Results from testing each surface
```
S: same
A: any
SF: same as first
~: not tested

|   func name    |depth|format|
|scale           |  S  |  S  |
|scale_by        |  S  |  S  |
|smoothscale     |  S  |  S  |
|smoothscale_by  |  S  |  S  |
|laplacian       |  S  |  A  |
|box_blur        |  S  |  S  |
|gaussian_blur   |  S  |  S  |
|invert          |  S  |  A  |
|grayscale       |  S  |  A  |
|average_surfaces|  SF |  A  |
```

MyreMylar (approved):
LGTM 👍 

The description is clear, and probably exposes a few places where we should tidy up how the functions work when passing destinations surface with non-matching pixel surfaces. I personally thing that we should be checking that the destination surface has a matching pixel format as I don't really want to add format conversion on top of what each function is doing - especially as passing a destination surface is supposed to be 'speeding up' the process.

docs/reST/ref/transform.rst
```diff
@@ -42,11 +42,9 @@ Instead, always begin with the original image and scale to the desired size.)
...
 
-   An optional destination surface can be used, rather than have it create a
-   new one. This is quicker if you want to repeatedly scale something. However
-   the destination must be the same size as the size (width, height) passed in. Also
-   the destination surface must be the same format.
-
+   An optional destination surface can be passed which is faster than creating a new
+   Surface. This destination surface must be the same as the size (width, height) passed
+   in, and the same depth and format as the source Surface.
    .. ## pygame.transform.scale ##
```
Starbuck5, line 48: There needs to be a space in between these two elements.

I tested it, without the space the paragraph doesn't generate at all:
Capture

Starbuck5:
Needs space added (as per comment), needs merge conflict resolution. 

Once those two small things are sorted this looks good for merge!

I'd also like a list of the functions that accept differing formats and then fail. Those should be changed. (Probably to just error instead)

Starbuck5 (approved):
Thanks!

PR #2064, opened 2023-03-27
Use PyMem_Free instead of PyMem_Del

Starbuck5:
Super simple PR, just moving PyMem_Del to the (equivalent) PyMem_Free.

Why? Because PyMem_Free is in the stable ABI, so we might as well use it instead of PyMem_Del. This change was suggested by https://github.com/python/pythoncapi-compat.

Docs backing up their equivalence: https://docs.python.org/3/c-api/memory.html#c.PyMem_Free


MyreMylar (approved):
LGTM 👍  We already use `PyMem_Free` in over 100 places to free memory in an identical fashion.

novialriptide (approved):
LGTM

PR #2067, opened 2023-03-28
Enforce dest_alpha->0 in opaque dest blitter

Starbuck5:
This fixes a blitter regression (😳🤕 AHHHHHHHH)
Reported by @bigwhoopgames on discord (Verified that this solution fixes it)

Regression between 2.1.4 and 2.2.0.

[initial_reproducer.zip](https://github.com/pygame-community/pygame-ce/files/11085251/initial_reproducer.zip) (takes color background behind text and turns it transparent for OP)

Minimal reproducer:
```py
import pygame

MAGENTA = (255, 0, 255, 255)

surf1 = pygame.Surface((8,1))
surf1.fill(MAGENTA)
print(surf1, surf1.get_shifts(), surf1.get_alpha(), surf1.get_colorkey())

surf2 = pygame.Surface((8,1), pygame.SRCALPHA)
surf2.set_at((0,0), MAGENTA)
print(surf2, surf2.get_shifts(), surf2.get_alpha(), surf2.get_colorkey())

surf1.blit(surf2, (0,0))
#print(surf1.get_at((0,0)))
print(surf1.get_buffer().raw[:4])
print(surf1.get_buffer().raw[3] == 0)

```

On 2.1.4 this prints True (and on this PR), on 2.2.0 it prints False. Why use the surface's raw bytes buffer instead of using `get_at`? Turns out, `get_at` returns A=255 on RGBX surfaces regardless of what the value of X is, which makes sense. So this is impossible to diagnose from get_at.

This issue pushed my debugging skills (and my sanity lol) because of that.

I still need to add a test case for this into the surface tests. Just putting this PR up for now to be quick.

### DIFF
--- a/src_c/simd_blitters_avx2.c
+++ b/src_c/simd_blitters_avx2.c
@@ -201,6 +201,11 @@ alphablit_alpha_avx2_argb_no_surf_alpha_opaque_dst(SDL_BlitInfo *info)
 
     __m256i src_alpha, temp;
 
+    /* The location of the destination's missing alpha channel can be masked
+     * out by composing all the other channels' masks. */
+    __m256i mask_out_alpha = _mm256_set1_epi32(
+        (info->dst->Rmask | info->dst->Gmask | info->dst->Bmask));
+
     /* Original 'Straight Alpha' blending equation:
         --------------------------------------------
         dstRGB = (srcRGB * srcA) + (dstRGB * (1-srcA))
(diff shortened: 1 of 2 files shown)


MyreMylar (approved):
LGTM 👍

Starbuck5:
Okay, this is ready now with a unit test.

I think this PR is low risk, because the SSE blitter is very explicit about setting alpha to zero. I just thought it wouldn't matter for an RGBX surface, and none of the existing tests caught this.

The question is how badly of a regression this is in 2.2 and how we should handle it --do we need to release a 2.2.1 within a few days with this fix, yank and rerelease 2.2?

MyreMylar:
> The question is how badly of a regression this is in 2.2 and how we should handle it --do we need to release a 2.2.1 within a few days with this fix, yank and rerelease 2.2? Wait until 2.3? Currently I'm in favor of trying to do a 2.2.1 patch with this and dependency updates by this weekend.


A 2.2.1 patch sounds like a good idea to me - just because the blitters are used all over the place.

ankith26 (approved):
Again, not super familiar with SIMD stuff but this fixes the regression and the unit test helps verify that. Thanks for the PR! :tada:

PR #2071, opened 2023-03-29
Create Surface.get_frect method

### LINKED ISSUE #2066: add get_frect as a method to surfaces
Since Pygame now has Rects as well as FRects it seems like a good idea to add a get_frect method. This would also create more consistency.

durkisneer1:
- Surface.get_frect()
- Implement FRect documentation

Fixes #2066


test/surface_test.py
Matiiss, line 1: Should add more tests for various keyword arguments to make sure those are set properly and test that the initial position is at (0, 0)

Matiiss (changes requested):

test/surface_test.py
```diff
@@ -569,6 +569,17 @@ def test_get_rect(self):
 
     ########################################################################
 
+    def test_get_frect(self):
+        """Ensure a surface's rect can be retrieved."""
+        size = (16, 16)
+        surf = pygame.Surface(size)
+        frect = surf.get_frect()
+
+        self.assertEqual(frect.size, size)
```
novialriptide, line 578: small nitpick: maybe check if the position is also `(0,0)`.

test/surface_test.py
```diff
@@ -569,6 +569,18 @@ def test_get_rect(self):
 
     ########################################################################
 
+    def test_get_frect(self):
+        """Ensure a surface's frect can be retrieved."""
+        size = (16, 16)
```
novialriptide, line 574: instead of the size being integers, could you make them into floats instead?

docs/reST/ref/surface.rst
```diff
@@ -709,6 +709,23 @@
...
+
+      | :sl:`get the rectangular area of the Surface`
+      | :sg:`get_frect(\**kwargs) -> FRect`
+      
+      This is the same as ``.get_rect()`` but returns an FRect. FRect is similar
+      to Rect, except it stores float values instead.
+
+      You can pass keyword argument values to this function. These named values
+      will be applied to the attributes of the FRect before it is returned. An
+      example would be ``mysurf.get_frect(center=(100, 100))`` to create a
```
VantaTree, line 722: make it: ``mysurf.get_frect(center=(100.0, 100.0))``  instead, might be better.

test/surface_test.py
```diff
@@ -569,6 +569,18 @@ def test_get_rect(self):
 
     ########################################################################
 
+    def test_get_frect(self):
+        """Ensure a surface's frect can be retrieved."""
+        size = (16.0, 16.0)
```
novialriptide, line 574: I meant something like `16.65` or something like that.

docs/reST/ref/surface.rst
```diff
@@ -709,6 +709,23 @@
...
+      
+      This is the same as ``.get_rect()`` but returns an FRect. FRect is similar
+      to Rect, except it stores float values instead.
+
+      You can pass keyword argument values to this function. These named values
+      will be applied to the attributes of the FRect before it is returned. An
+      example would be ``mysurf.get_frect(center=(100.0, 100.0))`` to create a
+      rectangle for the Surface centered at a given position.
+
+      :ref:`Surface.get_rect`
```
Starbuck5, line 725: This reference doesn't work, it generates a warning when I build the docs. It would need to be a `:meth:` tag instead of a ref tag, had to see what other pygame doc rst did.

Iirc, this was supposed to be a link in the text `.get_rect()` above.  So replace \`\`.get_rect()\`\` with :meth:\`Surface.get_rect\`

Starbuck5 (approved):

test/surface_test.py
Matiiss: Also test whether the correct exceptions are raised if, for example, `centre` is used instead of `center` or `aljdsfjdsa` instead of something sensible.

Test that using `center` actually sets the `center` of the rect properly.

Matiiss (approved):
LGTM, congrats on the first PR (otherwise, thanks for contributing again, much appreciated)

PR #2072, opened 2023-03-29
In examples/ import pygame as pg -> import pygame

### LINKED ISSUE #2033: Examples: use `import pygame` instead of `import pygame as pg`
This is one of the largest questions of our time.

What is the "preferred" way of importing pygame, what should we use in our example files?

Status quo: `import pygame as pg`
My preferred: `import pygame`

React with rocket for `import pygame as pg`, react with eyes for `import pygame`.

gresm:
Likely fixes #2033


Starbuck5 (approved):

Starbuck5:
I'm going to leave the issue open for removing `pg` from other places in the docs. Thanks for doing most of it.

PR #2074, opened 2023-03-30
Examples updates

Starbuck5:
When looking through #2072 to review it, I found several things I wanted to update in the example code.

I wanted to use the new object aliases, remove obsolete checks and things, use `custom_type()` instead of USEREVENT, and use booleans better.

If the different stories of changes get muddled in the overall diff, it may be easier to review each commit individually to see the changes broken up.


novialriptide (approved):
Makes pygame look more modern. I like it.

MyreMylar (approved):
LGTM 👍

Issue #2073, opened 2023-03-30
`mouse.get/set_relative_mode()` instead of `_sdl2.Window.relative_mouse`


yunline:
`_sdl2.Window.relative_mouse` is internally calling `SDL_GetRelativeMouseMode` and `SDL_SetRelativeMouseMode`, which are not a part of window api.

ankith26:
The linked PR fixes this issue

PR #2076, opened 2023-03-30
`mouse.get/set_relative_mode()` instead of `_sdl2.Window.relative_mouse`

yunline:
Issue: #2073


docs/reST/ref/mouse.rst
```diff
@@ -216,4 +216,31 @@ scroll, such as ``which`` (it will tell you what exact mouse device trigger the
...
+
+   .. ## pygame.mouse.get_relative_mode ##
+
+.. function:: set_relative_mode
+
+   | :sl:`set relative mouse mode`
+   | :sg:`set_relative_mode(enable) -> None`
+
+   Sets the relative mouse mode state.
+   SDL2 docs: *"While the mouse is in relative mode, the cursor is hidden,
```
zoldalma999, line 234: Instead of just quoting the sdl2 docs, you should probably write up something that explains it. Or at least remove the part that mentions sdl. Most people will probably not know what it is, and this would just confuse them.

docs/reST/ref/mouse.rst
```diff
@@ -216,4 +216,31 @@ scroll, such as ``which`` (it will tell you what exact mouse device trigger the
...
+   the mouse position is constrained to the window, and SDL will report
+   continuous relative mouse motion even if the mouse is at the edge of the
+   window.*
+
+   *This function will flush any pending mouse motion."*
+
+   Calling :func:`pygame.mouse.set_visible` with argument
+   ``True`` will exit relative mouse mode.
```
zoldalma999, line 242: The arg being named "enable" and passing in True to exit it seems weird. Probably naming it something like disable would be better

src_c/mouse.c
```diff
@@ -464,6 +464,25 @@ mouse_get_cursor(PyObject *self, PyObject *_null)
...
+    return PyBool_FromLong(SDL_GetRelativeMouseMode());
+}
+
+static PyObject *
+mouse_set_relative_mode(PyObject *self, PyObject *arg)
+{
+    SDL_bool mode = SDL_FALSE;
+    if (PyObject_IsTrue(arg)) {
+        mode = SDL_TRUE;
+    }
```
zoldalma999, line 479: This would probably be cleaner as 
```c
int mode = PyObject_IsTrue(arg);
if (mode == -1) {
    return NULL;
}
```
and cast mode if there is a compiler warning

docs/reST/ref/mouse.rst
yunline: But it says ‘Calling pygame.mouse.set_visible’, not calling this function.

docs/reST/ref/mouse.rst
yunline: Docs are copied from here https://github.com/pygame-community/pygame-ce/blob/e077dc7cd08be3b4f2946dc82aa603ab6573b0ef/docs/reST/ref/sdl2_video.rst?plain=1#L113-L128

zoldalma999: I know that you copied it from there, I saw it. But I also think that this would be the proper place to change it, since you are moving the implementation anyway, especially since _sdl2.video docs are not final, while these ones are supposed to be.

docs/reST/ref/mouse.rst
zoldalma999: Right, I misread that. Thanks for pointing it out!

docs/reST/ref/mouse.rst
yunline: I removed "SDL". Is that ok?

zoldalma999: Yep, thank you!

zoldalma999 (approved):
Looks good to me!

MyreMylar (changes requested):
OK, so there are a couple of things that come up because `SDL_SetRelativeMouseMode()` also hides the cursor and locks the mouse to the window.

As the event grab also locks the key events to the window I think it is fine that `pygame.event.get_grab()` doesn't return true while the mouse is in relative mode. 

However - `mouse.get_visible()` really should be false while we are in relative mode as the mouse cursor is not visible and this function does nothing else.

Anyway, it would be an easy fix, except these functions just got updated in the main branch for SDL3 so:

1. Please merge main into this PR.
2. If still needed (as the functions might have changed) please change line 218 of `mouse.get_visible()` to something like:
```
result = (PG_CursorVisible() || SDL_GetRelativeMouseMode());
``` 

so that `pygame.mouse.get_visible()` returns False while in relative mode.

Test program:

```python
import pygame


display_surf = pygame.display.set_mode((1280, 800))


pygame.init()

clock = pygame.time.Clock()

running = True

mouse_state_changed = False

while running:
        if event.type == pygame.QUIT:
            running = False

        if event.type == pygame.KEYDOWN and event.key == pygame.K_r:
            if pygame.mouse.get_relative_mode():
                pygame.mouse.set_relative_mode(False)
            else:
                pygame.mouse.set_relative_mode(True)
            mouse_state_changed = True

        if event.type == pygame.KEYDOWN and event.key == pygame.K_v:
            if pygame.mouse.get_visible():
                pygame.mouse.set_visible(False)
            else:
                pygame.mouse.set_visible(True)
            mouse_state_changed = True

        if event.type == pygame.KEYDOWN and event.key == pygame.K_g:
            if pygame.event.get_grab():
                pygame.event.set_grab(False)
            else:
                pygame.event.set_grab(True)
            mouse_state_changed = True

    if mouse_state_changed:
        print("mouse relative mode:", pygame.mouse.get_relative_mode())
        print("mouse visible:", pygame.mouse.get_visible())
        print("events grabbed:", pygame.event.get_grab(), "\n")
        mouse_state_changed = False

    pygame.display.flip()

    clock.tick(60)
```

src_c/mouse.c
```diff
@@ -215,7 +215,7 @@ mouse_get_visible(PyObject *self, PyObject *_null)
 
     VIDEO_INIT_CHECK();
 
-    result = PG_CursorVisible();
+    result = (PG_CursorVisible() || SDL_GetRelativeMouseMode());
```
MyreMylar, line 218:
```suggestion
    result = (PG_CursorVisible() && !SDL_GetRelativeMouseMode());
```

MyreMylar (approved):

test/mouse_test.py
```diff
@@ -340,6 +340,23 @@ def test_set_visible__invalid_value(self):
         for invalid_value in ((1,), [1, 2, 3], 1.1, "1", (1, "1"), []):
             with self.assertRaises(TypeError):
                 prev_visible = pygame.mouse.set_visible(invalid_value)
+    
+    def test_set_relative_mode(self):
+        """Tests that set_relative_mode hides the cursor."""
+        pygame.mouse.set_visible(True)
+        pygame.mouse.set_relative_mode(True)  # sets the mouse invisible
+        visible = pygame.mouse.get_visible()
```
MyreMylar, line 348:
```suggestion

    def test_set_relative_mode(self):
        """Tests that set_relative_mode hides the cursor."""
        pygame.mouse.set_visible(True)
        pygame.mouse.set_relative_mode(True)  # sets the mouse invisible
        visible = pygame.mouse.get_visible()
```

test/mouse_test.py
```diff
@@ -341,6 +341,23 @@ def test_set_visible__invalid_value(self):
...
+        pygame.mouse.set_relative_mode(False)  # sets the mouse visible
+        visible = pygame.mouse.get_visible()
+        self.assertEqual(visible, True)
+
+    def test_get_relative_mode(self):
+        """Tests that get_relative_mode correctly reports the relative mode"""
+        pygame.mouse.set_relative_mode(True)
+        self.assertEqual(pygame.mouse.get_relative_mode(), True)
+        pygame.mouse.set_relative_mode(False)
+        self.assertEqual(pygame.mouse.get_relative_mode(), False)
```
MyreMylar, line 359:
```suggestion
    @unittest.skipIf(
        os.environ.get("SDL_VIDEODRIVER", "") == "dummy",
        "mouse.set_relative_mode requires non-null video driver",
    )
    def test_set_relative_mode(self):
        """Tests that set_relative_mode hides the cursor."""
        pygame.mouse.set_visible(True)
        pygame.mouse.set_relative_mode(True)  # sets the mouse invisible
        visible = pygame.mouse.get_visible()
        self.assertEqual(visible, False)
        pygame.mouse.set_relative_mode(False)  # sets the mouse visible
        visible = pygame.mouse.get_visible()
        self.assertEqual(visible, True)

    @unittest.skipIf(
        os.environ.get("SDL_VIDEODRIVER", "") == "dummy",
        "mouse.set_relative_mode requires non-null video driver",
    )
    def test_get_relative_mode(self):
        """Tests that get_relative_mode correctly reports the relative mode"""
        pygame.mouse.set_relative_mode(True)
        self.assertEqual(pygame.mouse.get_relative_mode(), True)
        pygame.mouse.set_relative_mode(False)
        self.assertEqual(pygame.mouse.get_relative_mode(), False)
```

docs/reST/ref/mouse.rst
```diff
@@ -216,4 +216,31 @@ scroll, such as ``which`` (it will tell you what exact mouse device trigger the
...
+   the mouse position is constrained to the window, and pygame
+   will report continuous relative mouse motion even if the 
+   mouse is at the edge of the window.
+
+   *This function will flush any pending mouse motion."*
+
+   Calling :func:`pygame.mouse.set_visible` with argument
+   ``True`` will exit relative mouse mode.
+
+   .. ## pygame.mouse.set_relative_mode ##
```
ankith26, line 244: I know I'm late and the PR has been merged already, but this is missing version added tags. Can open a new PR to fix this

MyreMylar: https://github.com/pygame-community/pygame-ce/pull/2573

Issue #2077, opened 2023-03-30
`SDL_WINDOW_ALWAYS_ON_TOP` is always 0


yunline:
Marco `SDL_WINDOW_ALWAYS_ON_TOP` is defined as 0 instead of the real value.
![image](https://user-images.githubusercontent.com/31395137/228822853-99b5e253-cf88-4703-94b1-5477ed6d13bc.png)

The macro is defined here ↓
https://github.com/pygame-community/pygame-ce/blob/fdad67ad83c0ff6b8937b7bf228c90993a6bd888/src_c/include/pgcompat.h#L82-L84
However the real value of SDL_WINDOW_ALWAYS_ON_TOP is 0x00008000
![image](https://user-images.githubusercontent.com/31395137/228821755-75da2e28-ddd2-4af8-9f75-d21776077d83.png)

It seems this macro is for backward compatibility, but it does't work well on SDL 2.26.4.

yunline:
I see, SDL_WINDOW_ALWAYS_ON_TOP is enum, which can't be detected by `#ifdef`.

yunline:
These marcos are created for the backward compatibility for sdl2.0.4 and sdl2.0.2

See https://github.com/pygame-community/pygame-ce/blame/main/src_c/include/pgcompat.h#L34-L105

Time to remove them.

PR #2078, opened 2023-03-30
Remove out-dated macros

yunline:
Issue: https://github.com/pygame-community/pygame-ce/issues/2077


MyreMylar (approved):
LGTM 👍 

You can see these flags here: https://wiki.libsdl.org/SDL2/SDL_WindowFlags

And all should be present in our minimum SDL version, I would expect these defines to do nothing at all so it seems fine to remove them.

src_c/include/pgcompat.h
```diff
@@ -29,79 +29,7 @@ typedef uint8_t Uint8;
...
-#define SDL_WINDOW_TOOLTIP 0
-#endif
-
-#ifndef SDL_WINDOW_POPUP_MENU
-#define SDL_WINDOW_POPUP_MENU 0
-#endif
-
-#ifndef SDL_MOUSEWHEEL_FLIPPED
-#define NO_SDL_MOUSEWHEEL_FLIPPED
-#endif
```
Starbuck5, line 104: Unlike the window flags one, this may still be used.

Starbuck5:
I guess the `#ifndef` doesn't work because they are not macros, they're enum values? So the defines are then shadowing the real values.

Added a note about one we may want to keep.

Just fyi it's spelled "macros."

src_c/include/pgcompat.h
yunline: Solved

PR #2079, opened 2023-03-30
fix null bug on X11 but unset $XDG_SESSION_TYPE

robertpfeiffer:
Even on the console, $XDG_SESSION_TYPE is set on my machine (to "tty"), but I guess not all distros use freedesktop stuff and that's ok


novialriptide (approved):

MyreMylar (approved):
LGTM 👍

PR #2081, opened 2023-03-30
Add support for DPI aware applications

Temmie3754:
Adds a "DPIAWARE" flag to facilitate the creation of windows that do not have automatic DPI scaling applied to them by the OS.


MyreMylar:
What does this do that setting the `SDL_WINDOWS_DPI_SCALING` hint doesn't?

```
import pygame
import os

os.environ["SDL_WINDOWS_DPI_SCALING"] = "1"

pygame.init()

desktop_size = pygame.display.get_desktop_sizes()[0]
print("window size =", desktop_size)
display_surf = pygame.display.set_mode(desktop_size)

print("window size =", display_surf.get_size())
running = True
while running:
        if event.type == pygame.QUIT:
            running = False

    pygame.display.flip()
```

I'm curious, since this High DPI stuff is often confusing ([see this long thread on what SDL 3](https://github.com/libsdl-org/SDL/issues/7134) is going to about it).

Temmie3754:
SDL_WINDOWS_DPI_SCALING doesn't exist in the current version of SDL (I believe something like that will be coming with the 3.0 release) so the code you provided doesn't work, and since the current SDL_WINDOW_ALLOW_HIGHDPI flag doesn't work on windows, this makes the adjustments so that it does and can be set through pygame.

Starbuck5:
Always cool to see people experimenting with these things, good job.

I'm skeptical with the prospect of adding anything to make our DPI handling better before SDL3, because it seems like SDL3 has a big DPI rework. So anything we add now in that area will be a headache to maintain over the SDL3 transition.

For example SDL_WINDOW_ALLOW_HIGHDPI doesn't exist in SDL3. See the migration guide if you're curious for a bit of info: https://github.com/libsdl-org/SDL/blob/main/docs/README-migration.md. Maybe SDL will eventually release some sort of "DPI and coordinates and pixels and logical size" guide for SDL3 (once SDL3 is out), that could be very handy.

MyreMylar:
> SDL_WINDOWS_DPI_SCALING

This hint was actually added in SDL 2.24.0 released last august - see:

https://github.com/libsdl-org/SDL/releases/tag/release-2.24.0

```
Added the hint SDL_HINT_WINDOWS_DPI_AWARENESS to set whether the application is DPI-aware. This hint must be set before initializing the video subsystem
Added the hint SDL_HINT_WINDOWS_DPI_SCALING to control whether the SDL coordinates are in DPI-scaled points or pixels
```

My windows build of pygame-ce is currently one behind current SDL on 2.26.3 and I've not done anything special to obtain it so I assume the windows wheels are also that up to date.

These are windows only hints, but I expect they probably cover what you are doing in this block here:

```
#ifdef WIN32
            int (*SetProcessDPIAware)();
            SetProcessDPIAware = (void *)GetProcAddress(
                LoadLibrary("user32.dll"), "SetProcessDPIAware");
            if (SetProcessDPIAware)
                SetProcessDPIAware();
#endif
```

Temmie3754:
Huh weird, it didn't work for me when I tried it a few days ago. I'll update it to use that hint instead.

Temmie3754:
Oh it appears that setting SDL_WINDOWS_DPI_SCALING doesn't actually work even in 2.26.4, this is more obvious when you set the window to be in fullscreen and you will see that display_surf.get_size() will return a scaled value even though SDL_WINDOWS_DPI_SCALING has been set.

MyreMylar:
So to summarise, if I set my desktop scaling to 150% and run this:

```python
import pygame
import os

os.environ["SDL_WINDOWS_DPI_SCALING"] = "0"

pygame.init()

desktop_size = pygame.display.get_desktop_sizes()[0]
display_surf = pygame.display.set_mode(desktop_size)
print("display surf size =", display_surf.get_size())

running = True
while running:
        if event.type == pygame.QUIT:
            running = False

    pygame.display.flip()
```

I get:

```
display surf size = (1280, 720)
```

If I toggle `SDL_WINDOWS_DPI_SCALING` to "1" and rerun I get:

```
display surf size = (1920, 1080)
```

But if we add the FULLSCREEN flag then the display surface returned is always scaled to the smaller, scaled size. I wonder where pygame is getting that size from, as it is clearly not the size from desktop_sizes which gets scaled correctly based on the SDL_WINDOWS_DPI_SCALING flag.

Temmie3754:
Yes this occurs since SDL2 is yet to have proper support for dpi aware applications (at least on windows). When you set SDL_WINDOWS_DPI_SCALING to "1" while it will make get_desktop_sizes return the larger size, if you check, the scaling for the window will still be applied even when not in fullscreen (in which case it will extend beyond the borders of the monitor). When it is in fullscreen since the dpi scaling doesn't work it gets the scaled display size which in your case is (1280, 720).

```python
import pygame

pygame.init()

desktop_size = pygame.display.get_desktop_sizes()[0]
display_surf = pygame.display.set_mode(desktop_size, pygame.FULLSCREEN | pygame.DPIAWARE)
print("display surf size =", display_surf.get_size())

running = True
while running:
        if event.type == pygame.QUIT:
            running = False

    pygame.display.flip()
```

Running this code with the PR will show that the DPIAWARE flag correctly makes the window dpi aware and thus it will return the correct fullscreen size. 

An issue I have encountered with this is that get_desktop_sizes doesn't account for the pygame.DPIAWARE flag since it runs beforehand and thus will provide a smaller size than the available screen space even if the DPIAWARE flag is set in the following line, leading to non-fullscreen applications having a smaller window than anticipated. 
To remedy this I see two solutions:
1. add a DPIAWARE argument to get_desktop_sizes so it will scale the size correctly if dpi awareness is wanted for this function.
2. move the setting of dpiawareness to some other new function such as pygame.setdpiaware, etc. It could also be moved into the pygame.init function but adding arguments to that doesn't seem like a great choice.

MyreMylar:
Anyway after messing about with this for a bit I figured out that the `SDL_WINDOWS_DPI_SCALING` hint appears to be a bit of a blind alley. At least I couldn't get it to to work properly on it's own despite the hint description. It does seem to control the scaling on/off once you are in a DPI aware mode. Anyway...

I found it a lot easier to work the hint `SDL_WINDOWS_DPI_AWARENESS` which has two pretty useful states you can set in pygame like the below:

os.environ["SDL_WINDOWS_DPI_AWARENESS"] = "permonitorv2"
# os.environ["SDL_WINDOWS_DPI_AWARENESS"] = "unaware"

the top one `"permonitorv2"` disables using the scaling factor from windows display settings so you can have a scaled desktop and a game window that matches the monitor resolution, or is exactly the specified size in physical pixels. The other option "unaware" is the default where windows and surfaces are scaled up in physical pixels and all "pixels" used in pygame are actually scaled up by the desktop scale factor.

Does this hint solve your original issue @Temmie3754 ? If so I think we should just document it in the pygame environment variable docs.

MyreMylar:
My little DPI awareness hints test program:

```python
import os
import pygame

os.environ["SDL_WINDOWS_DPI_AWARENESS"] = "permonitorv2"  # turn on DPI awareness (pygame apps respect requested pixel resolution)
# os.environ["SDL_WINDOWS_DPI_AWARENESS"] = "unaware" # explicitly turn off DPI awareness (pygame apps scale with desktop)
# os.environ["SDL_WINDOWS_DPI_SCALING"] = "1" # with DPI awareness enabled this flag controls whether we scale with the desktop anyway

pygame.init()

desktop_size = (800, 600)  # pygame.display.get_desktop_sizes()[0] # switch this over and enable FULLSCREEN to test both modes
print("window request size =", desktop_size)
display_surf = pygame.display.set_mode(desktop_size)

chess_board = pygame.Surface(desktop_size)
chess_board.fill(pygame.Color("black"))
chess_board.fill(pygame.Color("white"), pygame.Rect(0, 0, 100, 100))
chess_board.fill(pygame.Color("white"), pygame.Rect(200, 0, 100, 100))
chess_board.fill(pygame.Color("white"), pygame.Rect(400, 0, 100, 100))
chess_board.fill(pygame.Color("white"), pygame.Rect(600, 0, 100, 100))

chess_board.fill(pygame.Color("white"), pygame.Rect(100, 100, 100, 100))
chess_board.fill(pygame.Color("white"), pygame.Rect(300, 100, 100, 100))
chess_board.fill(pygame.Color("white"), pygame.Rect(500, 100, 100, 100))
chess_board.fill(pygame.Color("white"), pygame.Rect(700, 100, 100, 100))

chess_board.fill(pygame.Color("white"), pygame.Rect(0, 200, 100, 100))
chess_board.fill(pygame.Color("white"), pygame.Rect(200, 200, 100, 100))
chess_board.fill(pygame.Color("white"), pygame.Rect(400, 200, 100, 100))
chess_board.fill(pygame.Color("white"), pygame.Rect(600, 200, 100, 100))

chess_board.fill(pygame.Color("white"), pygame.Rect(100, 300, 100, 100))
chess_board.fill(pygame.Color("white"), pygame.Rect(300, 300, 100, 100))
chess_board.fill(pygame.Color("white"), pygame.Rect(500, 300, 100, 100))
chess_board.fill(pygame.Color("white"), pygame.Rect(700, 300, 100, 100))

chess_board.fill(pygame.Color("white"), pygame.Rect(0, 400, 100, 100))
chess_board.fill(pygame.Color("white"), pygame.Rect(200, 400, 100, 100))
chess_board.fill(pygame.Color("white"), pygame.Rect(400, 400, 100, 100))
chess_board.fill(pygame.Color("white"), pygame.Rect(600, 400, 100, 100))

chess_board.fill(pygame.Color("white"), pygame.Rect(100, 500, 100, 100))
chess_board.fill(pygame.Color("white"), pygame.Rect(300, 500, 100, 100))
chess_board.fill(pygame.Color("white"), pygame.Rect(500, 500, 100, 100))
chess_board.fill(pygame.Color("white"), pygame.Rect(700, 500, 100, 100))

print("display surf size =", display_surf.get_size())
running = True
while running:
        if event.type == pygame.QUIT:
            running = False
            running = False
            print(event.pos)

    display_surf.blit(chess_board, (0, 0))

    pygame.display.flip()

pygame.quit()

```

Temmie3754:
Ah yes thanks that works and solves the issue from what I can tell, seeing as a docs change will be done rather than an underlying code change I'll close this PR.

PR #2086, opened 2023-03-31
Change renderer methods to use colors and remove draw_color

Temmie3754:
Changes the renderer draw methods to accept a color in their input with a default white colour when it isn't provided (and black in the case of clear()), relies on #2039 due to overlap.


yunline:
I think add a `set_default_color()` may be better.
It can prevent calling `SDL_SetRenderDrawColor` in every draw
like:
```python
renderer.set_default_color((127,255,127,255))
for i in range(99999):
    renderer.draw_rect( ... ) # draw with default color

renderer.draw_rect( ... ,color=(255,0,0,255)) # draw with specified color
```

Temmie3754:
I'm not sure that calling SDL_SetRenderDrawColor every draw is an issue, using a set_default_color() would require calling SDL_GetRenderDrawColor to see if it has changed and then calling SDL_SetRenderDrawColor if it has, with extra logic for comparison which seems less efficient and doesn't meet the goal of making the renderer draw methods behave like the normal pygame ones.
SDL_SetRenderDrawColor doesn't seem to have any significant overhead when compared to SDL_GetRenderDrawColor as can be seen in SDL here: https://github.com/libsdl-org/SDL/blob/b6ae281e97c4e4680a9010e7e13fe1222c0bcd4b/src/render/SDL_render.c#L2579

Temmie3754:
Alternatively it could accept a color input of "None" to mean that it uses the last colour assigned and thus doesn't update the renderer colour or check the renderer colour either. Could be a good middle ground?

yunline:
> I'm not sure that calling SDL_SetRenderDrawColor every draw is an issue, using a set_default_color() would require calling SDL_GetRenderDrawColor every draw instead to see if the current renderer colour is the same as the default color and then calling SDL_SetRenderDrawColor as well if it isn't, with extra logic for comparison which seems less efficient and doesn't meet the goal of making the renderer draw methods behave like the normal pygame ones.

Oh, my bad. I forget that default color shouldn't change after draw.

> Alternatively it could accept a color input of "None" to mean that it uses the last colour assigned and thus doesn't update the renderer colour or check the renderer colour either. Could be a good middle ground?

Yes, that's a good idea.

MyreMylar (commented):
I'm generally in favour of this API switch, I'm also not convinced that this will hurt draw performance that much, but it is tweaking my old instincts around draw call batching so why not be cautious?

How about we phase this in slowly, leave the existing `draw_color()` function as it was - just setting a draw colour on the renderer (that way we won't break a bunch of folks code). Then change the draw functions `color` arguments you've just added so the default argument is `None` instead of white. That way old code will still work and so will the new API you propose - of just setting the draw colour when you call a shape drawing function.

This would also make it easy to test them out side by side with some performance testing (i.e. draw green 10,000 rects with both styles of API) to see if 10,000 set draw colour calls to the same colour actually does anything to the performance.

MyreMylar (changes requested):
See comment above. Also, this has built up lots of conflicts in the intervening time and could no doubt use a merge with main.

Starbuck5:
This PR is not ready for merge, so I'm marking it as a draft.

zoldalma999:
This PR has been stale for 2 years now, with a lot of conflicts and Temmie (the author) has been inactive for quite a bit now, so I'll close it.

PR #2092, opened 2023-04-02
Rewrite pygame._sdl2.video docs, tweak some function signatures

Mega-JC:
This commit effectively rewrites all the documentation of the `pygame._sdl2.video` module, by either adding new information for undocumented content, or extracting information from existing docstrings in the codebase. Along with the documentation changes, some function signatures were modified to adhere to the snake case naming convention described in PEP 8 and for more consistency.


Starbuck5:
I would like a copy of this PR against main. It would be nice to update the _sdl2.video docs independently of any Window API work.

Issue #2084, opened 2023-03-31
Update example code that uses scrap


Starbuck5:
At least two files in the example code use the pygame scrap API for clipboard support. @oddbookworm recently created a new scrap API to supersede the old one, since the old one had many problems.

Code in examples/ should be updated to use the new API.

Issue #2050, opened 2023-03-24
Improve OpenCV camera backend


Starbuck5:
I believe there a few problems with it now.

Problems: 
- OpenCV needs an explicit backend passed in now?
- OpenCV can report webcam FPS as 0, leads to a divide by zero in my code.

I'd like someone to investigate these and verify functionality. You can select the opencv backend by passing "opencv" into pygame.camera.init's backend argument.

PurityLake:
After some testing with the `opencv-python`.

1.
Not much has changed really as regards initialisation so I don't think a backend *needs* to be passed in. We solely rely on `cv2.VideoCapture` and from what I am seeing it is possible to create a `cv2.VideoCapture` with and *optional* API preference. So when we use `cv2.VideoCapture(0)`, it is equivalent to using `cv2.VideoCapture(0, cv2.CAP_ANY)`.  

In most cases the default settings work for most people *but* I have read in certain places that the API can matter for certain machines. The last piece of information there I have not been able to confirm on my end, pretty much adding any supported and available API caused no difference in the capture rate of my webcam. Although I suppose it would be silly to assume for that for all webcams.

2.
The framerate issues seems to be mostly due to incorrect configurations of opencv although it may be worth adding a check for that with a specific warning or error thrown letting the user know the issue rather than leaving a not so informative `DivideByZeroError`.

PR #2097, opened 2023-04-04
Keyword implementation for font.render

narilee2006:
Fixes: https://github.com/pygame-community/pygame-ce/issues/1990

### DIFF
--- a/src_c/font.c
+++ b/src_c/font.c
@@ -475,7 +475,7 @@ font_set_strikethrough(PyObject *self, PyObject *arg)
 }
 
 static PyObject *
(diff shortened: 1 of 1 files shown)


Starbuck5:
There is another PR in progress for this issue, #2000. Progress on that has stalled recently, but that PR still gets priority on this issue since it's been worked on already. So I'm going to close this PR for now.

But thanks for working this up! I'm always impressed when someone new steps into the C code, since it's a difficult thing to get set up with developing.

If you're looking for an issue to tackle, #2050 and #2084 opened by me are both pretty actionable, in my opinion.

PR #2099, opened 2023-04-05
New title for the unittests README.

narilee2006:
Does not actually exist run_tests.py...

### DIFF
--- a/test/README.rst
+++ b/test/README.rst
@@ -1,5 +1,5 @@
-run_tests.py
-************
+Pygame Unit Tests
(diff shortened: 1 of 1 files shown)


MyreMylar:
> Does not actually exist run_tests.py...

It does exist under `test/test_utils`. However I agree that a particular filename is not a great title for a Readme which is really discussing the entire test module.

MyreMylar (approved):

novialriptide (approved):

PR #2101, opened 2023-04-07
Deprecate pygame.image.tostring and fromstring

### LINKED ISSUE #2085: Deprecate `image.tostring`, `image.fromstring`
It's been a while since we introduced `tobytes` and `frombytes` as the better named equivalents of these functions, and I've noticed people in the wild actually using the new names, so it's probably time to deprecate the old ones:

- [ ] Make functions generate a deprecation warning when called
- [ ] Mark functions deprecated in the docs

SSS-Says-Snek:
Fixes #2085

Since `to/fromstring`'s gonna be deprecated, I also replaced usage of them in tests to `to/frombytes`
Finally, like I mentioned on the discord, because to/frombytes and to/fromstring's docs were the exact same and made it hard to follow, I replaced it with a redirection to `to/frombytes`

### DIFF
--- a/src_c/image.c
+++ b/src_c/image.c
@@ -1230,12 +1230,7 @@ image_tostring(PyObject *self, PyObject *arg, PyObject *kwarg)
         return NULL;
     }
 
-    PyObject *result = image_tobytes(self, arg, kwarg);
-    if (result == NULL) {
-        return NULL;
-    }
-
-    return result;
+    return image_tobytes(self, arg, kwarg);
 }
 
 PyObject *
@@ -1247,12 +1242,7 @@ image_fromstring(PyObject *self, PyObject *arg, PyObject *kwarg)
         return NULL;
     }
 
-    PyObject *result = image_frombytes(self, arg, kwarg);
-    if (result == NULL) {
-        return NULL;
-    }
-
-    return result;
+    return image_frombytes(self, arg, kwarg);
 }
 
 static int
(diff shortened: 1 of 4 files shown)


Starbuck5:
Thanks for working on this!

My only review notes is for the docs:
- The functions with removed documentation should at least say something like "This is the same functionality as frombytes"
- The reference to the new function should be a sphinx link to the new function. For example, ":func:\`pygame.image.load()`" is a link to the function.

In case you haven't built the docs before, do `python setup.py docs` (with sphinx installed), and then access them with `python -m docs`

SSS-Says-Snek:
Alright, I'll fix it when I get back

SSS-Says-Snek:
Soo I won't be back for 6 more days💀

novialriptide:
🚗

Starbuck5 (approved):
Looks good to me!

src_c/image.c
```diff
@@ -1221,6 +1221,40 @@ image_frombytes(PyObject *self, PyObject *arg, PyObject *kwds)
...
+                     1) == -1) {
+        return NULL;
+    }
+
+    PyObject *result = image_frombytes(self, arg, kwarg);
+    if (result == NULL) {
+        return NULL;
+    }
+
+    return result;
```
ankith26, line 1255:
```suggestion
    return image_frombytes(self, arg, kwarg);
```
super minor nitpick. You can also do the same for the above function

ankith26 (approved):

Left a minor nitpick-ey review, resolve at will

PR #2103, opened 2023-04-07
Blur fix

### LINKED ISSUE #2102: `transform.box_blur` works incorrectly.
**Test Code**
```python
import pygame
import sys

sf=pygame.image.load("peppers3.tif")
pygame.display.init()

while 1:
    dsf.fill((0,150,0))

    dsf.blit(sf, (640,0,0,0))

    pygame.transform.box_blur(dsf,10,dest_surface=dsf)

    dsf.blit(sf,(0,0,1,1))
    pygame.display.update()
        if event.type==pygame.QUIT:
            sys.exit()

```
image

yunline:
fix #2102
image

### DIFF
--- a/src_c/transform.c
+++ b/src_c/transform.c
@@ -2959,9 +2959,10 @@ box_blur(SDL_Surface *src, SDL_Surface *dst, int radius, SDL_bool repeat)
     // Reference : https://blog.csdn.net/blogshinelee/article/details/80997324
 
     int inplace = 0;
+    void *tmp_pixels;
     if (dst == src) {
         inplace = 1;
-        void *tmp_pixels = malloc((src->h) * (src->pitch));
+        tmp_pixels = malloc((src->h) * (src->pitch));
         memcpy(tmp_pixels, src->pixels, (src->h) * (src->pitch));
         dst = SDL_CreateRGBSurfaceWithFormatFrom(
             tmp_pixels, src->w, src->h, src->format->BitsPerPixel, src->pitch,
@@ -3044,6 +3045,7 @@ box_blur(SDL_Surface *src, SDL_Surface *dst, int radius, SDL_bool repeat)
     if (inplace) {
         memcpy(srcpx, dstpx, h * dst_pitch);
         SDL_FreeSurface(dst);
+        free(tmp_pixels);
     }
     free(buf);
     free(sum_v);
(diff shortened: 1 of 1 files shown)


oddbookworm (approved):
LGTM! Fixes the linked issue nicely

src_c/transform.c
```diff
@@ -3031,7 +3037,13 @@ box_blur(SDL_Surface *src, SDL_Surface *dst, int radius, SDL_bool repeat)
             }
         }
     }
-
+    if (inplace) {
+        for (i = 0; i < src->h; i++) {
+            memcpy(srcpx + i * src_pitch, dstpx + i * dst_pitch,
+                   sizeof(Uint8) * MIN(dst_pitch, src_pitch));
+        }
```
ankith26, line 3044: Hmmm, instead of manual surface copying should we use something like `SDL_BlitSurface` instead?

yunline: The Surface `src` is locked and `SDL_BlitSurface` returns an error. I think copy manually may be better (?

Starbuck5: I wonder if surface pitches could be negative?

The sizeof(Uint8) is unnecessary.

If you used SDL_CreateRGBSurfaceWithFormatFrom in the setup you could make sure the new dst has the same pitch as the source. (You'd need to malloc the pixels yourself though, `src->BytesPerPixel * src->h  * src->pitch` worth. 
The advantage of making sure the pitch matches is that you could do a single malloc to copy the whole pixel region into the src/dest surface.

yunline:
Solved another problem when use gaussian_blur in place.
I think it can be a part of this PR.
```py
import pygame
import sys

sf=pygame.image.load("peppers3.tif")
pygame.display.init()
ddsf=dsf.subsurface((720,100,200,300))

while 1:
    dsf.fill((0,150,0))

    dsf.blit(sf, (640,0,0,0))

    pygame.transform.gaussian_blur(ddsf,50,dest_surface=ddsf)

    #dsf.blit(sf,(0,0,1,1))
    pygame.display.update()
        if event.type==pygame.QUIT:
            sys.exit()
```
**Before fix**
image

**After fix**
image

yunline:
It might be better to raise an error instead of creating a buffer when handling in-place operation.
People might think in-place operation is faster, but it's actually slower.

Starbuck5:
>  It might be better to raise an error instead of creating a buffer when handling in-place operation.
People might think in-place operation is faster, but it's actually slower.

So an error if the source and destination surfaces are the same? I'm totally okay with that.

yunline:
> So an error if the source and destination surfaces are the same? I'm totally okay with that.

OK, I'll open another PR for this

PR #2108, opened 2023-04-09
Start move away from format masks (prep for SDL3)

Starbuck5:
The purpose of this PR is to start moving away from format masks. In SDL3, `SDL_CreateRGBSurface` and `SDL_CreateRGBSurfaceWithFormat` are both replaced with `SDL_CreateSurface`. `SDL_CreateSurface` takes SDL pixelformat enums instead of {R, G, B, A}masks.

So this is a PR to make it slightly easier to port to SDL3 when that time comes. I think small PRs like this one are a good way of eventually getting to a working SDL3 integration, without huge unreviewable PRs.

The process of reasoning about replacing masks with pixelformat enum members is greatly complicated by having to think about endianness. I did not guess, nor did I try to think each through on my own. To translate the masks into pixelformats, I checked the source of `SDL_GetPixelFormatEnumForMasks` (SDL3's version of SDL_MasksToPixelFormatEnum), as well as checking the pixelformat definitions in SDL_pixels.h.

This should keep the exact same behavior before and after this PR, little endian or big endian.

I also removed some obsolete arguments. `SDL_CreateRGBSurface` and `SDL_ConvertSurface` both have a `flags` argument inherited from SDL1, which is ignored in SDL2. I set that flag to 0 in places it wasn't zero. I also replaced a couple uses of `SDL_SRCALPHA`, an SDL1 constant removed in SDL2-- defined in a pygame header as `0`.

### DIFF
--- a/src_c/display.c
+++ b/src_c/display.c
@@ -702,7 +702,7 @@ pg_ResizeEventWatch(void *userdata, SDL_Event *event)
                 pgSurfaceObject *display_surface =
                     pg_GetDefaultWindowSurface();
                 SDL_Surface *surf = SDL_CreateRGBSurface(
-                    SDL_SWSURFACE, w, h, 32, 0xff << 16, 0xff << 8, 0xff, 0);
+                    0, w, h, 32, 0xff << 16, 0xff << 8, 0xff, 0);
 
                 SDL_FreeSurface(display_surface->surf);
                 display_surface->surf = surf;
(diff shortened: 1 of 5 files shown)


ankith26 (approved):

Considering we have a big endian CI running RN, I have more confidence approving this

Starbuck5:
> Considering we have a big endian CI running RN, I have more confidence approving this

Forgot about that. This makes me more confident as well.

PR #2110, opened 2023-04-10
Rect and FRect swizzling

ScriptLineStudios:
Swizzling for Rects and FRects similar to Vector and Color swizzling


test/rect_test.py
```diff
@@ -2691,6 +2728,50 @@ def test_contains(self):
...
+        self.assertAlmostEqual(res[0], 10.34, 5)
+        self.assertAlmostEqual(res[1], 10.34, 5)
+
+        res = c.xyw
+        self.assertAlmostEqual(res[0], 10.34, 5)
+        self.assertAlmostEqual(res[1], 20.31, 5)
+        self.assertAlmostEqual(res[2], 30.39, 5)
+
+        with self.assertRaises(AttributeError):
+            c.rr = (10.3421784, 10.3421784)
```
oddbookworm, line 2752: Another copy/paste error?

test/rect_test.py
```diff
@@ -2334,6 +2334,43 @@ def test_collideobjectsall_call_variants(self):
...
+        c.xywh = c.hwyx
+        self.assertEqual(c.xywh, (40, 30, 20, 10))
+
+        with self.assertRaises(AttributeError):
+            c.rr = (10, 10)
+            c.bb = (10, 10)
+            c.gg = (10, 10)
+            c.aa = (10, 10)
+            c.aaa = (10, 10, 100)
+            c.rgb = (256, 256, 256)
```
oddbookworm, line 2355: Copy/Paste error? `c` is a `Rect`, so it doesn't have these properties in the first place

docs/reST/ref/rect.rst
```diff
@@ -32,6 +32,13 @@
    None but affects the original Rect. These "in-place" methods are denoted
    with the "ip" suffix.
 
+   Rect objects support swizzling for their ``XYWH`` attributes, which allows
+   the creation of new color objects with the corresponding swizzled attributes
```
oddbookworm, line 36: copy/paste error? "color"

src_c/rect.c
```diff
@@ -438,6 +442,402 @@ four_floats_from_obj(PyObject *obj, float *val1, float *val2, float *val3,
...
+    attr = PyUnicode_AsUTF8AndSize(attr_unicode, &len);
+    if (attr == NULL)
+        goto internal_error;
+    /* If we are not a swizzle, go straight to GenericGetAttr. */
+    if ((attr[0] != 'x') && (attr[0] != 'y') && (attr[0] != 'w') &&
+        (attr[0] != 'h')) {
+        goto swizzle_failed;
+    }
+
+    if (len == 4) {
```
oddbookworm, line 474: Like with swizzling colors, I'm not a fan of having a different return type in special cases. I think it should just return a tuple even when 4 values are requested

oddbookworm (changes requested):

src_c/rect.c
Matiiss: For `Vector3` swizzling two values returns a `Vector2` though and three values return a `Vector3`, also despite them being different types they can both be iterated over, the only loss here is immutability (but Vectors have it very similar).

test/rect_test.py
```diff
@@ -2334,6 +2334,36 @@ def test_collideobjectsall_call_variants(self):
         r.collideobjectsall(objects, key=lambda o: o.rect1)
         self.assertRaises(TypeError, r.collideobjectsall, objects)
 
+    def test_swizzle_get(self):
```
Starbuck5, line 2337: No test that a 4 element swizzle returns a Rect?

Starbuck5:
I'm assuming you got most of the implementation by checking the vector code, that should be good. Before adding this, I'd like to hear a use case for why this should be added.

MyreMylar:
> I'm assuming you got most of the implementation by checking the vector code, that should be good. Before merging this, I'd like to hear a use case for why this should be added.

I'm also struggling to think of a reason I would use this. I guess flipping coordinates around with rectangles is just less common in game development? 

If anybody out there has a use case for this then I am totally open to hearing it.

Matiiss:
The only somewhat useful case I can think of would be if the swizzling returned a `pygame.Vector2` when swizzling two values. That way you could get the center of the rect as a vector like so:
```py
center_vec = rect.xy + rect.wh / 2
```
compared to
```py
center_vec = pygame.Vector2(rect.center)
```
So yeah... not a very strong argument, especially given that we could add `_v` attributes to the rects, like `center_v` that would achieve the same thing

MyreMylar:
No clear uses cases for this, seem to have emerged. In a search I can't find any other game or graphic library that has added swizzling to geometry other than vectors I'm going to move that we close this PR. 

Thankyou for the work put into this @ScriptLineStudios, but I think it is not a good idea for us to add new functions that it seems like nobody will use as it just creates more code to maintain in the future and complicates the public API for pygame.

If I'm wrong, please re-open this with a compelling use case or demand for this feature and I'd be happy for it to go in.

PR #2111, opened 2023-04-10
vector2 support for Surface.get_at and set_at

ScriptLineStudios:
the ability to use vector2's to specify coordinates for surface set_ats and get_ats


oddbookworm (changes requested):
I think you should edit the typestub info in [buildconfig/stubs/pgyame/surface.pyi](https://github.com/pygame-community/pygame-ce/blob/main/buildconfig/stubs/pygame/surface.pyi) so that linters don't throw a fit when using a `Vector2` in these methods. Also, adding documentation for this new functionality would be a good idea IMO. Maybe make sure to mention that any floating point values will be truncated, not rounded

novialriptide:
`Vector2`s are not `Sequence[int]` as @oddbookworm has mentioned. I'm not entirely sure if this PR should go through as stuff like this becomes valid.
```python
import pygame
a = pygame.Surface((50,50))
a.set_at(pygame.Vector2(50,50.5), (25,25,25))
a.set_at((50,50.5), (25,25,25))
```
whereas before, this would've thrown an error
```python
import pygame
a = pygame.Surface((50,50))
a.set_at((50,50.5), (25,25,25))
```
```
TypeError: 'float' object cannot be interpreted as an integer
```
If we favor introducing this, we can mention in the docs that we now support float values for `get_at()` and `set_at()`.

narilee2006:
I agree

Starbuck5:
> this got me thinking that I'm not entirely sure if this PR should go through as stuff like this becomes valid

Many things in pygame support floats now, and didn't always. Transform.scale used to need integer dimensions, for example. I think it's perfectly fine to have integer things accept floats and just truncate them.

But you're right, this can be documented and hinted as support for a 2-sequence of floats.

src_c/surface.c
```diff
@@ -777,14 +777,20 @@ surf_get_at(PyObject *self, PyObject *args)
...
     Uint8 rgba[4] = {0, 0, 0, 255};
 
-    if (!PyArg_ParseTuple(args, "(ii)", &x, &y))
+    if (!PyArg_ParseTuple(args, "O", &position))
         return NULL;
     SURF_INIT_CHECK(surf)
 
+    if (!pg_TwoIntsFromObj(position, &x, &y)) {
+        return RAISE(PyExc_TypeError,
+                     "position must be a sequence containing 2 integers");
```
Starbuck5, line 791: Not really, since they can be floats.

`position must be a sequence of two numbers` ?

src_c/surface.c
```diff
@@ -916,13 +928,19 @@ surf_get_at_mapped(PyObject *self, PyObject *args)
...
     Uint8 *pix;
 
-    if (!PyArg_ParseTuple(args, "(ii)", &x, &y))
+    if (!PyArg_ParseTuple(args, "O", &position))
         return NULL;
     SURF_INIT_CHECK(surf)
 
+    if (!pg_TwoIntsFromObj(position, &x, &y)) {
+        return RAISE(PyExc_TypeError,
+                     "position must be a sequence containing 2 integers");
```
Starbuck5, line 941: Missed this one with the `position must be a sequence of two numbers` action.

buildconfig/stubs/pygame/surface.pyi
```diff
@@ -112,8 +113,8 @@ class Surface:
     def mustlock(self) -> bool: ...
     def get_locked(self) -> bool: ...
     def get_locks(self) -> Tuple[Any, ...]: ...
-    def get_at(self, x_y: Sequence[int]) -> Color: ...
-    def set_at(self, x_y: Sequence[int], color: ColorValue) -> None: ...
+    def get_at(self, x_y: Union[Sequence[int], Vector2]) -> Color: ...
+    def set_at(self, x_y: Union[Sequence[int], Vector2], color: ColorValue) -> None: ...
     def get_at_mapped(self, x_y: Sequence[int]) -> int: ...
```
Starbuck5, line 118: I don't think this is right.

In the code get_at_mapped is changed too. Also it's not just a sequent of ints or a Vector, it supports floats. See how `transform.scale` hints itself. It uses https://github.com/pygame-community/pygame-ce/blob/main/buildconfig/stubs/pygame/_common.pyi#L18

Starbuck5:
Also I'm curious what kind of performance impact this has. on get_at/set_at.

docs/reST/ref/surface.rst
```diff
@@ -445,20 +446,24 @@
       methods - or by using :mod:`pygame.surfarray`/:mod:`pygame.PixelArray`.
 
       This function will temporarily lock and unlock the Surface as needed.
-
-      .. versionaddedold:: 1.9
+      
+      .. note:: Any floating point values given in the specified pixel coordiante will be truncated.
```
Starbuck5, line 450: I feel like it's much important to say that it supports floats now than the specific way it gets from floats to ints.

I would just give it a `versionchanged:: 2.3.0 now can accept float coordinates (and hence Vectors2)` or something.

buildconfig/stubs/pygame/surface.pyi
```diff
@@ -112,8 +113,8 @@ class Surface:
     def mustlock(self) -> bool: ...
     def get_locked(self) -> bool: ...
     def get_locks(self) -> Tuple[Any, ...]: ...
-    def get_at(self, x_y: Sequence[int]) -> Color: ...
-    def set_at(self, x_y: Sequence[int], color: ColorValue) -> None: ...
+    def get_at(self, x_y: Union[Tuple[float, float], Sequence[float], Vector2]) -> Color: ...
+    def set_at(self, x_y: Union[Tuple[float, float], Sequence[float], Vector2], color: ColorValue) -> None: ...
```
Starbuck5, line 117: Rather than copy the value of the Coordinate value I linked, you can import it into the type stub file and use it here directly, like how the Surface typestubs do it.

docs/reST/ref/surface.rst
```diff
@@ -434,9 +434,10 @@
       | :sl:`get the color value at a single pixel`
       | :sg:`get_at((x, y)) -> Color`
 
-      Return a copy of the ``RGBA`` Color value at the given pixel. If the
-      Surface has no per pixel alpha, then the alpha value will always be 255
-      (opaque). If the pixel position is outside the area of the Surface an
+      Return a copy of the ``RGBA`` Color value at the given pixel. The pixel can
+      either be given as Sequence or as a Vector2. If the Surface has no per
```
Starbuck5, line 438: "The pixel can either be given as Sequence or as a Vector2"

Vector2s are sequences
Also I don't think the average person reading these docs knows what a sequence is

I would drop this sentence addition.

src_c/surface.c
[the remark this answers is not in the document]
ScriptLineStudios: changed it :+1:

buildconfig/stubs/pygame/surface.pyi
```diff
@@ -3,6 +3,7 @@ from typing import Any, List, Optional, Sequence, Tuple, Union, overload, Iterab
 from pygame.bufferproxy import BufferProxy
 from pygame.color import Color
 from pygame.rect import Rect, FRect
+from pygame.math import Vector2
```
Starbuck5, line 6: Now that you're using Coordinate this Vector2 is an unneeded import?

docs/reST/ref/surface.rst
```diff
@@ -445,12 +445,15 @@
       methods - or by using :mod:`pygame.surfarray`/:mod:`pygame.PixelArray`.
 
       This function will temporarily lock and unlock the Surface as needed.
-
-      .. versionaddedold:: 1.9
+      
+      .. versionchanged:: 2.3.0 now can accept both float coordinates and Vector2s
```
Starbuck5, line 449: 2.3.1 now (several other locations too)

docs/reST/ref/surface.rst
```diff
@@ -434,9 +434,9 @@
       | :sl:`get the color value at a single pixel`
       | :sg:`get_at((x, y)) -> Color`
 
-      Return a copy of the ``RGBA`` Color value at the given pixel. If the
-      Surface has no per pixel alpha, then the alpha value will always be 255
-      (opaque). If the pixel position is outside the area of the Surface an
+      Return a copy of the ``RGBA`` Color value at the given pixel. If the Surface has no per 
+      pixel alpha, then the alpha value will always be 255 (opaque). 
+      If the pixel position is outside the area of the Surface an
```
Starbuck5, line 439: Since these have the same content there shouldn't be a change in the diff.

src_c/surface.c
```diff
@@ -784,10 +783,16 @@ surf_get_at(PyObject *self, PyObject *args)
     Uint8 *pix;
     Uint8 rgba[4] = {0, 0, 0, 255};
 
-    if (!PyArg_ParseTuple(args, "(ii)", &x, &y))
-        return NULL;
+    if (!PySequence_Check(position))
+        return RAISE(PyExc_ValueError, "Argument must be a sequence type");
```
Starbuck5, line 787: The sequence_check is done in pg_TwoIntsFromObj. Relevant in the other method too.

Starbuck5:
I did some performance testing, and set_at with this PR is 1.5% faster now! (This is pretty low, it could be a fluke)
But this means that there isn't a big performance regression here we're not noticing. 

I'm not sure whether pg_TwoIntFromObj or that PyArg_ParseTuple is more expensive, but METH_O is less expensive than METH_VARARGS so maybe that helps it.

src_c/surface.c
ScriptLineStudios: done!

oddbookworm (approved):
LGTM!

buildconfig/stubs/pygame/surface.pyi
```diff
@@ -112,8 +112,8 @@ class Surface:
     def mustlock(self) -> bool: ...
     def get_locked(self) -> bool: ...
     def get_locks(self) -> Tuple[Any, ...]: ...
-    def get_at(self, x_y: Sequence[int]) -> Color: ...
-    def set_at(self, x_y: Sequence[int], color: ColorValue) -> None: ...
+    def get_at(self, x_y: Coordinate) -> Color: ...
+    def set_at(self, x_y: Coordinate, color: ColorValue) -> None: ...
     def get_at_mapped(self, x_y: Sequence[int]) -> int: ...
```
ankith26, line 117:
```suggestion
    def get_at_mapped(self, x_y: Coordinate) -> int: ...
```

ankith26 (changes requested):
The docs, tests and stubs for `get_at_mapped` aren't updated yet

Starbuck5 (approved):
Looks good!

Thanks for working on this. Getting stuff supporting floats when appropriate is a good initiative throughout the pygame-ce API.

PR #2113, opened 2023-04-11
added stub info for PixelArray.__getitem__

oddbookworm:


buildconfig/stubs/pygame/pixelarray.pyi
```diff
@@ -17,6 +17,7 @@ class PixelArray:
     def __init__(self, surface: Surface) -> None: ...
     def __enter__(self) -> PixelArray: ...
     def __exit__(self, *args, **kwargs) -> None: ...
+    def __getitem__(self, indices: Union[int, Sequence[int]]) -> Union[PixelArray, int]: ...
```
ankith26, line 20: Hmm, I'm not quite familiar with `PixelArray`, but a quick glance through the code says this could be more complicated. It looks like the code is only handles `tuple` and not generic `Sequence`s. The code also handles ellipsis and slicing.
It seems to return `int` in very few cases, might be worth separating that into a separate `overload` for this method

oddbookworm: @ankith26 ...

You're right, it's nasty.

Starbuck5 (approved):

buildconfig/stubs/pygame/pixelarray.pyi
Starbuck5: Overload added, so resolving this little subthread.

zoldalma999 (approved):
Looks good to me!

PR #2070, opened 2023-03-28
pygame.window.Window

yunline:
#2069
## Tasks
<details close>

**Functions**
- [x] *NEW* `window.get_windows()`
- [x] `window.get_grabbed_window()`

**Class**
- [x] `window.Window`

**Class Methods**
- [x] `window.Window.from_display_module()`
- [x] *NEW* `window.Window.from_existing_window()`

**Methods**
- [x] `window.Window.destroy()`
- [x] `window.Window.set_windowed()`
- [x] `window.Window.set_fullscreen()`
- [x] `window.Window.focus()`
- [x] `window.Window.hide()`
- [x] `window.Window.show()`
- [x] `window.Window.restore()`
- [x] `window.Window.maximize()`
- [x] `window.Window.minimize()`.
- [x] *NEW* `window.Window.flash()`
- [x] *NEW* `window.Window.flash_cancel()`
- [x] `window.Window.set_modal_for()`.
- [x] *NEW* `window.Window.update_from_surface()`

**Getters/setters**
- [x] `window.Window.set_icon()`
- [x] `window.Window.set_grab()`
- [x] `window.Window.get_grab()`
- [x] `window.Window.set_title()`
- [x] `window.Window.get_title()`
- [x] `window.Window.set_resizable()`
- [x] `window.Window.get_resizable()`
- [x] `window.Window.set_borderless()`
- [x] `window.Window.get_borderless()`
- [x] `window.Window.get_window_id()`
- [x] `window.Window.set_size()`
- [x] `window.Window.get_size()`
- [x] `window.Window.set_position()`
- [x] `window.Window.get_position()`
- [x] `window.Window.set_opacity()`
- [x] `window.Window.get_opacity()`
- [x] `window.Window.get_display_index()`
- [x] *NEW* `window.Window.get_always_on_top()`
- [x] *NEW* `window.Window.set_always_on_top()`
- [x] *NEW* `window.Window.get_surface()`
- [x] *NEW* `window.Window.get_wm_info()`
- [x] *NEW* `window.Window.get_foreign()`

**Others**
- [x] tests
- [x] docs
- [ ] example

</details>

## Compatibility
Notice that `window.Window` is not compatible with `_sdl2.video.Window`, but you can use `_sdl2.video.Window.from_window()` to  get a "_sdl2 window" from a "window_module window"
```python
_win=pygame.window.Window()
win=video.Window.from_window(_win)
```

PR #2114, opened 2023-04-12
Port video.Window to C

yunline:
Split from #2070
- Port video.Window to C
- `_sdl2.video.Window` (old) -> `_sdl2.video._Window` and `_sdl2.window.Window`


src_c/window.c
```diff
@@ -0,0 +1,667 @@
...
+
+#include "doc/window_doc.h"
+
+static PyTypeObject pgWindow_Type;
+
+#define pgWindow_Check(x) \
+    (PyObject_IsInstance((x), (PyObject *)&pgWindow_Type))
+
+static PyObject *
+get_grabbed_window(PyObject *self)
```
Starbuck5, line 17: The PyCFunction functions like this need to have a second PyObject* parameter to make it friendlier for webassembly, apparently.

See things like `mouse_get_rel` in mouse.c, it has a second parameter `PyObject *_null` (unused)

This needs to be done for a bunch of the functions in this file.

src_c/window.c
```diff
@@ -0,0 +1,667 @@
...
+{
+    int index = SDL_GetWindowDisplayIndex(self->_win);
+    if (index < 0) {
+        return RAISE(pgExc_SDLError, SDL_GetError());
+    }
+    return PyLong_FromLong(index);
+}
+
+static void
+window_dealloc(pgWindowObject *self)
```
Starbuck5, line 333: It's kind of suspicious this doesn't call SDL_DestroyWindow. I mean, I see that the _sdl2.video implementation doesn't either, but still.

src_c/window.c
```diff
@@ -0,0 +1,667 @@
...
+
+static PyObject *
+window_get_size(pgWindowObject *self, void *v)
+{
+    int w, h;
+    PyObject *out = PyTuple_New(2);
+
+    SDL_GetWindowSize(self->_win, &w, &h);
+    PyTuple_SetItem(out, 0, PyLong_FromLong(w));
+    PyTuple_SetItem(out, 1, PyLong_FromLong(h));
```
Starbuck5, line 260: The Python C-API stuff here needs to be error checked. But rather than error checking these three function calls, you could use `Py_BuildValue`. Then you only have to error check the one thing.

src_c/window.c
```diff
@@ -0,0 +1,667 @@
...
+}
+
+static PyObject *
+window_get_opacity(pgWindowObject *self, void *v)
+{
+    float opacity;
+    if (SDL_GetWindowOpacity(self->_win, &opacity)) {
+        return RAISE(pgExc_SDLError, SDL_GetError());
+    }
+    return PyFloat_FromDouble(opacity);
```
Starbuck5, line 319: I'm surprised the compiler didn't complain about this, minor type mismatch. Function expects a double and you give it a float. So you should explicitly cast to double. `PyFloat_FromDouble((double)opacity);`

src_c/window.c
```diff
@@ -0,0 +1,667 @@
...
+     (setter)window_set_borderless, NULL, NULL},
+    {"size", (getter)window_get_size, (setter)window_set_size, NULL, NULL},
+    {"position", (getter)window_get_position, (setter)window_set_position,
+     NULL, NULL},
+    {"opacity", (getter)window_get_opacity, (setter)window_set_opacity, NULL,
+     NULL},
+    {"display_index", (getter)window_get_display_index, NULL, NULL, NULL},
+    {"id", (getter)window_get_window_id, NULL, NULL, NULL},
+    {NULL, 0, NULL, NULL, NULL} /* Sentinel */
+};
```
Starbuck5, line 585: The defined setter functions all have the wrong signature and semantic. This is something I got tricked by when I was working on a similar thing before :)

https://docs.python.org/3/c-api/structures.html#c.PyGetSetDef
The setters are supposed to return ints. `0` on success, -1` on failure.

Right now if you call a setter you get a SystemError.

Starbuck5:
One thing I'd like to address besides the code is the namespace considerations.

Right now
```
_sdl2.video.Window -> _sdl2.video._Window
*** -> window.Window
```

I would like this new window to be in _sdl2 for now as well, until we decide it's ready for a public release (do a lot more integration work, add surface rendering functionality)

It also breaks every _sdl2 program to get rid of the Window class being where it is.

So instead, I suggest
```
_sdl2.video.Window -> _sdl2.video._Window
*** -> _sdl2.window.Window
    |  _sdl2.video.Window
```

I suggest aliasing a _sdl2.window.Window to _sdl2.video.Window. This way we can use existing _sdl2 programs and users to test the new Window implementation. Since the old window is still around (for the moment), we can also switch to that at runtime to test differences.

buildconfig/stubs/pygame/_sdl2/video.pyi
```diff
@@ -3,6 +3,7 @@ from typing import Any, Generator, Iterable, Optional, Tuple, Union
 from pygame.color import Color
 from pygame.rect import Rect
 from pygame.surface import Surface
+from pygame._window import Window
```
Starbuck5, line 6: Use the _sdl2 namespace of it please.

Starbuck5: Actually...

Why does this need a typing change at all?

You're replacing sdl2.video.Window with a new sdl2.video.Window

Is it really necessary to define all the type operations of the old code now at sdl2.video._Window?

docs/reST/ref/sdl2_video.rst
```diff
@@ -13,15 +13,15 @@
...
 
    | :sl:`pygame object that represents a window`
-   | :sg:`Window(title="pygame", size=(640, 480), position=None, fullscreen=False, fullscreen_desktop=False, keywords) -> Window`
+   | :sg:`_Window(title="pygame", size=(640, 480), position=None, fullscreen=False, fullscreen_desktop=False, keywords) -> Window`
 
    .. classmethod:: from_display_module
    
       | :sl:`Creates window using window created by pygame.display.set_mode().`
-      | :sg:`from_display_module() -> Window`
+      | :sg:`from_display_module() -> _Window`
```
Starbuck5, line 24: Why change this stuff?

Your changes are not supposed to be user facing. People can still use Window. Ideally without any behavior change.

docs/reST/ref/window.rst
```diff
@@ -0,0 +1,156 @@
+.. include:: common.txt
```
Starbuck5, line 1: This is out of date with the namespace changes.

But I don't think it should be included in the first place. This PR just ports _sdl2.video.Window written in Cython to _sdl2.video.Window written in C. Whatever module it comes from in the C code shouldn't be relevant, and it shouldn't have its own docs to get out of date from Mega's _sdl2 changes.

test/video_test.py
```diff
@@ -14,7 +14,7 @@ class VideoModuleTest(unittest.TestCase):
     )
     def test_renderer_set_viewport(self):
         """works."""
-        window = video.Window(title=self.default_caption, size=(800, 600))
+        window = video._Window(title=self.default_caption, size=(800, 600))
```
Starbuck5, line 17: Why change this?
The new window is at video.Window right? And that's what we should be testing.

docs/reST/ref/window.rst
yunline: hummm but the docs is link to the  doc string in window.c

Starbuck5: You can use sdl2_video_doc

src_c/window.c
```diff
@@ -0,0 +1,684 @@
...
+    if (index < 0) {
+        return RAISE(pgExc_SDLError, SDL_GetError());
+    }
+    return PyLong_FromLong(index);
+}
+
+static void
+window_dealloc(pgWindowObject *self, PyObject *_null)
+{
+    PyObject_Free(self);
```
Starbuck5, line 346: This is the wrong function to use. Check something like the Surface C code, it does `Py_TYPE(self)->tp_free(self);`.

PyObject_Free (https://docs.python.org/3/c-api/memory.html#c.PyObject_Free) is like malloc/free.

src_c/window.c
```diff
@@ -0,0 +1,684 @@
...
+        Py_DECREF(&pgWindow_Type);
+        Py_DECREF(module);
+        return NULL;
+    }
+    Py_INCREF(&pgWindow_Type);
+    if (PyModule_AddObject(module, "WindowType", (PyObject *)&pgWindow_Type)) {
+        Py_DECREF(&pgWindow_Type);
+        Py_DECREF(module);
+        return NULL;
+    }
```
Starbuck5, line 673: Let's not add this. The other things that do this, do this for backwards compatibility. With new code we don't need to be backwards compatible in this way.

src_c/window.c
```diff
@@ -0,0 +1,684 @@
+#define PYGAMEAPI_WINDOW_INTERNAL
+
+#include "pygame.h"
+
+#include "pgcompat.h"
+
+#include <SDL_syswm.h>
```
Starbuck5, line 7: Is this necessary?

src_c/window.c
```diff
@@ -0,0 +1,684 @@
...
+
+    return out;
+}
+
+static int
+window_set_position(pgWindowObject *self, PyObject *arg, void *v)
+{
+    int x, y;
+
+    if (Py_TYPE(arg) == &PyLong_Type) {
```
Starbuck5, line 276: Use `PyLong_Check`

src_c/window.c
```diff
@@ -0,0 +1,684 @@
...
+}
+
+static PyObject *
+window_get_position(pgWindowObject *self, void *v)
+{
+    int x, y;
+    PyObject *out = NULL;
+
+    SDL_GetWindowPosition(self->_win, &x, &y);
+    out = Py_BuildValue("(i,i)", x, y);
```
Starbuck5, line 300: I'd do it as `Py_BuildValue("(ii)", x, y)`. Seems like the comma is ignored.

src_c/window.c
```diff
@@ -0,0 +1,684 @@
...
+}
+
+static PyObject *
+window_get_size(pgWindowObject *self, void *v)
+{
+    int w, h;
+    PyObject *out = NULL;
+
+    SDL_GetWindowSize(self->_win, &w, &h);
+    out = Py_BuildValue("(i,i)", w, h);
```
Starbuck5, line 263: I'd do it as Py_BuildValue("(ii)", x, y). Seems like the comma is ignored.

src_c/window.c
yunline: _sdl2.video actually does
https://github.com/pygame-community/pygame-ce/blob/752b6c32d1683183a79ba043fc6139a8a63f2f02/src_c/cython/pygame/_sdl2/video.pyx#L517-L520

Starbuck5:
Tests are failing because of the relative mouse mode not being in the type stub?

src_py/__init__.py
```diff
@@ -340,6 +340,11 @@ def Overlay(format, size):  # pylint: disable=unused-argument
 except (ImportError, OSError):
     pass
 
+try:
+    import pygame._window
+except (ImportError, OSError):
+    system = MissingModule("_window", urgent=1)
```
Starbuck5, line 346: It shouldn't be urgent if it fails to import.

Should it even be default imported at all?

src_c/window.c
```diff
@@ -0,0 +1,702 @@
...
+                    if (_value_bool)
+                        flags |= SDL_WINDOW_POPUP_MENU;
+                }
+                else if (!strcmp(_key_str, "vulkan")) {
+                    if (_value_bool)
+                        flags |= SDL_WINDOW_VULKAN;
+                }
+                else {
+                    sprintf(_exc_str, "__init__ got an unexpected flag \'%s\'",
+                            _key_str);
```
Starbuck5, line 504: PyErr_Format instead?
https://docs.python.org/3/c-api/exceptions.html#c.PyErr_Format

Starbuck5 (approved):
This is a very important PR to get the Window API up and running, thank you so much for tackling it!

I think it looks good now.

Roadmap in my head:
- Two window APIs coexist for a release (for back testing purposes)
- Window tests!
- Old _sdl2.window cython code is removed
- New window features: (display interop, software rendering, etc) 
- Release window API publicly

MyreMylar (commented):
Things I've noticed:

1. No snake icon in the title bar top left corner of the new C window versus the other ways to create a window.
2. When expanding a resizable new window to the bottom right some white lines are left behind.
3. If you `.destroy()` a new window with `.grab = True` your mouse is still grabbed to the now destroyed window's area.
4. Going directly from minimised to maximised with the functions does not expand the black area of the window in the same way it does if you go, for example: `minimise()` -> `restore()` ->`maximise()`
5. calling `.show()` on a hidden window turns the background white.

All this is tested on a Windows 11 PC, results may vary on other platforms as we are dealing with window management here. Some of these issues may be pre-exisiting in the old cython window.

Starbuck5:
@MyreMylar Thanks for testing this out.

Number 1 is consistent with the cython implementation. Number 2, number 4, and number 5 are also consistent with cython, and seem like the product of not clearing the screen with a Renderer. Even display surfaces have some weird artifacts when you don't fill them every frame (when resized).

I can't replicate number 4.

MyreMylar:
> @MyreMylar Thanks for testing this out.
> 
> Number 1 is consistent with the cython implementation. Number 2, number 4, and number 5 are also consistent with cython, and seem like the product of not clearing the screen with a Renderer. Even display surfaces have some weird artifacts when you don't fill them every frame (when resized).
> 
> I want this PR to be consistent with the cython implementation. It's "port window," not "improve window."
> 
> I can't replicate number 4.

For me the snake icon does appear on the old cython window but never on the new C window.

This is my test program:

```python
import pygame
from pygame._sdl2.window import Window as OldWindow
from pygame._sdl2.video import Window as NewWindow

pygame.init()

old_window = OldWindow(title="Old Window", position=(20, 100))
new_window = NewWindow(title="New Window", position=(1000, 100))


running = True
while running:
        if event.type == pygame.QUIT or event.type == pygame.KEYDOWN and event.key == pygame.K_ESCAPE:
            running = False


```

I also figured out that it is the first window created that gets the black background, any subsequent windows get a white background so that explains those differences somewhat. This is consistent between cython window and C window. Not really an issue for this PR.

Anyway it basically works being the important thing 👍

yunline:
I copied some code from display.c to set the default icon.
I think copying is not a good choice, so I opened #2222, made a new API to get resource from pkgdata

Starbuck5:
I’m unconvinced that the old cython code did this. Also if I understand the imports correctly, Myre’s test code is testing the new window twice. Haven’t fully verified yet, just how it looks.

yunline:
> I’m unconvinced that the old cython code did this.

https://github.com/pygame-community/pygame-ce/blob/f5b674fff326465d03174a5ca9893a00b031ed3d/src_c/cython/pygame/_sdl2/video.pyx#L263-L269

It really does.

Starbuck5:
Ok, it's because my test code had the window go fullscreen first, which leads SDL to lose track of the icon and the resulting non fullscreen window not having any icon.

oddbookworm (approved):
I forgot to actually leave a review when I looked at it a couple days ago, LGTM!

PR #2120, opened 2023-04-17
Updated SCALED description

FinFetChannel:
The previous description wasn't very clear on what the flag does

### DIFF
--- a/docs/reST/ref/display.rst
+++ b/docs/reST/ref/display.rst
@@ -147,7 +147,8 @@ required).
 
    On high resolution displays(4k, 1080p) and tiny graphics games (640x480)
    show up very small so that they are unplayable. SCALED scales up the window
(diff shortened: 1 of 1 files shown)


Starbuck5:
"Closest integer **scale**"

Closest integer would be like "scaled to 1920x1080 because those are the closest integers to the resolution of 1920x1080"

Also is it always an integer scale, what if it's in resizable mode as well?

robertpfeiffer:
I don't think this should be merged. I _deliberately_ left out how this works in detail, because it's an implementation detail, The whole point of SCALED was that it's a "do what I mean" kind of thing.

Starbuck5:
Thanks for the PR @FinFetChannel 

I'm going to close this for now because it's been stale for a while and I'm not sure the way forward on it. I'm not opposed to being more specific in the SCALED docs though.

Issue #1710, opened 2023-02-12
Dark mode for docs (3460)


GalacticEmperor1:
#  **Issue &numero;3460 opened by *[illume](https://github.com/illume)* at 2022-09-19 17:13:31**

Honoring the browser preference, with a button to toggle it, and a cookie to remember it.<hr>

#  Comments

# # #  **[novialriptide](https://github.com/novialriptide)* commented at 2022-09-20 01:10:22*

I was thinking, is it a good idea to create a new documentation website? I'm particularly a big fan of [readthedocs.org's format](https://pytmx.readthedocs.io/en/latest/). We can set a toggle between the old and new docs just like https://old.reddit.com/
<hr>

# # #  **[illume](https://github.com/illume)* commented at 2022-10-02 13:44:00*

This issue is about adding a dark mode, not about replacing the docs with a new website.
<hr>

# # #  **[itzpr3d4t0r](https://github.com/itzpr3d4t0r)* commented at 2022-10-02 13:56:12*

I'm super excited for a dark mode, that green is historic but hurts the eyes a lot. With that being said i think we should show some examples before accepting one, as "dark mode" is too general.
<hr>

# # #  **[osinmv](https://github.com/osinmv)* commented at 2022-10-26 19:12:18*

<img width="1680" alt="Screenshot 2022-10-26 at 3 10 13 PM" src="https://user-images.githubusercontent.com/31288523/198115711-51e1af1a-bb2f-414a-b22e-d61782d39fa3.png">

Does it look any good? Just a concept
<hr>

# # #  **[novialriptide](https://github.com/novialriptide)* commented at 2022-10-27 15:58:17*

> <img alt="Screenshot 2022-10-26 at 3 10 13 PM" width="1680" src="https://user-images.githubusercontent.com/31288523/198115711-51e1af1a-bb2f-414a-b22e-d61782d39fa3.png">
> 
> Does it look any good? Just a concept

I feel that we would need to rewrite a lot of the CSS for a potential dark mode. The bright green text hurts my eyes a bit.
<hr>

# # #  **[osinmv](https://github.com/osinmv)* commented at 2022-10-27 16:01:43*

Thanks, I will also talk with some people in pygame discord, maybe they will have some input on this. Once I gather info from them I will make a PR

<hr>

# # #  **[Zireael07](https://github.com/Zireael07)* commented at 2022-10-27 16:43:41*

Agreed, bright green on black while a throwback to the earliest terminals, makes my eyes hurt.
[...]

PR #2122, opened 2023-04-19
Add dark theme for documentation

Notenlish:
This commit adds dark theme to pygame-ce documentation. It uses localStorage in js to save the theme that the user chose and sets the class of the body tag based on that. The theme.conf file now has dark versions of some settings. It also has some modifications to the way docs are generated in makeref.py, I had to do this because I couldnt figure out a better way for it. There probably is a better way to do it but i don't know how.


Starbuck5:
I took a small look at this tonight.

I was surprised to see that the scroll bar design had changed, and that the quick start header is now in the middle of the page on the docs main page.

Notenlish:
The scrollbar could be kept the default one, I'm not sure if it fits the theme

Starbuck5:
>  The scrollbar could be kept the default one, I'm not sure if it fits the documentation theme

What documentation theme? Specifically "dark mode" or just in general?

I'd rather use the built in scroll bar than this very custom looking thing. Looking at Github light vs dark, they both use the normal scroll bar, but the colors of it change slightly.

Notenlish:
Reverted the scrollbar

docs/reST/themes/classic/script.html
```diff
@@ -0,0 +1,19 @@
+// named as script.html to avoid a conflict with the pathto() function used in template
+console.log("hi from script.js");
```
Starbuck5, line 2: Can't have this in production 😄

Starbuck5:
Scrollbar back how it was, thanks.

I'm trying to minimize the changes in this PR to make it just about the dark mode so it is focused and easy to review :)

Next up: the searchbar. I'd rather you keep it how it was-- shorter height, text search button instead of icon. With how tall it is now it distorts the whole top row of the top bar. I'm thinking you did this to match the height of the svg icon you got? The svg should be scaled down.

docs/reST/themes/classic/script.html
Notenlish: I couldn't figure out how to put it in an external file while also having it named "script.js". The pathto() function adds ".html" to the filename. I tried to change it but couldnt figure out how it works. Would it be better to put it directly inside the html?

Starbuck5: I'm referring to the console.log

Notenlish: oh sorry 😅 didnt notice that, fixed

Notenlish:
> Scrollbar back how it was, thanks.
> 
> I'm trying to minimize the changes in this PR to make it just about the dark mode so it is focused and easy to review :)
> 
> Next up: the searchbar. I'd rather you keep it how it was-- shorter height, text search button instead of icon. With how tall it is now it distorts the whole top row of the top bar. I'm thinking you did this to match the height of the svg icon you got? The svg should be scaled down.

Should I change the search icon back to the text or keep it?

buildconfig/makeref.py
```diff
@@ -27,13 +28,41 @@ def run():
...
+                f.write(data)
+        original = r"{}".format(os.path.join(rst_source_dir, 'themes/classic') + '\script.html')
+        target = r"{}".format(f'{rst_build_dir}\script.html')
+        shutil.copyfile(original, target)
+    except Exception:
+        print('---')
+        print('Failed to add js files')
+        print('---')
+        print(Exception)
+        return Exception
```
gresm, line 64: Few issues: 
1. Why do you print ``Exception`` class into the console. Maybe you want to re-raise it again?
2. ``return Exception`` Again, why?
3. This part probably was intended to be for debugging reasons, but instead it hides the error.

buildconfig/makeref.py
```diff
@@ -27,13 +28,41 @@ def run():
...
+                data = f.read()
+                index = data.find(tag_to_find)
+                if index == -1:
+                    return Exception
+                index += len(tag_to_find) - 1
+                data = data[:index + 1] + '\n' + tag_to_add + data[index:]
+            with open(os.path.join(rst_build_dir, file),'w') as f:
+                f.write(data)
+        original = r"{}".format(os.path.join(rst_source_dir, 'themes/classic') + '\script.html')
+        target = r"{}".format(f'{rst_build_dir}\script.html')
```
gresm, line 57: Wrong backslashes: use ``/``. This script has to work on different operating systems.

buildconfig/makeref.py
```diff
@@ -27,13 +28,41 @@ def run():
         if full_generation_flag:
             subprocess_args.append('-E')
         print("Executing sphinx in subprocess with args:", subprocess_args)
-        return subprocess.run(subprocess_args).returncode
+        returncode = subprocess.run(subprocess_args).returncode
+        if returncode == 0:
+            add_js_files()
+        return returncode
```
gresm, line 34: If the ``add_js_files()`` would fail, it still would return error code 0.

gresm (changes requested):
buildconfig/makeref.py fails to include js files, so no extra widgets are working...

buildconfig/makeref.py
gresm: Actually, those backslashes are unnecessary.

gresm:
For proper sphinx way of including JavaScript files see [this](https://www.sphinx-doc.org/en/master/extdev/appapi.html#sphinx.application.Sphinx.add_js_file).

Notenlish:
> For proper sphinx way of including JavaScript files see [this](https://www.sphinx-doc.org/en/master/extdev/appapi.html#sphinx.application.Sphinx.add_js_file).

I'll try and do it

buildconfig/makeref.py
```diff
@@ -14,26 +15,60 @@
...
+        print("---")
         raise
 
+    try:
+        if returncode == 0:
+            add_js_files()
+    except Exception:
+        print("---")
+        print("Failed to add js files")
+        print("---")
```
novialriptide, line 43: Remove the `---` prints

buildconfig/makeref.py
```diff
@@ -14,26 +15,60 @@
...
+        raise
+
+    return returncode
+
+def add_js_files():
```
novialriptide, line 48: Could you add a brief docstring to explain what this method is doing?

docs/reST/themes/classic/elements.html
```diff
@@ -17,14 +17,28 @@ <h5>pygame-ce documentation</h5>
...
-          <input name="q" value="" type="text">
-          <input value="search" type="submit">
-        </form>
+		<div class="nav-top-wrapper">
+			<a href="{{ theme_home_uri }}">Pygame Home</a> ||
+			<a href="{{ pathto(master_doc) }}">Help Contents</a> ||
+			<a href="{{ pathto('genindex') }}">Reference Index</a>
+		</div>
+		<div class="nav-theme-container">
+			<svg class="theme-icon" data-theme="light-theme" version="1.1" xmlns="http://www.w3.org/2000/svg" width="32" height="32" viewBox="0 0 32 32">
```
novialriptide, line 26: If possible, could you put these svg files into its own file?

docs/reST/themes/classic/theme.conf
```diff
@@ -18,28 +18,82 @@ footertextcolor      = #000000
...
+dark_codedot         = #D4D4D4
+dark_codecomment     = #169d00
+dark_codeword         = #85E5FF
+dark_codekeyword     = #DE6C90
+dark_highlightbgcolor= #0E5114
+dark_logobgcolor     = transparent
+dark_tooltipcolor    = #00ba10
+dark_tooltipbgcolor  = #141414
+dark_tooltipbdrcolor = #00ba10
+dark_searchbarbgcolor= #292929
```
novialriptide, line 44: Code quality nitpick: Could you put a space before and after `=`?

novialriptide (commented):
I like this pull request a lot! I just have a few nitpicks about it.

In the future, we should probably instate a formatter like Prettier or something.

buildconfig/makeref.py
```diff
@@ -3,6 +3,7 @@
 import sys
 import os
 import subprocess
+import shutil
```
Starbuck5, line 6: There are a lot of pointless changes in this file in this PR. It should be left how it was before with whitespace.

docs/reST/_static/script.js
```diff
@@ -0,0 +1,16 @@
...
+
+const theme_icons = document.querySelectorAll('.theme-icon');
+theme_icons.forEach((icon) => {
+    icon.addEventListener('click', () => {
+        const theme = icon.getAttribute('data-theme');
+        document.body.classList.remove('light-theme', 'dark-theme');
+        document.body.classList.add(theme);
+        localStorage.setItem('theme', theme);
+    });
+});
```
Starbuck5, line 16: Needs a blank line at the end.

buildconfig/makeref.py
```diff
@@ -27,13 +28,41 @@ def run():
...
+    files_to_change = ['search.html', 'genindex.html']
+    tag_to_find = '<div class="body" role="main">'
+    tag_to_add = f'<script src="script.html"></script>'
+    try:
+        for file in files_to_change:
+            with open(os.path.join(rst_build_dir, file),'r') as f:
+                data = f.read()
+                index = data.find(tag_to_find)
+                if index == -1:
+                    return Exception
```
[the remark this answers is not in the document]
Starbuck5: Looks resolved?

buildconfig/makeref.py
Starbuck5: Looks resolved?

buildconfig/makeref.py
```diff
@@ -14,26 +15,26 @@
...
+                           "-d", rst_doctree_dir,
+                           "-D", f"headers_dest={c_header_dir}",
+                           "-D", "headers_mkdirs=0",
+                           rst_source_dir, rst_build_dir, ]
         if full_generation_flag:
-            subprocess_args.append('-E')
+            subprocess_args.append("-E")
         print("Executing sphinx in subprocess with args:", subprocess_args)
-        return subprocess.run(subprocess_args).returncode
+        returncode = subprocess.run(subprocess_args).returncode
```
Starbuck5, line 30: I don't understand why this is changed.

Matiiss:
Fixes https://github.com/pygame-community/pygame-ce/issues/1710

docs/reST/themes/classic/elements.html
```diff
@@ -17,14 +17,24 @@ <h5>pygame-ce documentation</h5>
...
+			<img class="theme-icon" data-theme="light-theme" src="{{ pathto('_static/light-theme-icon.svg', 1) }}" width="32" height="32" alt="Light Theme Icon">
+			<img class="theme-icon" data-theme="dark-theme" src="{{ pathto('_static/dark-theme-icon.svg', 1) }}" width="32" height="32" alt="Dark Theme Icon">
+		</div>
+		<div class="searchbar-container">
+			<form action="{{ pathto('search') }}" method="get" class="searchbar-form">
+				<input name="q" value="" type="text">
+				<svg class="searchbar-button" viewBox="0 0 32 32" version="1.1" xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink">
+					<path d="M31.715 28.953c0.381 0.381 0.381 0.999 0 1.381l-1.381 1.381c-0.382 0.381-1 0.381-1.381 0l-9.668-9.668c-0.105-0.105-0.175-0.229-0.222-0.361-1.983 1.449-4.418 2.314-7.063 2.314-6.627 0-12-5.373-12-12s5.373-12 12-12c6.627 0 12 5.373 12 12 0 2.645-0.865 5.080-2.314 7.063 0.132 0.047 0.256 0.116 0.361 0.222l9.668 9.668zM12 4c-4.418 0-8 3.582-8 8s3.582 8 8 8 8-3.582 8-8c0-4.418-3.582-8-8-8z"></path>
+					<input value="search" type="submit">
+				</svg>
```
Starbuck5, line 35: Since the other two svgs were moved out into files, it seems like this one should as well.

docs/reST/themes/classic/elements.html
Starbuck5, line 1: This PR changes the shape of the header sections by "inflating" the top bar, because the icons are taller than the previous content.

To minimize this effect, I think the icons should go down to 24x24 or a similar size.

Starbuck5 (approved):
Ok.

IDK why you switched the buttons. I think it made more sense before. If you're in light mode you click moon to go to dark mode, and vice versa.

Why have two identical SVGs one labeled dark one labeled light. Just use one.

You have a merge conflict to resolve, let people in the discord know if you need help with that.

Also, I'm still a bit skeptical of the dark mode colors themselves. But that shouldn't prevent this from getting merged and going out a bit experimental at first.

With all that in mind, I'm approving it anyway. Thanks for working on this!

Notenlish:
> Why have two identical SVGs one labeled dark one labeled light. Just use one.

I wasn't sure if the colors of the icons should be the same for both themes.

Starbuck5:
> I wasn't sure if the colors of the icons should be the same for both themes.

I believe it's conventional to define the shape of an icon in the svg file and the color in the css. So it would be one svg file anyways.

docs/reST/ext/boilerplate.py
```diff
@@ -45,6 +45,9 @@ def setup(app):
     # This extension uses indexer collected tables.
     app.setup_extension("ext.indexer")
 
+    # Add js files for theme changing in docs.
+    app.add_js_file("script.js", loading_method="defer")
```
Matiiss, line 49: This maybe could use a different loading method or a different approach that lets the dark theme load before displaying the page to the user

MyreMylar:
This 'colour_value' text seems much larger than I would expect:

image

docs/reST/ext/boilerplate.py
MyreMylar: Does this cause the brief flash of green I see when switching between pages?

MyreMylar (approved):
Looks pretty good. I think I would probably prefer some more pastel-y variations on the current colours which seem a little 80's teletext to me (old UK resident reference).

Noticed one area where the styling changes made a particular word really pop out - see comment above - which can probably be fixed in the documentation itself. Overall 👍  

I won't merge yet as there were some questions about the theme caching/loading which it might be nice to resolve, but none of it is crucial. No need to make the perfect the enemy of the good.

docs/reST/ext/boilerplate.py
Notenlish: Fixed it

Matiiss (approved):
For the purposes of this PR as noted by Starbuck:
LGTM 👍 
Thanks for the PR

PR #2123, opened 2023-04-20
Continue moving away from format masks

Starbuck5:
Legwork for easier SDL3 compatibility.

Continuing from #2108

### DIFF
--- a/src_c/rotozoom.c
+++ b/src_c/rotozoom.c
@@ -533,8 +533,8 @@ rotozoomSurface(SDL_Surface *src, double angle, double zoom, int smooth)
         /*
          * New source surface is 32bit with a defined RGBA ordering
          */
-        rz_src = SDL_CreateRGBSurface(0, src->w, src->h, 32, 0x000000ff,
-                                      0x0000ff00, 0x00ff0000, 0xff000000);
+        rz_src = SDL_CreateRGBSurfaceWithFormat(0, src->w, src->h, 32,
+                                                SDL_PIXELFORMAT_ABGR8888);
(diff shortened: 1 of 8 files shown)


ankith26 (approved):
LGTM! Thanks for the PR :tada:

src_c/rotozoom.c
```diff
@@ -533,8 +533,8 @@ rotozoomSurface(SDL_Surface *src, double angle, double zoom, int smooth)
         /*
          * New source surface is 32bit with a defined RGBA ordering
          */
-        rz_src = SDL_CreateRGBSurface(0, src->w, src->h, 32, 0x000000ff,
-                                      0x0000ff00, 0x00ff0000, 0xff000000);
+        rz_src = SDL_CreateRGBSurfaceWithFormat(0, src->w, src->h, 32,
+                                                SDL_PIXELFORMAT_ABGR8888);
```
MyreMylar, line 537: This (and the other rotozoom instance) is weird by the way. I know you are just porting the code here but it jumped out at me.

I see no good reason why 8 and 16bit source surfaces should be rotozoomed into 32bit ABGR surfaces rather than ARGB/XRGB surfaces like everything else seems to be.

I'm actually going to go and double check some code where I use rotozoom() now to make sure I'm converting any resulting output in case it is slowing any blitting stuff down.

MyreMylar (approved):
LGTM (apart from that weird pixel format from rotozoom).

👍

src_c/rotozoom.c
Starbuck5: The format enums make this stuff clearer than the masks :)

PR #2126, opened 2023-04-21
speedup for triangle rendering

ScriptLineStudios:
An ≈50% speedup for triangle rendering with pygame.draw.polygon using a specialized triangle rendering algorithm.

**Benchmark**
```python
import pygame
import timeit
import random

surf = pygame.Surface((600, 600))
def draw():
    p1 = (random.randrange(0, 600), random.randrange(0, 600))
    p2 = (random.randrange(0, 600), random.randrange(0, 600))
    p3 = (random.randrange(0, 600), random.randrange(0, 600))
    pygame.draw.polygon(surf, pygame.Color("white"), [p1, p2, p3])

print(timeit.timeit(draw))
```

**Results**
```bash
Before:
scriptline@pop-os:~/programming/pygame-ce$ python3 benchmark.py
pygame-ce 2.3.0.dev1 (SDL 2.0.20, Python 3.10.6)
40.24612309800068

After:
scriptline@pop-os:~/programming/pygame-ce$ python3 benchmark.py
pygame-ce 2.3.0.dev1 (SDL 2.0.20, Python 3.10.6)
23.257022812000287
```


Starbuck5:
I made a small test program with a 4 three element polygons.

After this PR, only two of them render at all.

```py
import pygame

pygame.init()
clock = pygame.Clock()

while True:
    screen.fill("lightblue")

    # Draws properly
    pygame.draw.polygon(screen, "forestgreen", [(20, 20), (150, 20), (20, 150)])
    pygame.draw.polygon(screen, "forestgreen", [(600, 20), (600, 120), (600, 120)])

    # Doesn't seem to draw at all
    pygame.draw.polygon(screen, "forestgreen", [(400, 20), (500, 20), (500, 20)])
    pygame.draw.polygon(screen, "forestgreen", [(30, 500), (500, 700), (400, 300)])

        if event.type == pygame.QUIT:
            pygame.quit()
            raise SystemExit

    pygame.display.flip()
    
    clock.tick(144)
```

ScriptLineStudios:
> I made a small test program with a 4 three element polygons.
> 
> After this PR, only two of them render at all.
> 
> ```python
> import pygame
> 
> pygame.init()
> screen = pygame.display.set_mode((1280, 720))
> clock = pygame.Clock()
> 
> while True:
>     screen.fill("lightblue")
> 
>     # Draws properly
>     pygame.draw.polygon(screen, "forestgreen", [(20, 20), (150, 20), (20, 150)])
>     pygame.draw.polygon(screen, "forestgreen", [(600, 20), (600, 120), (600, 120)])
> 
>     # Doesn't seem to draw at all
>     pygame.draw.polygon(screen, "forestgreen", [(400, 20), (500, 20), (500, 20)])
>     pygame.draw.polygon(screen, "forestgreen", [(30, 500), (500, 700), (400, 300)])
> 
>     for event in pygame.event.get():
>         if event.type == pygame.QUIT:
>             pygame.quit()
>             raise SystemExit
> 
>     pygame.display.flip()
>     
>     clock.tick(144)
> ```

fixed!

Starbuck5:
I put in some random point shuffling to see if there were any weird effects with different pointing ordering, and I found that the triangle `[(500, 20), (500, 20), (400, 20)]` renders nothing but the triangle `[(400, 20), (500, 20), (500, 20)]` renders a horizontal line. On main they both render as a horizontal line.

Requesting a review from @MightyJosip as our `draw` expert.

MightyJosip (changes requested):
Hey ScriptLine, nice change 👍 , triangles are by far the most important polygons so speeding them up is nice thing.

Few notes:
You actually don't need lines 1531-1532. Those values are used only in drawhorzlineclipbounding, but that method does the swap anyways (lines 1468-1472), so it does same thing twice, that will give you a bit of extra speed

Now about the error. 
Actually it doesn't (not draw) anything, it draws a single pixel. Algorithm fails every time it tries to draw `Triangle( (x1, y1), (x1, y1), (x2, y1) )`. And why exactly: because if the y values of all three points are the same, third point never gets used. 

Let's take another example that fails `[(400, 420), (400, 420), (69, 420)]`. Since y are the same for all of them lines 1503-1516 do nothing. Now those d values, they are actually useless for this example, no need to cover them. And now important part.
x1 gets set to p0x (in this example 400) in line 1523, (`+` part doesn't matter since y values are the same, they are 0, and that is why I said d values don't matter as well). Now next part.
x2 also gets set to p0x. 1526 line condition is false (they are ==) and you go to 1529 that sets x2 to p1x which is also 400. And then you draw the line between (400, 420) and (400, 420) and you "correctly" get a single pixel being colored. This algorithm never uses p2x which would result in drawing the line.
And then the function is over because there is only single y value.

What I would suggest to solve this is the following. In line 1517 add this condition
`if ( (p0y == p1) && (p1y == p2y) && (p0x == p1x) && (p1x != p2x) )`
If this is true, swap p1 and p2. 
Then p0 and p1 will be different and that example will work.

Another thing, after you remove 1531-1532, swap_int function will only ever be used in pair. So replace `swap_int` with `swap_coordinate` or something (which would take 4 arguments and replace both x and y values. which reduces the amount of function calls (but now I am talking about extreme micro-optimization)

TLDR:
1. Remove 1531-1532
2. Add the `if ( (p0y == p1) && (p1y == p2y) && (p0x == p1x) && (p1x != p2x) )` check that would swap second and third coordinate.
3. Replace swap_int with swap_coordinate

MightyJosip (approved):
It is ok now

MyreMylar (approved):
LGTM 👍 

I can confirm the observed speed up on my machine in the given test and that it looks the same in the other posted test program. Passes all tests on my machine and the code seems reasonable.

Issue #343, opened 2023-02-12
Downward-facing triangle edges are not correctly anti-aliased (328)


GalacticEmperor1:
#  **Issue &numero;328 opened by *[illume](https://github.com/illume)* at 2017-03-02 21:05:47**

Originally reported by: **SeanRBurton (Bitbucket: [SeanRBurton](https://bitbucket.org/SeanRBurton), GitHub: [SeanRBurton](https://github.com/SeanRBurton))**

----------------------------------------

When using draw.polygon with gfxdraw.aatrigon to emulate filled, anti-aliased triangles, downward-facing edges are jagged where upward-facing edges are correctly smoothed. I believe this is caused by an asymmetry between upwards and downwards-facing edges in draw.polygon, rather than a defect in aatrigon. Shrinking the solid triangle only partially solves this issue. Unfortunately, this is just one example of the difficulties faced when trying to draw filled, anti-aliased shapes with pygame.

Please see the attached program for an example of the problem.

----------------------------------------
- Bitbucket: https://bitbucket.org/pygame/pygame/issue/328

**Related Docs:** https://www.pygame.org/docs/ref/draw.html# pygame.draw.polygon<hr>

#  Comments

# # #  **[e1000](https://github.com/e1000)* commented at 2018-11-11 21:26:04*

The mentionned code is this : 

```python
import pygame
import pygame.gfxdraw

BLACK = (0, 0, 0)
WHITE = (255, 255, 255)
pygame.init()

h = 600
w = 1000

screen = pygame.display.set_mode((w, h))
screen.fill(WHITE)

x0 = w // 4
y0 = h // 8
y1 = h // 2
tri0 = ((x0, y0), (2 * x0, y1), (3 * x0, y0))
tri1 = tuple((x, h + 10 - y) for x, y in tri0)

for tri in (tri0, tri1):
    pygame.draw.polygon(screen, BLACK, tri)
    (x0, y0), (x1, y1), (x2, y2) = tri
    pygame.gfxdraw.aatrigon(screen, x0, y0, x1, y1, x2, y2, BLACK)

pygame.display.flip()

while True:
    for event in pygame.event.get():
        if event.type == pygame.QUIT:
            exit()
```

The problem stays when we replace `gfxdraw.aatrigon` with `draw.aalines` and is solved when moving one of the aa-lines one pixel down. So we have to investigate  `draw.aaline` vs `draw.polygon`

PR #2131, opened 2023-04-25
Fix incorrectly drawn edges with fillpoly

### LINKED ISSUE #515: Exact expected shape of draw.polygon? (644)
#  **Issue &numero;644 opened by *[e1000](https://github.com/e1000)* at 2018-11-27 20:39:29**

Investigating # 90 I found some y-axis-asymetry when drawing a rhombus (or diamond) shapes; we generate following shapes :

asymetric-diamonds

This was done with following code : 
```
def pxl_to_rect(x, y, size=20):
    return (100 + size * x, size * y, size, size)

def large_map(surface, bg_col, left, right, top, bottom, size=20):
    '''enlarge a piece of the image for immediate vizualization'''
    for x in range(left, right):
        for y in range(top, bottom):
            col = surface.get_at((x, y))
            if col != bg_col:
                pg.draw.rect(surface, col, pxl_to_rect(x, y, size), 0)

def main():
    surface = pg.display.set_mode((800, 600))
    surface.fill(BLACK)

    pg.draw.polygon(surface, GREEN, [(1, 3), (4, 5), (7, 3), (4, 1)])
    pg.draw.polygon(surface, RED, [(11, 4), (13, 7), (15, 4), (13, 1)])
    large_map(surface, BLACK, 0, 20, 0, 20)
```
So the polygon has a tendency to emphasize lower pixel. The reason is probably some (int)-casting in the polygon algorithm... 
So what's the expected shape ? Should we keep it like that or search some top-down-symetric solution ?

**Related Docs:** https://www.pygame.org/docs/ref/draw.html# pygame.draw.polygon<hr>

#  Comments

# # #  **[e1000](https://github.com/e1000)* commented at 2018-11-27 20:44:26*

Btw, the current *Python* and *C* drawing algorithm do not generate the same shape, so it would be nice to fix this with respect to @illume's remark in # 90 to have a library of drawing algorithms...
I would give it a try to see if some usage of `round` could make things symetric.
<hr>

# # #  **[snoyes](https://github.com/snoyes)* commented at 2020-06-19 13:46:54*

Note also that it is only filled polygons which do this.

Top is polygon with width 0. Middle is polygon with width=1. Bottom is draw.lines.
image

<hr>

# # #  **[snoyes](https://github.com/snoyes)* commented at 2020-06-19 16:11:24*

Filled polygons are created by drawing horizontal lines. The algorithm to figure out where the horizontal line should start and stop is:

https://github.com/pygame/pygame/blob/4562396964cca81bceb2d12d3f09aa96746cb3fa/src_c/draw.c# L1856

Integer division truncates towards 0.

For the top left line, (x1, y1) = (4, 1); (x2, y2) = (1, 3)
The second horizontal line starts at the x-coordinate:
(y - y1) * (x2 - x1) / (y2 - y1) + x1
=  (2 - 1) * (1 - 4) / (3 - 1) + 4
= -3 / 2 + 4
= -1 + 4
= 3

For the bottom left line, (x1, y1) = (1, 3); (x2, y2) = (4, 5)
The fourth horizontal line starts at the x-coordinate:
(y - y1) * (x2 - x1) / (y2 - y1) + x1
= (4 - 3) * (4 - 1) / (5 - 3) + 1
= 3 / 2 + 1
= 1 + 1
= 2

Hence the asymmetry. Lines with positive slope will round to the pixel on the right; lines with negative slope will round to the pixel on the left.
<hr>

# # #  **[snoyes](https://github.com/snoyes)* commented at 2020-06-22 00:04:15*

Also note that just adjusting the rounding will not fix this. When the Bresenham line algorithm would draw several horizontal pixels, this scanline method starts in the middle of those. As the edge becomes closer to horizontal, that difference will be increasingly exaggerated. Here the top and bottom rows of the filled polygon are only one pixel when they ought to be seven. Likewise the second row misses three pixels on each side:

image

I'd be more inclined to make a call to the outlined polygon code, and then do some sort of floodfill.

Temmie3754:
Resolves #515 by changing the intersection calculator to round up or down depending on if it is entering/exiting the shape.

### DIFF
--- a/src_c/draw.c
+++ b/src_c/draw.c
@@ -2284,6 +2284,7 @@ draw_fillpoly(SDL_Surface *surf, int *point_x, int *point_y,
     int y, miny, maxy;
     int x1, y1;
     int x2, y2;
+    float intersect;
     /* x_intersect are the x-coordinates of intersections of the polygon
      * with some horizontal line */
     int *x_intersect = PyMem_New(int, num_points);
@@ -2349,12 +2350,16 @@ draw_fillpoly(SDL_Surface *surf, int *point_x, int *point_y,
             if (((y >= y1) && (y < y2)) || ((y == maxy) && (y2 == maxy))) {
                 // add intersection if y crosses the edge (excluding the lower
                 // end), or when we are on the lowest line (maxy)
-                x_intersect[n_intersections++] =
-                    (y - y1) * (x2 - x1) / (y2 - y1) + x1;
+                intersect = (y - y1) * (x2 - x1) / (float)(y2 - y1);
+                if (n_intersections % 2 == 0) {
+                    intersect = floor(intersect);
+                }
+                else
+                    intersect = ceil(intersect);
+                x_intersect[n_intersections++] = intersect + x1;
             }
         }
         qsort(x_intersect, n_intersections, sizeof(int), compare_int);
-
         for (i = 0; (i < n_intersections); i += 2) {
             drawhorzlineclipbounding(surf, color, x_intersect[i], y,
                                      x_intersect[i + 1], drawn_area);


Temmie3754:
This also resolves #343 by making the behaviour consistent between the top and bottom.

ankith26 (commented):
Tested this locally, and while I can confirm that this resolves #515 it somehow makes #343 worse by making both upward and downward triangles have non-smooth edges. Though atleast it's consistent that way now, IG

I quickly tested swapping `floor` and `ceil` in your code, and it actually fixes #343 completely

Temmie3754:
> Tested this locally, and while I can confirm that this resolves #515 it somehow makes #343 worse by making both upward and downward triangles have non-smooth edges. Though atleast it's consistent that way now, IG
> 
> I quickly tested swapping `floor` and `ceil` in your code, and it actually fixes #343 completely

Yes both being non smooth is intended, the aaline has to be adjusted to be in the correct position. Swapping floor and ceil can't be done since it will then make the shapes drawn inconsistent with the non-fill methods (and thus leave the issue of #515).

MyreMylar (approved):
Can confirm that both issues are reasonably fixed.

For the AA outline of a filled shape issue, as the drawing algorithm is now consistent in upward and downwards direction you can adjust the outline in a predictable fashion for both directions, expanding the size of the AA version (so we can see it) and just flipping the y for up versus down. Adapting the original issue code we get:

```
import pygame
import pygame.gfxdraw

BLACK = (0, 0, 0)
RED = (255, 0, 0)
WHITE = (255, 255, 255)
pygame.init()

h = 600
w = 1000

screen = pygame.display.set_mode((w, h))
screen.fill(WHITE)

x0 = w // 4
y0 = h // 8
y1 = h // 2
tri0 = ((x0, y0), (2 * x0, y1), (3 * x0, y0))
tri1 = tuple((x, h + 10 - y) for x, y in tri0)

y_diff = 1
for tri in [tri0, tri1]:
    pygame.draw.polygon(screen, BLACK, tri)
    (x0, y0), (x1, y1), (x2, y2) = tri
    pygame.gfxdraw.aatrigon(screen, x0-2, y0-y_diff, x1, y1+y_diff, x2+2, y2-y_diff, BLACK)
    y_diff = -1


pygame.display.flip()

while True:
        if event.type == pygame.QUIT:
            exit()

```
And smooth triangles whether facing upwards or downwards. I maintain it is still easier to get AA triangles in most cases by just drawing the triangle at 2x resolution and then smooth scaling the surface down to the desired size with transform.

I'm hesitant to spend too much time chasing after gfxdraw as I think we should be trying to eventually deprecate it - moving any useful functionality into draw - to reduce the amount of 'duplicate concept' submodules. It makes little sense to users have a draw submodule, and then another draw submodule for no clear reason. 

For the other issue the shapes now look like this:

image

Which looks much more even in both horizontal and vertical directions to me.

dr0id (approved):
I would welcome if there where some automated tests for this (as there is already code provided in the issue, could just turn that into tests).

Issue #2021, opened 2023-03-14
Add `path` attribute to `pygame.Font`


Starbuck5:
Recently, we decided to add a `name` attribute to Font objects, with an API inspired by `pygame.freetype`. See #1969 and #1971.

The next obvious step is to also emulate the `path` attribute on `pygame.freetype.Font`. A read only attribute to get the path of the loaded font.

Matiiss:
`pygame.freetype.Font` already has a `path` attribute, according to docs at least: https://pyga.me/docs/ref/freetype.html#pygame.freetype.Font.path

Did you mean `pygame.font.Font`?

Starbuck5:
This is a feature request about emulating a property of freetype.Font on font.Font

dr0id:
@Matiiss 
Just saw this request, I have been working on this and some more the last few days, [see my branch in my repo](https://github.com/dr0id/pygame-ce/tree/enhancement/font/is_char_defined-stylename-path). I want to add a charmap example, that's why I haven't opened a PR yet.

dr0id:
PR is open

PR #2132, opened 2023-04-25
Enhancement/font/is char defined stylename path

dr0id:
PR for issue #2021


dr0id:
@Reviewers: please take a careful look for memory leaks, I'm no expert in ref counting. Thanks in advance for any hints.

src_c/font.c
```diff
@@ -942,6 +1030,15 @@ font_init(PyFontObject *self, PyObject *args, PyObject *kwds)
         fontsize = (int)(fontsize * .6875);
     }
 
+    path = pg_EncodeString(obj, "UTF-8", NULL, NULL);
```
Starbuck5, line 1033: Alright. So it seems like the big deal in this method is getting the path out.

Because the path is known in a consistently encoded manner when the SDL_RWop is created. It's just that there is no way to extricate that info.

There might be an elegant solution (for both font and to bolster freetype) with a new variant of `pgRWops_FromObject`. Imagine a `pgRWops_FromObjectWithName()` that has an out parameter for a string, then that can be used instead of encoding the path twice. That could also populate with filelike object logic automatically instead of burdening this code.

src_c/font.c
```diff
@@ -641,6 +696,26 @@ font_getter_name(PyObject *self, void *closure)
...
+font_getter_style_name(PyObject *self, void *closure)
+{
+    TTF_Font *font = PyFont_AsFont(self);
+    const char *font_name = TTF_FontFaceStyleName(font);
+
+    return PyUnicode_FromString(font_name);
+}
+
+static PyObject *
+font_getter_path(PyObject *self, void *closure)
```
Starbuck5, line 709: It seems strange to store it as pyobject bytes and then have to translate it to pyobject str.

Probably how freetype does it because of python2 legacy?

If you stored it as a pyobject str or pyobject None, this method could just be `return (PyFontObject *)self)->path;`

Starbuck5:
Thanks for this!

I've got to commend you for making this work with ftfont too, normally we just skip ftfont for new font features.

I could see this as 3 separate PRs:
- stylename
- is_char_defined
- path

I'm just thinking that could make this easier to review / get merged / move forward if it was broken up. Thoughts?

dr0id:
I will do 3 PRs, got carried away with the changes :D

dr0id:
Thanks for the review!

Starbuck5:
Since you've split out the PRs, is there a reason to keep this one open?

dr0id:
Nope, thanks for the reminder, closing since the changes will find its way via other PR

PR #2133, opened 2023-04-26
Update docs to use pygame-ce versioning

Temmie3754:
Changes the version format strings to use pygame-ce instead of pygame and changes a few static version change references.


Temmie3754:
Updated this to include tags for old pygame versioning and changed docs to reflect it.

Starbuck5 (approved):
Thanks!

PR #2135, opened 2023-04-27
Implement `_sdl2.Renderer.draw_circle` and `.fill_circle`

avaxar:
I haven't tested whether it works yet; builds can't even be imported on my machine, because of some `FRect` problem that can't find specific SDL2 functions, which makes it impossible to `import pygame`. ~~I've also made other `Renderer.draw_*` and `Renderer.fill_*` methods take in statically-typed Cython arguments, which would check the types automatically during runtime when arguments are fed in, therefore making Cython be sure to assume the types of the arguments in the function body, and thereby removing any Python-related overhead.~~ (Migrated to #2039)

If anyone would be so kind, please build and try this example script:
```py
import pygame
import pygame._sdl2


pygame.init()

window = pygame._sdl2.Window("amogus", (800, 600))
renderer = pygame._sdl2.Renderer(window, vsync=True)

running = True
while running:
        if event.type == pygame.QUIT:
            running = False
            break

    renderer.draw_color = (0, 0, 0, 255)
    renderer.clear()

    renderer.draw_color = (255, 255, 255, 255)
    renderer.fill_circle((400, 300), 200)

    renderer.draw_color = (255, 0, 0, 255)
    renderer.draw_circle((400, 300), 100)

    renderer.present()
```


avaxar:
It seems that emscripten's `-Wall` is making the Cython-generated C code impossible to compile, specifically the part where it handles Cython tuple arguments.

oddbookworm:
image
doesn't seem to be working properly

avaxar:
Just realized the problem. I've fixed it now, I think.

mcpalmer1980:
Hi there.

I'm wondering if you might be wasting time with this implementation. Since SDL2 doesn't offer any circle drawing functions, perhaps you should just use pg.draw.circle, copy it into a Texture, and then change the color and destrect values to render it at different sizes and colors.

Obviously this wont help with outlined circles. Anyway just my 1.5 cents.

yunline:
It seems the current fill_circle() algorithm works badly with OpenGL Anti-Aliasing.
image
image
Test Code:
```py
import pygame
import os

pygame.display.init()
os.environ["SDL_RENDER_DRIVER"]="opengl"
pygame.display.gl_set_attribute(pygame.GL_MULTISAMPLEBUFFERS,1)
pygame.display.gl_set_attribute(pygame.GL_MULTISAMPLESAMPLES,4)

win=video.Window()
ren=video.Renderer(win)
ren.draw_blend_mode=1
ren.draw_color=(255,255,255,255)
ren.clear()
ren.draw_color=(127,0,0,127)
ren.fill_circle((200,200),200)
ren.draw_color=(0,127,0,127)
ren.fill_circle((400,200),200)
ren.present()

while 1:
    pygame.event.get()
```
To solve this, try draw triangles instead of lines.

yunline:
It seems the current fill_circle() algorithm works badly with OpenGL Anti-Aliasing.
image
image
Test Code:
```py
import pygame
import os

os.environ["SDL_RENDER_DRIVER"]="opengl"
pygame.display.init()
pygame.display.gl_set_attribute(pygame.GL_MULTISAMPLEBUFFERS,1)
pygame.display.gl_set_attribute(pygame.GL_MULTISAMPLESAMPLES,4)

win=video.Window()
ren=video.Renderer(win)
ren.draw_blend_mode=1
ren.draw_color=(255,255,255,255)
ren.clear()
ren.draw_color=(127,0,0,127)
ren.fill_circle((200,200),200)
ren.draw_color=(0,127,0,127)
ren.fill_circle((400,200),200)
ren.present()

while 1:
    pygame.event.get()
```
To solve this, try draw triangles instead of lines.

avaxar:
@yunline How about now? I reckon that using triangles would be rather slow. In the fix, I used anti-aliased lines instead.

> I'm wondering if you might be wasting time with this implementation. Since SDL2 doesn't offer any circle drawing functions, perhaps you should just use pg.draw.circle, copy it into a Texture, and then change the color and destrect values to render it at different sizes and colors.
@mcpalmer1980 That approach would be slow, as it'd have to copy over pixels to the GPU, even ones that are outside the circle. Plus, that'd waste some memory in the RAM for storing the transit surface, and circles of variably-changing radii would especially be a pain.

yunline:
> How about now? I reckon that using triangles would be rather slow and especially not worth it, given that it'd only need to be strands of lines to fill the circle. In the fix, I used anti-aliased lines instead.

It's a bit better now, bur still not good enough.
**Current image:**
ODQ$VE(WFK0LR_ 57DMJS89
**What I expected:**
6C}HN91)9XIO8@QI9D`GO82

Btw, triangle should be more efficient than line because it fills more pixels per call.

avaxar:
> Btw, triangle should be more efficient than line because it fills more pixels per call.
@yunline In the context of achieving the same precision, I don't think so.

yunline:
> > Btw, triangle should be more efficient than line because it fills more pixels per call.
> 
> @yunline In the context of achieving the same precision, I don't think so.

Sorry about that. Just forget what I said

avaxar:
Actually, there might be a more logarithmically-efficient approach using quads.

image

The left shows how it's currently done using lines. The right shows what I thought of if we were to use quads; it would definitely save more calls, but I would need to research more of whether there are already such preexisting implementations.

yunline:
@avaxar I found another way.
Just use SDL_RenderGeometry. It can draw a polygon once. It can be faster than draw lines/triangles manually.

avaxar:
> Just use `SDL_RenderGeometry`. It can draw a polygon once.

@yunline `SDL_RenderGeometry` accepts triangles, not polygons. The polygon would have to be broken up into triangles which wouldn't solve the aforementioned inefficiency.

avaxar:
I decided to move the changes regarding making other `.draw_*` and `.fill_*` methods accept statically-typed Cython tuple arguments to @yunline's PR on #2039, as to prevent a potential merge conflict. After a private discussion with @yunline, the aforementioned changes here have been removed, and have been migrated and adjusted accordingly in #2039's goals.

MightyJosip:
Wait why exactly are you using stuff like srqt cos sin and similar slow functions

avaxar:
Any alternative methods that you propose which don't require such?

MightyJosip:
https://github.com/pygame-community/pygame-ce/blob/5924b0c6a8b9229edb1141904a2f314764a3bbee/src_c/draw.c#L1987

avaxar:
I don't think the performance could be worth it. I'll close this PR for now.

PR #2136, opened 2023-04-27
Test latest mac "x1" runners

ankith26:


dr0id (approved):

Starbuck5:
According to this article, it's `xl` rather than `x1`. https://github.blog/2023-03-01-github-actions-introducing-faster-github-hosted-x64-macos-runners/

Also says that these runners are not free even in public repositories.

ankith26:
Oh, thanks for pointing me to that link

PR #2140, opened 2023-04-30
french README

bilhox:
Hi,

I don't know if it's necessary, but I wrote a french version of the repos README.
I'm now looking for reviews.


oddbookworm (approved):
LGTM! I don't speak French, but translating it back into English does provide a good-enough result that I'm pretty comfortable approving it

Starbuck5 (approved):

PR #2141, opened 2023-04-30
added Vector2.from_polar and Vector3.from_spherical classmethods

oddbookworm:
Port of https://github.com/pygame/pygame/pull/3744


Starbuck5:
I'm surprised upstream chose to make the same name method have two different signatures in two different scenarios.

Is there a reason that `Vector2().from_polar` shouldn't return itself? That way it could be:
```
Vector2.from_polar((r, phi)) -> Vector2
Vector2().from_polar((r, phi)) -> Vector2
```

And maybe that would have simpler code? The implementation they chose works and is clever, but is a lot of complexity to add.

oddbookworm:
I don't see why we can't have the same signature for both. I'll have to dedicate more time to rewriting it, so it might have to wait until this weekend. I have at least found a way to clean up the stub for the current implementation and I'll push a commit for that

MyreMylar:
Just toggling this to re-run the CI as I think it is ready for review.

MyreMylar (approved):
LGTM 👍 

Good compatibility with upstream fixes.

MyreMylar:
Looks like this will need to merge with main to clear the CircleCI failure.

Starbuck5:
It would be good to get this rebased and have the original author of this be added as a coauthor to the commit.

MyreMylar (approved):

src_c/math.c
```diff
@@ -4223,6 +4273,151 @@ static PyMethodDef _math_methods[] = {
...
+static int
+com_init(PyObject *self, PyObject *args, PyObject *kwds)
+{
+    ClassObjectMethod *com = (ClassObjectMethod *)self;
+    PyObject *cls_callable, *obj_callable;
+
+    if (!PyArg_UnpackTuple(args, "ClassObjectMethod", 2, 2, &cls_callable,
+                           &obj_callable))
+        return -1;
+    if (!_PyArg_NoKeywords("ClassObjectMethod", kwds))
```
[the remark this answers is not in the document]
oddbookworm: Removed

itzpr3d4t0r:
Why not make these functions METH_FASTCALL? Instead of doing it in a separate PR doubling review time we should implement it here i think.

src_c/math.c
```diff
@@ -4351,6 +4442,172 @@ static PyMethodDef _math_methods[] = {
...
     {NULL, NULL, 0, NULL}};
 
+/********************************
+ * ClassObjectMethod Descriptor
+ ********************************/
+
+/* This is a descriptor for a method that have a different functionality
+ * when called from the class and when called from an object. Its
+ * funcionaliy and its use for implementing the from_polar method is
+ * equal to:
```
MyreMylar, line 4452:
```suggestion
/* This is a descriptor for a method that has a different functionality
 * when called from the class and when called from an object. Its
 * functionality and its use for implementing the from_polar method is
 * equal to:
```

MyreMylar (approved):
OK, still looks good to me after recent improvements 👍 

I have little idea of how popular usage of these coordinate systems actually is among pygame-ce users but improving compatibility with upstream where it is doing generally useful things is a plus.

src_c/math.c
```diff
@@ -33,6 +33,7 @@
 #define NO_PYGAME_C_API
 #include "doc/math_doc.h"
 
+#include "base.c"
```
pmp-p, line 36: i don't think this is a good idea

src_c/base.c
```diff
@@ -2296,8 +2366,10 @@ MODINIT_DEFINE(base)
     c_api[26] = pg_TwoDoublesFromFastcallArgs;
```
pmp-p, line 2366: but pg_TwoDoublesFromFastcallArgs is declared a static inline.

pmp-p (changes requested):

buildconfig/stubs/pygame/math.pyi
```diff
@@ -219,6 +219,11 @@ class Vector2(_GenericVector):
     xy: Vector2
     yx: Vector2
     yy: Vector2
+    @overload  # type: ignore
+    @classmethod
+    def from_polar(cls, value: Tuple[float, float], /) -> Vector2: ...
```
aatle, line 224:
```suggestion
    @overload
    def from_polar(value: Tuple[float, float], /) -> Vector2: ...  # type: ignore
```
It looks like the hack for typing fails for the class call. A bit of experimentation gives a different fix, in mypy.
This should also be applied to `from_spherical`.

P.S.: The new `Point` and `SequenceLike` types should be integrated into the PR if it is going to be merged.

bigwhoopgames:
Any update on this? / is this going to be part of the next release? I was reminded of the ticket I made for this recently while reviewing some code.

oddbookworm:
@bigwhoopgames I don't know that this will ever merge. Nobody wants this to merge except as a compat with upstream. The implementation is awful, but there's no better way to do it

bigwhoopgames:
> @bigwhoopgames I don't know that this will ever merge. Nobody wants this to merge except as a compat with upstream. The implementation is awful, but there's no better way to do it

That is unfortunate. Perhaps it could be worked into version 3.0 but in a better way.

Issue #2098, opened 2023-04-05
pygame.transform.skew


JiffyRob:
**Description**
I think it would be nice if pygame offered a `pygame.transform.skew` function (or similar) that would skew an image.  eg:
![image](https://user-images.githubusercontent.com/101531330/229976278-0ba204e5-2f92-46d9-a2cd-75dc056fde85.png)
to
![luxa org-skewed-items(small)](https://user-images.githubusercontent.com/101531330/229976412-d82835ab-8e26-4bf9-a646-d07161b890b2.png)

As for arguments, probably `pygame.transform.skew(surface, axis, amount, background=None)` where `surface` is the surface to skew, `axis` is "x" or "y" for what direction to skew it in, `amount` is amount to skew it by, and `background` is the background color, defaulting to transparent.

yunline:
You can use `_sdl2.video.Texture.draw_quad()` to get the same effect. But copying from Renderer to Surface may be time consuming.

So maybe `transform.textured_quad()` can be a better solution (?

simplyrohan:
I agree with @yunline. Instead of just skewing, I think it would be better if you can input four points for the image to match. I think this is called a perspective warp or transformation. If you are looking for an immediate solution, look at [this](https://github.com/davidpendergast/pygame-utils/blob/main/warp.py#L51) function.

tigercoding56:
> I agree with @yunline. Instead of just skewing, I think it would be better if you can input four points for the image to match. I think this is called a perspective warp or transformation. If you are looking for an immediate solution, look at [this](https://github.com/davidpendergast/pygame-utils/blob/main/warp.py?rgh-link-date=2023-04-05T22%3A16%3A52Z) function.

what if you need more than 4 points?

Issue #2125, opened 2023-04-21
UV coordinates for polygon rendering


ScriptLineStudios:
_**Proposal**_

The ability to specify both a surface and uv coordinates when rendering a polygon, perhaps this would be best moved into a new function:

```python
pygame.draw.polygon_uv(surf_render: pygame.Surface, color: pygame.Color, points: Sequence[Sequence(int, int)], uvs: points: Sequence[Sequence(int, int)], surf_blit: pygame.Surface)
```

gresm:
I want a few clarifications:
What is ``Coordinate`` exactly: a ``Vector2`` compatible object?
``UV coordinates`` terminology is rather used in context about 3d, if that's the case, I believe that functions for drawing 2d should be in separate sub-module than functions for drawing 3d.
What's the intent behind ``surf_render`` and ``surf_blit``? It isn't obvious from just their names.
`points: Sequence[Coordinate], uvs: points: Sequence[Coordinate]` I think something went wrong when typing this...

And likely the most important question for this proposal: use-case - when and how would you imagine that it would look (some example images).

simplyrohan:
@gresm, I have discussed with ScriptLine about this PR, so I will try to clarify.
While the terminology and use cases can be for 3D, this function should be thought about in 2D. Pygame is a 2D library, so a function like this (which works in 2D, not 3D) should stay in the normal library. The `Coordinate`s XY pairs. So yes, you could consider them a `Vector2` compatible object. From my understanding, `surf_render` is the source Surface that should be mapped, and `surf_blit` is the destination.

For use cases other than 3D, this could be useful for generating animation, and also any automatic polygon generation. I will also attach a video (from ScriptLine).
![Example](https://cdn.discordapp.com/attachments/772847140992712734/1097080004212445195/Peek_2023-04-16_10-43.gif)

tigercoding56:
is there a way i can archive this effect right now ?

simplyrohan:
@tigercoding56 , I wrote a triangle rasterization algorithm that performs pretty well. I also noticed your other post about needing more than 4 points. You can achieve that by doing triangulation, which is pretty simple if your shape's vertices are in order of the shape. I think a discussion on this thread may be off-topic. Are you a member of the pygame community discord server? If so I can help you there. Simply open a help post and send me a link

PR #2143, opened 2023-05-04
Adds skew functionality to pygame.transform

Temmie3754:
This adds a skew function to the transform module as requested by #2098 with modifications to support inputs of any 4 coordinates to map the surface to. Do note this is still a WIP so the full implementation for release has not been completed, feel free however to test out the current algorithm implementation and submit feedback on it.


gresm:
Just two comments regarding this pull request:
1. The cython files: scrolling through the diff, it seems that the changes are caused by regeneration of these files with a newer version of cython (0.29.30 -> 0.29.34). I assume that it was accidental, but if it were intentional, this should be done in a different pull request.
2. In #2098 there is discussion regarding how this function should be named, what's the final decision?

Temmie3754:
1. Fixed that.
2. I think using skew() would be most appropriate for this implementation since it doesn't perform a perspective warp operation or a transformation but rather attempts to map all horizontal lines in the original surface to new positions in the new surface by changing the angle and length of each line in the surface. I believe that a quad_transform() naming scheme would be more appropriate for a different implementation that mapped points using a transformation matrix. skew() also fits the naming scheme of the transform module better since quads are not mentioned outside of _sdl2 so using it here may bring about more confusion, especially since this does not work with the quads drawn by the _sdl2 module.

simplyrohan:
> it doesn't perform a perspective warp operation or a transformation but rather attempts to map all horizontal lines in the original surface to new positions in the new surface by changing the angle and length of each line in the surface.

I do see use cases for this, but it may be redundant. [Another thread](https://github.com/pygame-community/pygame-ce/issues/2125) involves mapping a surface to new points. This skew functionality could be achieved using that method. Although that thread hasn't seen much activity...

MyreMylar:
I like the name skew, it is familiar to me for this type of operation (and is the sort of word I'd look for in a transform submodule to get this effect). Photoshop and GIMP use `shear` for this operation, so that is another option if others would prefer that.

I'm happy having a function like this even if you could achieve the same thing with a more complicated warping function as it is likely that users would look for something like this and might not grasp that a more complicated function could be used for the same thing.

If needed we could eventually reduce underlying code duplication by making this API call the more complicated spatial transform with some fixed parameters.

docs/reST/ref/transform.rst
```diff
@@ -101,6 +101,24 @@ Instead, always begin with the original image and scale to the desired size.)
...
+   This maps an image to a new surface warping the image so that its corners
+   match the provided points. When no destination surface is provided, the
+   adjust_size option will change the size of the resulting surface to be the
+   smallest surface the points are capable of fitting in, otherwhise it will
+   use the original size of the provided surface. If a color for the background
+   is not provided then it will behave the same as the rotate function. If the
+   image has pixel alphas, the padded area will be transparent. Otherwise pygame
+   will pick a color that matches the Surface colorkey or the topleft pixel value.
+
+   .. versionadded:: 2.3.0
```
MyreMylar, line 118:
```suggestion
   .. versionadded:: 2.4.0
```

MyreMylar (commented):
Tried this out finally. Unfortunately, there is no test or usage example and when I tired to use it like so:

```python
import pygame


display_surf = pygame.display.set_mode((1280, 600))

walt_2 = pygame.transform.skew(walt, [[0, 0], [417, 0], [0, 364], [417, 364]])
walt_3 = pygame.transform.skew(walt, [[0, 0], [417, 0], [0, 364], [417, 364]])

pygame.init()

clock = pygame.time.Clock()

running = True

while running:
        if event.type == pygame.QUIT:
            running = False

    display_surf.blit(walt, (0, 100))
    display_surf.blit(walt_2, (417, 100))
    display_surf.blit(walt_3, (834, 100))

    pygame.display.flip()

    clock.tick(60)
```

I get a segmentation fault. I may be doing it wrong, but in any case we defintely need some clearer documentation on the four points and the order they are supplied in. I'm assuming it goes - `[top_left, top_right, bottom_left, bottom_right]` but I could be wrong. I'm also assuming the supplied points are in pixels.

Anyway, I think this could be cool if it worked.

docs/reST/ref/transform.rst
```diff
@@ -101,6 +101,24 @@ Instead, always begin with the original image and scale to the desired size.)
...
+   This maps an image to a new surface warping the image so that its corners
+   match the provided points. When no destination surface is provided, the
+   adjust_size option will change the size of the resulting surface to be the
+   smallest surface the points are capable of fitting in, otherwhise it will
+   use the original size of the provided surface. If a color for the background
+   is not provided then it will behave the same as the rotate function. If the
+   image has pixel alphas, the padded area will be transparent. Otherwise pygame
+   will pick a color that matches the Surface colorkey or the topleft pixel value.
+
+   .. versionadded:: 2.4.0
```
damusss, line 118: Should be updated to 2.5.2. You could also use backticks when referring to argument names and perhaps split this in two or three paragraphs.

src_c/transform.c
```diff
@@ -716,6 +1034,154 @@ surf_rotate(PyObject *self, PyObject *args, PyObject *kwargs)
...
+
+    if (surf->w < 1 || surf->h < 1) {
+        Py_INCREF(surfobj);
+        return (PyObject *)surfobj;
+    }
+    if (surf->format->format != newsurf->format->format) {
+        return RAISE(PyExc_ValueError, "surface formats do not match");
+    }
+
+    if (surf->format->BytesPerPixel == 0 || surf->format->BytesPerPixel > 4)
```
damusss, line 1118: Should probably use the `PG_SURF_BytesPerPixel` macro that was added some time ago for compatibility purposes. Might be needed in more places than this one.

src_c/transform.c
```diff
@@ -716,6 +1034,154 @@ surf_rotate(PyObject *self, PyObject *args, PyObject *kwargs)
...
+        if (!rgba_default)
+            return -1;
+        *color = SDL_MapRGBA(format, rgba_default[0], rgba_default[1],
+                             rgba_default[2], rgba_default[3]);
+    }
+    return 0;
+}
+
+static int
+check_inside(SDL_Surface *surf, SDL_Point p)
```
damusss, line 1063: Do you think it's worth to make this small function inline aswell? And perhaps to be consistent it could start with an underscore just like `_copy_pixel_to` or `_color_from_obj`.

damusss (commented):
I have reason to believe you are going to push some commits soon, this is just rather minor things I noticed I thought were worth mentioning.

buildconfig/stubs/pygame/transform.pyi
```diff
@@ -16,6 +16,10 @@ def scale_by(
     dest_surface: Optional[Surface] = None,
 ) -> Surface: ...
 def rotate(surface: Surface, angle: float) -> Surface: ...
+def skew(surface: Surface, points: SequenceLike[Point],
+         bg_color: Optional[Union[ColorLike, None]] = None,
+         adjust_size: Optional[bool] = False,
```
aatle, line 21: `Optional` is misused. (Note: `Optional[X]` is equivalent to `Union[X, None]`.)
For `bg_color`, remove the redundant `Union`, and for `adjust_size`, remove `Optional`.
(Also, format the file with black, for consistent styling.)

src_c/transform.c
Temmie3754: Inlined it 👍

Temmie3754:
Will probably need to write some tests for this and maybe some further changes to the docs but otherwise at this point I think it is fully functional, unless anyone else has other ideas for this PR.

src_c/transform.c
```diff
@@ -716,6 +1034,158 @@ surf_rotate(PyObject *self, PyObject *args, PyObject *kwargs)
...
+                    bgcolor = (((Uint8 *)surf->pixels)[2]) +
+                              (((Uint8 *)surf->pixels)[1] << 8) +
+                              (((Uint8 *)surf->pixels)[0] << 16);
+#endif
+            }
+            bgcolor &= ~surf->format->Amask;
+        }
+    }
+    else {
+        if (_color_from_obj(colorobj, surf->format, NULL, &bgcolor))
```
damusss, line 1169: That function does not handle string colors, making the ColorLike typehint false. To correctly handle all colors you should use this pygame api function.
```suggestion
        if (!pg_MappedColorFromObj(colorobj, surf->format, &bgcolor,
                                   PG_COLOR_HANDLE_ALL))
```

damusss (changes requested):
I think I can make a PR that fixes the same thing for threshold. Now you don't need to worry about the original `_color_from_obj` function.

src_c/transform.c
```diff
@@ -716,6 +1034,159 @@ surf_rotate(PyObject *self, PyObject *args, PyObject *kwargs)
...
+                    bgcolor = (((Uint8 *)surf->pixels)[2]) +
+                              (((Uint8 *)surf->pixels)[1] << 8) +
+                              (((Uint8 *)surf->pixels)[0] << 16);
+#endif
+            }
+            bgcolor &= ~surf->format->Amask;
+        }
+    }
+    else {
+        if (!pg_MappedColorFromObj(colorobj, surf->format, &bgcolor,
```
damusss, line 1169: A PR has been merged that requires the surface itself (not the format). Since you are not using _color_from_obj anymore, you should move it back from where you moved it, which should fix the merge conflicts.

damusss (approved):
LGTM :) (also on phone I can't dismiss my review so this is the only way to remove it. nothing in the code looks strange anyways, and I tested it previously)

PR #2145, opened 2023-05-06
added software rendering keyword to display.set_mode

oddbookworm:
Possible fix for [this upstream issue](https://github.com/pygame/pygame/issues/3835). I can't reproduce the problem though, so I'm not sure if this is what is actually needed. @Jung-PS would you be willing to test your code on a dev wheel for pygame-ce?
Template provided:
```py
import pygame

screen = pygame.display.set_mode((500, 500), software_render=True)

running = True
while running:
        if event.type == pygame.QUIT:
            running = False
    screen.fill("red")
    pygame.display.flip()
```


Starbuck5:
Looking through the SDL source, I think the solution to the upstream issue is to set `SDL_HINT_FRAMEBUFFER_ACCELERATION` to 0. I also think this is a regression in SDL 2.0.22. Specifically it might be https://github.com/libsdl-org/SDL/pull/5296.

They're not using vsync or SCALED, so I don't think a renderer is being created at all.

I don't really know how to get Jung-PS to test that, or how to get this reported to SDL upstream in an actionable way.

oddbookworm:
Closing because there isn't enough info to keep this open with

PR #2146, opened 2023-05-07
Add tests for xm music

Starbuck5:
@Kyuchumimo reported a pygame-ce 2.2.1 regression about xm files on discord. They no longer load (at all). Kyuchumimo and I were both seeing this on Windows. I tracked this down to our SDL_mixer version upgrade (2.6.2 -> 2.6.3), and reported that upstream https://github.com/libsdl-org/SDL_mixer/issues/518.

To resolve this on our end, I'm dropping back down to 2.6.2 on all platforms, and adding a unit test to make sure we support these files.

The unit test revealed that we didn't support these files properly on MacOS either, and I managed to fix that with a slight buildconfig adjustment.


Starbuck5:
@ankith26 I don't understand the ManyLinux CI error messages I'm getting, can you help me out with this?

Starbuck5:
Mysteries right now:
- Linux (debian multiarch and manylinux) fail to load the music from a file object. What does that mean for us? It works on Windows.
- Both ubuntu builds, on the other hand, are trying to push to docker and failing.

PR #2147, opened 2023-05-07
Font warning spellcheck

Sjmarf:
In #1970 a warning was added to `Sysfont` if it couldn't find a matching font:
```
The system font 'Ariel' couldn't be found. Using the default font instead.
```
I've added a simple typo-checker to the warning, which shows potential corrections if the font name that the user entered was one letter different to a correct fontname:
```
The system font 'Ariel' couldn't be found. Did you mean: 'arial'? Using the default font instead.
```

### DIFF
--- a/src_py/sysfont.py
+++ b/src_py/sysfont.py
@@ -18,7 +18,6 @@
 # Pete Shinners
 # pete@shinners.org
 """sysfont, used in the font module to find system fonts"""
-import time
 import warnings
 import os
 import sys
--- a/src_py/sysfont.py
+++ b/src_py/sysfont.py
@@ -391,16 +391,31 @@ def _find_fontname_typos(comparison_word: str) -> str:
     results = []
     comparison_word_len = len(comparison_word)
     for word in itertools.chain(Sysfonts, Sysalias):
-        amount = 0
         word_len = len(word)
-        if comparison_word_len != word_len:
+        if not (word_len - 1 <= comparison_word_len <= word_len + 1):
             continue
-        for i in range(word_len):
+
+        if comparison_word.startswith(word) or word.startswith(comparison_word):
+            results.append(word)
+            continue
+
+        amount = 0
+        for i in range(min(word_len, comparison_word_len)):
             if comparison_word[i] != word[i]:
                 amount += 1
-            if amount == 2:
-                break
-        if amount == 1:
+                if amount == 2:
+                    break
+                if (
+                    word[:i] + word[i + 1 :]
+                    == comparison_word[:i] + comparison_word[i + 1 :]
+                    or word[:i] + word[i + 1 :] == comparison_word
+                    or word == comparison_word[:i] + comparison_word[i + 1 :]
+                ):
+                    results.append(word)
+                    amount = 0
+                    break
+
+        if amount == 1 and comparison_word_len == word_len:
             results.append(word)
 
     if len(results) == 0:
--- a/src_py/sysfont.py
+++ b/src_py/sysfont.py
@@ -407,11 +407,11 @@ def _find_fontname_typos(comparison_word: str) -> str:
     if len(results) == 0:
         return ""
     if len(results) == 1:
-        return f"Did you mean '{results[0]}'? "
+        return f"Did you mean: '{results[0]}'? "
     elif len(results) == 2:
-        return f"Did you mean '{results[0]}' or '{results[1]}'? "
+        return f"Did you mean: '{results[0]}' or '{results[1]}'? "
     else:
-        return "Did you mean '" + "', '".join(results[:-1]) + f"' or '{results[-1]}'? "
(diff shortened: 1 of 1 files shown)


Starbuck5:
This seems like a cool idea.

I don't like the amount of code it adds. Would something like `difflib.get_close_matches` work instead?

Sjmarf:
> This seems like a cool idea.
> 
> 
> 
> I don't like the amount of code it adds. Would something like `difflib.get_close_matches` work with much less code?

Ah, I didn't know about that! I'll implement it later 👍

Sjmarf:
> This seems like a cool idea.
> 
> I don't like the amount of code it adds. Would something like `difflib.get_close_matches` work with much less code?

I've switched to using `difflib.get_close_matches`. I've set it to match fonts that are 2 or less character different to the font the user tried to use. Is this a good, or would you prefer a different value?

Sjmarf:
>>> src_py/sysfont.py:393:0: R0915: Too many statements (52/50) (too-many-statements)

Also Pylint gave me this error, so I've refactored it a little.

MyreMylar:
I've noticed CPython has started producing errors with these 'spelling' check suggestions in (probably several python versions ago - but I'm slow to catch on) so I support us also trying to do so where possible.

src_py/sysfont.py
```diff
@@ -465,9 +473,26 @@ def SysFont(name, size, bold=False, italic=False, constructor=None):
                     f"Using the default font instead."
                 )
             else:
```
Starbuck5, line 475: Minor thing, but it would be nice to have a comment here briefly summarizing what this code is about.

Starbuck5 (approved):
This looks good, thanks!

I left a nitpick for you to resolve if you'd like.

Before merging the commits should probably get squashed down a bit, to keep commit history clean. We could also do a squash and merge to squash when merging. (Also needs another approval before merging, fyi)

src_py/sysfont.py
```diff
@@ -429,30 +460,7 @@ def SysFont(name, size, bold=False, italic=False, constructor=None):
...
-                    if bold and style[0]:
-                        gotbold = True
-                    if italic and style[1]:
-                        gotitalic = True
-                elif not fontname:
-                    fontname = plainname
-                elif plainname != fontname:
-                    gotbold = bold
-                    gotitalic = italic
+            fontname, gotbold, gotitalic = _load_single_font(single_name)
```
novialriptide, line 463: It looks like you didn't make any new changes here.. could you change it back?

src_py/sysfont.py
```diff
@@ -465,9 +473,26 @@ def SysFont(name, size, bold=False, italic=False, constructor=None):
                     f"Using the default font instead."
                 )
             else:
+                matches = difflib.get_close_matches(
+                    _simplename(name[0]), itertools.chain(Sysfonts, Sysalias), 2
+                )
+                if len(matches) == 0:
```
novialriptide, line 479: This can be simplified to just
```python
match_text = f"Did you mean: {', '.join(matches)}"
```

Although the output would be different, so do it if you wish.

src_py/sysfont.py
Sjmarf: The linting told me to split it up like that. I can change it back if you prefer though.

novialriptide: Which linter?

Sjmarf: Pylint

https://github.com/pygame-community/pygame-ce/pull/2147#issuecomment-1546256401

novialriptide: I'm sorry.. I did not see that comment before sending this review. Looks good to me now.

src_py/sysfont.py
Sjmarf: Do you mean removing the quotation marks? In my opinion, we should keep them. I can simplify it a little though:

```
The system font 'comicsans2' couldn't be found. Did you mean: 'comicsans', 'comicsansms'?
```

Using:
```
if matches:
    match_text = f"Did you mean: '" + "', '".join(matches) + "'? "
else:
    match_text = ""
```

src_py/sysfont.py
Sjmarf: Sure 👍 

```py
# Identifies the closest matches to the font provided by the user from the list of available fonts.
```

Starbuck5:
Hey, I see the code needs a format run. Just in case you haven't seen how to do it through the pygame setup.py: https://github.com/pygame-community/pygame-ce/pull/2197#issuecomment-1562430356

Sjmarf:
Just to be clear, this now uses the following error format (there is no 'or') at Novial's request:

```
The system font 'comicsans2' couldn't be found. Did you mean: 'comicsans', 'comicsansms'?
```

I'm happy for this to be merged, unless anyone else has thoughts on the format.

MyreMylar (approved):
LGTM 👍 All seems to work in my testing!

Thanks for the PR 🚀

PR #2152, opened 2023-05-09
Fixes for tracker music support, add test

Starbuck5:
Continued from #2146 

Problem: pygame-ce 2.2.1 has a regression where XM music doesn't work on Windows.

Regression caused by SDL_mixer, switching back down to 2.6.2 fixes it. Reported upstream: https://github.com/libsdl-org/SDL_mixer/issues/518.

Solution: go back to using SDL_mixer 2.6.2 (on all platforms, for consistency), and add a unit test to ensure support on all platforms.

Problem 2: apparently xm music wasn't working on Mac? -> fixed by adjusting the build flags to disable shared libmodplug like all the other audio codec libraries.

Problem 3: manylinux build was doing something with docker that fails on PR branches unless they are direct branches of pygame-ce, so had to move my stuff to this new branch.


ankith26 (commented):
So if I understand what's going on here as of now, just removing mp3 tests somehow makes everything work again including `xm` loading? This is so suspicious hehe

Starbuck5:
> So if I understand what's going on here as of now, just removing mp3 tests somehow makes everything work again including xm loading? This is so suspicious hehe

Just to be clear, I removed mp3 tests after adding them also in this PR. In the overall diff of this PR, no mp3 tests were removed.  It's not that this PR removes mp3 tests.

I opened an issue about the mp3 loading failing (#2154), and cool-guy-dev confirmed this problem does not occur on real built pygame wheels. It seems to only be a problem when getting SDL_mixer from apt on the Ubuntu sdist runs.

This PR is totally ready for review, just focusing on `xm` music. (As a proxy for all the different tracker music formats, like `mod` and `it`).

novialriptide:
LGTM, all tests pass on my machine.
image

novialriptide (approved):

MyreMylar (approved):
LGTM 👍 

The code changes fixes the failing xm tests on Windows for me.

Issue #2153, opened 2023-05-09
`abs` support for getting vector length/magnitude


Matiiss:
Simple idea, just like in math `|v|` symbolizes a vector's magnitude, in pygame `abs(v)` could also return the magnitude of the vector. Essentially that would be equivalent to `v.length()` or `v.magnitude()` but shorter and more mathy (and might as well really).

I would implement it if this gets accepted.

gresm:
People use ``|v|`` with different meaning, there's even a notation ``||v||`` for length of a vector. ``|v|`` could also mean a vector with every value absolute. (Based on [this](https://math.stackexchange.com/a/1346795/1145683).)

Matiiss:
@JiffyRob  `abs(v.elementwise())` already exists and returns a vector with all components being absolute values of whatever they were previously

Matiiss:
Won't get implemented now as it seems unnecessary and just adds another method on top of the two existing ones for getting the length of a vector, besides it's ambiguous.

PR #2157, opened 2023-05-13
Added `abs` support for returning vector magnitude

Matiiss:
Fixes https://github.com/pygame-community/pygame-ce/issues/2153


Starbuck5:
I think it's weird for abs() to do magnitude instead of just absolute-ifying each element.

Matiiss:
But `abs(v.elementiwise())` already does that, this would be like duplicating an existing feature. And `|v|` tends to be used for vector magnitudes.

Starbuck5:
> But abs(v.elementiwise()) already does that, this would be like duplicating an existing feature. And |v| tends to be used for vector magnitudes.

Several ways to do something? What is this, pygame?

MyreMylar:
> I think it's weird for abs() to do magnitude instead of just absolute-ifying each element.

Googling around other Vector3/Vector2 implementations it seems about 50/50 whether `abs(vector)` (or `vector.abs()`) returns the length scalar, or another vector with each x,y,z element absifyed.

I don't really care about this, as I think I would personally use `.length' first before `.magnitude` and then `abs(vector)` if it existed dead last for achieving the goal of obtaining the length of a vector.

On the other hand, I can't think of a particularly compelling reason to ever get the elementwise `abs()` of a Vector right now. Is there one?

I would just side with whichever is most popular if there is any evidence either way.

PR #2161, opened 2023-05-15
Raise an error when source and destination surfaces are the same (in `box_blur` and `gaussian_blur`).

### LINKED ISSUE #2102: `transform.box_blur` works incorrectly.

yunline:
Continue from #2103
Fix #2102

### DIFF
--- a/src_c/transform.c
+++ b/src_c/transform.c
@@ -3144,6 +3144,11 @@ blur(pgSurfaceObject *srcobj, pgSurfaceObject *dstobj, int radius,
         retsurf = pgSurface_AsSurface(dstobj);
     }
 
+    if (retsurf->pixels == src->pixels) {
+        return RAISE(PyExc_ValueError,
+                     "Source and destination surfaces are the same.");
+    }
+
     if ((retsurf->w) != (src->w) || (retsurf->h) != (src->h)) {
         return RAISE(PyExc_ValueError,
                      "Destination surface not the same size.");
--- a/src_c/transform.c
+++ b/src_c/transform.c
@@ -3150,8 +3150,11 @@ blur(pgSurfaceObject *srcobj, pgSurfaceObject *dstobj, int radius,
     Uint8 *src_end = src_start + src->h * src->pitch;
     if ((ret_start <= src_start && ret_end >= src_start) ||
         (src_start <= ret_start && src_end >= ret_start)) {
-        return RAISE(PyExc_ValueError,
-                     "Source and destination surfaces are overlapping.");
+        return RAISE(
+            PyExc_ValueError,
+            "Blur routines do not support dest_surfaces that share pixels "
+            "with the source surface. Likely the surfaces are the same, one "
+            "of them is a subsurface, or they are sharing the same buffer.");
     }
 
     if ((retsurf->w) != (src->w) || (retsurf->h) != (src->h)) {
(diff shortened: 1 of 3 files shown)


Matiiss (changes requested):
This should probably have a note in the docs that passing the calling surface as destination surface would raise an exception. Otherwise looks good to me.

Matiiss (approved):
LGTM 👍

src_c/transform.c
```diff
@@ -3144,6 +3144,11 @@ blur(pgSurfaceObject *srcobj, pgSurfaceObject *dstobj, int radius,
         retsurf = pgSurface_AsSurface(dstobj);
     }
 
+    if (retsurf->pixels == src->pixels) {
```
Starbuck5, line 3147: Is this a viable way of testing that surfaces have the same data?

Like I believe subsurfaces might have a `->pixels` pointer within the existing surface data region.

Could it test if the pixels regions overlap? Like something around `if (retsurf->pixels <= src->pixels && src->pixels <= retsurf->pixels + retsurf->pitch * retsurf->h) {}`

I'm not telling you to do this, this is a genuine question.

Matiiss: My understanding is that since pixel addresses are compared no two different surfaces would have the same pixel address, right? So, only the same surface would succeed that check. As for subsurfaces, they also pass that check and an exception is raised. Which makes sense as that's still the same surface under the hood.

Starbuck5: As I suspected, subsurfaces do have different addresses:

```py
>>> import pygame
pygame-ce 2.3.0.dev3 (SDL 2.26.4, Python 3.9.5)
>>> a = pygame.Surface((500,50))
>>> b = a.subsurface((10,10,100,30))
>>> b
<Surface(100x30x32)>
>>> a
<Surface(500x50x32)>
>>> a._pixels_address
1743672180768
>>> b._pixels_address
1743672200808
>>>
```

yunline: > As I suspected, subsurfaces do have different addresses:

Yes, but they are in different size, which will be caught by here:
https://github.com/pygame-community/pygame-ce/blob/2a21a5db2ba5b4180f948fd952b8423221e6a552/src_c/transform.c#L3152-L3155

src_c/transform.c
```diff
@@ -3144,6 +3144,16 @@ blur(pgSurfaceObject *srcobj, pgSurfaceObject *dstobj, int radius,
...
     }
 
+    Uint8 *ret_start = retsurf->pixels;
+    Uint8 *ret_end = ret_start + retsurf->h * retsurf->pitch;
+    Uint8 *src_start = src->pixels;
+    Uint8 *src_end = src_start + src->h * src->pitch;
+    if ((ret_start <= src_start && ret_end >= src_start) ||
+        (src_start <= ret_start && src_end >= ret_start)) {
+        return RAISE(PyExc_ValueError,
+                     "Source and destination surfaces are overlapping.");
```
Starbuck5, line 3154: The fact that they're overlapping is not a good error message, imo.

"Blur routines do not support dest_surfaces that share pixels with the source surface. Likely the surfaces are the same, one of them is a subsurface, or they are sharing the same buffer."

^ Something like this?

I tried to put it in words that would make sense with the context of the pygame API.

Starbuck5 (approved):
Left some error message feedback to resolve before merge, otherwise looks good.

PR #2163, opened 2023-05-15
Reveal `.. note::` reST block background color again

Mega-JC:
The initial change was made while doing [docs style improvements](https://github.com/pygame/pygame/commit/a428c87b9644cf681742599b8954fd6b8195b815), however it is ultimately more beneficial and widespread to have `.. note::` reST blocks stand out. It makes them harder to overlook, thus making it less likely for users to miss important details.

### Before
image

### After
image

### DIFF
--- a/docs/reST/themes/classic/static/pygame.css_t
+++ b/docs/reST/themes/classic/static/pygame.css_t
@@ -694,7 +694,7 @@ p.admonition-title:after {
     content: ":";
 }
 
-dl.definition div.note, dl.definition div.seealso {
+dl.definition div.seealso {
     background: none;
     border: none;
 }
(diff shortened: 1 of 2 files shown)


Starbuck5:
This increases the amount of emphasis the notes get, which is interesting. It's good in some cases, but I feel in other cases the previous content is now emphasized too much. So they should be changed away "notes" in that case.

Some candidates that might not want to be notes after this PR:
draw: `see the pygame.gfxdraw...`
font: every getter/setter note
controller: `see the pygame.joystick...`

Starbuck5:
This also will have implications for #2122. Since I believe this PR will be merged before that one, that will need to be rebased and updated to support these changes. Heads up @Notenlish

MyreMylar (approved):
I like how this looks in general. I think we can handle demoting blocks from being attention grabbing notes in other PRs over time.

itzpr3d4t0r:
I think this doesn't work well with dark theme right now:
image
VS light theme:
image

Mega-JC:
I've now added custom theme colors for the dark mode as well.  

image

zoldalma999 (approved):
Looks good to me, thanks!

Issue #2162, opened 2023-05-15
Some idea about `video.Window.grab`


yunline:
`grab` is an attribute represent the status of grabbed mode and should never change unless user changes it.

However, when the window loses focus, the `grab` attribute change to `False` but still in the grabbed mode . When gaining focus, the `grab` attribute will change back to `True`.

This is not expected, because sometimes when `grab` is `False`, user still needs to call `grab=False` to quit grabbed mode.

So I suggest, remove `Window.grab`, add `Window.grab_mode` and `Window.grabbed`.

- `Window.grab_mode` is readable and writable. 
- `Window.grabbed` is read only.

|Condition|`Window.grab_mode` |`Window.grabbed`|
|-|-|-|
|`Window.grab_mode=False` |`False`|`False`|
|`Window.grab_mode=True`, but the mouse is not confined to the window |`True`|`False`|
|`Window.grab_mode=True`, and the mouse is confined to the window |`True`|`True`|

Getter/setter version:
- `Window.get_grabbed()`
- `Window.get_grab_mode()`
- `Window.set_grab_mode()`

yunline:
Current Test code:
```py
# Use Alt+Tab to switch window

import pygame
from pygame._sdl2 import video

w1=video.Window(position=(200,200),)
w2=video.Window(position=(300,300),size=(200,200))
w1.grab=1
w2.grab=1

while 1:
    pygame.event.get()
    print(f"w1_grab:{w1.grab},    w2_grab:{w2.grab}")
```

Starbuck5:
What are the SDL equivalents of these ideas?

yunline:
Fixed by #2379

PR #2164, opened 2023-05-16
Remove `Window.grab`, add `Window.grab_mode` and `Window.grabbed`

yunline:
Issue #2162
So I suggest, remove `Window.grab`, add `Window.grab_mode` and `Window.grabbed`.

* `Window.grab_mode` is readable and writable.
* `Window.grabbed` is read only.

|Condition|`Window.grab_mode`|`Window.grabbed`|
|-|-|-|
|`Window.grab_mode=False`|`False`|`False`|
|`Window.grab_mode=True`, but the mouse is not confined to the window|`True`|`False`|
|`Window.grab_mode=True`, and the mouse is confined to the window|`True`|`True`|


yunline:
I've made a video to demonstrate how this two attributes work.
The window on the left is set to grab_mode but not always grabbed.

https://github.com/pygame-community/pygame-ce/assets/31395137/9fb0546f-48c9-417c-b494-e9776d18435d

yunline:
Close because #2379 implemented this.

Issue #2127, opened 2023-04-23
Missing wrapLength argument in ``pygame.font.Font.size``


bilhox:
Hello,

Recently, you have introduced ``wraplength`` for fonts when rendering with ``pygame.font.Font.render``.
That's why, we should have a ``wraplength`` parameter for ``pygame.font.Font.size`` so it gives us the right height.

Something like this :
```py
pygame.font.Font.size(text : str, wraplength : int)
```

Matiiss:
This method should also take newline characters into account.

PurityLake:
Might be tricky enough, especially using utf-8 encoding. I can probably take a stab at it but I'll see what I can do.

yuiio:
> This method should also take newline characters into account.

I agree with @Matiiss . One point of the key points of `pygame.font.Font.size`  is being able to anticipate the length of a string before rendering it. The result of the size obtained and the size of the rendering should be 100% identical, even with line breaks, in order to operate correctly.

PR #2166, opened 2023-05-16
Add `Font.size_wrapped` to the class

PurityLake:
Refers to #2127 

I have currently implemented a way to get the width and height of a wrapped piece of text and put it in a method called `size_wrapped` in the `Font` class.

I referred heavily to the SDL TTF implementation of how `TTF_Render_Wrapped_Internal` gets it's height and width of a surface (minus the drawing of course). I have provided tests, documentation and stubs. 

I'd love to hear opinions on if anything needs to change.


Starbuck5:
When you were bringing this up in discord I never understood why this is going into a new function rather than `size()`. Multiline text didn't go into `render_wrapped()`, it went into `render()`.

And is this something that should be added back into SDL_ttf itself?

PurityLake:
So my reasoning was for supporting using a wrap length of 0, in the case the user wants to get the size of the text using new line characters. Although it would be just as possible to maybe default to None and try cast to int if it isn't None.

I mean I guess I could upstream it, I'll check if there isn't already an open PR for this and make an issue and PR for it myself.

PurityLake:
As an update to this, I will be going about adding my code to SDL_ttf over the next little while, will link the PR here when I create it.

MyreMylar (commented):
@PurityLake did you decide to add this to SDL_ttf or should we include it in pygame-ce based on this PR?

PurityLake:
@MyreMylar for now yes. I have had less time to be able to add the feature to SLD_ttf than anticipated. I still plan to add it but no guarantee when that will happen. When I do so I will cut this code out at a later date.

MyreMylar:
This has now been [added to SDL_TTF](https://github.com/libsdl-org/SDL_ttf/commit/d1b61f38fc0e521491883f7204294546afb12758) as `TTF_GetTextSizeWrapped()` so we can probably close this PR and open a new on to wrap that function once we are on the correct version of SDL_TTF some time in the future.

PurityLake:
Agreed

PR #2170, opened 2023-05-17
Clean up ScancodeWrapper

zoldalma999:
Back when PurityLake added errors to `ScancodeWrapper`'s iter functions, I noticed that the code for the class is repetitive at places, and could be cleaned up. So I did it (about 3 months ago, and then never made a pr out of it). Technically this implementation should also be faster at creating the object as well, but since this is a small optimization, it is not like this is going to speed up anything significantly. Accessing times should be around the same though (again, not that it matter much).


zoldalma999:
I'll be closing this for now, but I do plan to reopen something like this in the future. I will, however, definitely wait for #1912 to be merged.

PR #2172, opened 2023-05-18
Re-allow destsurf scale w/ diff but compatible pixelformats

### LINKED ISSUE #2116: Bug when I scale a surface to destSurface
The problem is very simple, I have a module for pygame that works with scales, and when I apply the scales, the following error appears:

image

You can see that where the arrows are, is the 'format' of the surface, both are 32, and the error is displayed on the screen.

To clarify, the Display is a surface that I draw to later scale to the Window, which is the surface of the "display" module.

[]
pygame-ce 2.2.1 (SDL 2.26.4, Python 3.10.11)

Starbuck5:
Fixes #2116 

So, in #1884 I replaced our bespoke scaling code with a call to SDL_SoftStretch. I also removed our explicit source bitdepth == dest bitdepth check when a dest surface is provided, because I knew SDL_SoftStretch would do that error for us.

What I failed to take into account is that there are in fact times when a dest surface would be used (with the same bitdepth) but with a different pixelformat.

For example, scaling a RGBA image onto an RGBX destination surface. It was able to handle that before, and a few people used that functionality.

So I had a few ideas:
- Revert my previous changes (This would get rid of the speedups too, and people like speedups)
- Trick SDL_SoftStretch into thinking it had a surface of one type when it didn't, by overwriting the pixelformat and then replacing it later (Hacky)
- Give SDL_SoftStretch a real surface to modify with the equivalent pixelformat.

I went with the third option, using CreateSurfaceFrom to share the pixel data with the existing destination surface.

### DIFF
--- a/src_c/transform.c
+++ b/src_c/transform.c
@@ -413,6 +413,7 @@ scale_to(pgSurfaceObject *srcobj, pgSurfaceObject *dstobj, int width,
 {
     SDL_Surface *src = NULL;
     SDL_Surface *retsurf = NULL;
+    SDL_Surface *modsurf = NULL;
     int stretch_result_num = 0;
 
     if (width < 0 || height < 0)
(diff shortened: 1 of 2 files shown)


src_c/transform.c
```diff
@@ -442,11 +463,15 @@ scale_to(pgSurfaceObject *srcobj, pgSurfaceObject *dstobj, int width,
         pgSurface_Lock(srcobj);
         Py_BEGIN_ALLOW_THREADS;
 
-        stretch_result_num = SDL_SoftStretch(src, NULL, retsurf, NULL);
+        stretch_result_num = SDL_SoftStretch(src, NULL, modsurf, NULL);
```
ankith26, line 466: As I understand, this modified `modsurf`. But `modsurf` is freed, and this function actually returns `retsurf`?

Starbuck5: Mod surf and retsurf use the same pixel buffer.

ankith26: Oh, I assumed the pixel buffer would be copied. But yeah, shared pixel buffer makes sense

ankith26 (approved):
cool fix, thankss :star2:

Temmie3754 (changes requested):
Whilst this does solve the case of RGBX sources it does not fix BGRA cases (and possibly others) which the method prior to 2.2.1 handled fine so it will likely need to be modified to support those as well.

Starbuck5:
What kind of BGRA cases?

If you're trying to scale a BGRA surface onto a RGBA surface it should fail. It "worked" before but would produce incorrect output.

Temmie3754 (approved):
Cases where the destination format was BGRA which would result in altered colours in the old method.
Though if that was unintentional behaviour of the old function then everything else should be good to go.

Starbuck5:
My thought is that it’s better to explicitly fail in that case rather than give such weird output.

Temmie3754:
It likely is, I can't see many scenarios where someone would intentionally make use of the scaling methods to also change the displayed colour at the same time, there's a greater chance someone would accidentally cause that problem when scaling which the error message would help with.

PR #2173, opened 2023-05-18
Manually bump actions/cache, minor usage updates

ankith26:
Should fix the deprecation warning, and ensure the action continues to run.
Also takes the opportunity to do a few minor usage updates to save a tiny bit of CI time, and make it a bit more robust.

Not really sure why dependabot missed this action while it takes care of all other actions, if someone could shed light on this would be nice


Starbuck5 (approved):
Glad to see those warnings resolved.

zoldalma999 (approved):
Good bo... I mean thanks Ankith!

PR #2174, opened 2023-05-18
Optimize vector coords memory

Starbuck5:
Why do a malloc (+error checking) and a free when we don't need to?

This also improves memory usage on Vector3s (by one double), while being neutral on Vector2s.

Before Vector2 (relevant bits):
1 double*   (size = 8 bytes)
double[2] allocated     (size = 16 bytes)

After Vector2 (relevant bits):
double[3] in struct    (size = 24 bytes)

Before Vector3 (relevant bits):
1 double*   (size = 8 bytes)
double[3] allocated     (size = 24 bytes)

Now Vector3 (relevant bits):
double[3] in struct    (size = 24 bytes)

Without adjusting the float store gcc setting to apply on more modern compilers (as I believe was the intent behind the original code), I got a single rounding-error type failure on manylinux i686.

### DIFF
--- a/src_c/math.c
+++ b/src_c/math.c
@@ -55,7 +55,7 @@
 #endif /* M_PI_2 */
 
 #define VECTOR_EPSILON (1e-6)
-#define VECTOR_MAX_SIZE (4)
+#define VECTOR_MAX_SIZE (3)
 #define STRING_BUF_SIZE_REPR (110)
 #define STRING_BUF_SIZE_STR (103)
 #define SWIZZLE_ERR_NO_ERR 0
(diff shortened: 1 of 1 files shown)


Starbuck5:
This is my test script:
```py
import time, pygame

start = time.time()
for _ in range(10000000):
    a = pygame.Vector2(1, 5)
    b = pygame.Vector3(25, 2, -3.2)

print(time.time() - start)
```

It's hard to say exactly because of variability, but this simple change gets 1.5-4% improvement on this.

Not a breathtaking optimization maybe, but it does this with less code complexity and sometimes less memory usage. Plus Vector creation can easily happen millions of times in someone's game.

MyreMylar (approved):
LGTM 👍

PR #2175, opened 2023-05-18
add style_name property to font and freetype font

dr0id:
partial PR as requested in this PR #2132


src_c/font.c
```diff
@@ -845,6 +854,8 @@ font_set_direction(PyObject *self, PyObject *arg, PyObject *kwarg)
  */
 static PyGetSetDef font_getsets[] = {
     {"name", (getter)font_getter_name, NULL, DOC_FONT_FONT_NAME, NULL},
+    {"stylename", (getter)font_getter_style_name, NULL,
```
narilee2006, line 857: `style_name` instead?

dr0id: @narilee2006  renamed

docs/reST/ref/font.rst
```diff
@@ -209,6 +209,17 @@ solves no longer exists, it will likely be removed in the future.
 
       .. ## Font.name ##
 
+   .. attribute:: style_name
+
+      | :sl:`Gets the font's style_name.`
+      | :sg:`style_name -> str`
+
+      Read only. Returns the font's style name.
```
Starbuck5, line 217: It would be nice to give some example outputs to contextualize what a style name is. 

I wonder if there's a complete list of style names in freetype? (Nope: https://freetype.org/freetype2/docs/reference/ft2-base_interface.html#ft_facerec)

From testing I got things like `Regular`, `Narrow`, `Bold`.

src_c/_freetype.c
```diff
@@ -1100,6 +1104,16 @@ _ftfont_getname(pgFontObject *self, void *closure)
     return PyObject_Repr((PyObject *)self);
 }
 
+static PyObject *
+_ftfont_getstylename(pgFontObject *self, void *closure)
+{
+    if (pgFont_IS_ALIVE(self)) {
+        const char *stylename = _PGFT_Font_GetStyleName(self->freetype, self);
+        return stylename ? PyUnicode_FromString(stylename) : 0;
```
Starbuck5, line 1112: If stylename is NULL, the method returns NULL?

That will raise a Python SystemError because it will be returning NULL without setting an error.

But it raises the question of what to return if the Font doesn't have a stylename-- `None`? empty string?

src_c/_freetype.c
```diff
@@ -1100,6 +1104,16 @@ _ftfont_getname(pgFontObject *self, void *closure)
...
+    }
+    return PyObject_Repr((PyObject *)self);
```
Starbuck5, line 1114: I don't get this line.

src_c/font.c
```diff
@@ -645,6 +645,15 @@ font_getter_name(PyObject *self, void *closure)
     return PyUnicode_FromString(font_name);
 }
 
+static PyObject *
+font_getter_style_name(PyObject *self, void *closure)
+{
+    TTF_Font *font = PyFont_AsFont(self);
+    const char *font_name = TTF_FontFaceStyleName(font);
```
Starbuck5, line 652: In SDL_ttf:  https://github.com/libsdl-org/SDL_ttf/blob/SDL2/SDL_ttf.c#L3030, this is the same as the freetype call you make in freetype.

So it should also be NULL checked the way the other one is.

Also we technically support down to SDL_ttf 2.0.11 right now, and this function was added in SDL_ttf 2.0.12. We might as well just raise the dependency version. (You can do that in this PR, the dep versions are in the README)

Starbuck5:
Thanks for the PR!

One additional thing I noticed is that this should have some small unit tests added.

src_c/_freetype.c
dr0id: I think it returns just the string repr() of self.

src_c/_freetype.c
dr0id: done

docs/reST/ref/font.rst
dr0id: updated

dr0id:
I have added a test for the new property

src_c/_freetype.c
Starbuck5: I mean why it's there. You're emulating the other getter I guess? I still wonder why it's there at all.

src_c/font.c
```diff
@@ -642,7 +642,16 @@ font_getter_name(PyObject *self, void *closure)
...
+    return font_name ? PyUnicode_FromString(font_name) : NULL;
+}
+
+static PyObject *
+font_getter_style_name(PyObject *self, void *closure)
+{
+    TTF_Font *font = PyFont_AsFont(self);
+    const char *font_style_name = TTF_FontFaceStyleName(font);
+
+    return font_style_name ? PyUnicode_FromString(font_style_name) : NULL;
```
Starbuck5, line 654: If font_style_name is NULL this will return NULL, which will cause Python to have a SystemError because you've returned NULL without an exception set.

src_c/font.c
Starbuck5: Dependency version still needs to be updated.

src_c/font.c
dr0id: might this be the reason why the name property returned 0 instead of NULL? I have been copying that method. Not sure what or which error I should set.

src_c/_freetype.c
dr0id: yes, I copied from the name getter. I dunno, ask the original author? But I agree, its probably not a good idea. Should it raise an exception instead (same as in the NULL case?)?

src_c/font.c
dr0id: updated

src_c/font.c
dr0id: Maybe I should just return an empty string as I do in the freetype code.

dr0id:
@Starbuck5 @narilee2006 @novialriptide @ankith26  
updated, please review again

docs/reST/ref/font.rst
```diff
@@ -211,6 +211,22 @@ solves no longer exists, it will likely be removed in the future.
...
+      | :sg:`style_name -> str`
+
+      Read only. Returns the font's style name. Style names are arbitrary, here are some examples:
+
+        'Black', 'Bold', 'Bold Italic', 'BoldOblique', 'Book', 'BookOblique', 'Condensed', 'Condensed Oblique',
+        'ExtraLight', 'Italic', 'Light', 'LightOblique', 'Medium', 'MediumOblique', 'Oblique', 'Regular',
+        'Semibold', 'Semilight', 'Slanted'
+
+
+      .. versionadded:: 2.3
```
Starbuck5, line 226: 2.3.1 now, same with the other versionadded.

src_c/_freetype.c
Starbuck5: Yes, raising an exception there seems like the best idea.

src_c/font.c
```diff
@@ -656,7 +656,26 @@ font_getter_name(PyObject *self, void *closure)
...
+{
+    TTF_Font *font = PyFont_AsFont(self);
+    const char *font_style_name = TTF_FontFaceStyleName(font);
+
+    if (font_style_name) {
+        return PyUnicode_FromString(font_style_name);
+    }
+    else {
+        return RAISE(pgExc_SDLError, TTF_GetError());
+    }
```
Starbuck5, line 678: I think this error checking is unnecessary. TTF_FontFaceStyleName doesn't mention any error condition, and on SDL2_ttf's side it's just a direct access into the freetype font face struct.

src_c/font.c
```diff
@@ -656,7 +656,26 @@ font_getter_name(PyObject *self, void *closure)
...
+    if (font_name) {
+        return PyUnicode_FromString(font_name);
+    }
+    else {
+        return RAISE(pgExc_SDLError, TTF_GetError());
+    }
+}
+
+static PyObject *
+font_getter_style_name(PyObject *self, void *closure)
```
Starbuck5, line 668: I've just noticed this will segfault if you quit the font module. Also font_getter_name will as well.

```py
import pygame
pygame.init()
a = pygame.Font(None, 32)
pygame.font.quit()
a.style_name
```

Honestly this is such an edge case I'd be fine merging this without fixing it, especially because it seems like a pain to fix it. Just as long as we open an issue to track it and come back deliberately.

src_c/font.c
dr0id: if you say so, removed

src_c/font.c
dr0id: hmm, why should 'a' (in your example) still be a valid pointer after 'pygame.font.quit()'? Which code part does 'own' the pointer? The module code or the users code?

I agree, it is probably an edge case, not sure how often it will happen.

zoldalma999: Isn't this just #1506 again? The linked PR does add a check for this to some functions, but probably all of them need it (or most anyway).

dr0id: @zoldalma999 thanks for the hint. Yes it looks like it would be the same, will implement it the same way then.

dr0id: @Starbuck5 @zoldalma999 
would these kind of tests make sense?

```
    def test_font_method_should_raise_exception_after_quit(self):
        methods = [
            ("get_ascent", tuple()),
            ("get_bold", tuple()),
            ("get_descent", tuple()),
            ("get_height", tuple()),
            ("get_italic", tuple()),
            ("get_linesize", tuple()),
            ("get_strikethrough", tuple()),
            ("get_underline", tuple()),
            ("metrics", ("any text",)),
            ("render", ("hello", True, "white")),
            ("size", ("any text",)),
            # ("set_bold", (True,)),
            # ("set_direction", ("is it text",)),
            # ("set_italic", (True,)),
            # ("set_script", ("is it other text",)),
            # ("set_strikethrough", (True,)),
            # ("set_underline", (True,)),
        ]
        font = pygame_font.Font(None, 10)
        actual_names = [n for n in sorted(dir(font)) if callable(getattr(font, n)) and not n.startswith("__")]
        expected_names = [n for n,a in sorted(methods)]
        self.assertEqual(len(expected_names), len(actual_names)-6, msg=f"ensure all methods are checked {expected_names} != {actual_names}")

        for name, args in methods:
            with self.subTest(name=name):
                # arrange
                font = pygame_font.Font(None, 16)
                pygame_font.quit()
                pygame_font.init()

                # act / assert
                self.assertRaises(pygame.error, getattr(font, name), *args)
                # self.assertRaises(pygame.error, font.render, "hello", True, "white")


    def test_font_property_should_raise_exception_after_quit(self):
        properties = [
            ("align", 1),
            ("bold", True),
            ("italic", True),
            ("name", None),
            ("strikethrough", True),
            ("style_name", None),
            ("underline", True),
        ]

        font = pygame_font.Font(None, 10)
        actual_names = [n for n in sorted(dir(font)) if not callable(getattr(font, n)) and not n.startswith("__")]
        expected_names = [n for n,a in sorted(properties)]
        self.assertEqual(len(expected_names), len(actual_names), msg=f"ensure all methods are checked {expected_names} != {actual_names}")

        for name, arg in properties:
            with self.subTest(name=name):
                # arrange
                font = pygame_font.Font(None, 16)
                pygame_font.quit()
                pygame_font.init()

                # act / assert
                self.assertRaises(pygame.error, getattr, font, name) # read
                if arg is not None:
                    self.assertRaises(pygame.error, setattr, font, name, arg) # write
```

Which gives me 13 FAILS like

```
FAIL: test_font_method_should_raise_exception_after_quit (__main__.FontTypeTest) (name='get_underline')
  File "...\Python310\lib\site-packages\pygame\tests\font_test.py", line 794, in test_font_method_should_raise_exception_after_quit
    self.assertRaises(pygame.error, getattr(font, name), *args)
AssertionError: error not raised by get_underline

FAIL: test_font_property_should_raise_exception_after_quit (__main__.FontTypeTest) (name='align')
  File "...\Python310\lib\site-packages\pygame\tests\font_test.py", line 820, in test_font_property_should_raise_exception_after_quit
    self.assertRaises(pygame.error, getattr, font, name) # read
AssertionError: error not raised by getattr
```

 and following do segfault:

```
            # ("set_bold", (True,)),
            # ("set_direction", ("is it text",)),
            # ("set_italic", (True,)),
            # ("set_script", ("is it other text",)),
            # ("set_strikethrough", (True,)),
            # ("set_underline", (True,)),
```

dr0id: this should probably also work for ftfont....hmm will need different expected lists I guess

Starbuck5: If you'd like to add these tests it should probably be in a separate PR.

You could add the tests and then mark them expected to fail in unittest. It might make more sense to open an issue that's like "validate properties of font after de-init" and put that forward as starter code for the issue.

dr0id: ok, I will add them in another PR (uff). But first I would like to finish this one.

src_c/_freetype.c
dr0id: returning an exception, not sure if I have done it right.

dr0id:
@Starbuck5 @zoldalma999 @narilee2006 
updated, I would like to finish this PR within this week.

buildconfig/stubs/pygame/freetype.pyi
```diff
@@ -45,6 +45,7 @@ class Font:
     ascender: int
     descender: int
     style: int
+    style_name: str
```
ankith26, line 48: since this is a property that is only `get`-able, it should be typed as that. Same above too.

Maybe this mistake is a common theme in this file, in that case you might as well leave your changes as is, and the typing stuff can be fixed together in a future PR

docs/reST/ref/freetype.rst
```diff
@@ -224,6 +224,21 @@ loaded. This module must be imported explicitly to be used. ::
...
+      | :sl:`Gets the font's style_name.`
+      | :sg:`style_name -> str`
+
+      Read only. Returns the font's style name. Style names are arbitrary, here are some examples:
+
+        'Black', 'Bold', 'Bold Italic', 'BoldOblique', 'Book', 'BookOblique', 'Condensed', 'Condensed Oblique',
+        'ExtraLight', 'Italic', 'Light', 'LightOblique', 'Medium', 'MediumOblique', 'Oblique', 'Regular',
+        'Semibold', 'Semilight', 'Slanted'
+
+      .. versionadded:: 2.3
```
ankith26, line 238: outdated now

buildconfig/stubs/pygame/freetype.pyi
dr0id: after looking into it for 10 minutes I still don't see what should be changed. Can you give me an example or explicit pointers what you expect? Could not find an example either in pygame (maybe I have overlooked it).

docs/reST/ref/freetype.rst
dr0id: updated to 2.3.1

buildconfig/stubs/pygame/freetype.pyi
ankith26: I'll go ahead and resolve this (to be fixed in future work along with the rest of the file)

src_c/font.c
```diff
@@ -653,10 +655,31 @@ font_size(PyObject *self, PyObject *text)
...
+{
+    if (!PgFont_GenerationCheck(self)) {
+        return RAISE_FONT_QUIT_ERROR();
+    }
+
+    TTF_Font *font = PyFont_AsFont(self);
+    const char *font_style_name = TTF_FontFaceStyleName(font);
+    return PyUnicode_FromString(font_style_name);
```
ankith26, line 682: `font_style_name` can be `NULL` when unknown. Just like `freetype`, we should return an empty string in that case. Passing `NULL` to `PyUnicode_FromString` may lead to a segfault.

The same fix should also be applied to the above function.

Additionally, a small addition to the docs that state that these parameters can be empty stings too would also be nice

ankith26 (changes requested):

src_c/font.c
dr0id: returning "" now if the style_name is null

test/font_test.py
```diff
@@ -581,6 +581,19 @@ def test_set_name():
 
         self.assertRaises(AttributeError, test_set_name)
 
+    def test_font_style_name(self):
+        f = pygame_font.Font(None, 20)
+        self.assertNotEqual(f.style_name, "")
```
ankith26, line 586: This isn't right anymore. Maybe this line could be replaced with `self.assertIsInstance(f.style_name, str)`

src_c/_freetype.c
```diff
@@ -1095,9 +1099,23 @@ _ftfont_getname(pgFontObject *self, void *closure)
 {
     if (pgFont_IS_ALIVE(self)) {
         const char *name = _PGFT_Font_GetName(self->freetype, self);
-        return name ? PyUnicode_FromString(name) : 0;
+        return PyUnicode_FromString(name ? name : "");
```
ankith26, line 1102: actually, this change isn't required in freetype I think, only font. `_PGFT_Font_GetName` is already returning `""` when it has to, but it is also returning `NULL` when it is actually setting exceptions

src_c/_freetype.c
```diff
@@ -1095,9 +1099,23 @@ _ftfont_getname(pgFontObject *self, void *closure)
...
+    PyErr_SetString(PyExc_AttributeError, "<uninitialized Font object>");
+    return NULL;
+}
+
+static PyObject *
+_ftfont_getstylename(pgFontObject *self, void *closure)
+{
+    if (pgFont_IS_ALIVE(self)) {
+        const char *stylename = _PGFT_Font_GetStyleName(self->freetype, self);
+        return PyUnicode_FromString(stylename ? stylename : "");
```
ankith26, line 1114: same review as `_ftfont_getname` applies here

test/font_test.py
dr0id: this is still valid since following code
`f = pygame.font.Font(None, 10)
print(f.style_name)
`
prints 'Bold' for me.

But I can add the instance check.

src_c/_freetype.c
dr0id: but returning 0 is a bad choice if a string is expected?

dr0id: I see, the error is already set, so what should I return then? 
NULL? might cause a segfault, no? 
0? is not a string

I have difficulties testing this because it does not happen.

dr0id: changed back to original code

src_c/_freetype.c
dr0id: see comments on _ftfont_getname

ankith26 (approved):
LGTM, thanks for the PR :+1:

Starbuck5 (approved):
Let's get this merged!

PR #2179, opened 2023-05-18
Optimize `Surface.get_(f)rect`

Matiiss:
Now uses `METH_FASTCALL | METH_KEYWORDS`
Also created a generic function for less redundancy in code.

Test code I used:
```py
from timeit import timeit

import pygame

N = 10_000_000
G = globals()

surf = pygame.Surface((10, 10))

print(timeit("""
surf.get_rect()
""", number=N, globals=G))

print(timeit("""
surf.get_frect()
""", number=N, globals=G))

print(timeit("""
surf.get_rect(center=(5, 5))
""", number=N, globals=G))

print(timeit("""
surf.get_frect(center=(5, 5))
""", number=N, globals=G))
```

I managed to get an approximately 30% increase in speed for `get_rect` with a single kwarg and an approximately 20% increase in speed for `get_frect` with a single kwarg (no clue why they were different, perhaps, just some variations somewhere). (w/o kwargs was about the same as expected, didn't try with multiple kwargs but this should in theory get faster the more kwargs get added compared to what was previously)

### DIFF
--- a/src_c/surface.c
+++ b/src_c/surface.c
@@ -2536,7 +2536,7 @@ static PyObject *
 surf_generic_get_rect(PyObject *self, PyObject *const *args, Py_ssize_t nargs,
                       PyObject *kwnames, char *type)
 {
-    PyObject *rect;
+    PyObject *rect = NULL;
     SDL_Surface *surf = pgSurface_AsSurface(self);
 
     if (nargs > 0) {
--- a/src_c/surface.c
+++ b/src_c/surface.c
@@ -2534,9 +2534,12 @@ surf_get_height(PyObject *self, PyObject *_null)
     return PyLong_FromLong(surf->h);
 }
 
+#define GET_RECT_TYPE_RECT 1
+#define GET_RECT_TYPE_FRECT 2
+
 static PyObject *
 surf_generic_get_rect(PyObject *self, PyObject *const *args, Py_ssize_t nargs,
-                      PyObject *kwnames, char *type)
+                      PyObject *kwnames, int type)
 {
     PyObject *rect = NULL;
     SDL_Surface *surf = pgSurface_AsSurface(self);
@@ -2548,18 +2551,20 @@ surf_generic_get_rect(PyObject *self, PyObject *const *args, Py_ssize_t nargs,
 
     SURF_INIT_CHECK(surf)
 
-    if (strcmp(type, "rect") == 0) {
-        rect = pgRect_New4(0, 0, surf->w, surf->h);
-    }
-    else if (strcmp(type, "frect") == 0) {
-        rect = pgFRect_New4(0.f, 0.f, (float)surf->w, (float)surf->h);
+    switch (type) {
+        case GET_RECT_TYPE_RECT:
+            rect = pgRect_New4(0, 0, surf->w, surf->h);
+            break;
+        case GET_RECT_TYPE_FRECT:
+            rect = pgFRect_New4(0.f, 0.f, (float)surf->w, (float)surf->h);
+            break;
     }
 
     if (rect && kwnames) {
         Py_ssize_t i, sequence_len;
         PyObject **sequence_items;
         sequence_items = PySequence_Fast_ITEMS(kwnames);
-        sequence_len = PySequence_Fast_GET_SIZE(kwnames);
+        sequence_len = PyTuple_GET_SIZE(kwnames);
 
         for (i = 0; i < sequence_len; ++i) {
             if ((PyObject_SetAttr(rect, sequence_items[i], args[i]) == -1)) {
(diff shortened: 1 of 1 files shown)


src_c/surface.c
```diff
@@ -2533,25 +2535,34 @@ surf_get_height(PyObject *self, PyObject *_null)
...
+        rect = pgFRect_New4(0.f, 0.f, (float)surf->w, (float)surf->h);
+    }
 
-        while (PyDict_Next(kwargs, &pos, &key, &value)) {
-            if ((PyObject_SetAttr(rect, key, value) == -1)) {
+    if (rect && kwnames) {
+        Py_ssize_t i, sequence_len;
+        PyObject **sequence_items;
+        sequence_items = PySequence_Fast_ITEMS(kwnames);
+        sequence_len = PySequence_Fast_GET_SIZE(kwnames);
```
itzpr3d4t0r, line 2562: This can just be PyTuple_GET_SIZE.
```suggestion
        sequence_len = PyTuple_GET_SIZE(kwnames);
```

src_c/surface.c
```diff
@@ -2533,25 +2535,34 @@ surf_get_height(PyObject *self, PyObject *_null)
...
+                            "get_%s only accepts keyword arguments", type);
     }
 
     SURF_INIT_CHECK(surf)
 
-    rect = pgRect_New4(0, 0, surf->w, surf->h);
-    if (rect && kwargs) {
-        PyObject *key, *value;
-        Py_ssize_t pos = 0;
+    if (strcmp(type, "rect") == 0) {
```
dr0id, line 2551: If it is about speed: Isn't a string compare slow compared to a int comparison? Probably use an internal enum or just two constants?
Or to avoid the if entirely have either two functions or pass in a creation method pointer (not sure if that is possible...).

Matiiss: I don't know, I switched to `switch` and it didn't really improve the difference I think, in fact, before `get_frect` being slower somewhat made sense since it had to do `strcmp` twice to get there, but now, `get_frect` is faster for some reason, but yeah, I switched to some defines which also seems better for avoiding internal errors.

Matiiss:
Reworking some calculations because speedup is much nicer to present to people, so
based on the previous data

`get_rect` got a 32% speedup (increase in speed)
`get_frect` got a 43% speedup (increase in speed)

Due to testing externalities and whatnot, I guess it can be said that _`get_(f)rect` got a 32% to 43% speedup_ (which is pretty cool, ngl)

src_c/surface.c
```diff
@@ -2532,26 +2534,40 @@ surf_get_height(PyObject *self, PyObject *_null)
...
-    PyObject *rect;
+    PyObject *rect = NULL;
     SDL_Surface *surf = pgSurface_AsSurface(self);
 
-    if (PyTuple_GET_SIZE(args) > 0) {
-        return RAISE(PyExc_TypeError,
-                     "get_rect only accepts keyword arguments");
+    if (nargs > 0) {
+        return PyErr_Format(PyExc_TypeError,
+                            "get_%s only accepts keyword arguments", type);
```
Starbuck5, line 2549: Looks like this still relies on the previous strategy of the type being a string.

Starbuck5:
I'm skeptical of checking the type in the function call.

What about something like this? I think this would be both elegant code and performant, but I haven't tried to compile or test it or anything. Also made the helper func `inline` so it would hopefully get compiled into the caller.
```c
static inline PyObject *
_get_rect_helper(PyObject *rect, PyObject *const *args, Py_ssize_t nargs,
                      PyObject *kwnames, char *type)
{
    if (nargs > 0) {
        Py_DECREF(rect);
        return PyErr_Format(PyExc_TypeError,
                            "get_%s only accepts keyword arguments", type);
    }

    if (rect && kwnames) {
        Py_ssize_t i, sequence_len;
        PyObject **sequence_items;
        sequence_items = PySequence_Fast_ITEMS(kwnames);
        sequence_len = PyTuple_GET_SIZE(kwnames);

        for (i = 0; i < sequence_len; ++i) {
            if ((PyObject_SetAttr(rect, sequence_items[i], args[i]) == -1)) {
                Py_DECREF(rect);
                return NULL;
            }
        }
    }
    return rect;
}

static PyObject *
surf_get_rect(PyObject *self, PyObject *const *args, Py_ssize_t nargs,
              PyObject *kwnames)
{
    SURF_INIT_CHECK(surf)
    SDL_Surface *surf = pgSurface_AsSurface(self);
    PyObject *rect = pgRect_New4(0, 0, surf->w, surf->h);

    return _get_rect_helper(rect, args, nargs, kwnames, "rect");
}
```

Starbuck5 (approved):
Thanks!

Speedup looks good, code looks good.

zoldalma999 (approved):
Looks good to me!

PR #2180, opened 2023-05-18
font: ftfont should pass same tests as font

dr0id:
partial PR as requested in this PR https://github.com/pygame-community/pygame-ce/pull/2132
this PR depends upon the 'stylename' property from PR https://github.com/pygame-community/pygame-ce/pull/2175


ankith26 (approved):
LGTM, thanks!

I don't think there are many users of `ftfont` but it's nice to get it consistent anyways

PR #2182, opened 2023-05-20
Bump SDL_image to 2.6.3

yunline:
To support animation


Starbuck5:
You're discovering the problems that have prevented this dependency upgrade for so long.

The biggest problem is that SDL_image now uses stb_image for PNGs (and JPEGs?), instead of libpng or libjpeg. This means that loading/saving is slower, and png saving doesn't respect the image's pixelformat nearly as well.

yunline:
#2198

PR #2193, opened 2023-05-22
Updated Chimp Example Explanation for #2033 

that1guy232:
#2033

### DIFF
--- a/docs/reST/tutorials/en/chimp-explanation.rst
+++ b/docs/reST/tutorials/en/chimp-explanation.rst
@@ -77,9 +77,7 @@ It also checks for the availability of some of the optional pygame modules. ::
 First, we import the standard "os" python module. This allow
 us to do things like create platform independent file paths.
 
-In the next line, we import the pygame package. In our case, we import
-pygame, all of the functionality of pygame is able to
-be referenced from this package.
+In the next line, we import the pygame package. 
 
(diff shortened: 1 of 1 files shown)


docs/reST/tutorials/en/chimp-explanation.rst
```diff
@@ -78,8 +78,8 @@ First, we import the standard "os" python module. This allow
 us to do things like create platform independent file paths.
 
 In the next line, we import the pygame package. In our case, we import
-pygame as ``pg``, so that all of the functionality of pygame is able to
-be referenced from the namespace ``pg``.
+pygame, all of the functionality of pygame is able to
+be referenced from this package.
```
Matiiss, line 82: I think it's better to just leave it at
"On the next line, we import the pygame package."

The other sentence made sense before, but now it feels out of place.

that1guy232: I agree, when I made the change i wanted to keep it as close to the original but reading it again, It does feel out of place.

that1guy232: Updated with the changes.

Matiiss (approved):
LGTM
Although not sure what to do with "In the next line" as it seems that "On the next line" would fit better, but I'll leave that up to you.

Starbuck5 (approved):
Thanks!

PR #2194, opened 2023-05-23
A bit of pygame-ce-ification

Starbuck5:


yunline:
LGTM
These may also need to be changed:
https://github.com/pygame-community/pygame-ce/blob/899410188114d38bf77f0f4353b9589a565e86cc/setup.py#L154-L159

Starbuck5:
@yunline You can just push that change to this branch if you'd like.

It's something you can do easily with Github desktop, when you check out a PR branch.

yunline (approved):
LGTM

docs/reST/conf.py
```diff
@@ -39,8 +39,8 @@
 master_doc = 'index'
 
 # General information about the project.
-project = 'pygame'
-copyright = '2000-2022, pygame developers'
+project = 'pygame-ce'
+copyright = '2023, pygame-ce developers'
```
gresm, line 43: I don't think removing old copyright text is a good idea, it could be a copyright violation. Way safer would be `'2000-2022, pygame developers, 2023 pygame-ce developers'`

docs/reST/index.rst
gresm, line 1: Should pygame-ce be completely unaffiliated package to pygame, or rather an extension/fork/improved version of it?

Starbuck5: I don't understand the question.

docs/reST/conf.py
Starbuck5: Done.

ankith26 (approved):
LGTM, thanks for the PR :tada: :cake:

Issue #2189, opened 2023-05-22
Don't dealloc types using PyObject_Free


Starbuck5:
Some pygame-ce types use `Py_TYPE(self)->tp_free(self)`, some use `PyObject_Free`

The dealloc docs say you can use `PyObject_Del`, and `PyObject_Free` = `PyObject_Del` (Thanks Zoldalma for pointing this out), but only if the class isn't subtypeable.

But I think we might as well just as use `Py_TYPE(self)->tp_free(self)` everywhere, because it works in all cases and standardizes the code.

PR #2196, opened 2023-05-23
Use `Py_TYPE(self)->tp_free(self)` instead of `PyObject_Free`

yunline:
Issue: #2189


Starbuck5:
event.c also uses PyObject_Free in its error handling in the Event type init.

Starbuck5 (approved):

PR #2197, opened 2023-05-25
Updated scrap code example.

that1guy232:
Most functions in the old example were deprecated, I removed those from the example.
Also added the instructions to the window it's self along with the string that is on the clip bored. 
Currently, only strings are supported with the current scrap module so I wanted to reflect that in the example. 
#2084

### DIFF
--- a/examples/scrap_clipboard.py
+++ b/examples/scrap_clipboard.py
@@ -14,14 +14,6 @@ Keyboard Controls
 """
 
 
-def usage():
-    print("Press the 'v' key to view the current string in the clipboard")
-    print("Press the 'c' key to put a string into the clipboard")
-    print("Press the 'escape' key to quit")
-
-
-usage()
-
 
 def render_instructions():
 
@@ -50,6 +42,8 @@ font = pygame.font.Font(None, 20)
 # Initialize the scrap module and use the clipboard mode.
 pygame.scrap.init()
 pygame.scrap.set_mode(pygame.SCRAP_CLIPBOARD)
+
+
 clipboard_text = ''
 
 while running:
@@ -95,9 +89,6 @@ while running:
 
             elif event.key == pygame.K_ESCAPE:
                 running = False
-
-            else:
-                usage()
 
 
         pygame.display.flip()
--- a/examples/scrap_clipboard.py
+++ b/examples/scrap_clipboard.py
@@ -30,7 +30,7 @@ def render_instructions():
 
 pygame.init()
 pygame.display.set_caption("Clipboard Example")
-screen = pygame.display.set_mode((800, 800))
+screen = pygame.display.set_mode((960, 540))
 clock = pygame.Clock()
 font = pygame.Font(None, 30)
 
--- a/examples/scrap_clipboard.py
+++ b/examples/scrap_clipboard.py
@@ -37,9 +37,6 @@ running = True
 pygame.font.init()
 font = pygame.font.Font(None, 20)
 
-# Initialize the scrap module and use the clipboard mode.
-pygame.scrap.init()
-pygame.scrap.set_mode(pygame.SCRAP_CLIPBOARD)
 
 
 clipboard_text = ""
--- a/examples/scrap_clipboard.py
+++ b/examples/scrap_clipboard.py
@@ -24,7 +24,6 @@ usage()
 
 
 def render_instructions():
-    screen.fill((0, 0, 0))
 
     text = font.render("Press 'v' to view the clipboard", True, (255, 255, 255))
     screen.blit(text, (10, 10))
@@ -51,8 +50,22 @@ font = pygame.font.Font(None, 20)
 # Initialize the scrap module and use the clipboard mode.
 pygame.scrap.init()
 pygame.scrap.set_mode(pygame.SCRAP_CLIPBOARD)
+clipboard_text = ''
 
 while running:
+    screen.fill((0, 0, 0))
+
+    render_instructions()
+    
+    text = "Text on the clipboard:"
+    text = font.render(text, True, (255, 255, 255))
+    screen.blit(text, (10, 80))
+    
+    text = font.render(clipboard_text, True, (255, 255, 255))
+    screen.blit(text, (10, 100))
+
+
+
     for event in pygame.event.get():
         if event.type == pygame.QUIT:
             running = False
@@ -61,10 +74,13 @@ while running:
             if event.key == pygame.K_v:
                 # Look for any text data in the clipboard.
                 print("Looking for text data in the clipboard..")
+               
                 if pygame.scrap.has_text():
                     print("Text found:")
                     text = pygame.scrap.get_text()
                     print(text)
+                    clipboard_text = text
+
                 else:
                     print("No text data in the clipboard.")
                 pass
@@ -75,6 +91,7 @@ while running:
                 pygame.scrap.put_text(
                     "Hello World! This is some text from the pygame scrap example.",
                 )
+                clipboard_text = pygame.scrap.get_text()
 
             elif event.key == pygame.K_ESCAPE:
                 running = False
@@ -82,7 +99,6 @@ while running:
             else:
                 usage()
 
-        render_instructions()
 
         pygame.display.flip()
         clock.tick(30)
--- a/examples/scrap_clipboard.py
+++ b/examples/scrap_clipboard.py
@@ -47,7 +47,7 @@ while running:
     text = font.render(text, True, "white")
     screen.blit(text, (10, 80))
 
-    text = font.render(clipboard_text, True, "white", None, 480)
+    text = font.render(clipboard_text, True, "white", None, 790)
     screen.blit(text, (10, 110))
 
     for event in pygame.event.get():
--- a/examples/scrap_clipboard.py
+++ b/examples/scrap_clipboard.py
@@ -18,15 +18,13 @@ import pygame
 
 
 def render_instructions():
-    text = font.render("Press 'v' to view the clipboard", True, (255, 255, 255))
+    text = font.render("Press 'v' to view the clipboard", True, "white")
     screen.blit(text, (10, 10))
 
-    text = font.render(
-        "Press 'c' to copy some text to the clipboard", True, (255, 255, 255)
-    )
+    text = font.render("Press 'c' to copy some text to the clipboard", True, "white")
     screen.blit(text, (10, 30))
 
-    text = font.render("Press 'escape' to quit", True, (255, 255, 255))
+    text = font.render("Press 'escape' to quit", True, "white")
     screen.blit(text, (10, 50))
 
 
@@ -46,10 +44,10 @@ while running:
     render_instructions()
 
     text = "Text on the clipboard:"
-    text = font.render(text, True, ("white"))
+    text = font.render(text, True, "white")
     screen.blit(text, (10, 80))
 
-    text = font.render(clipboard_text, True, ("white"), None, 480)
+    text = font.render(clipboard_text, True, "white", None, 480)
     screen.blit(text, (10, 100))
 
     for event in pygame.event.get():
--- a/examples/scrap_clipboard.py
+++ b/examples/scrap_clipboard.py
@@ -38,7 +38,6 @@ pygame.font.init()
 font = pygame.font.Font(None, 20)
 
 
-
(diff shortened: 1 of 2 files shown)


Starbuck5:
In case you haven't come across it before, you can format your code so it passes the formatting checks running `python setup.py format` in your local pygame folder. To run it I believe needs `pip install black` and `pip install clang-format`

that1guy232:
> In case you haven't come across it before, you can format your code so it passes the formatting checks running `python setup.py format` in your local pygame folder. To run it I believe needs `pip install black` and `pip install clang-format`

Yeah, Never did that before, Was just using black, Thanks.

examples/scrap_clipboard.py
```diff
@@ -8,89 +8,80 @@
...
-    print("Press the 'g' key to get all of the current clipboard data")
-    print("Press the 'p' key to put a string into the clipboard")
-    print("Press the 'a' key to get a list of the currently available types")
-    print("Press the 'i' key to put an image into the clipboard")
+    text = font.render("Press 'escape' to quit", True, (255, 255, 255))
+    screen.blit(text, (10, 50))
 
 
-main_dir = os.path.split(os.path.abspath(__file__))[0]
+import pygame
```
Starbuck5, line 30: Imports should be above functions.

examples/scrap_clipboard.py
```diff
@@ -8,89 +8,80 @@
...
+                pygame.scrap.put_text(
+                    "Hello World! This is some text from the pygame scrap example.",
+                )
+                clipboard_text = pygame.scrap.get_text()
+
+            elif event.key == pygame.K_ESCAPE:
+                running = False
+
+        pygame.display.flip()
+        clock.tick(30)
```
Starbuck5, line 86: 30 is pretty low, seems like it should be 60 at least.

examples/scrap_clipboard.py
```diff
@@ -8,89 +8,80 @@
...
-                r = scrap.get(t)
-                if r and len(r) > 500:
-                    print(f"Type {t} : (large {len(r)} byte buffer)")
-                elif r is None:
-                    print(f"Type {t} : None")
+screen = pygame.display.set_mode((500, 500))
+clock = pygame.time.Clock()
+running = True
+
+pygame.font.init()
```
Starbuck5, line 37: Pygame.font.init is unnecessary because you already called pygame.init

examples/scrap_clipboard.py
```diff
@@ -8,89 +8,80 @@
...
-            # This means to look for data.
-            print("Getting the different clipboard data..")
-            for t in scrap.get_types():
-                r = scrap.get(t)
-                if r and len(r) > 500:
-                    print(f"Type {t} : (large {len(r)} byte buffer)")
-                elif r is None:
-                    print(f"Type {t} : None")
+screen = pygame.display.set_mode((500, 500))
+clock = pygame.time.Clock()
```
Starbuck5, line 34: New in pygame-ce you can do this as `pygame.Clock`, and the font below could be `pygame.Font`.

examples/scrap_clipboard.py
```diff
@@ -8,89 +8,80 @@
...
-i - put image into the clipboard
 """
-import os
 
-import pygame
-import pygame.scrap as scrap
 
-from io import BytesIO
+def render_instructions():
+    text = font.render("Press 'v' to view the clipboard", True, (255, 255, 255))
```
Starbuck5, line 18: Totally a matter of preference but you could use "white" instead of (255, 255, 255) in all the color fields. Thoughts?

examples/scrap_clipboard.py
```diff
@@ -8,89 +8,80 @@
...
+while running:
+    screen.fill((0, 0, 0))
+
+    render_instructions()
+
+    text = "Text on the clipboard:"
+    text = font.render(text, True, (255, 255, 255))
+    screen.blit(text, (10, 80))
+
+    text = font.render(clipboard_text, True, (255, 255, 255))
```
Starbuck5, line 52: If I have a long string in the clipboard it just goes offscreen. This is the perfect opportunity to use new pygame-ce multiline rendering, with `wraplength`.

examples/scrap_clipboard.py
```diff
@@ -8,89 +8,80 @@
...
-        elif e.type == pygame.KEYDOWN and e.key == pygame.K_g:
-            # This means to look for data.
-            print("Getting the different clipboard data..")
-            for t in scrap.get_types():
-                r = scrap.get(t)
-                if r and len(r) > 500:
-                    print(f"Type {t} : (large {len(r)} byte buffer)")
-                elif r is None:
-                    print(f"Type {t} : None")
+screen = pygame.display.set_mode((500, 500))
```
Starbuck5, line 33: Giving the example window a nice title is a nice way to instantly make it look way more polished.

Starbuck5:
I was tempted to just approve, since this is a great improvement anyways. But I saw a couple good avenues for improvement that shouldn't be too difficult, so I added some comments.

Thanks for working on this!

examples/scrap_clipboard.py
Starbuck5: You put white in 2 places out of 5?
Also it should be "white" not ("white").

Starbuck5:
Another bit of feedback, the font size and window size could be a bit larger to make the text clearer.

examples/scrap_clipboard.py
that1guy232: My bad, I missed the 3 in the render_instructions had it collapsed in my IDE. Fixed.

examples/scrap_clipboard.py
```diff
@@ -8,89 +8,78 @@
...
+screen = pygame.display.set_mode((800, 800))
+clock = pygame.Clock()
+font = pygame.Font(None, 30)
+
+
+clipboard_text = ""
+running = True
+
+while running:
+    screen.fill((0, 0, 0))
```
Starbuck5, line 42: Oh, and this might as well be "black" if the other ones are "white".

Starbuck5 (approved):
Looks good!

Note to other reviewers: this PR should be a squash and merge rather than a merge commit, since the history is messy at this point.

examples/scrap_clipboard.py
```diff
@@ -8,89 +8,79 @@
...
-def usage():
-    print("Press the 'g' key to get all of the current clipboard data")
-    print("Press the 'p' key to put a string into the clipboard")
-    print("Press the 'a' key to get a list of the currently available types")
-    print("Press the 'i' key to put an image into the clipboard")
+    text = font.render("Press 'c' to copy some text to the clipboard", True, "white")
+    screen.blit(text, (10, 30))
 
+    text = font.render("Press 'escape' to quit", True, "white")
+    screen.blit(text, (10, 50))
```
Matiiss, line 28: This could take advantage of multiline text rendering and just use one call to `render` to render the text.

examples/scrap_clipboard.py
```diff
@@ -8,89 +8,79 @@
...
+                    print("No text data in the clipboard.")
+                pass
+
+            elif event.key == pygame.K_c:
+                # put some text into the clipboard.
+                print("Putting some text into the clipboard")
+                pygame.scrap.put_text(
+                    "Hello World! This is some text from the pygame scrap example.",
+                )
+                clipboard_text = pygame.scrap.get_text()
```
Matiiss, line 79: Should this automatically change the text on the screen? Because you're not viewing it yet and if you have copied some of your own text from somewhere and you display it, then pressing C clears it and there's this sort of discontinuity in the flow of the controls. So, perhaps, this line could be removed.

examples/scrap_clipboard.py
```diff
@@ -8,89 +8,79 @@
...
-            buf = fp.read()
-            scrap.put("image/bmp", buf)
-            fp.close()
-
-        elif e.type in (pygame.KEYDOWN, pygame.MOUSEBUTTONDOWN):
-            usage()
-    pygame.display.flip()
-    c.tick(40)
+                    print("No text data in the clipboard.")
+                pass
```
Matiiss, line 71: This `pass` doesn't do anything, can be removed.

Matiiss:
Also, for consistency, what about all the "logging" messages also getting displayed on the screen, because the instructions are, but if this is supposed to be quick, I guess someone else can tackle it in some other PR.

Starbuck5:
> Also, for consistency, what about all the "logging" messages also getting displayed on the screen, because the instructions are, but if this is supposed to be quick, I guess someone else can tackle it in some other PR.

I'd say that's a large change to ask from this PR at this point. So save it as a future improvement if that wants to be pursued.

examples/scrap_clipboard.py
that1guy232: I wanted to always show what was on the clipboard.
Pressing C is intended to copy that string to the clipboard and immediately display it.
so I than grab the text i just put on the clipboard and display it.

I felt having the user see that we copyed the text and than grabbed from clipboard to be displayed was a good showing of how it worked.

We do not need to show that we copyed it to the clipboard immediately but I would like some more input on this.

Matiiss: I'm just saying that if the user was to copy the text from some other place, it wouldn't immediately get displayed, they'd have to return to the window and press V for the new text to appear, this is not the case when you press C inside the window, the text gets immediately displayed and that makes for a discontinuity in the flow of controls, they differ based on how you got the text into the clipboard.

that1guy232: Oh i understand now, I'll see what I can do thabk you.

examples/scrap_clipboard.py
```diff
@@ -8,26 +8,13 @@
 
 Keyboard Controls
 -----------------
-    v - View the current clipboard data.
```
Matiiss: The V control should stay

examples/scrap_clipboard.py
```diff
@@ -42,41 +29,32 @@ def render_instructions():
...
-                print("Looking for text data in the clipboard..")
-
-                if pygame.scrap.has_text():
-                    print("Text found:")
-                    text = pygame.scrap.get_text()
-                    print(text)
-                    clipboard_text = text
-
-                else:
-                    print("No text data in the clipboard.")
```
Matiiss: So this should remain in the example
(I guess minus `print(text)`)

examples/scrap_clipboard.py
```diff
@@ -42,41 +29,32 @@ def render_instructions():
...
-                pass
-
-            elif event.key == pygame.K_c:
-                # put some text into the clipboard.
-                print("Putting some text into the clipboard")
+            if event.key == pygame.K_c:
                 pygame.scrap.put_text(
                     "Hello World! This is some text from the pygame scrap example.",
                 )
-                clipboard_text = pygame.scrap.get_text()
```
Matiiss: What I meant was to just remove this line which you have done, so that's great

examples/scrap_clipboard.py
```diff
@@ -42,41 +29,32 @@ def render_instructions():
...
 
-    text = "Text on the clipboard:"
-    text = font.render(text, True, "white")
-    screen.blit(text, (10, 80))
+    text = font.render("Text on the clipboard:", True, "white")
+    screen.blit(text, (10, 100))
 
     text = font.render(clipboard_text, True, "white", None, width - 20)
-    screen.blit(text, (10, 110))
+    screen.blit(text, (10, 120))
```
Matiiss: This could be combined into one `font.redner` since multiline support so something like
```py
text = font.render(f"Text on the clipboard:\n{clipboard_text}", True, "white", None, width - 20)
screen.blit(text, (10, 100))
```

examples/scrap_clipboard.py
```diff
@@ -42,41 +29,32 @@ def render_instructions():
 while running:
     screen.fill("black")
 
-    render_instructions()
+    instruction = (
+        "Keyboard Controls: \n c - Copy some text into the clipboard. \n escape - Quit"
```
Matiiss, line 33: There are some whitespaces here that shouldn't be there, like anywhere after `\n` there shouldn't be a space, nor it should be before a `\n`

Matiiss (commented):
Also the `print` stuff should remain in the example (except for that one I mentioned)

Matiiss (approved):
LGTM 👍 
Thanks for the PR!

PR #2202, opened 2023-05-27
Issue redirection

oddbookworm:
Now that all of the issues are visible, here's this PR that I've had waiting since the fork opened


Starbuck5:
The manylinux build is failing because you have changed the content in the manylinux buildconfig, causing it to automatically regenerate the docker build. Since your branch is on your fork rather than the pygame-ce repo itself, your branch does not have permissions to push up the regenerated image, hence it fails.

oddbookworm:
Closing due to permission issues with the manylinux runs because I modified the manylinux buildconfig. Opened a new PR that should not have that problem

PR #2203, opened 2023-05-28
Issue redirection (correct one)

oddbookworm:
Continuation of #2202 so that the manylinux runs can pass

### DIFF
--- a/buildconfig/manylinux-build/README.rst
+++ b/buildconfig/manylinux-build/README.rst
@@ -12,7 +12,7 @@ build on an old Linux distribution in a docker container.
 manylinux is an older linux with a fairly compatible ABI, so you can make linux binary
 wheels that run on many different linux distros.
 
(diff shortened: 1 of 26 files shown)


Starbuck5:
Is it in the scope of this PR to retarget things that aren't full links, but say `issue #xyz`?

For example, https://github.com/pygame-community/pygame-ce/blob/c4095e69f7ce54f85f1b7f93093c6cc9ab9a1221/test/color_test.py#L588

oddbookworm:
> Is it in the scope of this PR to retarget things that aren't full links, but say `issue #xyz`?

I believe it would be. I need to step away for a bit, but I can do a more thorough search for these when I come back

oddbookworm:
I'm pretty sure I got them all

Starbuck5 (approved):
Thanks!

zoldalma999 (approved):
Thanks!

PR #2204, opened 2023-05-28
updated scrap related code.

that1guy232:
#2084 I think this is the last example that uses scrap?


Starbuck5 (approved):
Thanks!

Starbuck5:
If you'd like to do some additional improvements to this example, you could use the pygame-ce multiline text implementation (plus font.align for centering) instead of the bespoke multiline text in this example.

Also it could use a title (set_caption).

Matiiss (approved):
LGTM
Thanks for contributing again 👍

Issue #1660, opened 2023-02-12
Remove obsolete vidcapture camera backend (3357)


GalacticEmperor1:
#  **Issue &numero;3357 opened by *[Starbuck5](https://github.com/Starbuck5)* at 2022-07-31 05:49:08**

Pygame's camera module, (src_py/camera.py) uses different strategies to try to get a working webcam backend.

At one point, videocapture / vidcapture was usable and used, another python library that could provide webcam functionality for Windows. However, that package is now obsolete (last updated in 2012), and we have a new camera backend for Windows now.

I'd like someone to remove this backend and relevant code.

- [ ] update documentation ( camera docs, documented env variable in "pygame" doc? )
- [ ] remove _camera_vidcapture.py source
- [ ] remove vidcapture support from camera.py
- [ ] figure out if there's something else I'm forgetting in this list that should be done<hr>

#  Comments

# # #  **[MyreMylar](https://github.com/MyreMylar)* commented at 2022-07-31 13:22:06*

What version of Windows does the new version of the camera backend support? IIRC it was 8 or was it 8.1?

Googling online suggests Windows 8.1 is EOL from Jan 2023.
<hr>

# # #  **[Starbuck5](https://github.com/Starbuck5)* commented at 2022-08-02 03:00:59*

> What version of Windows does the new version of the camera backend support? IIRC it was 8 or was it 8.1?

It was unclear to me as I wrote it. It's been tested by Matiis, who runs some iteration of 8. (EDIT: confirmed, it was 8.1)

Seeing your comment reminds me that you actually were able to get vidcapture to work, I had forgotten anybody had in recent years. I was never able to get it to work locally.
<hr>

# # #  **[MyreMylar](https://github.com/MyreMylar)* commented at 2022-08-02 06:35:53*

Yeah I got it working a while back, I think I just downloaded binary versions from that random python windows binary website on a university server - that is now being shut down. 

Since then I've updated to a new PC and then Windows 11 and the new backend has been working fine on both Windows 10 and Windows 11 so I'm pretty sure it is safe to deprecate the old backend for new versions. 

I expect the intersection of people who run pre-windows 8, managed to download the binary vidcapture backend before the website was shut down (or built it themselves from source) and still want to keep up to date on the latest versions of pygame is a very small number.
<hr>

# # #  **[MVBarros](https://github.com/MVBarros)* commented at 2022-08-08 23:06:00*

[...]

PR #2207, opened 2023-05-29
Remove obsolete vidcapture camera backend (#1660)

that1guy232:
#1660
I very much imagine I am missing things.


Starbuck5:
This looks pretty good to go from me. The only thing I notice straight away is that the docs should have a note that support was removed in this pygame version.

See `..versionchanged:` tags in other bits of the codebase.

that1guy232:
> The only other file I can find mentioning the Videocapture back end is here:
> 
> https://github.com/pygame-community/pygame-ce/blob/main/docs/reST/tutorials/en/intro-to-camera.rst
> 
> In the second paragraph/first 'Note' block.
> 
> > As of Pygame 1.9, the camera module offers native support for cameras that use v4l2 on Linux. There is support for other platforms via Videocapture or OpenCV, but this guide will focus on the native module.

Good catch, I Think I got the last of it!

Starbuck5 (approved):
Thanks!

novialriptide (approved):
LGTM.

PR #2210, opened 2023-05-31
Use builtin broadcasting intrinsics in blitters

Starbuck5:
Just a minor code style thing for the SSE and AVX blitters.

The bonus is that it makes the AVX code a bit more compact.


MyreMylar (approved):
LGTM 👍

ankith26 (approved):
LGTM, thanks! :cake:

PR #2212, opened 2023-05-31
Fix MAC_ARCH environment variable on CI

MyreMylar:
For whatever reason the `MAC_ARCH` environment variable was implemented in a non-githubby way in our CI and this eventually give me a warning, treated as an error while working on #2042. This changes it to work like all the other environment variables in the file.

In the interests of making that work easier to review I'm splitting off this small change which is largely unrelated to SIMD.

### DIFF
--- a/.github/workflows/build-macos.yml
+++ b/.github/workflows/build-macos.yml
@@ -117,6 +117,7 @@ jobs:
           }
 
     env:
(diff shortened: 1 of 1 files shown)


ankith26 (approved):
Looks aight, thanks! :star2:

Starbuck5 (approved):

PR #2214, opened 2023-05-31
Add SIMD version of grayscale

MyreMylar:
Splitting #2042 into two parts. This will just be the grayscale implementation excluding the setup #2042 should be resolved first.


src_c/simd_transform_sse2.c
```diff
@@ -0,0 +1,112 @@
...
+        ((0x4C << src->format->Rshift) | (0x96 << src->format->Gshift) |
+         (0x1D << src->format->Bshift));
+
+    __m128i mm_src, mm_dst, mm_alpha, mm_zero, mm_two_five_fives,
+        mm_rgb_weights, mm_alpha_mask, mm_rgb_mask;
+
+    mm_zero = _mm_setzero_si128();
+    mm_alpha_mask = _mm_cvtsi32_si128(amask);
+    mm_rgb_mask = _mm_cvtsi32_si128(rgbmask);
+    mm_two_five_fives = _mm_set_epi64x(0x00FF00FF00FF00FF, 0x00FF00FF00FF00FF);
```
Starbuck5, line 69: Could be `_mm_set1_epi64x`

src_c/simd_transform_sse2.c
```diff
@@ -0,0 +1,112 @@
...
+    mm_rgb_weights =
+        _mm_unpacklo_epi8(_mm_cvtsi32_si128(rgb_weights), mm_zero);
+
+    while (num_batches--) {
+        pixel_batch_counter = pixel_batch_length;
+        while (pixel_batch_counter--) {
+            mm_src = _mm_cvtsi32_si128(*srcp);
+            /*mm_src = 0x000000000000000000000000AARRGGBB*/
+            mm_alpha = _mm_subs_epu8(mm_src, mm_rgb_mask);
```
Starbuck5, line 78: I've never been super clear about what saturation means in the intrinsics, so this is a very interesting example.

https://en.wikipedia.org/wiki/Saturation_arithmetic tells me it is arithmetic clamped to a range.

So alpha could be anything, but you subtract 255 with saturation--> meaning it will always end as zero. Very clever stuff. I wonder if there's masking in other SIMD stuff that could be converted to this for a small efficiency gain.

src_c/simd_transform_sse2.c
```diff
@@ -0,0 +1,112 @@
...
+            mm_dst = _mm_mullo_epi16(mm_src, mm_rgb_weights);
+            /*mm_dst = 0x00000000000000000000RRRRGGGGBBBB*/
+            mm_dst = _mm_add_epi16(mm_dst, mm_two_five_fives);
+            /*mm_dst = 0x00000000000000000000RRRRGGGGBBBB*/
+            mm_dst = _mm_srli_epi16(mm_dst, 8);
+            /*mm_dst = 0x0000000000000000000000RR00GG00BB*/
+
+            mm_dst = _mm_adds_epu8(
+                _mm_adds_epu8(
+                    _mm_shufflelo_epi16(mm_dst, _MM_SHUFFLE(0, 0, 0, 0)),
```
Starbuck5, line 92: Maybe we should provide our own version of `_MM_SHUFFLE` so we're not dependent on this implementation detail of sse2neon.

It took me a bit to parse whether this was something by us, by the intrinsics header, or by sse2neon.

PR #2219, opened 2023-06-01
Added multiline text rendering to eventlist example

Martinus2004:
Instead of a list, history is now a string that will contain the multine text. The for loops have been replaced with a single font.render call. Also did some string operations for properly showing the "usage" text and to only display the last -13 elements. As you can see the pace between lines has become a bit bigger due to not manually setting the lines' position)

Before after eventlist

### DIFF
--- a/examples/eventlist.py
+++ b/examples/eventlist.py
@@ -100,8 +100,14 @@ def drawhistory(win, history):
     img = font.render("Event History Area", 1, (155, 155, 155), (0, 0, 0))
     win.blit(img, (2, 132))
 
-    img = font.render(history, 1, (50, 200, 50), (0, 0, 0), wraplength=640)
+    img = font.render("\n".join(history), 1, (50, 200, 50), (0, 0, 0), wraplength=640)
     win.blit(img, (10, 250))
+
+
+def draw_usage_in_history(history, text):
+    for line in text.split("\n"):
+        if "===" not in line and line != "":
+            history.append(line)
 
 
 def main():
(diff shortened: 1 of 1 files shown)


yunline:
run `python setup.py format` to format the code

examples/eventlist.py
```diff
@@ -193,4 +177,4 @@ def main():
 
 
 if __name__ == "__main__":
-    main()
+    main()
```
yunline: A new-line is needed at the end of file

Martinus2004: All formatting issues should be fixed now. (I hope)

examples/eventlist.py
```diff
@@ -99,23 +99,9 @@ def drawstatus(win):
...
-
-def draw_usage_in_history(history, text):
-    lines = text.split("\n")
-    for line in lines:
-        if line == "" or "===" in line:
-            continue
-        img = font.render(line, 1, (50, 200, 50), (0, 0, 0))
-        history.append(img)
 
+    img = font.render(history, 1, (50, 200, 50), (0, 0, 0))
```
yunline, line 103: You can add `warplength=640` here to enable auto warp
image

examples/eventlist.py
```diff
@@ -131,23 +118,22 @@ def main():
     img_on_off.append(font.render("Off", 1, (0, 0, 0), (255, 50, 50)))
     img_on_off.append(font.render("On", 1, (0, 0, 0), (50, 255, 50)))
 
-    # stores surfaces of text representing what has gone through the event queue
-    history = []
```
yunline, line 135: You can still use list for the `histrory` since `append` method is way faster than `+=` for string

yunline: then use `'\n'.join(history)` to convert list as string, which you have already done.

yunline (approved):
LGTM

Starbuck5 (approved):
Going through this example code I just see lots of strange things.

For example the width is hardcoded places, there's a pointless `raise SystemExit`, the code has a boolean variable if running on SDL2 or not that it determines in a strange way, window resizing doesn't even try to adjust the text window, etc...

But this is not your fault, this PR is an improvement but it doesn't have to overhaul the example. In fact it's maybe better it doesn't.

Anyway, the part you made looks good. Looks like it makes the example's code a bit more straightforward?

PR #2222, opened 2023-06-02
Add new C API `pg_GetPkgdataResource`

### LINKED ISSUE #2221: Add `pgPkgdata_GetResource()` to C API
To load default icon for window/display or default font for font/freetype.

yunline:
fixes #2221

- New API `pg_GetPkgdataResource`
- New API `pgImage_LoadBasic`


narilee2006:
What is pg_GetPkgdataResource ?

yunline:
> Do these changes provide any benefits that I'm missing? I can see that it makes the code a bit cleaner

You are right. It just cleans up some code.

src_c/display.c
```diff
@@ -1323,7 +1248,8 @@ pg_set_mode(PyObject *self, PyObject *arg, PyObject *kwds)
 
     /*set the window icon*/
     if (!state->icon) {
-        state->icon = pg_display_resource(icon_defaultname);
+        PyObject *icon_fileobj = pg_GetPkgdataResource(icon_defaultname);
```
Starbuck5, line 1251: This isn't a "file object," I believe, so the variable is misleadingly named.

src_c/image.c
```diff
@@ -66,12 +67,12 @@ find_extension(const char *fullname)
 }
 
 static PyObject *
-image_load_basic(PyObject *self, PyObject *obj)
+pgImage_LoadBasic(PyObject *fileobj)
```
Starbuck5, line 70: Same comment about `fileobj`

yunline: I don't think this one is misleading.
`fileobj` can be a str filename or a file-like object.

src_c/display.c
yunline: fixed

src_c/base.c
```diff
@@ -2064,6 +2064,79 @@ pg_uninstall_parachute(void)
...
+    else if (encoded != Py_None) {
+        Py_DECREF(fresult);
+        fresult = encoded;
+    }
+    else {
+        Py_DECREF(encoded);
+    }
+
+    // return fresult
+    goto get_resource_end;
```
ankith26, line 2131: this is a redundant `goto`, even without it, control is going to flow into the block below

src_c/window.c
```diff
@@ -742,7 +665,8 @@ window_init(pgWindowObject *self, PyObject *args, PyObject *kwargs)
 
     SDL_SetWindowData(_win, "pg_window", self);
 
-    PyObject *icon = pg_display_resource(icon_defaultname);
+    PyObject *icon_path_obj = pg_GetPkgdataResource(icon_defaultname);
+    PyObject *icon = pgImage_LoadBasic(icon_path_obj);
```
ankith26, line 669: Needs check for `icon_path_obj` being `NULL` (and exit early with error forward) and if it is not `NULL` it needs to be `DECREF`d after `pgImage_LoadBasic` finishes

src_c/display.c
```diff
@@ -1334,7 +1259,8 @@ pg_set_mode(PyObject *self, PyObject *arg, PyObject *kwds)
 
     /*set the window icon*/
     if (!state->icon) {
-        state->icon = pg_display_resource(icon_defaultname);
+        PyObject *icon_path_obj = pg_GetPkgdataResource(icon_defaultname);
+        state->icon = pgImage_LoadBasic(icon_path_obj);
```
ankith26, line 1263: same as above review (except here maybe just ignore the error and don't even call `pgImage_LoadBasic` in that case)

ankith26 (changes requested):

MyreMylar:
This could use merging with main as the conflicts are making the diff hard to understand now.

MyreMylar (changes requested):
Please merge with main and resolve the conflicts, to make it easier to review.

yunline:
It‘s not worth to create two C APIs and just simplify code a little bit.
The simplifying works will contunue in #2552

PR #2226, opened 2023-06-02
fix search page green background in dark theme

Notenlish:
fixes the weird bug that @yunline reported
image

Here's how it looks currently 
image

here's the light theme version: 
image

### DIFF
--- a/docs/reST/themes/classic/static/pygame.css_t
+++ b/docs/reST/themes/classic/static/pygame.css_t
@@ -76,12 +76,20 @@ body {
     color: {{ theme_textcolor }};
     font-style: normal;
     font-family: {{ theme_bodyfont }};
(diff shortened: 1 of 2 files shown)


NOTMEE12:
maybe make that searchbar longer to the length of the text (`Searching for multiple words...`) or make it centered, cause it's looks a lil' weird.

Notenlish:
> maybe make that searchbar longer to the length of the text (`Searching for multiple words...`) or make it centered, cause it's looks a lil' weird.

I don't think centering it would make sense
image

NOTMEE12:
Now I saw that the text is not centered.

yunline (approved):
LGTM

Starbuck5 (approved):
Fixes the problem.

PR #2230, opened 2023-06-05
Add unitest for `Window`

yunline:


test/window_test.py
```diff
@@ -0,0 +1,122 @@
...
+
+
+class WindowTypeTest(unittest.TestCase):
+    def setUp(self):
+        self.win = Window()
+
+    @unittest.skipIf(
+        os.environ.get("SDL_VIDEODRIVER") == "dummy",
+        "requires the SDL_VIDEODRIVER to be a non dummy value",
+    )
```
Starbuck5, line 17: It would be great to get as many of these running without this check if possible.

This check means the test won't run on CI.

MyreMylar (approved):
LGTM. Tests all run locally and seem reasonable 👍 

Looks like about 60% of them can now run on the CI. kind of inevitable that some of them won't be runnable with the display-less CI machines. Another reminder of why it is good to regularly test PRs locally.

dr0id (approved):
run tests too locally

test/window_test.py
```diff
@@ -0,0 +1,176 @@
...
+    def test_opacity(self):
+        self.win.opacity = 0.5
+        self.assertEqual(self.win.opacity, 0.5)
+        self.win.opacity = 0
+        self.assertEqual(self.win.opacity, 0)
+        self.win.opacity = 1
+        self.assertEqual(self.win.opacity, 1)
```
ankith26, line 114: I think these tests can be improved for testing `TypeError`

yunline: Fixed.

test/window_test.py
Starbuck5: I tested some of these skipped methods without SLD_VIDEODRIVER=dummy set and they didn't crash or anything, so we should not skip them.

Starbuck5 (approved):
Alright, I added some additional tests to get around the CI videodriver, for some added peace of mind when running tests on CI.

Thanks for working on this yunline! If my edits look good to you, you're free to merge.

Issue #2206, opened 2023-05-29
Make use of multiline text rendering in `pygame.examples`


Matiiss:
A lot of examples use multiple calls to `pygame.Font.render` to render different lines of text and then blit each returned surface at a different position, however, with the advent of multiline text rendering since `pygame-ce` version `2.1.3` that's not necessary anymore. 

So, all of these examples that have these multiple lines of text rendered this way, could switch over to a single call to `pygame.Font.render` and utilize `wraplength` and `\n` characters. Also those examples that implement alignment, could use `pygame.Font.align`.

A list of examples where this needs to be "fixed":
- [x] `dropevent.py`, simply add `wraplength` argument to the two `render` calls in the event loop so large dropped text gets wrapped (https://github.com/pygame-community/pygame-ce/pull/2235)
- [x] `eventlist.py`, a couple places where there's some iteration over lines of text instead of using a single call to `render`, also maybe add `wraplength` to `showtext` (https://github.com/pygame-community/pygame-ce/pull/2219)
- [x] `font_viewer.py` (https://github.com/pygame-community/pygame-ce/pull/2235)
- [x] `joystick.py`, `TextPrint` should use strings or lists for stuff that it does (lists for easier modifications on different lines, but render it using `str.join`) (https://github.com/pygame-community/pygame-ce/pull/2235)
- [x] `music_drop_fade.py` (https://github.com/pygame-community/pygame-ce/pull/2235)
- [x] `textinput.py` (https://github.com/pygame-community/pygame-ce/pull/2588 ~~https://github.com/pygame-community/pygame-ce/pull/2209~~)
- [x] `playmus.py`, switch over to `pygame.font` and utilize multiline rendering (should it be touched at all?) (https://github.com/pygame-community/pygame-ce/pull/2581)

JiffyRob:
On it

bilhox:
Hey ! Everything should be done here, it should be fine to close it now.

PR #2235, opened 2023-06-08
Multiline examples

JiffyRob:
This contains a commit for every example listed and not covered in issue #2206 except for playmus.  The default wrapping behaviour may not be desired for dropevent though:
image
I tried not to change too much but the joystick example got the TextPrint class removed and the music fade example's constants and globals were put to the top (because I just could not wrap my head around the code else).  The font viewer is also left aligned now because multiline rendering does not center text.  I'm open to change this if necessary.

### DIFF
--- a/examples/font_viewer.py
+++ b/examples/font_viewer.py
@@ -100,7 +100,13 @@ class FontViewer:
         # display instructions at the top of the display
         font = pygame.font.SysFont(pygame.font.get_default_font(), font_size)
         instructions = f"Use the scroll wheel or click and drag to scroll up and down.  Fonts that don't use the Latin Alphabet might render incorrectly.  Here are your {len(fonts)} fonts"
-        surf = font.render(instructions, True, instruction_color, self.back_color, self.screen_size[0] - 20)
+        surf = font.render(
+            instructions,
+            True,
+            instruction_color,
+            self.back_color,
+            self.screen_size[0] - 20,
+        )
         font_surfaces.append((surf, total_height))
         total_height += surf.get_height() + padding
         max_width = max(max_width, surf.get_width())
@@ -113,7 +119,9 @@ class FontViewer:
                 continue
             line = text.replace("&N", name)
             try:
-                surf = font.render(line, 1, color, self.back_color, self.screen_size[0] - 20)
+                surf = font.render(
+                    line, 1, color, self.back_color, self.screen_size[0] - 20
+                )
             except pygame.error as e:
                 print(e)
                 break
--- a/examples/font_viewer.py
+++ b/examples/font_viewer.py
@@ -99,7 +99,12 @@ class FontViewer:
         # display instructions at the top of the display
         font = pygame.font.SysFont(pygame.font.get_default_font(), font_size)
         font.align = pygame.FONT_CENTER
-        instructions = f"Use the scroll wheel or click and drag to scroll up and down.  Fonts that don't use the Latin Alphabet might render incorrectly.  Here are your {len(fonts)} fonts"
+        instructions = (
+            "Use the scroll wheel or click and drag to scroll up "
+            "and down.  Fonts that don't use the Latin Alphabet "
+            "might render incorrectly.  Here are your "
+            f"{len(fonts)} fonts"
+        )
         surf = font.render(
             instructions,
             True,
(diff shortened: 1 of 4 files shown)


examples/font_viewer.py
```diff
@@ -155,7 +155,7 @@ def display_fonts(self):
                     bottom >= self.y_offset
                     and top <= self.y_offset + display.get_height()
                 ):
-                    x = center - surface.get_width() // 2
+                    x = (display.get_width() - surface.get_width()) // 2
```
Starbuck5, line 158: Why'd you choose to change this? This change makes the `center` variable above unused.

examples/font_viewer.py
```diff
@@ -99,19 +98,18 @@ def render_fonts(self, text="A display of font &N"):
...
         font = pygame.font.SysFont(pygame.font.get_default_font(), font_size)
-        lines = (
-            "Use the scroll wheel or click and drag",
-            "to scroll up and down.",
-            "Fonts that don't use the Latin Alphabet",
-            "might render incorrectly.",
-            f"Here are your {len(fonts)} fonts",
-            "",
+        font.align = pygame.FONT_CENTER
+        instructions = f"Use the scroll wheel or click and drag to scroll up and down.  Fonts that don't use the Latin Alphabet might render incorrectly.  Here are your {len(fonts)} fonts"
```
Starbuck5, line 102: This should be broken onto multiple lines.

examples/joystick.py
```diff
@@ -3,41 +3,22 @@
...
+    screen = pygame.display.set_mode(size)
     pygame.display.set_caption("Joystick example")
 
     # Used to manage how fast the screen updates.
     clock = pygame.Clock()
 
     # Get ready to print.
-    text_print = TextPrint()
+    font = pygame.font.SysFont(None, 25)
+    wraplength = size[1] - 20
```
Starbuck5, line 21: wraplength is the height minus 20, not the width?

examples/joystick.py
```diff
@@ -79,63 +60,66 @@ def main():
...
-        text_print.reset()
+        indentation = 0
+        lines = []
 
         # Get count of joysticks.
         joystick_count = pygame.joystick.get_count()
-
-        text_print.tprint(screen, f"Number of joysticks: {joystick_count}")
-        text_print.indent()
+        lines.append((indentation, f"Number of joysticks: {joystick_count}"))
```
Starbuck5, line 68: If you're going to embed the indentation information and a string in each line, why not just do the indent at that time and only add the indented string to the list of lines?

examples/font_viewer.py
JiffyRob: Probably because I had changed it to left align and then back to center when I heard about `pygame.font.align`.  I'll change this back.

examples/joystick.py
JiffyRob: Typo.  Will fix.

Starbuck5: This hasn't been resolved yet, fyi.

Starbuck5 (approved):
This looks good to me, thanks for working through this.

I'd like someone with joystick hardware to review the joystick example, if possible. I can't really test it in action because I don't have any joysticsks.

JiffyRob:
I did test it with a USB gamepad and AFAIK it did everything properly.  That said, I can't give a screenshot and a maintainer ought to test it anyway

oddbookworm (approved):
This looks good to me, but like Starbuck5, I can't test it myself. I'm not merging it because someone else should test it first

zoldalma999 (commented):
Tested the joystick example, and left some notes.

Also, you can set font.align to be centered for the dropevent example, I think that way it does look better.

examples/joystick.py
```diff
@@ -3,41 +3,22 @@
...
+    size = (500, 700)
+    screen = pygame.display.set_mode(size)
     pygame.display.set_caption("Joystick example")
 
     # Used to manage how fast the screen updates.
     clock = pygame.Clock()
 
     # Get ready to print.
-    text_print = TextPrint()
+    font = pygame.font.SysFont(None, 25)
```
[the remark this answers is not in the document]
JiffyRob: Well this won't be published out until the next version, where this won't be a problem anymore, right?  I don't think that this is something that we should do in the example if passing None is more consistent and works just as well.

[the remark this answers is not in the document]
Starbuck5: Moving to freesansbold also would change the font size because of the font constructor weirdness with None/freesansbold, and it emits a warning because sysfont thinks it can find it. Lets stick with None for now.

Actually, pygame.Font(None, 25) would be more conventional, imo.

Starbuck5:
It seems like the joystick example notes have been resolved, so I'm going to go ahead and merge. Thanks for working on this JiffyRob!

examples/joystick.py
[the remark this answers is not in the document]
Starbuck5: I think this is fine as it is right now.

PR #2238, opened 2023-06-10
Possible optimization for blit functions

itzpr3d4t0r:
Today I remembered of a comment made on discord, this one:
image
and thought that maybe there was something weird going on with blit rect clipping. Ran an ultra quick check and wanted to test something: adding a rect collision test between the destination rect and the surface clip_rect. While this seemed to work there might be better ways to do this, but i didn't have the time to check the actual blit function. This is also true for blits.

EDIT: i remade the test program to make it more compact and clearer, so i also had to change the results shown.
```Python
import pygame
from timeit import repeat
from statistics import mean


pygame.init()
WIDTH, HEIGHT = 400, 400

surf = pygame.Surface((WIDTH, HEIGHT))
surf.fill((255, 0, 0))

lst_inside = [
    (surf, (350, 350)) for _ in range(1000)
]

lst_outside = [
    (surf, (1000, 1000)) for _ in range(1000)
]

G = globals()

def test(name: str, number=1000, rep=10) -> float:
    return mean(repeat(name, globals=G, number=number, repeat=rep))


print("fblits - outside", test("win.fblits(lst_outside)"))
print("blits - outside", test("win.blits(lst_outside)"))
print("blit - outside", test("win.blit(surf, (1000, 1000))"))
print()
print("fblits - inside", test("win.fblits(lst_inside)"))
print("blits - inside", test("win.blits(lst_inside)"))
print("blit - inside", test("win.blit(surf, (350, 350))"))
```

**WITH THE CHANGES**
```
fblits - outside 0.03345977000008134
blits - outside 0.10718321000003925
blit - outside 0.00023007999989204108

fblits - inside 0.6032612599999994
blits - inside 0.704143540000041
blit - inside 0.0008147099999405328
```
**MAIN**
```
fblits - outside 0.14072892000003775
blits - outside 0.208359400000154
blit - outside 0.00032422000003862195

fblits - inside 0.599861510000028
blits - inside 0.6935029799999939
blit - inside 0.0007882900000367954
```
So this change made not drawing shapes 4X faster and didn't slow down visible blits.

### DIFF
--- a/src_c/surface.c
+++ b/src_c/surface.c
@@ -2347,6 +2347,10 @@ surf_fblits(pgSurfaceObject *self, PyObject *const *args, Py_ssize_t nargs)
             dest_rect.w = src->w;
             dest_rect.h = src->h;
 
+            if (!SDL_HasIntersection(&dest_rect, &dest->clip_rect)) {
+                continue;
+            }
+
             /* Perform the blit */
             if (pgSurface_Blit(self, (pgSurfaceObject *)src_surf, &dest_rect,
                                NULL, blend_flags)) {
--- a/src_c/surface.c
+++ b/src_c/surface.c
@@ -1924,6 +1924,11 @@ surf_blit(pgSurfaceObject *self, PyObject *args, PyObject *keywds)
     dest_rect.w = src_rect->w;
     dest_rect.h = src_rect->h;
 
+    if (!SDL_HasIntersection(&dest_rect, &dest->clip_rect)) {
+        dest_rect.w = dest_rect.h = 0;
+        return pgRect_New(&dest_rect);
+    }
+
     if (!blend_flags)
         blend_flags = 0;
 
@@ -2088,6 +2093,10 @@ surf_blits(pgSurfaceObject *self, PyObject *args, PyObject *keywds)
         dest_rect.y = dy;
         dest_rect.w = src_rect->w;
         dest_rect.h = src_rect->h;
+
+        if (!SDL_HasIntersection(&dest_rect, &dest->clip_rect)) {
+            continue;
+        }
 
         if (special_flags) {
             if (!pg_IntFromObj(special_flags, &blend_flags)) {
@@ -2259,6 +2268,10 @@ surf_fblits(pgSurfaceObject *self, PyObject *const *args, Py_ssize_t nargs)
 
             dest_rect.w = src->w;
             dest_rect.h = src->h;
+
+            if (!SDL_HasIntersection(&dest_rect, &dest->clip_rect)) {
+                continue;
+            }
 
             /* Perform the blit */
             if (pgSurface_Blit(self, (pgSurfaceObject *)src_surf, &dest_rect,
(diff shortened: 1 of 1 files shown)


itzpr3d4t0r:
Changed test program again to include one more case, here is the new one and new results:
**WITH THE CHANGES**
```
OUTSIDE
fblits) 0.036678159999974015
blits doreturn) 0.1063232300000891
blits) 0.08050288000013098
blit) 0.00020791000006283866

INSIDE
fblits) 0.605369150000115
blits doreturn) 0.7010963499998979
blits) 0.6715423200004806
blit) 0.0008256900002379553
```
**MAIN**
```
OUTSIDE
fblits) 0.13856487999983075
blits doreturn) 0.208694860000287
blits) 0.18958259999963048
blit) 0.0003108700000666431

INSIDE
fblits) 0.5962242999994487
blits doreturn) 0.703472459999648
blits) 0.6794426200000089
blit) 0.000789950000216777
```

```Python
import pygame
from timeit import repeat
from statistics import mean


pygame.init()
WIDTH, HEIGHT = 400, 400

surf = pygame.Surface((WIDTH, HEIGHT))
surf.fill((255, 0, 0))

lst_inside = [
    (surf, (350, 350)) for _ in range(1000)
]

lst_outside = [
    (surf, (1000, 1000)) for _ in range(1000)
]

G = globals()

def test(name: str, number=1000, rep=10) -> float:
    return mean(repeat(name, globals=G, number=number, repeat=rep))

print("OUTSIDE")
print("fblits)", test("win.fblits(lst_outside)"))
print("blits doreturn)", test("win.blits(lst_outside)"))
print("blits)", test("win.blits(lst_outside, False)"))
print("blit)", test("win.blit(surf, (1000, 1000))"))

print("\nINSIDE")
print("fblits)", test("win.fblits(lst_inside)"))
print("blits doreturn)", test("win.blits(lst_inside)"))
print("blits)", test("win.blits(lst_inside, False)"))
print("blit)", test("win.blit(surf, (350, 350))"))
```

Starbuck5:
That is quite an interesting result, cool work.

SDL_BlitSurface and pygame_Blit both do bounds checking, so it's not like the blit is happening, it's just the set up and tear down of the blit operation that is taking time.

In SDL, there is SDL_UpperBlit and SDL_LowerBlit. UpperBlit is the same as BlitSurface, and does bounds checking. LowerBlit does no bounds checking.

So maybe a good approach to implement this would be to make pgSurface_Blit do the early exit bounds checking itself, then assuming it has a blit to do, call a new function pgSurface_BlitLower. This function could call `SDL_LowerBlit` or a new function `pygame_BlitLower` (`pygame_Blit` without bounds checking) so that it doesn't replicate the bounds check.

MyreMylar:
I did a quick test swapping in `SDL_LowerBlit()` for `SDL_BlitSurface()` in this PR branch, but it segfaults so clearly `SDL_BlitSurface()` is still doing something useful we are not capturing at the higher level yet.

MyreMylar (commented):
I feel like this one needs a little more investigation time.

Issue #2236, opened 2023-06-09
`pg.mouse.get_pos()` is unfunctional after 2.3.0.dev2 


yolowex:
**Environment:**

You can get some of this info from the text that pops up in the console when you run a pygame program.

- **Operating system**:
* OS: Kali GNU/Linux Rolling x86_64
* Kernel: 6.1.0-kali9-amd64
- **Pygame, SDL and python version**:
* pygame-ce 2.3.0.dev2 (SDL 2.26.4, Python 3.11.2)
- **Relevant hardware** :
* Bus 001 Device 005: ID 18f8:0f99 [Maxxter] Optical gaming mouse ( this is the mouse, I got it with lsusb, its brand is tsco )

**Current behavior:**

I always update my pygame to the latest dev version. but after I installed dev2, I noticed that `pg.mouse.get_pos` does not 
work anymore and it only returns `[-1,-1]`. 

**Expected behavior:**

I want that method to return the actual position of the mouse.

**Steps to reproduce:**
this issue is only reproducible on my own computer, it is mostly likely a problem in my Linux machine.

**Test code**
```python
import pygame as pg
from pygame.locals import *
pg.init()
screen = pg.display.set_mode([800,600] )
run = True
while run:
    for i in pg.event.get():
        if i.type == QUIT or i.type == KEYDOWN and i.key == K_ESCAPE:
            run = False
        if i.type == MOUSEMOTION:
            print(pg.mouse.get_pos())
    pg.display.update()
```

# important notation
1. This problem does **NOT** occur while using the SCALED flag for the window. ( JoKing found this out )

2. I also checked multiple desktop sessions to see if this is a `wayland` or `plasma` problem, but it's not.
![sessions](https://github.com/pygame-community/pygame-ce/assets/122750743/2b440bcc-9a63-4689-aae5-6cb135179539)
I use Plasma (Wayland) as my primary desktop session. 
I tried both Plasma (X11) and Xfce Session, but the issue persisted regardless of the desktop session.  

I can check my GNOME session too, and also send any data that is needed to solve this problem.

Starbuck5:
I think this is a product of https://github.com/pygame-community/pygame-ce/pull/2088 with the "Renderer being created even though no renderer was created" issues.

Starbuck5:
I think this may fix it: #2247 

@yolowex I'd appreciate if you could give that PR a test.

PR #2088, opened 2023-03-31
fix #2087 until SDL catches up

robertpfeiffer:

PR #2247, opened 2023-06-13
Try reverting PR #2088 to solve #2236

Starbuck5:
This reverts commit 5924b0c6a8b9229edb1141904a2f314764a3bbee, reversing changes made to 16a17de867bb8957aac572f1dc808bb215c51640.

The problem is that sometimes SDL creates a renderer internally. So when we get a renderer, we can't assume that we created it (we can't assume we set the logical size in SCALED).

This could be fixed more legitimately adding a way to get the scaled renderer into the C api or something, but for the purposes of 2.3.0 I think we should just revert to the 2.2.1 status quo.

### DIFF
--- a/src_c/mouse.c
+++ b/src_c/mouse.c
@@ -76,7 +76,6 @@ mouse_get_pos(PyObject *self, PyObject *_null)
         if (sdlRenderer != NULL) {
             SDL_Rect vprect;
             float scalex, scaley;
-            int w, h;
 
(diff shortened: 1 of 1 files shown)


rethanon:
Fixed the issue for me - tested on:

Intel Mac, python 3.11
Windows 11 PC, python 3.11

ankith26 (approved):
Reverting the PR does seem like the way to go now, thanks for the PR :tada:

PR #2254, opened 2023-06-15
Add lerp & smoothstep to math module

robertpfeiffer:
e.g.

```python
import pygame

a = 0.0
b = 10.0
pygame.math.lerp(a, b, 0.5)

>>> 5.0
```


src_c/math.c
```diff
@@ -595,6 +595,7 @@ pgVector_NEW(Py_ssize_t dim)
 static void
 vector_dealloc(pgVector *self)
 {
+    PyMem_Free(self->coords);
```
Starbuck5, line 598: Coords is not dynamic memory anymore on pygame-ce main.

It doesn't need to be allocated or freed, it can't be allocated or freed. See #2174

robertpfeiffer: Can you guess I wrote this in March and rebased it? Sorry to cause you trouble.

Starbuck5:
I'm surprised that these functions do vectors and numbers. It seems reasonable to have the module functions just do numbers and the vector methods do vectors. There's already a way to lerp two vectors together, with vector.lerp.

MyreMylar:
I agree with Starbuck. We should add a `.smoothstep()` method to the two vector classes and then have scalar/number only versions at the submodule function level. 

That way there is one clear way to do it with either type of thing.

MyreMylar (changes requested):
As commented above. Please make the module level functions for scalars only, and - if you like add the vector code for smoothstep as a method for the Vector classes.

This is to keep it consistent with the already existing `Vector2.lerp()` & `Vector3.lerp()`.

It has been a long ass time, for me to get around to reviewing this so also happy to take this over if you don't want to make any changes.

docs/reST/ref/math.rst
```diff
@@ -61,6 +61,33 @@ Multiple coordinates can be set using slices or swizzling
 
    .. ## math.clamp ##
 
+.. function:: lerp
+
+   | :sl:`returns value interpolated between a and b`
+   | :sg:`lerp(a, b, value, do_clamp=True) -> float`
+
+   Returns a Vector or number which is a linear interpolation between ``a``
+   and ``b``. The third parameter determines how far between ``a`` and
```
dr0id, line 70: please add an explanation for the range of the `third parameter`

docs/reST/ref/math.rst
```diff
@@ -61,6 +61,33 @@ Multiple coordinates can be set using slices or swizzling
...
+   .. versionadded:: 2.4.0
+
+   .. ## math.lerp ##
+
+ .. function:: smoothstep
+
+   | :sl:`returns value smoothly interpolated between a and b.`
+   | :sg:`smoothstep(a, b, value) -> float`
+
+   Returns a Vector or number which is a smooth interpolation between ``a``
```
dr0id, line 83: what is a smooth interpolation?

docs/reST/ref/math.rst
```diff
@@ -61,6 +61,33 @@ Multiple coordinates can be set using slices or swizzling
...
+   and ``b``. The third parameter determines how far between ``a`` and
```
dr0id, line 84: also here, define the range of the parameter I guess its `[0, 1]`

src_c/math.c
```diff
@@ -4193,6 +4193,127 @@ math_clamp(PyObject *self, PyObject *const *args, Py_ssize_t nargs)
...
+
+    double t = PyFloat_AsDouble(value);
+    if (t <= 0.0) {
+        t = 0;
+    }
+    else if (t >= 1.0) {
+        t = 1;
+    }
+    else {
+        t = t * t * (3.0f - 2.0f * t);
```
dr0id, line 4280: `(3.0f - 2.0f * t);`  this part needs some explanation I guess. Otherwise I wonder why 3 -2 = 1 => `1.0f * t` should be different than just `t`

docs/reST/ref/math.rst
dr0id: maybe even write the formula: `a * value + (1-value) * b`
looking at this I would suggest to rename value to `t`

docs/reST/ref/math.rst
dr0id: maybe also include the formula? like `a * t + (1-t) * b where t = t*t*(3 - 2*t) if t in [0, 1]`

suggestion: rename value to t

MyreMylar (approved):

MyreMylar:
Tidied this up a bit to respond to reviews.

Matiiss (approved):
Code looks great, seems to run fine, these are wonderful additions. Although I'm not sure why the vector methods don't use `METH_FASTCALL` too, but that can be changed later if necessary.

PR #2255, opened 2023-06-16
Fix segfault when rendering fonts

yunline:
Fixes: #1506
**Test code:**
```py
import pygame

pygame.init()
font = pygame.font.Font(None, 40)

pygame.quit()
pygame.init()

font.render("hello", True, "white")
#font.size("hello")
#font.metrics("hello")
```
Now it will raise an error instead of segfault
```
pygame-ce 2.3.1.dev1 (SDL 2.26.4, Python 3.10.6)
  File "x:\python\pygame_test\segfault.py", line 10, in <module>
    font.render("hello", True, "white")
pygame.error: Invalid font.
```

### DIFF
--- a/docs/reST/ref/font.rst
+++ b/docs/reST/ref/font.rst
@@ -57,6 +57,8 @@ solves no longer exists, it will likely be removed in the future.
 
    It is safe to call this function even if font is currently not initialized.
 
+   Previously created font objects will be invalid after the font module is quit.
+
    .. ## pygame.font.quit ##
 
 .. function:: get_init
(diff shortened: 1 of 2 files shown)


novialriptide (approved):
I see nothing out of the ordinary. LGTM!

Starbuck5:
This looks great!

I'm wondering if we can be more clear with the "invalid font" error message. "Invalid font (font module quit since font created)" ?

It would also be good to add some documentation about this whole generation constraint. I think it would go well as another line in pygame.font.quit, mentioning that previously created font objects become invalid after the font module is quit.

Starbuck5 (approved):
Thanks!

PR #2258, opened 2023-06-18
Update joystick.rst - Switch Joy-cons mappings wrong

bilhox:
Hello,

After using Nintendo Switch Joy-cons since yesterday, I found out that button IDs given for Nintendo switch Joy-cons in the documentation page of joystick module are wrong.

I don't know who wrote the button ids, but if it was the right IDs for them, then it should be checked again so we can be sure which ones are correct, and if it depends of OS.

Furthermore, the other button IDs for other controllers should be checked as well.

Bellow the code to check button IDs : 

```py
import pygame
import sys

screen = pygame.display.set_mode([800, 500])

pygame.init()
joysticks : dict[str , pygame.joystick.Joystick] = {}

while True:

    screen.fill([0,0,0])

        if event.type == pygame.QUIT:
            pygame.quit()
            sys.exit(0)        

        elif event.type == pygame.JOYBUTTONDOWN:
            
            joystick = joysticks[event.instance_id]
            print(joystick.get_name(), event.button)

            if joystick.rumble(0, 0.5, 200):
                print(f"Rumble effect played on joystick {joystick.get_name()}")

        elif event.type == pygame.JOYDEVICEADDED:
            joystick = pygame.joystick.Joystick(event.device_index)
            joysticks[joy.get_instance_id()] = joystick
            print(f"Joystick {joystick.get_instance_id()} connected")
            print("Joystick name : " + joystick.get_name())
            print(f"Number of axis : {joystick.get_numaxes()}")
            print(f"Power level : {joystick.get_power_level()}") # Doesn't seem to work


        elif event.type == pygame.JOYDEVICEREMOVED:
            if event.instance_id in joysticks:
                del joysticks[event.instance_id]
                print(f"Joystick {event.instance_id} disconnected")

    pygame.display.flip()
```


novialriptide:
I wrote them and I got them verified by @Mega-JC, SDL might've had an update that made them change.

novialriptide:
I have checked again between pygame-ce 2.1.3 and pygame-ce 2.3.0, the controller mappings are different.

bilhox:
You mean, they are different between versions, or they are different of mines ?

novialriptide:
> You mean, they are different between versions, or they are different of mines ?

They're different between versions.

bilhox:
Okay, also I forgot to mention that if we have the Nintendo Joy-con L and Nintendo Joy-con R connected in the same time, the two get removed and now a Nintendo Joy-con (L/R) is added. 
Before updating docs, I would like to mention 2 issues with this device :

- The rumbling effect only works on the Right Joy-con but not the Left one
- ZL and ZR buttons don't work (is it intended ?)

It would be a pleasure if you can test it and see if you have the same issue.

Starbuck5:
If they're different between versions this should note in the documentation that these numbers have varied in the past and were last verified for pygame-ce 2.3.0.

It seems like the real solution for this from a user perspective is to use the controller module for these, rather than joystick.

zoldalma999:
Not sure what is the best way to document this, since this is not tied to pygame version, but sdl version... of which we support a lot of different ones. And a specific pygame version does not mean a specific sdl version.

Whenever controller releases, I reckon (as Starbuck mentioned) we just remove all of the mappings and just say "use the controller module instead for this purpose". Until then though, not sure. However I don't think this is the best solution. It is just that I don't know what *would* be better either.

bilhox:
hello,

I understand your POV, is dropping older SDL versions possible ? Is it because of other OS we still support older SDL versions ?

MyreMylar:
>  Is it because of other OS we still support older SDL versions ?

Exactly that. We've just had a new PR to re-add support for SDL 2.0.9 because that is what ships with a still-in-use version of Debian released back in 2018. Sadly this sometimes means bugs will exist for some users in a given version of pygame-ce that do not exist for others simply due to the version of system SDL they have installed for their OS.

A large part of pygame/pygame-ce is wrapping calls to underlying SDL functions.

I expect we could probably update the joystick docs to match the behaviour of the newest version of SDL in the interim as that is most likely to be accurate. Perhaps add a note under the mapping saying that they were checked with a specific SDL version.

bilhox:
Hello,

After researches, I found out in which version the mappings changed. The wiki page got updated :

- Fixed old mappings and added a new column of mappings (Before and After SDL 2.24)
- Added a new section about Nintendo Switch Joy-con (L/R) device

ZR and ZL buttons don't work, I don't know why, however, after a test with SDL 3.0, the buttons worked, so I just wrote in the Wiki a '?' for now.

zoldalma999 (approved):
I don't have the necessary hardware to test the claims here, but looks good to me. At some point this will be removed in favor of controller module anyway. Thanks! 🎉

Starbuck5 (approved):
Thanks!

Issue #2259, opened 2023-06-19
Documentation relook or rephrase in certain sections


simplyrohan:
Looking into many of the sections in the current pygame-ce documentation from a beginner's viewpoint, it can seem absolutely terrifying. Some of the most fundamental ~~and simple~~ parts have been overcomplicated and the basics have been muddied. One of the most evident examples is the [`pygame.display.flip` doc](https://pyga.me/docs/ref/display.html#pygame.display.flip).
This is virtually the most important line in an app, but the description is extremely complicated. One sentence is dedicated to the main functionality! One!!! The rest is into the specifics. For a beginner looking into pygame for the first time, this is impossible!

I think these should retain their current content but also get a few more sentences for the basics of what these methods do.

`pygame.display`, `pygame.math`, `pygame.image`,  and `pygame.mixer`

I know that this is vague, and doesn't give any specifics. But simply adding a sentence or two about the basic operation of these methods would do great.

Footnote:
I know reference docs aren't a great place to start as a beginner, but it is the most obvious place to go.

Starbuck5:
Lots of content has accumulated in the documentation over time. I feel like more eyes on the docs with the current status of everything could rework descriptions to be more clear.

PRs welcome.
Change could be as simple as doing one function/method.

itzpr3d4t0r:
I had a similar thought recently. I would be glad to revise some documentation, such as the blit documentation, to make it more beginner-friendly. Additionally, we should consider removing references to older versions of Pygame, as Pygame-ce's first version is 2.1.3. Listing previous versions wouldn't be very relevant and is just confunsing/of no use to new users. Here's an example that illustrates my point:
![image](https://github.com/pygame-community/pygame-ce/assets/103119829/dda75726-e983-4a0e-92e8-17f65ac8023a)
In my opinion, it no longer serves a purpose to include versions other than Pygame-ce's versions beyond 2.1.3.

Starbuck5:
I still agree with you Rohan, but this issue isn't super actionable and it's gotten old at this point, so I'm going to close this. If you or anyone else in the future want to refactor parts of the docs feel free to open more (potentially more targeted) issues.

PR #2262, opened 2023-06-19
Blit docs rewrite

itzpr3d4t0r:
Follows #2259. This one aims to improve the blit docs. I heavily encourage feedback as this is just a starting point. I tried to make docs easy to read and comprehend to a beginner but i may have rephrased some things wrong and/or made big mistakes.


itzpr3d4t0r:
Special flags section got changed to include more information and now looks like this:
image

Starbuck5:
Here's an idea to chew on. What if the special flags explanation was a completely separate page, like the list of default colors in pygame.color. Right now the function is more daunting to look at because the description is so detailed, and a lot of that length is now coming from the special_flags documentation.

docs/reST/ref/surface.rst
```diff
@@ -92,41 +92,118 @@
...
+
+              -) Adding white to any color produces white.
+
+              -) Adding any color to white produces white.
+
+          ``BLEND_SUB`` / ``BLEND_RGB_SUB``
+              Subtracts the source color channels from the destination color channels, clamped to a minimum of 0.
+              The result color is always a darker color.
+
+              -) Subtracting any color from black leaves the color unchanged.
```
Starbuck5, line 133: Subtracting black from any color leaves the color unchanged.

docs/reST/ref/surface.rst
```diff
@@ -92,41 +92,118 @@
...
       | :sg:`blit(source, dest, area=None, special_flags=0) -> Rect`
 
-      Draws a source Surface onto this Surface. The draw can be positioned with
-      the dest argument. The dest argument can either be a pair of coordinates representing the position of
-      the upper left corner of the blit or a Rect, where the upper left corner of the rectangle will be used as the
-      position for the blit. The size of the destination rectangle does not
-      effect the blit.
+      Draws a Surface onto this Surface. The placement of the ``source`` surface on this surface
+      can be specified by passing either a coordinate (x, y) or a Rect as the ``dest``
+      parameter. The ``dest`` parameter denotes the upper left corner from where the blit operation
```
Starbuck5, line 97: `The placement of the ``source`` surface on this surface can be specified by passing either a coordinate (x, y) or a Rect as the ``dest`` parameter.`

This reminds me of "passive voice" I learned about in school, and could perhaps be rewritten to be more direct.

`The ``dest`` parameter specifies the placement of the ``source`` surface onto this Surface, and takes a coordinate (x, y) or a Rect.`

Better maybe? I'm curious what your take on this will be, look up passive voice so you get what I'm talking about.

docs/reST/ref/surface.rst
```diff
@@ -92,41 +92,118 @@
...
+      will commence. In the case of a Rect, its top-left corner will be used.
```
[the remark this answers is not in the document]
Starbuck5: Same with commece.

docs/reST/ref/surface.rst
```diff
@@ -92,41 +92,118 @@
...
+              Works like ``BLEND_RGB_MIN``, but also minimizes the alpha channel.
+          ``BLEND_RGBA_MAX``
+              Works like ``BLEND_RGB_MAX``, but also maximizes the alpha channel.
+
+      **Special Alpha Blending (RGBA)**
+          ``BLEND_PREMULTIPLIED``
+              Uses premultiplied alpha blending for faster and accurate results when
+              the color channels are already multiplied by the image alpha channel.
+              You should only use this blend mode if you previously premultiplied the source
+              surface with :meth:`premul_alpha()`.
```
Starbuck5, line 178: I like the callout and link to premul_alpha here!

docs/reST/ref/surface.rst
itzpr3d4t0r: Yeah that was kinda passive and convoluted from the beginning tbh.

docs/reST/ref/surface.rst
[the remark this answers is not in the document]
itzpr3d4t0r: I swapped them with "begins" and "represents" respectively.

Starbuck5:
Just for reference this is what I'm talking about: https://pyga.me/docs/ref/color_list.html Whole separate page not in the modules top bar.

itzpr3d4t0r:
This is how the docs would look like without flags, i still feel like i should at least provide a basic list of available flags and note that for detailed explaination you should check wathever.
image

docs/reST/ref/surface.rst
Starbuck5: x - y != y - x

itzpr3d4t0r:
This is how it looks now after the changes:
First the special_flags section inside Surface():
image
image

The blit docs:
image
image

itzpr3d4t0r:
image
image

itzpr3d4t0r:
> Here's an idea to chew on. What if the special flags explanation was a completely separate page, like the list of default colors in pygame.color.

Should this be implemented in this PR or a followup?

Starbuck5:
I just don't like that this PR seems to make the blit docs significantly longer, when part of the issue being brought up by OP was the docs being too long and overcomplicated for entry level users. I'm not sure how you'd like to tackle this in PRs.

itzpr3d4t0r:
I don't find these docs too overcomplicated, maybe a bit too long still but mind that writing docs isn't an easy task, as is demonstrated by how poorly our docs are written (my fblits docs as well). With this PR I think I added a lot more information that wasn't even there before, and organised it in a way in which if you're a beginner you can simply read the first 6 lines and you are fine but if you're an advanced user you can go more in-depth further down. Sorry but aside from it beeing a bit too long and the possibility of making a separate page for special flags i just don't see how these docs are bad. They are better structured, have a clear begginner at the top and advanced at the bottom structure, add meaningful information that you could've only gotten with errors/debug.

itzpr3d4t0r:
Latest revision:
image

docs/reST/ref/surface.rst
```diff
@@ -87,46 +87,152 @@
...
+      **Return**
+
+          A ``Rect`` object representing the affected area of this ``Surface`` that was modified
+          by the blit operation. This area includes only the pixels within this ``Surface`` or
+          its clipping area (see :meth:`set_clip`).
+
+      **Notes**
+
+          .. note:: If the ``dest`` parameter is a ``Rect`` object, only its ``x`` and ``y`` coordinates
+                    are used, its will not affect the blit.
```
oddbookworm, line 221: Did part of the sentence get missed here?

docs/reST/ref/surface.rst
```diff
@@ -87,46 +87,152 @@
...
+              - ``special_flags`` is a special flag that controls how the source Surface is drawn.
 
-      .. versionaddedold:: 1.8.1
-         Optional ``special_flags``: ``BLEND_RGBA_ADD``, ``BLEND_RGBA_SUB``,
-         ``BLEND_RGBA_MULT``, ``BLEND_RGBA_MIN``, ``BLEND_RGBA_MAX``
-         ``BLEND_RGB_ADD``, ``BLEND_RGB_SUB``, ``BLEND_RGB_MULT``,
-         ``BLEND_RGB_MIN``, ``BLEND_RGB_MAX``.
+          The ``source`` parameter is the ``Surface`` object that will be drawn onto this Surface.
 
-      .. versionaddedold:: 1.9.2
```
oddbookworm, line 114: Are none of these being kept?

oddbookworm (changes requested):

docs/reST/ref/surface.rst
itzpr3d4t0r: Thanks, fixed that.

docs/reST/ref/surface.rst
itzpr3d4t0r: Where they useful to a pygame.ce user in any way?

itzpr3d4t0r: Anyway I've opened an issue on this #2281.

itzpr3d4t0r: It seems the sentiment is to keep them so I've readded them.

Starbuck5:
Went over this a bit again.

- Weird to have a bulleted list followed by 4 paragraphs explaining in further detail. All the detail could be in the bulleted list of parameters in the first place.
- Lets do a separate page for the flags, like `color_list`
- The notes section needs to get restructured, because the notes all look very high importance, but they're not that important. I suggested having them as a paragraph of bits and bobs after the list of parameters.

docs/reST/ref/special_flags_list.rst
```diff
@@ -0,0 +1,112 @@
...
+
+----
+
+    .. versionaddedold:: 1.9.2
+
+       - ``BLEND_PREMULTIPLIED``
+           Uses premultiplied alpha blending for faster and accurate results when
+           the color channels are already multiplied by the image alpha channel.
+           You should only use this blend mode if you previously premultiplied the Surface with
+           :meth:`premul_alpha()`. Check out :doc:`surface`'s :meth:`premul_alpha()` method for more
```
Starbuck5, line 96: Rather than referencing a class and then referencing a method, you could just reference Surface.premul_alpha directly.

Starbuck5:
So now the list of parameters is not a bulleted list at all, which looks weird to me. The pygame functions usually have a bulleted list of parameters (draw) or paragraphs.

There is now a bulleted list of notes, which is better than before. But it's just weird that almost all the notes directly map to one parameter. Like why not have the information in the parameter descriptions?

docs/reST/ref/surface.rst
```diff
@@ -92,41 +92,36 @@
       | :sl:`draw one image onto another`
```
MyreMylar, line 92:
```suggestion
      | :sl:`draw one surface onto another`
```

docs/reST/ref/special_flags_list.rst
```diff
@@ -0,0 +1,111 @@
...
+
+**Special Alpha Blending (RGBA)**
+
+----
+
+    .. versionaddedold:: 1.9.2
+
+       - ``BLEND_PREMULTIPLIED``
+           Uses premultiplied alpha blending for faster and accurate results when
+           the color channels are already multiplied by the image alpha channel.
```
MyreMylar, line 94:
```suggestion
           Uses premultiplied alpha blending for slightly faster blits and more 
           accurate blending results when the color channels are already multiplied 
           by the image alpha channel.
```

docs/reST/ref/surface.rst
```diff
@@ -92,41 +92,36 @@
...
-
-      The return rectangle is the area of the affected pixels, excluding any
-      pixels outside the destination Surface, or outside the clipping area.
-
-      Pixel alphas will be ignored when blitting to an 8 bit Surface.
-
-      For a surface with colorkey or blanket alpha, a blit to self may give
-      slightly different colors than a non self-blit.
+      Draws a Surface onto this Surface.
+
```
MyreMylar, line 96:
```suggestion

      **Common usage**
          - window_surface.blit(my_surface, (100, 100))
```

I feel like we need something a bit like this right at the top to clearly indicate the basic, standard use of blit that most new users will try out before delving into anything like special_flags, or blitting sub areas. Maybe in a code block instead of just loose like this.

docs/reST/ref/surface.rst
```diff
@@ -92,41 +92,36 @@
...
+              If the ``Rect`` has negative position, the final blit position will be
+              ``dest`` - ``Rect.topleft``.
+          - ``special_flags`` *(optional)*
+              Controls how the colors of the ``source`` are combined with this Surface.
+              If not provided it defaults to ``BLENDMODE_NONE`` (``0``).
+              See :doc:`special_flags_list` for a list of possible values.
+      **Return**
+          A :doc:`rect` object representing the affected area of this ``Surface`` that was modified
+          by the blit operation. This area includes only the pixels within this ``Surface`` or
+          its clipping area (see :meth:`set_clip`).
```
MyreMylar, line 118: I'd be tempted to mention that in most cases you generally do not need to keep this return value as it is rarely used for anything (it is more of a legacy optimisation when updating the screen pixels was expensive enough that it was worth calculating sub regions of it to update). Just because most new programmers are trained to care about return values of functions and may be wondering why they need this rectangle.

MyreMylar (commented):
This seems better overall to me, I think we are closing in on something that addresses the new user to pygame.

I added a few suggestions you can take or leave.

docs/reST/ref/surface.rst
itzpr3d4t0r: I like this suggestion but I'm a bit worried that it would break the docs structure, tho I could add a "usage" section for that

itzpr3d4t0r: Like do you mean a full program like this one?
image
The main issue is that it's taking the same space as the parameters and return sections combined...

MyreMylar:
No that is way too much just a short one line code snippet.

docs/reST/ref/surface.rst
MyreMylar: No I was just thinking of a code snippet. Like so:

**Example usage:**
```python
display_surface.blit(player_surface, (100, 100)
```

MyreMylar: I think what you've added looks fine.

docs/reST/ref/special_flags_list.rst
```diff
@@ -0,0 +1,112 @@
...
+=======================
+
+    **Special flags are a means of controlling how a** :class:`Surface` **is drawn onto another**.
+    They can be used to create visual effects, such as glowing particles, or to perform surface
+    masking or manipulation.
+    They are used with the following methods:
+
+        - :meth:`pygame.Surface.blit`
+        - :meth:`pygame.Surface.blits`
+        - :meth:`pygame.Surface.blits`
```
MyreMylar, line 17:
```suggestion
```
`pygame.Surface.blits` in here twice.

docs/reST/ref/special_flags_list.rst
```diff
@@ -0,0 +1,112 @@
...
+----
+
+    .. versionaddedold:: 1.9.2
+
+       - ``BLEND_PREMULTIPLIED``
+           Uses premultiplied alpha blending for slightly faster blits and more
+           accurate blending results when the color channels are already multiplied
+           by the surface alpha channel.
+           You should only use this blend mode if you previously premultiplied the Surface with
+           :meth:`pygame.Surface.premul_alpha()`.
```
MyreMylar, line 97:
```suggestion
           You should only use this blend mode if you previously premultiplied the Surface with
           :meth:`pygame.Surface.premul_alpha()`, or if you know that the Surface was already created or loaded in
           with premultiplied alpha colors. You can read more about the advantages of `premultiplied alpha blending 
           here <https://en.wikipedia.org/wiki/Alpha_compositing>`_.
```

MyreMylar (approved):
Looks good over all - I added a couple of small suggestions.

Thanks for the hard work on this 👍

docs/reST/ref/special_flags_list.rst
itzpr3d4t0r: Oops, was supposed to be referencing fblits.

MyreMylar:
Looks like this one needs a merge with main to clear the CircleCI fail.

dr0id (approved):

PR #2263, opened 2023-06-19
Improved `blits()` docs

itzpr3d4t0r:
Follows #2259. This one aims to improve the `Surface.blits()` method docs.

**Added:**
- A better and more natural explaination of what the function parameters are and what they do.
- A thorough explaination of what each item in the blit_sequence can be composed of and its resulting behaviour.
- A series of notes regarding suggestions about its use/when not to use blits over blit or fblits.

**Removed:**
- The compact parameters version, as they weren't used extensively in the surface module and forced having shorter descriptions.

Before:
image

After:
image

### DIFF
--- a/docs/reST/ref/surface.rst
+++ b/docs/reST/ref/surface.rst
@@ -133,22 +133,54 @@
    .. method:: blits
 
       | :sl:`draw many images onto another`
-      | :sg:`blits(blit_sequence=((source, dest), ...), doreturn=1) -> [Rect, ...] or None`
+      | :sg:`blits(blit_sequence=((source, dest), ...), doreturn=True) -> [Rect, ...] or None`
       | :sg:`blits(((source, dest, area), ...)) -> [Rect, ...]`
       | :sg:`blits(((source, dest, area, special_flags), ...)) -> [Rect, ...]`
 
-      Draws many surfaces onto this Surface. It takes a sequence as input,
-      with each of the elements corresponding to the ones of :meth:`blit()`.
-      It needs at minimum a sequence of (source, dest).
+      The ``blits`` method efficiently draws a sequence of surfaces onto this ``Surface``.
 
-      :param blit_sequence: a sequence of surfaces and arguments to blit them,
-         they correspond to the :meth:`blit()` arguments
-      :param doreturn: if ``True``, return a list of rects of the areas changed,
-         otherwise return ``None``
+      **Parameters**
 
-      :returns: a list of rects of the areas changed if ``doreturn`` is
-         ``True``, otherwise ``None``
-      :rtype: list or None
+              - ``blit_sequence``: The sequence of surfaces to blit.
+              - ``doreturn``: (optional) Modifies the return value of the method.
+
+          The ``blit_sequence`` parameter is a sequence that contains each surface to be drawn
+          along with its associated blit arguments. See the **Sequence Item Formats** section below
+          for the possible formats.
+
+          The ``doreturn`` parameter controls the return value. When set to ``True``, it returns
+          a list of rectangles representing the changed areas. When set to ``False``, returns
+          ``None``.
+
+      **Return**
+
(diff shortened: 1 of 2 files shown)


docs/reST/ref/surface.rst
```diff
@@ -137,20 +137,67 @@
...
-         otherwise return ``None``
+      The method has two parameters:
 
-      :returns: a list of rects of the areas changed if ``doreturn`` is
-         ``True``, otherwise ``None``
-      :rtype: list or None
+          - ``blit_sequence`` is the sequence of surfaces to blit
+          - ``doreturn`` modifies the return value of the method from a list of rectangles to ``None``
 
-      .. versionaddedold:: 1.9.4
```
oddbookworm, line 153: I feel like this should be kept

oddbookworm (changes requested):
I don't see any problems with this other than the comment I left

docs/reST/ref/surface.rst
```diff
@@ -137,20 +137,53 @@
...
+            - ``dest``: Position where the source Surface should be blitted.
+
+          ``(source, dest, area)``
+            - ``area``: (optional) Specific area of the source Surface to be drawn.
+
+          ``(source, dest, area, special_flags)``
+            - ``special_flags``: (optional) Controls the blending mode for drawing colors.
+
+      **Notes**
+
```
Matiiss, line 171: Since you're indicating that what is below are notes, what about using bullet points instead of `Note:`? It would be sort of less repetitive and not as "noisy" IMO.

docs/reST/ref/surface.rst
```diff
@@ -137,20 +137,53 @@
...
-      :rtype: list or None
+              - ``blit_sequence``: The sequence of surfaces to blit.
+              - ``doreturn``: (optional) Modifies the return value of the method.
 
-      .. versionaddedold:: 1.9.4
+          The ``blit_sequence`` parameter is a sequence where each element represents a surface to
+          be drawn and its associated blit arguments. See the **Sequence Item Formats** section below
+          for the possible formats.
+
+          The ``doreturn`` parameter is a bool that controls the method's return value.
```
Matiiss, line 151: This is a bit repetitive of what you describe it does above in the parameter section, it's just longer, so it might as well be removed? Because it gets explained again in the Return section too.

Matiiss (commented):
Looks pretty cool.
One more thing that should be changed is on line 136 it should probably be `doreturn=True` in the signature instead of `1`, since, well, `True` is more Pythonic.

docs/reST/ref/surface.rst
Matiiss: Talking about `doreturn`

docs/reST/ref/surface.rst
itzpr3d4t0r: The thing is that imho doing so makes notes less readable and too compact:
image
image
So there's repetition but it's more readable.

docs/reST/ref/surface.rst
itzpr3d4t0r: I agree that there's repetition there but i don't think i should remove it, instead i'm trying to significantly shorten the return description and make the parameter description more dense.

docs/reST/ref/surface.rst
itzpr3d4t0r: I've readded the tag

docs/reST/ref/surface.rst
Matiiss: Would it be possible to add space between bullet points?

itzpr3d4t0r: Ok found a way to achieve that:
image
Already pushed the changes

Matiiss (approved):
LGTM

MyreMylar (approved):
LGTM 👍

MyreMylar (approved):
LGTM 👍 

Documentation seems clearer.

PR #2271, opened 2023-06-23
Fix SysFont None issue

Sjmarf:
In Pygame CE 2.3.0, a bug was introduced in which an error is raised by Pygame when initialising a `SysFont` object with `None` as shown below.

```py
font = pygame.font.SysFont(None, 40)
```

This PR fixes the problem, and adds a test for it. Apologies, the error was almost certainly introduced in my PR #2147. It might be worth pushing this fix out in a minor version.

### DIFF
--- a/src_py/sysfont.py
+++ b/src_py/sysfont.py
@@ -488,6 +488,8 @@ def SysFont(name, size, bold=False, italic=False, constructor=None):
                     f"found. {match_text}"
                     "Using the default font instead."
                 )
+    else:
+        fontname = None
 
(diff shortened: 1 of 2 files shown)


Starbuck5:
I'm shocked nobody found this in the dev release that was out for a week with it, and it's only now shown up 10 days after release.

Yeah maybe we'll do a quicker than usual 2.3.1 to get this fix out.

Starbuck5 (approved):

oddbookworm (approved):

PR #2272, opened 2023-06-24
Restore SDL 2.0.9 support

Starbuck5:
This is important for people using Debian Buster, which is on SDL 2.0.9 (https://packages.debian.org/buster/libsdl2-2.0-0)
-- AKA Kyuchumimo's raspberry pi.
You can see on piwheels (https://www.piwheels.org/project/pygame-ce/) we lost support for 2.0.9 when pygame-ce 2.2.0 was released (FRects)

The oldest supported Ubuntu is 20.04, Focal, which is on SDL 2.0.10 (https://packages.ubuntu.com/focal/libsdl2-dev)

Therefore we should probably drop our stated support for SDL 2.0.8.

SDL 2.10.0 was released July 23rd 2019
SDL 2.0.9 was released October 30th 2018
SDL 2.0.8 was released March 1st 2018

### DIFF
--- a/src_c/_sdl2/video.c
+++ b/src_c/_sdl2/video.c
@@ -772,6 +772,14 @@ static CYTHON_INLINE float __PYX_NAN() {
                 SDL_Color color;
                 _pgsdlFPoint tex_coord;
             } _pgsdlVertex;
+
+            int SDL_RenderGeometry(SDL_Renderer *renderer, SDL_Texture *texture,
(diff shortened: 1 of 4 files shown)


oddbookworm (approved):

MyreMylar (approved):
All seems reasonable to me from a code skim - I'm not going back to 2018-era installs of SDL to test it though. Since it's not going to break anything for 90% of our users, as all this code is gated - Approved!

PR #2274, opened 2023-06-27
Enabling users to set the soundfont on midi playback

AlexanderGroeger:
This PR adds a new `set_soundfont` function to the `pygame.mixer.music` module. This new feature allows game developers to leverage to pay a lower cost for storing music in their game by using smaller midi-formatted music tied with any soundfont they choose. 

**Technical Notice**
This PR relies on SDL_mixer's [Mix_SetSoundFonts](https://wiki.libsdl.org/SDL2_mixer/Mix_SetSoundFonts) function. At the initial commit, tests showed that setting the soundfont before loading and playing a MID file did not yield using the specified soundfont but instead fell back on the system default soundfont. This needs to be addressed before a merge will contribute anything.

UPDATE
A new `get_soundfont` function was added to `pygame.mixer.music` to verify whether SDL_mixer's internal soundfonts variable was correctly being updated. It turns out both functions work correctly, but there is still something in the SDL_mixer that is not immediately using the updated soundfont variable. 

UPDATE 2
It appears the feature works on Raspbian, but not Windows. This means the compilation of SDL_Mixer needs to be considered in the compilation of pygame-ce because it is not compiled to support soundfonts for all operation systems. At this time, the only known working platform is Raspbian 64-bit. 

UPDATE 3
`set_soundfont` now allows passing an empty string or no argument to clear the `soundfont_paths` variable in the SDL_Mixer. A stub has been added for both functions. I didn't see a MIDI used in the tests, so I wasn't sure if one existed, so both tests are left as todo tests. The docs now reflect this update too.

UPDATE 4
The todo tests have now been replaced with real tests that validate setting and getting soundfont paths will yield expected behavior.

UPDATE 5
More quick fixes. Some extra testing has been reported.
@oddbookworm and @rethanon have done some testing. On `main` (for both manjaro linux and mac), loading a midi yields a pygame error `Couldn't open timidity.cfg` which is the alternative to fluidsynth. Installing `timidity` with `yay` removes the error, but does not produce music. On this PR, (for both manjaro linux and mac), setting the soundfont before loading a midi works as intended. It seems Windows is still an issue.

UPDATE 6
Moved API to `pygame.mixer`, fixed some typos, and allowed `set_soundfont` to accept a `None` argument.


Starbuck5:
For the two failing runs you asked about clarification about:
- The code needs to be linted. Run `python setup.py format` after installing clang-format and black.
- Type stubs need to be added for this new API. See buildconfig/stubs/pygame/

src_c/music.c
```diff
@@ -280,6 +280,42 @@ music_get_pos(PyObject *self, PyObject *_null)
...
+    int paths_set;
+    const char *paths = PyUnicode_AsUTF8(arg);
+    if (PyErr_Occurred()) {
+        PyErr_Clear();
+        return RAISE(PyExc_TypeError, "set_soundfont expects 1 string argument");
+    }
+
+    MIXER_INIT_CHECK();
+
+    Py_BEGIN_ALLOW_THREADS;
```
Starbuck5, line 295: The BEGIN/END allow threads in these functions are unnecessary. They are very quick SDL functions based on the source, so there's no need to release the GIL for them.

src_c/music.c
```diff
@@ -280,6 +280,42 @@ music_get_pos(PyObject *self, PyObject *_null)
     return PyLong_FromUnsignedLongLong(ticks);
 }
 
+static PyObject *
+music_set_soundfont(PyObject *self, PyObject *arg)
```
Starbuck5, line 284: This function doesn't have the SDL_mixer functionality of passing NULL to un-set soundfont settings.

Additionally, would it be better to accept a list (sequence) of strings rather than a semicolon separated list? It would make the implementation more complex, but it seems more Python-normal.

Starbuck5: Implementation idea for joining strings into semicolon separated list.

Use PyUnicode_FromString to make a python string (";")
Use PyObject_CallMethod on the semicolon string to call string.join

Difficult part is error handling and making sure to avoid leaking objects by messing up reference counting.

docs/reST/ref/music.rst
```diff
@@ -219,6 +219,47 @@ MP3 in most cases.
...
+   This sets the soundfont file to be used in the playback of midi music.
+   The soundfont only affects the playback of ``MID``, ``MIDI``, and ``KAR`` file formats.
+   The ``path`` argument, a string (or multiple strings separated by a semi-colon),
+   must point to the soundfont file(s) to be searched for in order given if some
+   are missing.
+
+   Note that while SDL_Mixer stores the soundfont path, whether the soundfont is
+   used in the playback is a different story. SDL_Mixer needs to be compiled with 
+   soundfont support for a given platform. Soundfonts work on Raspbian 64-bit, but 
+   not Windows 11. Other operation systems have not been tested.
```
Starbuck5, line 236: *operating systems

Also Windows 11 is too specific, just say Windows. The SDL_mixer binary is the same for all Windows.

Starbuck5:
As well as the type stubs the CI complained about, this PR also needs an entry into the pygame-ce test suite. See the tests/ directory.

docs/reST/ref/music.rst
```diff
@@ -219,6 +219,49 @@ MP3 in most cases.
...
+
+   Note that while SDL_Mixer stores the soundfont path, whether the soundfont is
+   used in the playback is a different story. SDL_Mixer needs to be compiled with 
+   soundfont support for a given operation system. Soundfonts work on Raspbian 64-bit, 
+   but not Windows. Other operation systems have not been tested.
+
+   Function :func:`set_soundfont` calls underlying SDL_mixer function
+   ``Mix_SetSoundFonts``.
+
+   .. versionaddedold:: 2.3.1
```
Starbuck5, line 242: Versionaddedold is for pre-fork things. This should be versionadded.

test/mixer_music_test.py
```diff
@@ -417,6 +417,35 @@ def todo_test_set_pos(self):
 
         self.fail()
 
+    def todo_test_set_soundfont(self):
```
Starbuck5, line 420: These should be real tests rather than todo tests, since they're new.

For example, you could test that if you set soundfont and then get soundfont they're the same. (self.assertEqual the things)
You could could test that invalid arguments raise an error. (using self.assertRaises)
You could test that return types match expected.

docs/reST/ref/music.rst
Starbuck5: Still says "operation systems" several places, which is a typo.

test/mixer_music_test.py
Starbuck5: They also need to be renamed so they'll run, and because they're not "to do" if you've done them.

methods starting with test_ get run, so these won't run at all I believe.

test/mixer_music_test.py
```diff
@@ -417,6 +417,25 @@ def todo_test_set_pos(self):
...
+        self.assertEqual(pygame.mixer.music.get_soundfont(), None)
+
+        self.assertEqual(pygame.mixer.music.set_soundfont(""), None)
+        self.assertEqual(pygame.mixer.music.get_soundfont(), None)
+
+        self.assertEqual(pygame.mixer.music.set_soundfont("test1.sf2;test2.sf2"), None)
+        self.assertEqual(pygame.mixer.music.get_soundfont(), "test1.sf2;test2.sf2")
+
+        # There is no reasonable way to call this function so it raises pgExc_SDLError
+        # because this requires the memory is full
```
Starbuck5, line 433: What?

AlexanderGroeger: SDL_Mixer can only raise this error is there is no memory left to set the soundfont paths. See [Return Value](https://wiki.libsdl.org/SDL2_mixer/Mix_SetSoundFonts)

ankith26 (commented):
First of all, thanks for contributing to pygame-ce! :cake: 

Left a couple of typing related reviews. Also I'd like to discuss the API a bit more? Should we go with SDL's semicolon-separated style? I wouldn't mind making the API more pythonic and having the users deal with sequences of strings

buildconfig/stubs/pygame/mixer_music.pyi
```diff
@@ -15,6 +15,8 @@ def get_volume() -> float: ...
 def get_busy() -> bool: ...
 def set_pos(pos: float) -> None: ...
 def get_pos() -> int: ...
+def set_soundfont(paths: Optional[str] = "") -> None: ...
+def get_soundfont() -> str: ...
```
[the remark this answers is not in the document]
AlexanderGroeger: My pleasure. I like pygame-ce. I have been digging through source code to solve my Raspberry Pi 4 Optimizations discord help post, and while waiting for it to be resolved, I found something interesting to work on.

Thank you for correcting these stubs. I haven't done _that part_ of programming much.

In regards to the first inquiry, @Starbuck5 also brought this up consideration (on Discord I think?). My main reasons against passing a list of strings are two-fold: I wasn't comfortable with the programming of list arguments, and the API (intentionally) suggests there is only 1 soundfont, which ultimately only 1 will be used by SDL_Mixer because it uses the first valid soundfont file in the list of semi-colon-separated filepaths. I'm still unsure why this extra functionality would be helpful or ever used by developers, so I didn't want them to even bother thinking about it or being confused about how setting multiple soundfonts works internally, nevertheless, this fact is documented under `pygame.mixer.music.set_soundfont.`

test/mixer_music_test.py
Starbuck5: I read it as `There is no reasonable way to call this function, so it raises pgExc_SDLError`

You meant it as `There is no reasonable way to call this function that results in it raising pgExc_SDLError`

test/mixer_music_test.py
ankith26: This review thread isn't resolved yet, the functions still have the `todo_` prefix that need removing, as mentioned above. The tests themselves look good so far, maybe for good measure we can throw in some typeerror/valueerror validator subtests?

AlexanderGroeger: Okay. Forgot about this. I think commit [cc8058b](https://github.com/pygame-community/pygame-ce/pull/2274/commits/cc8058bd9719b3cfcc49e3386af2b22682f08d6a) addresses this.

docs/reST/ref/music.rst
Starbuck5: This is not resolved.

AlexanderGroeger:
> Question: why is this in mixer.music? Does this not apply to midi added as sounds (I'm on Windows so I can't test)? If it applies to all midi I'd think it should go in the mixer music.

I'm not sure what you're asking. It is in mixer.music, and you're suggesting it should go in mixer music....

Starbuck5:
Sorry I mistyped. I meant why does it not go in mixer.

AlexanderGroeger:
I hadn't considered this. I tested `mixer.Sound` with the soundfont selected and confirmed it works on Raspbian. I can move the new API over to the mixer instead of within its music submodule.

src_c/_pygame.h
```diff
@@ -281,6 +281,17 @@ supported Python version. #endif */
...
 
+/* Python macro for comparing to Py_None
+ * Py_IsNone is naturally supported by
+ * Python 3.10 or higher
+ * so this macro can be removed after the minimum
+ * supported
+ * Python version reaches 3.10
+ */
+#ifndef Py_IsNone
+#define Py_IsNone(x) x == Py_None
```
oddbookworm, line 292:
```suggestion
#define Py_IsNone(x) (x == Py_None)
```

src_c/mixer.c
```diff
@@ -1467,6 +1467,53 @@ mixer_find_channel(PyObject *self, PyObject *args, PyObject *kwargs)
     return pgChannel_New(chan);
 }
 
+static PyObject *
+mixer_set_soundfont(PyObject *self, PyObject *args)
+{
+    int paths_set;
+    PyObject *path = Py_None;
+    const char *string_path = "";
```
oddbookworm, line 1475:
```suggestion
    char *string_path = "";
```

src_c/mixer.c
```diff
@@ -1467,6 +1467,53 @@ mixer_find_channel(PyObject *self, PyObject *args, PyObject *kwargs)
...
+    if (paths_set == 0)
+        return RAISE(pgExc_SDLError, SDL_GetError());
+
+    Py_RETURN_NONE;
+}
+
+static PyObject *
+mixer_get_soundfont(PyObject *self, PyObject *_null)
+{
+    const char *paths;
```
oddbookworm, line 1505:
```suggestion
    char *paths;
```

src_c/music.c
```diff
@@ -280,6 +280,43 @@ music_get_pos(PyObject *self, PyObject *_null)
...
+    MIXER_INIT_CHECK();
+
+    paths = Mix_GetSoundFonts();
+
+    if (paths)
+        return PyUnicode_FromString(paths);
+
+    Py_RETURN_NONE;
+}
+
```
oddbookworm, line 319: If moving to mixer, please get rid of the music versions

src_c/mixer.c
```diff
@@ -1467,6 +1467,53 @@ mixer_find_channel(PyObject *self, PyObject *args, PyObject *kwargs)
...
+    if (paths_set == 0)
+        return RAISE(pgExc_SDLError, SDL_GetError());
+
+    Py_RETURN_NONE;
+}
+
+static PyObject *
+mixer_get_soundfont(PyObject *self, PyObject *_null)
+{
+    char *paths;
```
oddbookworm, line 1505: I guess you should put the `const` back...

src_c/doc/music_doc.h
```diff
@@ -13,6 +13,8 @@
 #define DOC_MIXER_MUSIC_GETBUSY "get_busy() -> bool\ncheck if the music stream is playing"
 #define DOC_MIXER_MUSIC_SETPOS "set_pos(pos) -> None\nset position to play from"
 #define DOC_MIXER_MUSIC_GETPOS "get_pos() -> time\nget the music play time"
+#define DOC_MIXER_MUSIC_SETSOUNDFONT "set_soundfont(path) -> None\nset the soundfont for playing midi music"
+#define DOC_MIXER_MUSIC_GETSOUNDFONT "get_soundfont() -> paths\nget the soundfont for playing midi music"
```
oddbookworm, line 17: This documentation should be removed because it's no longer in `mixer.music`. I think `python setup.py docs --fullgeneration` should handle that for you and you just need to commit the result

docs/reST/ref/mixer.rst
```diff
@@ -230,6 +230,49 @@ change the default buffer by calling :func:`pygame.mixer.pre_init` before
...
+   The soundfont only affects the playback of ``MID``, ``MIDI``, and ``KAR`` file formats.
+   The optional ``path`` argument, a string (or multiple strings separated by a semi-colon),
+   must point to the soundfont file(s) to be searched for in order given if some
+   are missing. If ``path`` is an empty string or not specified, any specified soundfont paths
+   will be cleared from SDL_Mixer.
+
+   Note that while SDL_Mixer stores the soundfont path, whether the soundfont is
+   used in the playback is a different story. Soundfonts work on Linux and Mac, 
+   but not Windows because the SDL_Mixer has not been compiled with soundfont 
+   support for a that operation system.
```
oddbookworm, line 248: grammar nitpick: "for a that operation system" -> "for that operating system"

docs/reST/ref/music.rst
```diff
@@ -219,6 +219,49 @@ MP3 in most cases.
...
+   Note this function only retrieves ``soundfont_paths`` that is stored in the SDL_Mixer 
+   which may not be the soundfont used in the playback (reasons in :func:`set_soundfont`). 
+
+   Function :func:`get_soundfont` calls underlying SDL_mixer function
+   ``Mix_GetSoundFonts``.
+
+   .. versionadded:: 2.3.1
+
+   .. ## pygame.mixer.music.get_soundfont ##
+
```
oddbookworm, line 264: docs in `mixer.music` are unneeded because it no longer exists there

src_c/mixer.c
```diff
@@ -1467,6 +1467,54 @@ mixer_find_channel(PyObject *self, PyObject *args, PyObject *kwargs)
...
 
+static PyObject *
+mixer_set_soundfont(PyObject *self, PyObject *args)
+{
+    int paths_set;
+    PyObject *path = Py_None;
+    const char *string_path = "";
+
+    if (!PyArg_ParseTuple(args, "|O", &path))
+        return NULL;
```
oddbookworm, line 1478: I understand that braces aren't needed here because it's a single line, but I still suggest adding them to be more explicit

src_c/mixer.c
```diff
@@ -1467,6 +1467,54 @@ mixer_find_channel(PyObject *self, PyObject *args, PyObject *kwargs)
...
+            PyErr_SetString(PyExc_TypeError,
+                            "Must pass string or None to set_soundfont");
+            return NULL;
+        }
+    }
+
+    if (strlen(string_path) == 0)
+        paths_set = Mix_SetSoundFonts(NULL);
+    else
+        paths_set = Mix_SetSoundFonts(string_path);
```
oddbookworm, line 1495: Same comment about braces

src_c/mixer.c
```diff
@@ -1467,6 +1467,54 @@ mixer_find_channel(PyObject *self, PyObject *args, PyObject *kwargs)
...
+
+    if (paths_set == 0)
+        return RAISE(pgExc_SDLError, SDL_GetError());
```
oddbookworm, line 1498: Same comment about braces

src_c/mixer.c
```diff
@@ -1467,6 +1467,54 @@ mixer_find_channel(PyObject *self, PyObject *args, PyObject *kwargs)
...
+mixer_get_soundfont(PyObject *self, PyObject *_null)
+{
+    const char *paths;
+
+    MIXER_INIT_CHECK();
+
+    paths = Mix_GetSoundFonts();
+
+    if (paths)
+        return PyUnicode_FromString(paths);
```
oddbookworm, line 1513: same comment about braces

test/mixer_test.py
```diff
@@ -116,6 +116,25 @@ def test_set_num_channels(self):
...
+        self.assertEqual(mixer.get_soundfont(), None)
+
+        self.assertEqual(mixer.set_soundfont(None), None)
+        self.assertEqual(mixer.get_soundfont(), None)
+
+        self.assertEqual(mixer.set_soundfont(""), None)
+        self.assertEqual(mixer.get_soundfont(), None)
+
+        self.assertEqual(mixer.set_soundfont("test1.sf2;test2.sf2"), None)
+        self.assertEqual(mixer.get_soundfont(), "test1.sf2;test2.sf2")
```
oddbookworm, line 133: Where you assert that `set_soundfont` returns `None` is a bit redundant. If it doesn't return `None`, then an error was raised and the test is failed anyway. I think you can just use the function

oddbookworm (changes requested):
Overall, looks good. I just have a few (mostly minor) nitpicks

src_c/_pygame.h
```diff
@@ -281,6 +281,17 @@ supported Python version. #endif */
...
+/* Python macro for comparing to Py_None
+ * Py_IsNone is naturally supported by
+ * Python 3.10 or higher
+ * so this macro can be removed after the minimum
+ * supported
+ * Python version reaches 3.10
+ */
+#ifndef Py_IsNone
+#define Py_IsNone(x) (x == Py_None)
+#endif
```
itzpr3d4t0r, line 293: Instead of having a separate macro that will later need to be removed, why couldn't you just put `path != Py_None` where you have the `!Py_IsNone(path)`?

oddbookworm: That was actually my suggestion. That macro is defined for python 3.10+ and I was going to do a followup PR using this macro and potentially some similar ones in many places throughout the codebase. Specifically, `Py_IsNone`, `Py_IsTrue`, and `Py_IsFalse` rather than the direct comparisons everywhere in the codebase. When 3.10 becomes our minimum supported version, it would be super simple just to remove 3 macros rather than do the PR then to update the code

oddbookworm: Maybe the true/false ones wouldn't be as useful because there aren't really many places we expect a boolean literal, but I know there are a large number of instances of `___ == Py_None` or `___ != Py_None` in the codebase that can be cleaned up this way

itzpr3d4t0r: Alright, was just curious, your explaination makes sense.

src_c/mixer.c
```diff
@@ -1467,6 +1467,60 @@ mixer_find_channel(PyObject *self, PyObject *args, PyObject *kwargs)
...
+    if (!Py_IsNone(path)) {
+        if (PyUnicode_Check(path)) {
+            string_path = PyUnicode_AsUTF8(path);
+        }
+        else {
+            PyErr_SetString(PyExc_TypeError,
+                            "Must pass string or None to set_soundfont");
+            return NULL;
+        }
+    }
```
Starbuck5, line 1492: Rewritten with less nesting:

```
if (PyUnicode_Check(path)) {
    string_path = PyUnicode_AsUTF8(path);
}
else if (!Py_IsNone(path) {
    PyErr_SetString
}
```

src_c/mixer.c
```diff
@@ -1902,6 +1956,10 @@ static PyMethodDef _mixer_methods[] = {
     {"get_busy", (PyCFunction)get_busy, METH_NOARGS, DOC_MIXER_GETBUSY},
     {"find_channel", (PyCFunction)mixer_find_channel,
      METH_VARARGS | METH_KEYWORDS, DOC_MIXER_FINDCHANNEL},
+    {"set_soundfont", (PyCFunction)mixer_set_soundfont, METH_VARARGS,
+     DOC_MIXER_SETSOUNDFONT},
```
Starbuck5, line 1960: Could be `METH_O` instead of varargs, since it's just one arg.

docs/reST/ref/mixer.rst
```diff
@@ -230,6 +230,49 @@ change the default buffer by calling :func:`pygame.mixer.pre_init` before
 
    .. ## pygame.mixer.find_channel ##
 
+.. function:: set_soundfont
```
Starbuck5, line 233: These docs look good. But I do think it seems heavy on "implementation detail" mentions about SDL_mixer.

src_c/mixer.c
AlexanderGroeger: This change fails the test `TypeError: pygame.mixer.set_soundfont() takes exactly one argument (0 given)`

src_c/mixer.c
AlexanderGroeger: [4d35782](https://github.com/pygame-community/pygame-ce/pull/2274/commits/4d3578287f8ae553421917b56278cef44df570dc)

docs/reST/ref/mixer.rst
AlexanderGroeger: [1560c3a](https://github.com/pygame-community/pygame-ce/pull/2274/commits/1560c3ac3c1007de226cc0d81a953cc1672fb261)

AlexanderGroeger:
One final note here. Perhaps we can learn something from this [issue](https://github.com/pygame/pygame/issues/2980) to get soundfonts working on windows.

src_c/mixer.c
Starbuck5: Ah, ok. Doesn't work because it's also supposed to take 0 args.

docs/reST/ref/mixer.rst
Starbuck5: That isn't really what I was concerned about.

Each function has an entire (short) paragraph saying exactly what function this refers to in SDL_mixer
Each function makes sure to tell the user the soundfont setting is stored in SDL_mixer

IMO this is unnecessary information.

AlexanderGroeger:
I tried including fluidsynth in the windows build config but didn't have the compiling experience or intuition to make the SDL_Mixer use soundfonts on windows, so I'd prefer to wrap up this PR and make an issue regarding this.

docs/reST/ref/mixer.rst
```diff
@@ -230,6 +230,43 @@ change the default buffer by calling :func:`pygame.mixer.pre_init` before
...
+
+   | :sl:`set the soundfont for playing midi music`
+   | :sg:`set_soundfont(path) -> None`
+
+   This sets the soundfont file to be used in the playback of midi music.
+   The soundfont only affects the playback of ``MID``, ``MIDI``, and ``KAR`` file formats.
+   The optional ``path`` argument, a string (or multiple strings separated by a semi-colon),
+   must point to the soundfont file(s) to be searched for in order given if some
+   are missing. If ``path`` is an empty string or not specified, any specified soundfont paths
```
ankith26, line 242:
```suggestion
   are missing. If ``path`` is an empty string or the default (``None``), any specified soundfont paths
```
minor update to the wording, basically to mention that ``None`` is allowed, and gives same behavior as the empty string

ankith26 (approved):

PS: to the person merging this, please squash and merge this PR

Starbuck5 (approved):
Nice work getting this up to be fully ready!

PR #2276, opened 2023-06-28
fixed pixelarray indexing to prevent segfault reported in #2275

### LINKED ISSUE #2275: Segfault when using PixelArray with 0 height/width surface.
To start off, I know that I shouldn't use Surfaces with a height or width of 0, but just hear me out.

Here is an MRE for the issue:
```py
import pygame

s = pygame.Surface((5, 0))
pixarr = pygame.PixelArray(s)
for r in pixarr: pass
```
It simply throws a Segmentation Fault on the final line.

I believe it should throw an error, rather than a Segfault (as Segfaults cannot be caught).

I am using `pygame-ce 2.3.0`, `SDL 2.26.4`, and `Python 3.11.1` on MacOs Ventura

oddbookworm:
fixes #2275


oddbookworm:
Further note:
It is also possible to segfault by trying to just index into that created pixelarray. This fixes both and the regression test has both cases

Starbuck5:
I think this may be more complicated.

If I create a PixelArray from a normal Surface and access the shape attribute, I get `(w, h)`. With the example code in the issue, I get `(w,)`. However, if I flip the dimension that it is zero dimensional in, everything works perfectly fine and it raises an IndexError as I would expect.

```py
>>> weird_surface = pygame.Surface((0,5))
>>> pixarray = pygame.PixelArray(weird_surface)
>>> pixarray.shape
(0, 5)
>>> pixarray[0]
IndexError: array index out of range
```

So the problem is that the code is using 0 in `shape[1]` as a special value in places to indicate the dimensionality of the data? And this just happens to be slotting into that? Maybe the special value should be replaced with `-1`, which would never come from a Surface's dimensions.

simplyrohan:
Another thing I found was that when the 0d value is in the Y value, then the shape of the pixel array does not include the Y value at all...
```py
import pygame
test_surf1 = pygame.Surface((0, 5))
test_surf2 = pygame.Surface((5,0))
pixarr1 = pygame.PixelArray(test_surf1)
pixarr2 = pygame.PixelArray(test_surf2)
print(pixarr1.shape, pixarr2.shape)
```
Outputs:
 `(0, 5) (5,)`

oddbookworm:
My most recent commit makes it impossible to create that pixelarray in the first place. I'll keep looking at how to solve the issue in a less strict way, but this should at least prevent any segfault conditions for now

ankith26 (approved):
thanks :+1:

PR #2277, opened 2023-06-29
Port Rect.scale_by(|_ip)

Starbuck5:
Credit goes to @dr0id for implementation and @jonotassia for keywords

This is mostly a straightforward code port, as is allowed under LGPL.

It did take me a while to get everything into the Rect/FRect system we've got, and to make sure the FRect tests demonstrated float-ness 😅

I'm not sure who wrote it, but this method has a huge test suite.


oddbookworm (approved):
LGTM!

ankith26 (commented):
Even though you have correctly credited original authors in the PR comment, it would be nice to add co-authors in the git commit itself

PS: FYI I did some digging, and the tests seem to be written by the 2 people you have already credited

Starbuck5:
I added the two original authors as co-authors on my commit.

docs/reST/ref/rect.rst
```diff
@@ -138,6 +138,34 @@
 
       .. ## Rect.inflate_ip ##
 
+   .. method:: scale_by
+
+      | :sl:`scale the rectangle by given a multiplier`
+      | :sg:`scale_by(scale_by) -> Rect`
+      | :sg:`scale_by(scalex, scaley) -> Rect`
```
ankith26, line 145: minor nit: I just noticed, this should be `x`/`y` not `scalex`/`scaley`

ankith26 (approved):
Approving this PR as it's just a port of an upstream one. The minor nitpick I mentioned can also be left for later

docs/reST/ref/rect.rst
Starbuck5: Good catch, updated.

PR #2278, opened 2023-06-29
Remove unused code from test utils

Starbuck5:


ankith26 (approved):
My (unverified) guess is that there is more unused code in testutils, but for now this PR LGTM, Thanks!

oddbookworm (approved):

PR #2284, opened 2023-07-01
Add pg_post_event and DictProxy C API to release GIL, fix timer issues and speedups

ankith26:
A port of my ol PR https://github.com/pygame/pygame/pull/3062 (I am skipping re-copying the details here, the link should have all the necessary info)

It's been a while but I just remembered I had this PR lying around for a while and it happens to fix a nasty bug that has been reported a couple of times over the years (I noticed upstream also seems to have patched the bug, but without any of my changes)

Anyways this PR does other improvements apart from the fix (hence the large diff). MyreMylar approved this PR in the past, and RN nothing much has changed since (though I took in nthykier's reviews while re-making this PR as I said earlier on)

Personally I would consider this PR semi-important for the next release

### DIFF
--- a/src_c/_pygame.h
+++ b/src_c/_pygame.h
@@ -46,7 +46,7 @@
  * state of number of events on queue and whether the owner of this struct
  * wants this dict freed. This DictProxy is only to be freed when there are no
  * more instances of this DictProxy on the event queue. Access to this is
- * safeguarded with a per-proxy spinlock, which is more optiminal than having
+ * safeguarded with a per-proxy spinlock, which is more optimal than having
  * to hold GIL in case of event timers */
 typedef struct _pgEventDictProxy {
     PyObject *dict;
--- a/src_c/_pygame.h
+++ b/src_c/_pygame.h
@@ -42,6 +42,19 @@
 
 #include <SDL.h>
 
+/* DictProxy is useful for event posting with an arbitrary dict. Maintains
+ * state of number of events on queue and whether the owner of this struct
+ * wants this dict freed. This DictProxy is only to be freed when there are no
+ * more instances of this DictProxy on the event queue. Access to this is
+ * safeguarded with a per-proxy spinlock, which is more optiminal than having
+ * to hold GIL in case of event timers */
+typedef struct _pgEventDictProxy {
+    PyObject *dict;
+    SDL_SpinLock lock;
+    int num_on_queue;
+    Uint8 do_free_at_end;
+} pgEventDictProxy;
+
 /* SDL 1.2 constants removed from SDL 2 */
 typedef enum {
     SDL_HWSURFACE = 0,
--- a/src_c/time.c
+++ b/src_c/time.c
@@ -28,43 +28,153 @@
 
 #define WORST_CLOCK_ACCURACY 12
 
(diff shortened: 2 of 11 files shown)


src_c/_pygame.h
```diff
@@ -42,6 +42,19 @@
 
 #include <SDL.h>
 
+/* DictProxy is useful for event posting with an arbitrary dict. Maintains
+ * state of number of events on queue and whether the owner of this struct
+ * wants this dict freed. This DictProxy is only to be freed when there are no
+ * more instances of this DictProxy on the event queue. Access to this is
+ * safeguarded with a per-proxy spinlock, which is more optiminal than having
```
Starbuck5, line 49: typo: optiminal -> optimal

Starbuck5:
Devlog of me wading through these changes:
I like the simplifications this seems to enable in the mixer and music code, it's a nice API to just say `pg_post_event`

Do you think this will have an impact that is possible to measure in user facing APIs? Since you're talking about this being an optimization. It seems like the main motivation is really to make things more convenient for the external callers. But the external callers could just call a function to get the GIL, doesn't seem so inconvenient.

I think I followed everything I've read so far, but I now need to go into the timer code, and that is a complex diff.

src_c/time.c
```diff
@@ -28,128 +28,223 @@
...
+    PG_TIMER_NO_ERROR,
+    PG_TIMER_SDL_ERROR,
+    PG_TIMER_MEMORY_ERROR,
+} pgSetTimerErr;
+
+/* This is a doubly linked list of event timers. Access to this linked list
+ * must be protected with an SDL mutex.
+ * An SDL mutex would be redundant if python GIL was unconditionally held in
+ * the C callback function. But acquiring GIL can be costlier than acquiring a
+ * dedictated mutex for event timers, because the GIL is usually held more
```
Starbuck5, line 42: typo: dedicated

src_c/time.c
```diff
@@ -300,50 +407,93 @@ time_set_timer(PyObject *self, PyObject *args, PyObject *kwargs)
...
+        case PG_TIMER_NO_ERROR:
+            Py_RETURN_NONE;
+        case PG_TIMER_SDL_ERROR:
+            return RAISE(pgExc_SDLError, SDL_GetError());
+        case PG_TIMER_MEMORY_ERROR:
+            return PyErr_NoMemory();
+        default:
+            return RAISE(
+                pgExc_SDLError,
+                "Unknown and unhandled internal error occured while handling "
```
Starbuck5, line 493: typo: occurred

ankith26:
While fixing the typos you pointed out, I just realised that there is a bit of an inconsistency with this PR. This PR makes it clear that events posted by any of the event posting functions, post a shallow-copy of the event (the dict is a common reference) *however* this PR also adds special-casing for the empty dict which would mean any event posted with an empty dict is actually true-copied. I'm inclined to fixing this up and removing the special-case, I'm also pretty sure it's not going to significantly affect performance. However before I do that I'd like opinions on this, as this change would have user-facing impacts.

ankith26:
While adding more event related tests, I also a small issue in pypy specific code, as it was a one-line fix I did it here.

PR #2287, opened 2023-07-05
Add window hiddenimport

Starbuck5:
In #2114, the Window class was moved to a hidden C module.

When distributing with PyInstaller, this C module is not picked up, so any program that uses Window fails at runtime only after building.

The solution to this is to add it as a hiddenimport to our pyinstaller hook.


yolowex:
This is great, but I can't test this on the os that encountered this issue, since it's not my pc and I can't have them build pygame from source.

Starbuck5:
Why can't you test this?

This is not OS specific. This is pygame-ce version specific. Make a pyinstaller build of a program that uses Window. With this patch it works, without it it fails at runtime.

ankith26 (approved):
Thanks, LGTM :tada:

yunline (approved):
LGTM 🎉

Issue #2286, opened 2023-07-04
Coordinate and _CanBeRect type aliases too restrictive


NALStudio:
TL;DR: The minimum requirement to be Coordinate-compliant is to conform to `collection.abc.Sequence` which is way too restrictive. This applies for _CanBeRect as well.

If I want to implement my own Vector2 implementation, due to the restrictions by the Coordinate type alias in _common.pyi, I have to at least support the functionality of `collections.abc.Sequence` for my linter to not complain which I find too restrictive as it requires functions like `index()` and `count()` to be implemented which make no sense in a Vector2 class; especially since the runtime requirements for the Coordinate type are much less strict than what the Coordinate type alias requires, as pygame _seems_ to run perfectly fine even when only `__len__` and `__getitem__` are implemented.
 
There are two ways I see which could solve this problem.

### Solution 1
Create a `typing.Protocol` for the Coordinate type alias which would require everything needed during runtime (Which presumably are `__len__` and `__getitem__`, although I'm not completely sure whether this is true). This would look something like this:
```python
class Coordinate(Protocol):
    def __getitem__(self, index: int) -> Union[int, float]: ...
    def __len__(self) -> int: ...
    # This is just an example as I'm not 100 % sure what
    # the Coordinate class' implementation requires
```

### Solution 2
Drop the requirement for `typing.Sequence` to a type which is less strict, like `typing.Collection` which would require much less to be implemented to be compliant. There is just one caveat, as `typing.Collection` doesn't require `__getitem__` to be implemented, it might necessitate other changes so that the Collection's values can be extracted from its iterator. I would see it looking something like this:
```python
Coordinate = Union[Tuple[float, float], Collection[float], Vector2]

# Or since Tuple[float, float] satisfies Collection[float]
Coordinate = Union[Collection[float], Vector2]
```

I also mentioned _CanBeRect in the title since I feel that its too restrictive as well, the solutions above can also be adapted to that.

I also thought about the possibility of me implementing this myself, but I decided against it since I'm not really familiar with the codebase of pygame and I'm a newbie contributor as I haven't contributed to open source before.

Starbuck5:
Thanks for opening an issue.

I'm not sure what the best way forward on this is, since I'm not one of the most typing-involved people. Maybe @ankith26 or @oddbookworm would have stronger opinions.

> If I want to implement my own Vector2 implementation

[...]

PR #2292, opened 2023-07-06
Introduce new custom 'Sequence' ABC, fix issues

ankith26:
PR to tackle #2286 and related issues that arise from the usage of `collections.abc.Sequence`


buildconfig/stubs/pygame/mask.pyi
```diff
@@ -41,9 +41,7 @@ class Mask:
         output: Optional[Mask] = None,
         offset: Coordinate = (0, 0),
     ) -> Mask: ...
-    def connected_component(
-        self, pos: Union[Sequence[int], Tuple[int, int]] = ...
-    ) -> Mask: ...
+    def connected_component(self, pos: Coordinate = ...) -> Mask: ...
```
Starbuck5, line 44: This hint is kinda weird, why is it defaulting to `...` ?

Starbuck5 (approved):

buildconfig/stubs/pygame/mask.pyi
ankith26: According to the docs,
> By default this method will return a [Mask](https://pyga.me/docs/ref/mask.html#pygame.mask.Mask) containing the largest connected component in the mask. Optionally, a bit coordinate can be specified and the connected component containing it will be returned. If the bit at the given location is not set, the returned [Mask](https://pyga.me/docs/ref/mask.html#pygame.mask.Mask) will be empty (no bits set).

So passing nothing is it's own behaviour, you cannot achieve the same behaviour by passing any type as the first argument. Hence the ellipsis (using any other type would mislead a user making them think passing that triggers no-argument behaviour)

ankith26:
Force pushed with only a merge-conflict fix

buildconfig/stubs/pygame/_common.pyi
```diff
@@ -1,40 +1,104 @@
...
-    Tuple[int, int, int, int],
-    Tuple[float, float, float, float],
-    Tuple[Coordinate, Coordinate],
-    Sequence[int],
-    Sequence[float],
-    Sequence[Coordinate],
-]
+ColorValue = Union[int, str, Sequence[int]]
+
+_CanBeRect = Sequence[Union[float, Coordinate]]
```
oddbookworm, line 40: `Sequence[Coordinate]`? I'm afraid I don't quite grasp how that fits the `_CanBeRect` type

ankith26: Because a 2-wide sequence of 2 coordinates is interpreted as `((x, y), (w, h))`

oddbookworm: 🤦 I'm dumb sometimes

oddbookworm (approved):
LGTM

PR #2294, opened 2023-07-07
Remove the old Window cython sources

yunline:


src_c/cython/pygame/_sdl2/video.pyx
```diff
@@ -144,380 +144,8 @@ def messagebox(title, message,
...
-        """
-        if SDL_SetWindowModalFor(self._win, parent._win):
-            raise error()
-
-    def __dealloc__(self):
-        if self._is_borrowed:
-            return
-        self.destroy()
+globals()["Window"] = Window
+globals()["_Window"] = Window
```
Starbuck5, line 148: Why? Window is already in the global namespace here, and _Window does not to be a preserved API.

Starbuck5 (approved):

buildconfig/stubs/pygame/_sdl2/window.pyi
```diff
@@ -1 +1,3 @@
-from pygame._sdl2.video import _Window as Window
+from pygame._sdl2 import video
+
+Window: type[video.Window]
```
ankith26, line 3: Hmm this is a bit of a weird pattern, I'd suggest something like (untested, but should work)

```suggestion
from pygame._sdl2.video import Window as Window
```
This actually has another advantage apart from being concise, the current code would error on python < 3.9

yunline: I thought this solution causes linter error, but it doesn't. (

ankith26 (approved):
I'm not really familiar with `Window` and `_sdl2` but approving this nonetheless, thanks!

PR #2297, opened 2023-07-08
Improved Rect and FRect typing to be more generic and accurate

oddbookworm:
dickerdackel (Discord) pointed out that rect positioning properties have lint problems with anything except a tuple, but the code accepts more. So, this PR is to fix that issue for both `Rect` and `FRect`. I also updated `size`

Edit: this has become more of an "all-around improvement of rect typestubs"


buildconfig/stubs/pygame/rect.pyi
```diff
@@ -33,18 +33,68 @@ class Rect(Collection[int]):
...
+    def topleft(self) -> Tuple[int, int]: ...
+    @topleft.setter
+    def topleft(self, value: Coordinate) -> None:
+        """The top left corner of the rect. Floats will be truncated to ints"""
+        ...
+    @property
+    def bottomleft(self) -> Tuple[int, int]: ...
+    @bottomleft.setter
+    def bottomleft (self, value: Coordinate) -> None:
+        """The bottom left corner of the rect. Floats will be truncated to ints"""
```
zoldalma999, line 46: We don't leave documentation in stub files

Also, minor nitpick, there is a space after the function name here. Same goes for the FRect version

oddbookworm: fixed

ankith26 (approved):

buildconfig/stubs/pygame/rect.pyi
```diff
@@ -33,18 +33,48 @@ class Rect(Collection[int]):
     left: int
     bottom: int
     right: int
```
ankith26, line 35: Actually I just remembered, these should technically be properties too. Even in plain ol' `Rect`, you are allowed to do something like `rect_obj.right = 8.6` and this doesn't error, but the argument is truncated/rounded

ankith26 (changes requested):

buildconfig/stubs/pygame/rect.pyi
oddbookworm: Pretty big commit incoming due to a conversation between us on Discord

buildconfig/stubs/pygame/rect.pyi
```diff
@@ -102,11 +170,11 @@ class Rect(Collection[int]):
     @overload
```
ankith26, line 170: A few lines above, `scale_by` function should return `Self` and not `Rect`

buildconfig/stubs/pygame/rect.pyi
```diff
@@ -188,202 +256,20 @@ class Rect(Collection[int]):
     # Also undocumented: the dict collision methods take a 'values' argument
     # that defaults to False. If it is False, the keys in rect_dict must be
     # Rect-like; otherwise, the values must be Rects.
-    @overload
     def collidedict(
-        self, rect_dict: Dict[RectValue, _V], values: bool = ...
+        self, rect_dict: Dict[RectTuple[_N], _V], values: bool = ...
```
ankith26, line 260: I don't get why this is being changed. I guess `RectValue` is correct here? I don't see the need for `RectTuple[_N]`, it's just a narrower-than-needed type

buildconfig/stubs/pygame/rect.pyi
```diff
@@ -188,202 +256,20 @@ class Rect(Collection[int]):
...
     ) -> Tuple[RectValue, _V]: ...
-    @overload
-    def collidedict(
-        self, rect_dict: Dict[_K, "Rect"], values: bool
-    ) -> Tuple[_K, "Rect"]: ...
```
ankith26, line 198: I don't get why this overload is being removed. According to the code, this is a valid signature

buildconfig/stubs/pygame/rect.pyi
ankith26: Yeah, but I'd still prefer it to be a `RectValue`. The user could still have a custom sequence-like object that is hashable (that isn't a tuple - weird usage, I'm not saying this would be common). As for non-hashable types that can pass under `RectValue`, I'm pretty sure mypy catches that error anyways, regardless of what the key is type'd as

oddbookworm: Discussed in Discord (if you can even call it an actual discussion):
I’m still not a fan of leaving this as possibly accepting a nonhashable type, but I agree with you that it can be more closely looked at sometime in the future.

buildconfig/stubs/pygame/rect.pyi
```diff
@@ -190,200 +258,28 @@ class Rect(Collection[int]):
...
     def collidedict(
-        self, rect_dict: Dict[RectValue, _V], values: bool = ...
+        self, rect_dict: Dict[RectValue, _V], values: bool = False
     ) -> Tuple[RectValue, _V]: ...
     @overload
     def collidedict(
-        self, rect_dict: Dict[_K, "Rect"], values: bool
-    ) -> Tuple[_K, "Rect"]: ...
+        self, rect_dict: Dict[_K, RectValue], values: bool = False
+    ) -> Tuple[_K, RectValue]: ...
```
ankith26, line 266: Not really your PRs issue, but I also noticed another fix you can do while you are at fixing this section. The return value of `collidedict` is allowed to be `None`, so wrap it in an `Optional`

buildconfig/stubs/pygame/rect.pyi
ankith26: This is not really your PRs fault, but if you want to fix it, there's a small issue in `scale_by` typing here. The `y` argument should be optional (so, something like `y: float = ...` should do)

buildconfig/stubs/pygame/rect.pyi
```diff
@@ -190,200 +258,28 @@ class Rect(Collection[int]):
...
     @overload
     def collidedict(
-        self, rect_dict: Dict[RectValue, _V], values: bool = ...
+        self, rect_dict: Dict[RectValue, _V], values: bool = False
     ) -> Tuple[RectValue, _V]: ...
     @overload
     def collidedict(
-        self, rect_dict: Dict[_K, "Rect"], values: bool
-    ) -> Tuple[_K, "Rect"]: ...
+        self, rect_dict: Dict[_K, RectValue], values: bool = False
```
ankith26, line 265: AFAIK there is no convenient way to properly convey "use this overload if this flag is falsey else use the next one" so next the best option we got RN is to actually leave this as `values: bool` as this somewhat conveys that the above overload is default, and not this.

oddbookworm: I think I have found a solution, but I'd welcome feedback. Adding a new commit with the solution included

buildconfig/stubs/pygame/rect.pyi
```diff
@@ -28,6 +30,8 @@ _K = TypeVar("_K")
 _V = TypeVar("_V")
 _T = TypeVar("_T")
 
+RectLike = TypeVar("RectLike", bound=RectValue, covariant=True)
```
ankith26, line 33: This is useful, I guess it would be useful in more places in this file, now that you've added this. One place I can think of is `Sequence[RectValue]` should actually be `Sequence[RectLike]`. While you are at it, maybe also change the naming a bit would help to not confuse `RectValue` and `RectLike`. Typically typevars are prefixed with an underscore, and another convention in the python ecosystem is to suffix covariant typevars with `_co`

oddbookworm: How's the new version?

buildconfig/stubs/pygame/rect.pyi
oddbookworm: I wasn't able to come up with a better solution for general "truthiness". And honestly, I don't see it being a huge problem, I can't imagine many scenarios where an explict bool wouldn't be the cleanest option anyway

buildconfig/stubs/pygame/rect.pyi
ankith26: LGTM

buildconfig/stubs/pygame/rect.pyi
ankith26: Fair enough

ankith26 (approved):
I like how this PR started with one area improvement and quickly spiralled into all round general improvements. 
Thanks a lot! :sunglasses: 

PS to the person merging this: Please squash and merge this PR

zoldalma999 (approved):
Looks good to me! Thanks!

PR #2303, opened 2023-07-11
Fix round rect with negative values

### LINKED ISSUE #2285: rounding corners with pg.draw.rect is broken with negative values
**Environment:**
(Those are probably irrelevant, but still)
- **Operating system**: MacOS
- **Python version**: 3.9.9
- **SDL version**: 2.26.4
- **PyGame version**: pygame-ce 2.3.0

**Current behavior:**
`pygame.draw.rect()` has a set of parameters, responsible for rounding the corners. There is a `border_radius` parameter, which is responsible for all 4 corners, and then there are 4 parameters, that are responsible for a single corner assigned to them. Default value of each parameter is `-1`, however, if every parameter, responsible for rounding the corners, is negative,  and one of them is positive, the shape becomes weird (as in the image)

Screen Shot 2023-07-03 at 22 15 47

This behaviour is NOT present, when all values are negative, no matter what they are, as long as they are negative.

**Expected behavior:**

no broken corners

**Steps to reproduce:**

1) type `pg.draw.rect(surface, color, rect, 0, -n, -n, +n, -n, -n)`
where `n` is any number
2) run the script and view the result

NOTE: only the "master" border_radius parameter sets the size of the drawn corners, each individual parameter responsible for only a single corner does not affect the outcome.

**Test code**

```
import pygame as pg
pg.init()
running = True
display = pg.display.set_mode((600, 600))


while running:
    for event in pg.event.get():
        if event.type == pg.QUIT:
            pg.quit()
            quit()

    display.fill("black")

    pg.draw.rect(display, (255, 255, 255), (100, 100, 50, 50), 0, -50, -10, 10, 10, -10)

    pg.display.update()

```

JoKing-1999:
Fix #2285.

During this I discovered drawing round rect and circle quadrant have another issue not related to this one, but that would require another PR.

### DIFF
--- a/src_c/draw.c
+++ b/src_c/draw.c
@@ -2478,10 +2478,14 @@ draw_round_rect(SDL_Surface *surf, int x1, int y1, int x2, int y2, int radius,
 {
     int pts[16], i;
     float q_top, q_left, q_bottom, q_right, f;
-    if (!top_left) top_left = radius;
-    if (!top_right) top_right = radius;
-    if (!bottom_left) bottom_left = radius;
-    if (!bottom_right) bottom_right = radius;
+    if (!top_left)
+        top_left = radius;
+    if (!top_right)
+        top_right = radius;
+    if (!bottom_left)
+        bottom_left = radius;
+    if (!bottom_right)
+        bottom_right = radius;
     if ((top_left + top_right) > (x2 - x1 + 1) ||
         (bottom_left + bottom_right) > (x2 - x1 + 1) ||
         (top_left + bottom_left) > (y2 - y1 + 1) ||


src_c/draw.c
```diff
@@ -972,9 +972,15 @@ rect(PyObject *self, PyObject *args, PyObject *kwargs)
             rect->y += rect->h;
             rect->h = -rect->h;
         }
+        // Ignore negative values
+        radius = MAX(radius, 0);
```
ankith26, line 976: Instead of simply ignoring negative values here, we could raise an error if this is negative instead

src_c/draw.c
```diff
@@ -972,9 +972,15 @@ rect(PyObject *self, PyObject *args, PyObject *kwargs)
...
+        top_left_radius = MAX(top_left_radius, 0);
+        top_right_radius = MAX(top_right_radius, 0);
+        bottom_left_radius = MAX(bottom_left_radius, 0);
+        bottom_right_radius = MAX(bottom_right_radius, 0);
 
         if (width > rect->w / 2 || width > rect->h / 2) {
-            width = MAX(rect->w / 2, rect->h / 2);
+            width = 0;
```
ankith26, line 983: I'm not sure how this change is helping fix the issue, it looks like it will change behaviour against what is documented

ankith26 (commented):
Hello, thanks for the PR.

As I'm not experienced in these parts of the docs, I'm stuggling to understand even the smallest diffs :sweat_smile: 

I left some questions as reviews

src_c/draw.c
```diff
@@ -2472,13 +2478,13 @@ draw_round_rect(SDL_Surface *surf, int x1, int y1, int x2, int y2, int radius,
...
+    if (!top_left)
         top_left = radius;
-    if (top_right < 0)
+    if (!top_right)
         top_right = radius;
-    if (bottom_left < 0)
+    if (!bottom_left)
         bottom_left = radius;
-    if (bottom_right < 0)
+    if (!bottom_right)
```
ankith26, line 2487: As I understand from the docs

if any of these values are 0, the corner should have no rounded, and if the value is negative (default) it should "inherit" radius. This fix seems to be breaking whatever is documented in my understand, correct me if I'm missing something

src_c/draw.c
MightyJosip: That could be another solution indeed

src_c/draw.c
MightyJosip: It doesn't do much. I do it just because of this line https://github.com/pygame-community/pygame-ce/blob/2e864cdd13d44b9318dcc7e9cbd1d078afc2173b/src_c/draw.c#L2497 Basically if width == 0, draw filled rect. With old width=MAX(...) you would still get filled rect, but it wouldn't go in https://github.com/pygame-community/pygame-ce/blob/2e864cdd13d44b9318dcc7e9cbd1d078afc2173b/src_c/draw.c#L2497 but rather in https://github.com/pygame-community/pygame-ce/blob/2e864cdd13d44b9318dcc7e9cbd1d078afc2173b/src_c/draw.c#L2525

src_c/draw.c
MightyJosip: Wait a second, you are probably right. Gonna check tomorrow.

src_c/draw.c
dr0id: silently clamping it to 0 vs raising an error

I'm not sure which is preferable when using the code. I tend to be against silent things. Make is explicit, not implicit.

src_c/draw.c
MyreMylar: He is right. This way a `border_top_right_radius` radius of 0 would inherit the value of `border_radius` whereas before it would set that corner to have no rounding. 

I could imagine this breaking usage if somebody used the popular GUI button pattern of three rounded corners (set via `border_radius` and one at right angles set with `border_top_right_radius`.

MyreMylar (commented):
Looks like this needs more work to finish off.

PR #2304, opened 2023-07-11
transform.solid_overlay

### LINKED ISSUE #1847: pygame.transform.solid_overlay()
The purpose of this function is to replace every single non transparent pixel with the inputted color.

dasd

ScriptLineStudios:
closes #1847

```python
import pygame

dis = pygame.display.set_mode((600, 600))

img = pygame.image.load("transparent_walt.png")
img = pygame.transform.solid_overlay(img, (0, 0, 0))

while True:
    dis.fill((100, 100, 100))

        if event.type == pygame.QUIT:
            raise SystemExit

    dis.blit(img, (30, 30))
    pygame.display.update()
```

produces:

Screenshot from 2023-07-11 14-01-26


oddbookworm:
Any idea what's going on here?
```python
import pygame

dis = pygame.display.set_mode((600, 600))

img = pygame.image.load("examples/data/alien1.png").convert_alpha()
img = pygame.transform.scale(img, (600, 600))

overlay_rect = pygame.Rect(200, 200, 100, 100)
subsurface = img.subsurface(overlay_rect)
subsurface = pygame.transform.solid_overlay(subsurface, (0, 0, 0))

while True:
    dis.fill((100, 100, 100))

        if event.type == pygame.QUIT:
            pygame.image.save(dis, "output.png")
            raise SystemExit

    dis.blit(img, (0, 0))
    pygame.draw.rect(dis, "red", overlay_rect, 2)
    pygame.display.update()
```
output

oddbookworm:
Also, when trying to use a string for the color, an error is raised:
```
pygame-ce 2.3.1.dev1 (SDL 2.26.4, Python 3.11.1)
  File "c:\Users\andre\Projects\pygame-ce\test.py", line 10, in <module>
    subsurface = pygame.transform.solid_overlay(subsurface, "black")
                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
TypeError: invalid search_color argument
```

src_c/doc/transform_doc.h
```diff
@@ -18,4 +18,5 @@
 #define DOC_TRANSFORM_AVERAGECOLOR "average_color(surface, rect=None, consider_alpha=False) -> Color\nfinds the average color of a surface"
 #define DOC_TRANSFORM_INVERT "invert(surface, dest_surface=None) -> Surface\ninverts the RGB elements of a surface"
 #define DOC_TRANSFORM_GRAYSCALE "grayscale(surface, dest_surface=None) -> Surface\ngrayscale a surface"
+#define DOC_TRANSFORM_SOLIDOVERLAY "grayscale(surface, color, dest_surface=None) -> Surface\ngrayscale a surface"
```
MyreMylar, line 21: This doc string is incorrect.

docs/reST/ref/transform.rst
```diff
@@ -336,6 +336,28 @@ Instead, always begin with the original image and scale to the desired size.)
...
+   depth as the source Surface.
+   
+   :param pygame.Surface surf: The target surface.
+
+   :param pygame.Color color: Color which all non transparent within the target surface must be set to.
+
+   :param dest_surface: Optional destination surface to which the changes will be applied.
+   :type dest_surf: pygame.Surface or None
+
+   .. versionadded:: 2.3.1
```
MyreMylar, line 357:
```suggestion
   .. versionadded:: 2.4.0
```

docs/reST/ref/transform.rst
```diff
@@ -336,6 +336,28 @@ Instead, always begin with the original image and scale to the desired size.)
 
    .. ## pygame.transform.grayscale ##
 
+.. function:: solid_overlay
+
+   | :sl:`grayscale a surface`
+   | :sg:`grayscale(surface, color, dest_surface=None) -> Surface`
```
MyreMylar, line 342: These slug lines are wrong.

src_c/transform.c
```diff
@@ -2118,6 +2118,99 @@ surf_grayscale(PyObject *self, PyObject *args, PyObject *kwargs)
...
+            SURF_GET_AT(pixel, src, x, y, (Uint8 *)src->pixels, src->format,
+                        pix);
+            SDL_GetRGBA(pixel, src->format, &r, &g, &b, &a);
+
+            if (a != 0) {
+                SURF_SET_AT(replace_pixel, newsurf, x, y,
+                            (Uint8 *)newsurf->pixels, newsurf->format, pix);
+            }
+        }
+    }
```
MyreMylar, line 2169: Question - would this be better if it set only the RGB values of a pixel and left the alpha alone so you could get feathered edges on your single colour cut outs?

I'm just trying to imagine the use cases for this.

MyreMylar: I notice the first post in the original issue had feathered edges in their example.

Temmie3754: I also think feathered edges are important for this feature, perhaps it can be added as an argument to the function.

src_c/transform.c
```diff
@@ -2118,6 +2118,99 @@ surf_grayscale(PyObject *self, PyObject *args, PyObject *kwargs)
...
+        return (SDL_Surface *)(RAISE(
+            PyExc_ValueError,
+            "Destination surface must be the same size as source surface."));
+    }
+
+    if (src->format->BytesPerPixel != newsurf->format->BytesPerPixel) {
+        return (SDL_Surface *)(RAISE(
+            PyExc_ValueError,
+            "Source and destination surfaces need the same format."));
+    }
```
itzpr3d4t0r, line 2146: Checking for `BytesPerPixel` isn't enough to make sure the surfaces have the same format. You should check that the Rmask, Gmask and Bmask are equal between the surfaces.

itzpr3d4t0r:
As per @ScriptLineStudios on Discord I'll take over this PR.

src_c/transform.c
itzpr3d4t0r: I'll add this. Still a bit unsure on how to call the kwarg between `maintain_alpha` or `feathered_edges`. I think maintain_alpha is more clear.

src_c/doc/transform_doc.h
itzpr3d4t0r: This should be fixed now.

itzpr3d4t0r:
Ltl graph to show improvement for 32bit surfaces:
image

itzpr3d4t0r:
One test on `Debian Multiarch / Debian (Bullseye - 11) [s390x]` is failing but it's unclear why and it's the only platform exhibiting this behaviour. Hopefully is just a blip but correct me if i'm wrong here.

src_c/transform.c
```diff
@@ -2211,6 +2211,141 @@ surf_grayscale(PyObject *self, PyObject *args, PyObject *kwargs)
...
+        Uint32 *dstp = (Uint32 *)newsurf->pixels;
+
+        const int src_skip = src->pitch - src->w * 4;
+        const int dst_skip = newsurf->pitch / 4 - newsurf->w;
+        int n, height = src->h;
+
+        while (height--) {
+            LOOP_UNROLLED4(
+                {
+                    if ((a = *srcp)) {
```
MyreMylar, line 2271: What is this line doing? The source pixel is being assigned to `a` then we continue only if the source pixel was not equal to 0?

MyreMylar:
> One test on `Debian Multiarch / Debian (Bullseye - 11) [s390x]` is failing but it's unclear why and it's the only platform exhibiting this behaviour. Hopefully is just a blip but correct me if i'm wrong here.

Likely to be an endianness issue - s390x is big endian. I would first change this failing test to an `assertEqual()` type so it will show the values that are different on the CI.

src_c/transform.c
itzpr3d4t0r: In the optimized version `srcp` is a `Uint8 *`, I then shift that pointer to point to the first pixel's alpha value and then icrement that pointer by 4 each time, that way we jump alpha by alpha and only check that instead of the whole pixel. The increment is happening anyway, the if only checks if the alpha != 0 then substitute the destination color with the desired color.

MyreMylar: ah, so srcp is really `srcp_alpha` at this point. I see perhaps we need a comment here as this is pretty dense code logic.

itzpr3d4t0r: I've added an explainer comment.

MyreMylar (approved):
OK, LGTM. 

All works locally for me, I like the `keep_alpha` flag and I can think of lots of uses for this in games. Also everything is passing CI now and the docs make sense.

I'm not sure there is much scope for SIMD in this transform either as it is a per pixel branch rather than any mathematical operation on every pixel. 

Happy to see this go in.

docs/reST/ref/transform.rst
```diff
@@ -341,14 +341,39 @@ Instead, always begin with the original image and scale to the desired size.)
...
+   | :sg:`solid_overlay(surface, color, dest_surface=None, keep_alpha=False) -> Surface`
+
+   Returns a new version of the original surface with all non transparent pixels set to the color provided.
+
+   An optional destination surface can be passed which is faster than creating a new
+   Surface. 
+   This destination surface must have the same dimensions (width, height) and
+   depth as the source Surface.
+
+   :param pygame.Surface surf: The target surface.
```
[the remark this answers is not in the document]
itzpr3d4t0r: Thanks for noticing!

damusss (commented):
I think the message I sent on discord got lost, so I'm typing it here instead.
Basically providing a string color such as `"green"` raises a `TypeError: invalid search_color argument`, which is not expected as the argument supports `ColorValue`, and the error is also referring to a non existent parameter `search_color`

On a side note, I'd like for this function to have some kind of alpha threshold parameter, but it's fine if it doesn't (the first time I tested my image had an alpha of 6 which I wasn't aware of, and I had to open an image editing program to bring it to 0 before trying again.)

damusss (approved):
the search_color in the error is still present but this is such a small nitpick that I'm approving anyways. I'm gonna let some SR or SC member merge it as this is new api. LGTM, and works locally.

buildconfig/stubs/pygame/transform.pyi
```diff
@@ -20,6 +20,7 @@ def rotate(surface: Surface, angle: float) -> Surface: ...
 def rotozoom(surface: Surface, angle: float, scale: float) -> Surface: ...
 def scale2x(surface: Surface, dest_surface: Optional[Surface] = None) -> Surface: ...
 def grayscale(surface: Surface, dest_surface: Optional[Surface] = None) -> Surface: ...
+def solid_overlay(surface: Surface, color: ColorValue, dest_surface: Optional[Surface] = None, keep_alpha: bool = False) -> Surface: ...
```
ankith26, line 23: This line is too long, maybe you could re-format this file with black/ruff?

src_c/transform.c
```diff
@@ -2215,6 +2215,176 @@ surf_grayscale(PyObject *self, PyObject *args, PyObject *kwargs)
...
+                    SURF_SET_AT(
+                        keep_alpha ? color_p | (a << dst_ashift) : color_p,
+                        newsurf, x, y, (Uint8 *)newsurf->pixels,
+                        newsurf->format, pix);
+                }
+            }
+        }
+    }
+
+    SDL_UnlockSurface(newsurf);
```
ankith26, line 2341: Why is it an unmatched lock here?

src_c/transform.c
```diff
@@ -2215,6 +2215,176 @@ surf_grayscale(PyObject *self, PyObject *args, PyObject *kwargs)
...
+        src->format->Amask != newsurf->format->Amask) {
+        return (SDL_Surface *)(RAISE(
+            PyExc_ValueError,
+            "Source and destination surfaces need the same format."));
+    }
+
+    Uint8 c_R, c_G, c_B, c_A;
+    Uint8 a;
+    SDL_GetRGBA(color, src->format, &c_R, &c_G, &c_B, &c_A);
+    Uint32 color_p = SDL_MapRGBA(newsurf->format, c_R, c_G, c_B, c_A);
```
ankith26, line 2253: why is this being done if you are checking that source and destination have the same format?

buildconfig/stubs/pygame/transform.pyi
itzpr3d4t0r: This should be fixed now thanks.

src_c/transform.c
itzpr3d4t0r: This is the original sin lol, always been there so it became invisible to my eyes, should be fixed now.

src_c/transform.c
itzpr3d4t0r: Right. thanks for noticing. Fixed ig.

damusss:
@itzpr3d4t0r after typing.py, ColorValue isn't available anymore. If you can you can update the branch and replace ColorValue with ColorLike from pygame.typing

itzpr3d4t0r:
> @itzpr3d4t0r after typing.py, ColorValue isn't available anymore. If you can you can update the branch and replace ColorValue with ColorLike from pygame.typing

I'm not sure i can do it right atm. I've always committed from github because of commit permissions being broken in my ide for some reason... so I'm afraid someone else has to do it.

PR #2307, opened 2023-07-12
Add `always_on_top` attribute for video.Window

yunline:
Split from #2070


src_c/window.c
```diff
@@ -308,6 +308,34 @@ window_get_borderless(pgWindowObject *self, void *v)
                            SDL_WINDOW_BORDERLESS);
 }
 
+#if SDL_VERSION_ATLEAST(2, 0, 16)
```
Starbuck5, line 311: The `#if` block can be moved within the scope of the function to avoid writing out the function name and arguments twice.

docs/reST/ref/sdl2_video.rst
```diff
@@ -146,6 +146,16 @@
       Gets or sets whether the window is borderless.
 
       .. note:: You can't change the border state of a fullscreen window.
+   
+   .. attribute:: always_on_top
```
Starbuck5, line 150: Should this get a `versionadded` notice?

Maybe not because it's an unstable module? I think lets do it just to be clear.

docs/reST/ref/sdl2_video.rst
```diff
@@ -146,6 +146,16 @@
...
+
+      | :sl:`Get or set whether the window is always on top`
+      | :sg:`always_on_top -> bool`
+
+      Get or set whether the window is always on top.
+
+      .. note:: Setting the always-on-top mode requires SDL2.0.16+.
```
Starbuck5, line 157: On main now `note`s are huge and important looking, and I don't think this is significant enough to make it look this important. (Since July 8th, looks like this branch was branched July 6th)

Also nitpick: `SDL2.0.16+` looks wrong to me. `SDL 2.0.16+` seems more conventional.

Starbuck5 (approved):
This works, seems straightforward enough.

Thanks!

ankith26 (approved):
LGTM, thanks! :cake:

src_c/window.c
```diff
@@ -308,6 +308,30 @@ window_get_borderless(pgWindowObject *self, void *v)
...
+    int enable = PyObject_IsTrue(arg);
+    if (enable == -1)
+        return -1;
+
+    SDL_SetWindowAlwaysOnTop(self->_win, enable);
+
+    return 0;
+#else
+    PyErr_SetString(pgExc_SDLError, "'always_on_top' requires SDL 2.0.16+");
```
Starbuck5, line 323: I was just thinking about this some more. We could have this ignore on SDL < 2.0.16, rather than error. Or raise a warning instead.

That way if someone gives their game to their friend on raspberry pi (which uses older SDL versions), it will still mostly work.

PR #2308, opened 2023-07-12
Port SDL_cpuinfo.h

yunline:
**Add functions:**
- `get_cpu_instruction_sets()`
- ~~`get_cpu_count()`~~
- ~~`get_cpu_cache_line_size()`~~
- `get_system_ram()`

**Test code:**
```py
import pygame

info = pygame.system.get_cpu_instruction_sets()

for name in info:
    line = "{:<12}{}".format(name, info[name])
    print(line)

print(f"System ram: {pygame.system.get_system_ram()} MiB")
```
```
pygame-ce 2.3.1.dev1 (SDL 2.26.4, Python 3.10.6)
RDTSC       True
ALTIVEC     False
MMX         True
SSE         True
SSE2        True
SSE3        True
SSE41       True
SSE42       True
AVX         True
AVX2        True
AVX512F     False
NEON        False
ARMSIMD     False
LSX         False
LASX        False
System ram: 15590 MiB
```


src_c/system.c
```diff
@@ -4,6 +4,80 @@
...
+    INSERT_INSTRUCTIONSET_INFO("LSX", SDL_HasLSX);
+    INSERT_INSTRUCTIONSET_INFO("LASX", SDL_HasLASX);
+#else
+    if (PyDict_SetItemString(instruction_sets, "LSX", Py_False))
+        goto error;
+    if (PyDict_SetItemString(instruction_sets, "LASX", Py_False))
+        goto error;
+#endif
+
+#undef INSERT_INSTRUCTIONSET_INF
```
Starbuck5, line 54: INF -> INFO

Starbuck5:
Why should we add these things?

`get_cpu_instruction_sets()` -> Doesn't seem to be in Python stdlib, which helps make a case to add it. But what's the use case? Should this tie in with pygame-ce instruction set support in other places somehow?
`get_cpu_count()` -> Duplicates functionality of `os.cpu_count()`
`get_cpu_cache_line_size()` -> Is that relevant for a Python program? That's something we don't even track in our own optimized C code.
`get_system_ram()` -> Doesn't seem to be in Python stdlib, which helps make a case to add it. But what's the use case? Maybe loading lower quality settings on boot if the amount of ram is too small?

src_c/system.c
yunline: Fixed

ankith26 (commented):
I don't think this info is super useful on the python end as others also pointed out, but I don't have any reasons against this PR either. It seems like a more useful usecase for `get_cpu_instruction_sets` is in issue reports, when we need user's runtime info of stuff that do depend on this. As for `get_system_ram`, it might have its niche too.

I'd also like if you added basic unit tests for these two functions

docs/reST/ref/system.rst
```diff
@@ -16,6 +16,51 @@ open just in case something obvious comes up.
...
+          'LASX': False
+     }
+
+   .. Note:: The values of ``LSX`` and ``LASX`` will be always False if
+      SDL version<2.24.0. The value of ``ARMSIMD`` will be always False
+      if SDL version<2.0.12.
+   
+   .. versionadded:: 2.3.1
+
+.. function:: get_system_ram
```
Starbuck5, line 55: `pygame.system.get_system_ram()` says "system" twice, which seems strange.

`get_total_ram()` ?

docs/reST/ref/system.rst
```diff
@@ -16,6 +16,51 @@ open just in case something obvious comes up.
...
+          'AVX2': True, 
+          'AVX512F': False,
+          'NEON': False, 
+          'ARMSIMD': False,
+          'LSX': False, 
+          'LASX': False
+     }
+
+   .. Note:: The values of ``LSX`` and ``LASX`` will be always False if
+      SDL version<2.24.0. The value of ``ARMSIMD`` will be always False
```
Starbuck5, line 50: nitpick: these two SDL things should have spaces in their inequalities.

SDL version < 2.24.0

docs/reST/ref/system.rst
```diff
@@ -16,6 +16,51 @@ open just in case something obvious comes up.
...
 .. versionadded:: 2.2.0
 
+.. function:: get_cpu_instruction_sets
+
+   | :sl:`get the information of CPU instruction sets`
+   | :sg:`get_cpu_instruction_sets() -> instruction_sets`
+
+   Returns a dict of the information of CPU instruction sets. The keys of
+   the dict are the names of instruction sets and the values determine 
+   whether the instruction set is available.
```
Starbuck5, line 26: You could add a note here about envisioned use as an information source about extension sets that pygame could be using.

Starbuck5 (approved):
I left some feedback on docs, and a suggestion to possibly rename one of the functions.

Feel free to resolve those as you feel fit.

Right now I'm comfortable enough with it to still approve.

ankith26 (approved):
LGTM, thanks :star2:

PR #2313, opened 2023-07-14
Fix FRect slice subscripting

ankith26:
Simple PR to fix a minor issue in `FRect` that I happened to notice.

Slicing an `FRect` should give floats in the list returned, not truncated integers. I'm assuming this was just a mistake, if it was a conscious decision feel free to correct me


Starbuck5 (approved):
Thanks!

That definitely is a bug.

oddbookworm (approved):
LGTM

PR #2171, opened 2023-05-18
Raise macOS deployment target to 10.11

Starbuck5:
For SDL 2.26.5 and above. https://github.com/libsdl-org/SDL/releases/tag/release-2.26.5

PR #2314, opened 2023-07-15
Move to SDL 2.26.5

Starbuck5:
This is mostly straightforward dep update.

The weird stuff is in the MacOS workflow, which is just a simplification and improvement of #2171 (getting our low mac target 10.11 for SDL 2.26.5)


Starbuck5:
Ugh, forgot dep updates need to be on a direct pygame-ce branch.

PR #2315, opened 2023-07-15
Replace Py_FileSystemDefaultEncoding

Starbuck5:
It's deprecated in 3.12. So deprecated they've just removed it from the docs without leaving a notice of deprecation there.

Anyways, the 3.12 whatsnew page links Py_FileSystemDefaultEncoding to PyConfig.filesystem_encoding (which I don't understand how could be used to get the encoding), but which also links to sys.getfilesystemencoding (python code)

So my approach was just to replace it with a C-API call to `sys.getfilesystemencoding`
Also had to relocate this SyntaxError setter to after the error flag would get reset by my new C API calls

I'm unsure if the function `pygame.encode_file_path` is even being used by anyone or whether it's useful, so maybe it should be deprecated and eventually removed. But the purpose of this PR is just to get wheels properly building in 3.12.

### DIFF
--- a/src_c/rwobject.c
+++ b/src_c/rwobject.c
@@ -243,8 +243,39 @@ pg_EncodeString(PyObject *obj, const char *encoding, const char *errors,
 static PyObject *
 pg_EncodeFilePath(PyObject *obj, PyObject *eclass)
 {
-    PyObject *result = pg_EncodeString(obj, Py_FileSystemDefaultEncoding,
-                                       UNICODE_DEF_FS_ERROR, eclass);
+    /* All of this code is a replacement for Py_FileSystemDefaultEncoding,
+     * which is deprecated in Python 3.12
+     *
+     * But I'm not sure of the use of this function, so maybe it should be
+     * deprecated. */
+
+    PyObject *sys_module = PyImport_ImportModule("sys");
+    if (sys_module == NULL) {
+        return NULL;
+    }
+    PyObject *system_encoding_obj =
+        PyObject_CallMethod(sys_module, "getfilesystemencoding", NULL);
+    if (system_encoding_obj == NULL) {
+        Py_DECREF(sys_module);
+        return NULL;
+    }
+    Py_DECREF(sys_module);
+    const char *encoding = PyUnicode_AsUTF8(system_encoding_obj);
+    if (encoding == NULL) {
+        Py_DECREF(system_encoding_obj);
+        return NULL;
+    }
+
+    /* End code replacement section */
+
+    if (obj == NULL) {
+        PyErr_SetString(PyExc_SyntaxError, "Forwarded exception");
+    }
+
+    PyObject *result =
+        pg_EncodeString(obj, encoding, UNICODE_DEF_FS_ERROR, eclass);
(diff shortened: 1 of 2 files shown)


oddbookworm (approved):
LGTM! Although, I do agree that deprecating `pygame.rwobject` should be considered. It looks like it's an artifact of python 2

Starbuck5:
The rwobject.c code is actually very important, every thing that is loaded in pygame (images, sounds, etc) all goes through this layer.

However, some of the python utilities it provides are of questionable use.

Starbuck5:
*Force pushed to rebase on the more recent CIBW version on main, so I can branch from this to 3.12 wheels generation.

src_c/rwobject.c
```diff
@@ -243,8 +243,39 @@ pg_EncodeString(PyObject *obj, const char *encoding, const char *errors,
...
+    if (encoding == NULL) {
+        Py_DECREF(system_encoding_obj);
+        return NULL;
+    }
+
+    /* End code replacement section */
+
+    if (obj == NULL) {
+        PyErr_SetString(PyExc_SyntaxError, "Forwarded exception");
+    }
```
ankith26, line 273: I don't get why this was moved here? I believe `pg_EncodeFilePath` can be called `obj == NULL` with more useful python errors already set.

Starbuck5: Because if I didn't move it there, the unit tests failed. Because the sys module I stuff I do messes with the error flags as a side effect.

I'm aware this changes the semantic of `pg_EncodeFilePath`, but it is only called in one place, the place that previously had this syntax error setter.

I was considering dropping this function out of the C-api to signify that it's semantics should not be considered important to the rest of pygame-ce, but I thought that would make the PR appear more complicated.

ankith26: The two functions were exposed here to just be used in our own unit tests. Hence the weird syntaxerror business, this is just there to check on the python side that the C level functions forward errors. But since you correctly pointed out that this `pg_EncodeFilePath` is not used anywhere else, it is true that changing any semantics isn't a big deal. 

However, IMHO A simpler solution is to just leave `pg_encode_file_path` as it is, but do a `if (!obj) return NULL;` at the top of this function.

ankith26 (approved):
Approving this PR, but left an alternate idea in a review thread. It ain't a big deal, so I'm leaving it up to you whether you want to do it before merging the PR

It might be worth documenting these functions as private to reduce any confusion among users

Starbuck5:
> It might be worth documenting these functions as private to reduce any confusion among users

What functions? The C api is not user facing anyway, and the Python function did not have any changed semantic.

ankith26:
I was referring to the python functions. Yes they don't have a change in semantics, but I don't think they are useful to end users anyways and that's why I want them marked internal/private in the docs. But that need not be an item for this PR to check

src_c/rwobject.c
Starbuck5: Hmm, that might work. I thought it wouldn't be that simple, which is why I did it the current way in the first place.

Right now I'm going to leave it how I currently have it.

PR #2325, opened 2023-07-17
3.12 wheels

Starbuck5:
Branched from #2315, which should be merged first. Now done ✅

### DIFF
--- a/.github/workflows/build-windows.yml
+++ b/.github/workflows/build-windows.yml
@@ -44,6 +44,12 @@ jobs:
       matrix:
         include:
           - {
+            name: "CPython 3.12 (64 bit)",
+            winarch: AMD64,
+            msvc-dev-arch: x86_amd64,
+            pyversions: "cp312-*"
+          }
+          - {
             name: "CPython 3.11 (64 bit)",
             winarch: AMD64,
             msvc-dev-arch: x86_amd64,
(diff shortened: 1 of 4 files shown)


.github/workflows/build-windows.yml
```diff
@@ -153,12 +167,14 @@ jobs:
...
         shell: cmd
+        # Use CIBW 2.13.1 on Python 3.7 because later versions don't support running on 3.7.
+        # They support generating 3.7 wheels, just not running on it.
         run: |
           set DISTUTILS_USE_SDK=1
           set MSSdk=1
           python -m pip install setuptools wheel requests numpy Sphinx
           python setup.py docs
-          python -m pip --disable-pip-version-check install cibuildwheel==2.12.0
+          python -m pip --disable-pip-version-check install cibuildwheel==${{ contains(matrix.pyversions, 37) && '2.13.1' || '2.14.1' }}
```
ankith26, line 177: This can be unconditionally 2.14.1, I think.
Yes, cibuildwheel 2.14.1 has dropped support for "running on python 3.7" but in my understanding it just means that they dropped cibuildwheel source support for 3.7. It must still support the 3.7 interpreter for making wheels and running tests on it.

Starbuck5: I don't think that's true. If cibuildwheel is running on python 3.7, it will no longer work. Look at the PR: https://github.com/pypa/cibuildwheel/pull/1526. Now the CIBW source code does stuff like import Literal from typing, which is new in 3.8.

ankith26: The `python` that invokes `cibuildwheel` in this case, is always going to be Python 3.9 regardless of the wheel being made

Starbuck5: Oh shoot you're totally right.

ankith26 (approved):
LGTM, thanks :tada:

novialriptide (approved):
LGTM

PR #2326, opened 2023-07-17
Corrected MacOS get_pref_path() documentation

IrvKalb:
The path generated by get_pref_path for MacOS should have included an extra folder using the "org" parameter.  (In the example, this has the value of "My Company").  I tested this call on my Mac, and the generated path includes the value of "org" as an additional folder.

### DIFF
--- a/docs/reST/ref/system.rst
+++ b/docs/reST/ref/system.rst
@@ -41,7 +41,7 @@ open just in case something obvious comes up.
         C:\\Users\\bob\\AppData\\Roaming\\My Company\\My Program Name\\
 
         On macOS, it would resemble
-        /Users/bob/Library/Application Support/My Program Name/
+        /Users/bob/Library/Application Support/My Company/My Program Name/
 
         And on Linux it would resemble
         /home/bob/.local/share/My Program Name/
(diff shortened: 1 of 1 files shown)


docs/reST/ref/system.rst
```diff
@@ -41,7 +41,7 @@ open just in case something obvious comes up.
         C:\\Users\\bob\\AppData\\Roaming\\My Company\\My Program Name\\
 
         On macOS, it would resemble
-        /Users/bob/Library/Application Support/My Program Name/
+        /Users/bob/Library/Application Support/My Company/My Program Name/
 
         And on Linux it would resemble
         /home/bob/.local/share/My Program Name/
```
ankith26, line 47:
```suggestion
        /home/bob/.local/share/My Company/My Program Name/
```

ankith26 (commented):
Hi, thanks for making the PR.

The docs here was originally copied from the [SDL docs](https://wiki.libsdl.org/SDL2/SDL_GetPrefPath) for the underlying function (which also has this issue). I tested this on linux, and even on linux the orgname is taken into account as expected.

docs/reST/ref/system.rst
IrvKalb: To be perfectly honest, I never use Linux, so I cannot comment on the accuracy of this additional change.  But it makes sense that this should work the same way on all supported operating systems.

Starbuck5: @IrvKalb I went ahead and committed Ankith's suggestion to this branch.

Starbuck5 (approved):
Thanks!

zoldalma999 (approved):
Thanks!

PR #2328, opened 2023-07-19
Continue moving away from format masks again

Starbuck5:
Since SDL3 is all about pixelformat instead.

Continuing from #2123


ankith26 (approved):
LGTM, I also like how these changes help in making the code simpler. Thanks! :balloon:

zoldalma999 (approved):
Looks good to me, thanks!

PR #2330, opened 2023-07-20
Made `Surface.set_at()` fastcall

itzpr3d4t0r:
This PR aims to get a quick and easy perf improvement with fastcall as well as removing a redundant check.

Results:
TLDR, about 34-40% improvement on my machine.
```
OLD
mean: 0.226506186666514
mean: 0.8316974133334346
mean: 3.5231991799999682
NEW
mean: 0.16023285333330933
mean: 0.6178482066666523
mean: 2.6118607199999437
```
Code Used:
```Py
from statistics import mean
from timeit import repeat

import pygame

pygame.init()

screen = pygame.display.set_mode((10, 10))

c = (255, 23, 1)
G = globals()
reps = 15
positions = [
    (0, 0), (1, 0), (0, 1), (1, 1),
    (0, 0), (1, 0), (0, 1), (1, 1),
    (0, 0), (1, 0), (0, 1), (1, 1),
    (0, 0), (1, 0), (0, 1), (1, 1),
]

color = pygame.Color(255, 23, 1)


def test(stmt: str):
    data = repeat(stmt, globals=G, repeat=reps)
    print(f"mean: {mean(data)}")


test("screen.set_at((0, 0), color)")
test_str = """
screen.set_at((0, 0), color)
screen.set_at((1, 0), color)
screen.set_at((0, 1), color)
screen.set_at((1, 1), color)
"""
test(test_str)
test_str = """
for pos in positions:
    screen.set_at(pos, color)
"""
test(test_str)

pygame.quit()
```


src_c/surface.c
```diff
@@ -843,12 +844,19 @@ surf_set_at(PyObject *self, PyObject *args)
...
+    if (nargs != 2) {
+        return PyErr_Format(PyExc_TypeError,
+                            "set_at takes exactly 2 arguments (%zd given)",
+                            nargs);
+    }
 
-    format = surf->format;
+    if (!pg_TwoIntsFromObj(args[0], &x, &y)) {
+        return RAISE(PyExc_TypeError, "invalid position argument");
+    }
```
ankith26, line 855: This change actually leads to a minor change in semantics. Previously the code would only take a sequence of ints. Now it can also take a sequence of floats. This isn't a big deal (most of pygame code actually does things this way), but with this change you also need to update typestubs for this function

ankith26 (commented):
`set_at` is typically called in tight loops, so I think this optimisation is pretty good.

The CI is failing on a warning, which I'm sure you figured by now, just needs an explicit cast.

src_c/surface.c
itzpr3d4t0r: Thanks, didn't think of that, resolved it (at least i think).

ankith26 (approved):
LGTM now, thanks! :+1:

Starbuck5 (dismissed):
This conflicts with Scriptline's PR #2111.

Lets try to get that in first, then you can rebase this against that?

Edit: resolved, earlier PR merged.

zoldalma999:
I'll convert it draft for now, feel free to open it back when the mentioned pr is merged

src_c/surface.c
```diff
@@ -871,11 +873,6 @@ surf_set_at(PyObject *self, PyObject *args)
         if (PyErr_Occurred() && (Sint32)color == -1)
             return RAISE(PyExc_TypeError, "invalid color argument");
     }
-    else if (PyLong_Check(rgba_obj)) {
-        color = (Uint32)PyLong_AsUnsignedLong(rgba_obj);
-        if (PyErr_Occurred() && (Sint32)color == -1)
-            return RAISE(PyExc_TypeError, "invalid color argument");
-    }
```
Starbuck5, line 878: Looking through the Git history, I can see the previous check used to be a PyInt_Check from python2, so maybe we should look into using the intended PyLong_AsUnsignedLong. Could the other way cause issues with high numbers?

src_c/surface.c
```diff
@@ -834,29 +834,31 @@ surf_get_at(PyObject *self, PyObject *position)
...
+        return PyErr_Format(PyExc_TypeError,
+                            "set_at takes exactly 2 arguments (%zd given)",
+                            nargs);
+    }
 
-    if (!pg_TwoIntsFromObj(position, &x, &y)) {
-        return RAISE(PyExc_TypeError,
-                     "position must be a sequence of two numbers");
+    if (!pg_TwoIntsFromObj(args[0], &x, &y)) {
+        return RAISE(PyExc_TypeError, "invalid position argument");
```
ankith26, line 856: Minor nitpick: I prefer the older error message here.

src_c/surface.c
ankith26: I am in process of doing another PR to address this, and many other places with a similar issue.

src_c/surface.c
```diff
@@ -867,11 +870,6 @@ surf_set_at(PyObject *self, PyObject *args)
     }
 
     if (PyLong_Check(rgba_obj)) {
-        color = (Uint32)PyLong_AsLong(rgba_obj);
-        if (PyErr_Occurred() && (Sint32)color == -1)
-            return RAISE(PyExc_TypeError, "invalid color argument");
-    }
-    else if (PyLong_Check(rgba_obj)) {
```
ankith26, line 874: I'd like to keep this change out of this PR, I have a fix for this in a separate PR.

If this is removed, this PR becomes fairly easy to merge, and we can get this in by the dev release IMO

ankith26:
To the person merging this: squash and merge this PR so that it doesn't conflict with my other PR. It would probably add me as a co-author of this PR, which you can remove if you want

zoldalma999 (approved):
Looks good to me, thanks! Nice speed ups!

PR #2335, opened 2023-07-21
Drop pypy3.7, add wheels for pypy3.10

Starbuck5:


ankith26 (commented):
I think we can drop PyPy 3.8 too (based on the download stats we discussed on discord, we don't really have users on it)

Also, the current distribution on mac is really suboptimal. Since PyPy builds are much slower in comparison to CPython builds, builds with PyPy have to be really minimal (it is fine to bundle 3-4 Cpython builds in one job, but don't bundle 2 PyPy builds in one)

Starbuck5:
> Also, the current distribution on mac is really suboptimal. Since PyPy builds are much slower in comparison to CPython builds, builds with PyPy have to be really minimal (it is fine to bundle 3-4 Cpython builds in one job, but don't bundle 2 PyPy builds in one)

Ok, I'll rearrange it. Maybe back closer to how it was before.
I was trying to balance by number of jobs in general, which is why I split the arm64 build.

Starbuck5:
I did the one thing you said not to, I put two pypy builds in the same run.

But I think it works out okay leveling everything off.

ankith26 (approved):
LGTM, thanks :partying_face:

.github/workflows/build-macos.yml
```diff
@@ -85,35 +85,35 @@ jobs:
         # of these builds take roughly the same time
         include:
           - { 
-            name: "x86_64 (CPython 3.10 and above)",
+            name: "x86_64 (CPython 3.9 - 3.12)",
             macarch: x86_64,
             # pattern matches any 2 digit number
```
Matiiss, line 90: This comment doesn't seem to be accurate anymore

Starbuck5: Fixed.

Matiiss (approved):
LGTM

dr0id (commented):
I'm not expert but why not add just such a snippet for each version added or remove one snippet instead of to have to adapt that many lines every time?

Starbuck5:
> I'm not expert but why not add just such a snippet for each version added or remove one snippet instead of to have to adapt that many lines every time?

@dr0id Are you referring the macOS changes?

It would've been a simpler diff but I also needed to rebalance the runs so all take similar times. There are 5 runs because that's the maximum concurrent we can do with Github Actions.

PR #2337, opened 2023-07-23
Draw dotted line

oddbookworm:
TODO:
[ ] write tests

### DIFF
--- a/src_c/draw.c
+++ b/src_c/draw.c
@@ -238,12 +238,12 @@ line(PyObject *self, PyObject *arg, PyObject *kwargs)
         return pgRect_New4(startx, starty, 0, 0);
 }
 
-/* Draws a dashed line 
+/* Draws a dashed line
  *
  * returns a Rect bounding the affected area
  *
  */
-static PyObject*
+static PyObject *
 dashed_line(PyObject *self, PyObject *args, PyObject *kwargs)
 {
     pgSurfaceObject *surfobj;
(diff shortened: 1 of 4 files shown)


src_c/draw.c
```diff
@@ -238,6 +238,107 @@ line(PyObject *self, PyObject *arg, PyObject *kwargs)
...
+        }
+
+        currentx = startx + i * xstep;
+        currenty = starty + i * ystep;
+        stopx = currentx + xstep;
+        stopy = currenty + ystep;
+
+        draw_line_width(surf, color, (int)currentx, (int)currenty, (int)stopx,
+                        (int)stopy, width, drawn_area);
+    }
```
MyreMylar, line 324: I think the dashed line might feel a bit better if the drawing code handled the remainder here. that way dashes would be the fixed length specified in the parameter and the last dash/space would just be incomplete. Right now line with a long dash length can vary quite a lot in length in practice - both of these lines have the same dash length for example:

image

image


But I think it would be better if it did something like this:

image

As the line gets longer.

MyreMylar (commented):
This is progressing nicely. Still needs tests as the To do list indicates.

I've also noticed that the segments in a line can be drawn with different top/bottom angles to one another. Like so:

image

Which is not a particularly terrible problem.

I guess you could extend it to take a list of lengths which would make up a repeating pattern of segment lengths (counting the spaces as segments). See:

https://developer.mozilla.org/en-US/docs/Web/API/CanvasRenderingContext2D/setLineDash

for inspiration.

src_c/draw.c
bilhox: Hello,

This feature is really interesting and helpful, good job oddbook !
Following what MyreMylar, I want to suggest one more thing, it could be a good idea to define an offset from where the first visible segment should be drawn from ``begin_pos``.

oddbookworm:
Will revive this soon, I promise

src_c/draw.c
oddbookworm: both comments implemented

oddbookworm:
> I've also noticed that the segments in a line can be drawn with different top/bottom angles to one another. Like so:

I think this is a problem with `draw_line_width` in `draw.c`. Not sure I can fix it without messing with that function, and I'm not sure that's the best idea

src_c/draw.c
```diff
@@ -223,6 +223,184 @@ line(PyObject *self, PyObject *arg, PyObject *kwargs)
...
+        return RAISE(PyExc_ValueError, "delay argument cannot be negative");
+    }
+
+    const int dx = end_x - start_x;
+    const int dy = end_y - start_y;
+
+    const double total_length = sqrt(dx * dx + dy * dy);
+    if (total_length == 0) {
+        return pgRect_New4(start_x, start_y, 0, 0);
+    }
```
damusss, line 309: About my latest comment, to add the second approach of the first problem, this snipped seems to do the trick:
```c
if (delay_gap > total_length) {
    delay_gap = delay_gap % (int)total_length;
}
```

PR #2339, opened 2023-07-24
aliased pygame.cursors as pygame.cursor

Python-ZZY:
aliased pygame.cursors as pygame.cursor

### DIFF
--- a/docs/reST/ref/midi.rst
+++ b/docs/reST/ref/midi.rst
@@ -208,7 +208,7 @@ midi events to pygame events.
    .. method:: pitch_bend
 
       | :sl:`modify the pitch of a channel.`
-      | :sg:`set_instrument(value=0, channel=0) -> None`
+      | :sg:`pitch_bend(value=0, channel=0) -> None`
 
       Adjust the pitch of a channel. The value is a signed integer
       from -8192 to +8191. For example, 0 means "no change", +4096 is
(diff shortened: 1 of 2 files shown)


oddbookworm (approved):
It's a super small change, but LGTM! It's always nice to see new contributors getting involved 🥳

Starbuck5:
I see that this PR was discussed on discord first.

If we want to move/duplicate the entire naming of a module, there needs to be a plan. Deprecations? Rationale? Documentation?

The rationale on discord was that other modules are not plural. I appreciate the drive for consistency, but I personally don't think this change is worth it.

Python-ZZY:
Fixed a clerical error in the documentation.
It seems inappropriate to send it here.  I can't seem to recreate a new pull request, so I just write it here (probably because I didn't create a new branch).

MyreMylar:
> Fixed 2 clerical errors in the documentation. It seems inappropriate to send it here. I can't seem to recreate a new pull request, so I just write it here (probably because I didn't create a new branch).

Yes, this is because you didn't create a new branch. How about we close this PR as it has got a bit unfocused and you can try making a branch and re-opening a PR for the midi documentation error?

I agree with Starbuck that it seems like a bit of a large effort to consider fully renaming the cursors module (including deprecations, examples etc), the examples submodule is also plural. We certainly would not jump straight to the renaming part. I'm also not sure I'd want singular and plural versions of all the modules in the autocomplete namespace as it would likely just be confusing. I think we can just live with this small inconsistency.

You are of course free to import pygame submodules as whatever name you please!

PR #2341, opened 2023-07-24
Reset window data on borrowed windows on dealloc

### LINKED ISSUE #1889: Seg fault when you call sdl2.Window.from_display_module after initializing a Color object
**Operating system**: Linux Mint 20.3, haven't tested on others

**Current behavior:**

I discovered a bizarre segfault that occurs when you initialize a `pygame.Color` and call `_sdl2-Window.from_display_module` in a very specific way. Rearranging the code even slightly seems to fix it (e.g. moving the `Color(...)` call to its own line), but with this configuration I get a Segmentation Fault every time:

```
import pygame
import pygame._sdl2 as sdl2


def get_window(var):
    sdl2.Window.from_display_module()  # without this line it doesn't crash


if __name__ == "__main__":
    pygame.init()

    fill_color = (92, 64, 92)
    pygame.display.set_mode((256, 128))

    # if this Color is constructed (basically) anywhere but right here, it doesn't crash
    get_window(pygame.Color(fill_color))

    clock = pygame.time.Clock()

    while True:
        for e in pygame.event.get():  # <-- segfault usually happens here, on first iteration
            if e.type == pygame.QUIT:  # <-- sometimes happens here as well
                raise SystemExit()

        pygame.display.get_surface().fill(fill_color)

        pygame.display.flip()
        clock.tick(60)
```

Output:
```
/usr/bin/python3 /home/david/Coding/python/pygame-utils/scaled_test.py
pygame-ce 2.1.3 (SDL 2.0.22, Python 3.8.10)
Fatal Python error: (pygame parachute) Segmentation Fault

Current thread 0x00007fbd0fb47740 (most recent call first):
  File "/home/david/Coding/python/pygame-utils/scaled_test.py", line 22 in <module>

Process finished with exit code 134 (interrupted by signal 6: SIGABRT)
```

zoldalma999:
Fix #1889

If the window is borrowed, don't point to an invalid location after deallocating, but reset it to NULL.

Also changed the `from_display_module` function to return an existing window if there is already one created and added a proper error string if `from_display_module` is called before `display.set_mode`, instead of an empty one (or whatever unrelated error `SDL_GetError` would give).

Test script (slightly modified from the one Ghast provided):
```py
import pygame
import pygame._window

pygame.init()

pygame.display.set_mode((256, 128))

x = [
    pygame._window.Window.from_display_module(),
    pygame._window.Window.from_display_module(),
]
print(x[0] is x[1])
del x

clock = pygame.time.Clock()

while True:
    for e in pygame.event.get():
        if hasattr(e, "window"):
            print(e)
        if e.type == pygame.QUIT:
            raise SystemExit()

    pygame.display.get_surface().fill("white")
    pygame.display.flip()
    clock.tick(60)

```

### DIFF
--- a/src_c/window.c
+++ b/src_c/window.c
@@ -476,8 +476,13 @@ mouse_set_relative_mode(pgWindowObject *self, PyObject *arg, void *v)
 static void
 window_dealloc(pgWindowObject *self, PyObject *_null)
 {
-    if (!self->_is_borrowed && self->_win) {
-        SDL_DestroyWindow(self->_win);
+    if (self->_win) {
+        if (!self->_is_borrowed) {
+            SDL_DestroyWindow(self->_win);
+        }
+        else if (SDL_GetWindowData(self->_win, "pg_window") != NULL) {
+            SDL_SetWindowData(self->_win, "pg_window", NULL);
+        }
     }
     Py_TYPE(self)->tp_free(self);
 }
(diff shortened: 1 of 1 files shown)


yunline (approved):
LGTM

ankith26 (approved):

PR #1981, opened 2023-03-04
Fix pygame.draw.arc

MightyJosip:
This should fix pygame.draw.arc. Old PR (https://github.com/pygame/pygame/pull/3068). It is possible this doesn't work right now, there were some other changes in the meantime, and I just speedrunned through the old PR, I will fix it in case it needs repair. Test code https://github.com/pygame/pygame/issues/3020#issuecomment-1064505060

PR #2344, opened 2023-07-25
Fix pygame.draw.arc

Cyuria:
Fix a rendering bug with pygame.draw.arc

This bug already has a pull request open (#1981), however there has been no activity on that PR for four months (except for a comment asking about its revival a few days ago).

I used the following sources for testing code: https://github.com/pygame/pygame/issues/3020#issuecomment-1064505060 as well as https://github.com/pygame-community/pygame-ce/pull/1981#pullrequestreview-1336063364

I intend to make a few more performance based commits, as there are a number of different ways I can think of to significantly improve performance.
I figured I should open the PR now as I have a working solution and I figured it would be good to do so at this point.

EDIT:

I have finished making performance improvements and am now just waiting for a review before we can merge this PR. I'd be happy to implement any other changes anyone can think of


Cyuria:
difference:
old:
pygame old arc
pygame_arc_test2_old
new:
pygame new arc 2
pygame_arc_test2_new

Cyuria:
I noticed quite quickly that it fails a few tests, which in hindsight I probably should have run before submitting the PR, so I'll do that before anything else

Cyuria:
new tests:
pygame_arc_test2_new2
pygame new arc 3

src_c/draw.c
```diff
@@ -48,8 +48,13 @@ static void
 draw_aaline(SDL_Surface *surf, Uint32 color, float startx, float starty,
             float endx, float endy, int blend, int *drawn_area);
 static void
-draw_arc(SDL_Surface *surf, int x, int y, int radius1, int radius2,
-         double angle_start, double angle_stop, Uint32 color, int *drawn_area);
+old_draw_arc(SDL_Surface *surf, int x, int y, int radius1, int radius2,
+             double angle_start, double angle_stop, Uint32 color,
+             int *drawn_area);
```
oddbookworm, line 53: Why keep this around?

src_c/draw.c
```diff
@@ -1676,50 +1677,59 @@ draw_line(SDL_Surface *surf, int x1, int y1, int x2, int y2, Uint32 color,
...
-         double angle_start, double angle_stop, Uint32 color, int *drawn_area)
+draw_arc(SDL_Surface *surf, int x_center, int y_center, int radius1,
+         int radius2, int width, double angle_start, double angle_stop,
+         Uint32 color, int *drawn_area)
 {
-    double aStep;  // Angle Step (rad)
-    double a;      // Current Angle (rad)
-    int x_last, x_next, y_last, y_next;
+    // ensure stop angle is greater than start angle, this allows for proper
+    // inverse fill handling
```
oddbookworm, line 1685: If you're going to require this, then you should enforce it, but it seems like it's probably easy enough to tweak in code if needed

Cyuria:
Could I get someone else's opinion on this?

The reason the tests are failing is because of some differences between the new arc algorithm and the old one when the difference between the start and stop angles is greater than 2 Pi radians, or 360 degrees.

At this point I will probably add a clause to my algorithm to satisfy the tests.

However, I think my algorithm's way of handling the edge case could be preferable if documented. Here's a comparison:
Old:
oldpygamearc
New:
pygamearc
The code I used to test this:
```py
import math
import time

import pygame

pygame.init()
font = pygame.font.Font(size=32)

def drawarc(screen):
    t = time.time() % (4 * math.pi)
    screen.fill((0, 0, 0))
    pygame.draw.arc(screen, pygame.Color('white'), (50, 50, 300, 300), 0, t, 50)
    screen.blit(font.render(str(round(t / math.pi, 1)) + ' Pi radians', True, pygame.Color('white')), (0, 0))
    pygame.display.flip()

def runanim():
    screen = pygame.display.set_mode((400, 400))
    while True:
        event = pygame.event.poll()
        drawarc(screen)
        if event.type == pygame.NOEVENT:
            continue
        if event.type == pygame.QUIT:
            break
            print(time.time() % (4 * math.pi))
            continue
            print(event.pos)
            continue


if __name__ == "__main__":
    runanim()
```

oddbookworm (commented):
Haven't dug too deep, but here are a couple comments

oddbookworm:
> The reason the tests are failing is because of some differences between the new arc algorithm and the old one when the difference between the start and stop angles is greater than 2 Pi radians, or 360 degrees.
>
>At this point I will probably add a clause to my algorithm to satisfy the tests.

Generally, we try to avoid changes that might break someone's project. I also would like feedback from other reviewers about which behavior should be the "correct" behavior. IMO if there is a difference of more than 2pi radians, then a full circle should be drawn because it has important information in some cases. I could potentially be persuaded otherwise though

oddbookworm:
In particular, @MightyJosip is our resident draw expert iirc (maybe @zoldalma999 too?). But one of our steering committee should probably chime in about the behavior diff

src_c/draw.c
Cyuria: Thanks for mentioning that, I had a copy of the old algorithm laying around just in case I broke something while I was writing the new one, but it looks like I didn't clean everything up. The comment is similar, I had some extra code which I got rid of but didn't clean up the comment

Cyuria:
I've been working a bit more and I believe I have run into a place where the old algorithm (as well as one of the tests) are either inconsistent with the docs or just broken, if I fix the start angle at -17 radians and the stop angle at -5.5, (the only currently failing test), the following happens:
pygame arc oldnew compare
Unless I am making a mistake, the documentation says otherwise:
https://github.com/pygame-community/pygame-ce/blob/6ebeb191f5af2927a73a74b2c3388d2c4f4225ab/docs/reST/ref/draw.rst?plain=1#L270-L278

Cyuria:
wait nvm I'm an idiot

Cyuria:
seeing as all tests passed, I started working a bit on performance, using the following benchmark script:

```py
import math
import time

import pygame

def drawarc(surface, t):
    pygame.draw.arc(surface, pygame.Color('white'), (0, 0, 1000, 1000), 0, t, 300)

def runbench():
    screen = pygame.display.set_mode((50, 50))
    surface = pygame.Surface((1000, 1000))
    starttime = time.time_ns()

    for c in range(100):
        for i in range(200):
            drawarc(surface, i * math.pi / 100)

        # give the user some kind of feedback
        screen.fill((256 * c // 100,)*3)
        pygame.display.flip()

    endtime = time.time_ns()

    print(f"starttime:  {starttime / 1000000000}s")
    print(f"endtime:    {endtime / 1000000000}s")
    print(f"difference: {(endtime - starttime) / 1000000000}s")

if __name__ == "__main__":
    runbench()

```

The following are my results for the entire test

| algorithm | time |
|-----------|------|
|    old    |  47s |
|  new (1)  |  69s |
|  new (2)  |  54s |

(1) the first full implementation which passed all tests (commit a9522a69a1f18636d700a047f78e78356277915a)
(2) with bounds checking before running the algorithm (commit a31b616ff9e94c5b88192dd43f09c692d0535553)

It is also possible to visualize the bounds checking by adding the following code to the draw_arc function anywhere after the minima and maxima are calculated:
```c
    draw_line(surf, minx + x_center, miny + y_center,
              minx + x_center, maxy + y_center, color, drawn_area);
    draw_line(surf, minx + x_center, maxy + y_center,
              maxx + x_center, maxy + y_center, color, drawn_area);
    draw_line(surf, maxx + x_center, maxy + y_center,
              maxx + x_center, miny + y_center, color, drawn_area);
    draw_line(surf, maxx + x_center, miny + y_center,
              minx + x_center, miny + y_center, color, drawn_area);
```

doing so gives the following result on my machine with the animated test program (albeit with red for the bounding box lines)

pygamearc_bounding

There are still two more performance improvements I can think of which I will try to implement soon.
- reducing the amount of per pixel checking by using alternate methods to calculate the easier parts of the arc
- directly setting the pixels instead of updating drawn area and using the bounding box given values to calculate the drawn area

Cyuria:
The latest performance enhancements are documented in the two above commits. Here is the updated benchmark table:

| algorithm | time |
|-----------|------|
|    old    |  47s |
|  new (1)  |  69s |
|  new (2)  |  54s |
|  new (3)  |  48s |
|  new (4)  |  35s |

(1) the first full implementation which passed all tests (commit https://github.com/pygame-community/pygame-ce/commit/a9522a69a1f18636d700a047f78e78356277915a)
(2) with bounds checking before running the algorithm (commit https://github.com/pygame-community/pygame-ce/commit/a31b616ff9e94c5b88192dd43f09c692d0535553)

(3) Use the calculated bounding box to obtain the drawn_area instead of calculating it for each drawn pixel with `set_and_check_rect` (commit 2e3585308911a5c92ac2d89a95c690214f3a9e10) -- NOTE this commit fails the draw arc bounding box test due to an out by one error on some arcs, will be fixed in the next commit
(4) Bypass `set_at` to avoid checking surface clip rect and check it when calculating the bounding box instead (commit bef7f717c44391b37bff85533125a8d2234e5fe1) -- NOTE this commit fails the draw arc clip rect box due to an out by one error, will be fixed in the next commit

Starbuck5:
So this takes the old algorithm from mostly unusable (because of the holes) to twice as fast? Amazing!

Cyuria, are you on the pygame community discord?

Cyuria:
> Cyuria, are you on the pygame community discord?

I am now :)

Also there are a few weird interactions with the test edge cases which I'm still trying to fix, currently zero width rects are doing weird stuff

oddbookworm:
> I am now :)

You should check out the contributing channel under the "development" category :)

Cyuria:
I fixed the test cases, here's the updated performance table:

| algorithm | time |
| --------- | ---- |
|    old    |  47s |
|  new (1)  |  69s |
|  new (2)  |  54s |
|  new (3)  |  48s |
|  new (4)  |  35s |
|  new (5)  |  36s |

(5) up to the latest commit with test cases (commit f77a8912e5954c45094bc5ca3efef2400be4db06)

The timings can vary pretty badly, I ran some of the tests a few times and got up to a few seconds difference for each run, so don't necessarily trust the extra +1s (although there definitely should be slightly lower performance because I'm running quite a few extra checks, especially on arcs which are not fully inside the clipping rect of the drawn surface)

Cyuria:
I think I'm done with this PR unless someone has anything else to say, I still need a review.

The performance table for each commit is as follows:

| algorithm | time |
|-----------|------|
|    old    |  47s |
|  new (1)  |  69s |
|  new (2)  |  54s |
|  new (3)  |  48s |
|  new (5)  |  36s |
|  new (6)  |  30s |

(1) the first full implementation which passed all tests (commit https://github.com/pygame-community/pygame-ce/commit/a9522a69a1f18636d700a047f78e78356277915a)

(2) with bounds checking before running the algorithm (commit https://github.com/pygame-community/pygame-ce/commit/a31b616ff9e94c5b88192dd43f09c692d0535553)

(3) Use the calculated bounding box to obtain the drawn_area instead of calculating it for each drawn pixel with set_and_check_rect (commit https://github.com/pygame-community/pygame-ce/commit/2e3585308911a5c92ac2d89a95c690214f3a9e10) -- NOTE fails some tests, was fixed in the commit after

(5) Bypass set_at to avoid checking surface clip rect and check it when calculating the bounding box instead (commit https://github.com/pygame-community/pygame-ce/commit/bef7f717c44391b37bff85533125a8d2234e5fe1) -- Fixes issues with previous commit

(6) commit a7b3cd3b82e5193d447cd46e7760cc2952c2e05e, redo algorithm somewhat to use faster circle line drawing techniques

itzpr3d4t0r:
I suppose there should be some tests to check the behaviour stays fixed in the future, or do we already have those elsewhere?

Cyuria:
> I suppose there should be some tests to check the behaviour stays fixed in the future, or do we already have those elsewhere?

If the behaviour ever somehow gets reverted to the "loading screen" behaviour, a couple of the tests will fail because they assume that arcs of length 5.5 and 10 iirc are full circles. The relevant sections of the code are also commented, saying which section of the docs they relate to.

I don't think it's a problem but if you want I can add some tests.

src_c/draw.c
```diff
@@ -1675,51 +1672,325 @@ draw_line(SDL_Surface *surf, int x1, int y1, int x2, int y2, Uint32 color,
...
+    const int y_start_inner = (int)(y_start * inner_radius2 + 0.5);
+    const int x_stop_inner = (int)(x_stop * inner_radius1 + 0.5);
+    const int y_stop_inner = (int)(y_stop * inner_radius2 + 0.5);
+    const int x_start_outer = (int)(x_start * radius1 + 0.5);
+    const int y_start_outer = (int)(y_start * radius2 + 0.5);
+    const int x_stop_outer = (int)(x_stop * radius1 + 0.5);
+    const int y_stop_outer = (int)(y_stop * radius2 + 0.5);
+
+    // calculate maximums, accounting for each quadrant
+    // We can't just fidn the maximum and minimum points because the arc may
```
MyreMylar, line 1755:
```suggestion
    // We can't just find the maximum and minimum points because the arc may
```

src_c/draw.c
```diff
@@ -1675,51 +1672,325 @@ draw_line(SDL_Surface *surf, int x1, int y1, int x2, int y2, Uint32 color,
...
+    // span multiple quadrants, resulting in a maxima at the edge of the circle
+    // also account for the surfaces clip rect. This allows us to bypass the
```
MyreMylar, line 1757:
```suggestion
    // also account for the surface's clip rect. This allows us to bypass the
```

src_c/draw.c
```diff
@@ -1675,51 +1672,325 @@ draw_line(SDL_Surface *surf, int x1, int y1, int x2, int y2, Uint32 color,
...
-            aStep = 1.0;
-        }
-        else {
-            aStep = asin(2.0 / radius1);
+    // Return whether the angle of the point is within the accepted range
+    return x * x_middle + y * y_middle >= min_dotproduct * sqrt(x * x + y * y);
+}
+
+static inline void
+unsafe_set_at(SDL_Surface *surf, int x, int y, Uint32 color)
```
MyreMylar, line 1697: Probably should add a comment here indicating why this is an 'unsafe' version of `set_at()`. It seems to me like you have removed the surface clip boundary detection because your arc algorithm already dealt with surface clipping at a higher level, so I would say that and that the function is otherwise identical to `set_at()`.

MyreMylar (approved):
Overall this looks good to me. It passes the unit tests locally, seems to have fixed the original issue, runs faster and passes all the test cases I could come up with fiddling around with the function.

I added some minor comment & spelling things you could fix but nothing that should stop this from being merged. Whoever does merge this - don't forget to also close #1981 that this supercedes.

src_c/draw.c
Cyuria: thanks for picking up on those. Fixed in the latest commit

dr0id (approved):
LGTM, all tests pass as they are! Thanks for the PR.

src_c/draw.c
```diff
@@ -1675,51 +1672,329 @@ draw_line(SDL_Surface *surf, int x1, int y1, int x2, int y2, Uint32 color,
...
+
+            // Check if the point is within the arc for each quadrant
+            if (pos_y & pos_x & (x_dot + y_dot >= cmp))
+                unsafe_set_at(surf, px_offset, py_offset, color);
+            if (pos_y & neg_x & (-x_dot + y_dot >= cmp))
+                unsafe_set_at(surf, nx_offset, py_offset, color);
+            if (neg_y & pos_x & (x_dot - y_dot >= cmp))
+                unsafe_set_at(surf, px_offset, ny_offset, color);
+            if (neg_y & neg_x & (-x_dot - y_dot >= cmp))
+                unsafe_set_at(surf, nx_offset, ny_offset, color);
```
Starbuck5, line 1990: Shouldn't this be using logical ands instead of bitwise ands?

I see it would be the same because it's results of comparisons, but logical and seems to better represent the operation.

dr0id: I agree, these condition are all about logical choices and not about single bits. So logical ands would express the intention better.

Cyuria: thanks for finding that, I don't know what I was doing when I was writing that but it was probably 3am. It should be fixed now

Starbuck5 (approved):
I checked a bunch of before/after behavior on possible edge cases locally, and this got a positive note from JoKing (our draw expert) on discord, so let's go ahead and merge it.

Starbuck5:
I did a squash and merge since there were so many commits on the branch, to keep `main` a bit more tidy.

Great work on this @Cyuria, this was an impressive PR!

Issue #2273, opened 2023-06-25
Documentation: Add 'color name/str' to list of allowed color values for pygame.draw.* and pygame.Surface.fill()


DickerDackel:
**Environment:**

- **PyGame version** (e.g. 2.0.0.dev10, 1.9.6): 2.3.1

**Documentation:**

I just learned, that the (all?) draw functions now allow a color name.  Docs still only say, Color, int and tuple are allowed, but not str.  Same for `pygame.surface.Surface`

**Expected behavior:**

str should be in the list of acceptable color arguments.

Starbuck5:
PR is welcome :)

You can see how color.rst references color_list to link to the list of named colors.

rethanon:
I can tackle this - the issue says it's affecting version 2.3.1, is that a typo? We don't' have 2.3.1 yet do we?

DickerDackel:
Sorry, yes, 2.3.1 is a typo:

```
In [1]: import pygame
pygame-ce 2.3.0 (SDL 2.26.4, Python 3.11.4)

In [2]:
```

PR #2346, opened 2023-07-25
Add Named Colors to docs for pygame.draw and pygame.Surface 

rethanon:
I've updated the docs to reflect the changes suggested in #2273 which asked for the pygame.draw and pygame.Surface pages to more clearly articulate that strings for named colors can be used.

PLEASE NOTE - I'm not experienced with color palettes but I updated `set_palette` and `set_palette_at` in line with the rest as I could see in the code that they should also accept a ColorValue. Happy to amend as necessary if I've misunderstood.


docs/reST/ref/surface.rst
```diff
@@ -461,10 +462,10 @@
       | :sl:`set the color value for a single pixel`
       | :sg:`set_at((x, y), Color) -> None`
 
-      Set the ``RGBA`` or mapped integer color value for a single pixel. If the
-      Surface does not have per pixel alphas, the alpha value is ignored.
-      Setting pixels outside the Surface area or outside the Surface clipping
-      will have no effect.
+      Set the color of a single pixel at the speficified co-ordinates to be an ``RGB``,
```
Starbuck5, line 465: `speficified` typo

Starbuck5: Also co-ordinates should be coordinates

docs/reST/ref/draw.rst
```diff
@@ -469,7 +470,7 @@ object around the draw calls (see :func:`pygame.Surface.lock` and
       along the y-axis instead of the x-axis (using ``y`` instead of ``x``,
       top instead of left and bottom instead of right).
 
-   .. note::
+str()   .. note::
```
Starbuck5, line 473: ?

docs/reST/ref/surface.rst
rethanon: I fixed these typo's (although co-ordinates is a perfectly valid spelling here in Europe!)

docs/reST/ref/draw.rst
rethanon: No idea how this got in there, I wasn't making any changes to that part ! Sorry, fixed now.

docs/reST/ref/draw.rst
```diff
@@ -24,6 +24,7 @@ All the drawing functions accept a color argument that can be one of the
 following formats:
 
    - a :mod:`pygame.Color` object
+   - a string (for :doc:`color_list`)
```
ankith26, line 27: names from the color_list aren't the only kind of colors allowed, stuff like `#RRGGBB[AA]` are also allowed

docs/reST/ref/surface.rst
```diff
@@ -529,11 +530,14 @@
...
       | :sl:`set the color palette for an 8-bit Surface`
-      | :sg:`set_palette([RGB, RGB, RGB, ...]) -> None`
+      | :sg:`set_palette([color, color, color, ...]) -> None`
 
       Set the full palette for an 8-bit Surface. This will replace the colors in
       the existing palette. A partial palette can be passed and only the first
       colors in the original palette will be changed.
+      
+      The color can be an ``RGB`` sequence, an ``RGBA`` sequence, a string
+      (for :doc:`color_list`), or a mapped color index.
```
ankith26, line 540: As of now, this is incorrect. `set_palette` (and even `set_palette_at`) only handled `RGB[A]` sequences (and not strings and ints). I have a fix coming up for this, and for now you could keep that out of this PR

docs/reST/ref/draw.rst
rethanon: The types of color object accepted are all detailed in the pygame.Color page, but being able to use a string for named colors is kind of hidden away in there amongst all the long paragraphs of text. The idea here was just to add a line to show anyone reading the docs that a string representing a named color could be used, as it's a very handy and commonly used feature. I assume we wouldn't need to list all color object types here (hex values, html values, etc.)?

If you'd rather I remove it, or add all the other color object types I can do that?

rethanon: Line 26 is a link to the pygame.Color page, just for clarity.

ankith26: Hmmm, because I was also working on something related, I actually have updated docs for this section in a WIP branch. There I went with
```
- if an `int` value is passed, it must be mapped to the pixel format of the surface
  (see :func:`pygame.Surface.map_rgb` and :func:`pygame.Surface.unmap_rgb`)
- a :mod:`pygame.Color` object, or a type (apart from `int`) that can be passed to 
  its constructor
```
feel free to improvise or extend as per your judgement, but this mentions the 2 main points I want to convey here in this part of the docs

rethanon: OK thanks, perhaps I don't need to update this part in this PR given it already points to the pygame.Color page? The main issue (#2273) I'm trying to cover in this PR is the feedback that it's not clear to users that strings for named colors are available to use, even though there are links to the pygame.Color page that shows all the available color types. I've updated the individual functions to state a string can be used for a named color so maybe this is enough?

rethanon: Actually I'll just update this part to be very similar to your description above, sounds good.

Starbuck5 (approved):
Thanks!

PR #2350, opened 2023-07-26
Add `get_surface()` for Window class

yunline:
- `Window.get_surface()`
- `Window.update_from_surface()`
- Auto resize

```py
import pygame

pygame.init()

win = video.Window()
sf = win.get_surface()

sf.fill((255,0,0))
win.update_from_surface()

while 1:
    pygame.event.get()
```
<details>

<summary>Screenshot</summary>

图片
</details>


docs/reST/ref/sdl2_video.rst
```diff
@@ -190,6 +190,33 @@
...
+   .. method:: get_surface
+
+      | :sl:`Get the window surface`
+      | :sg:`get_surface() -> Surface`
+
+      Return a reference to the surface associated with the window.
+      
+      .. seealso:: :func:`update_from_surface`
+
+      .. versionaddedold:: 2.4.0
```
Starbuck5, line 202: Should be `versionadded`. `versionaddedold` is for pygame, versionadded is for pygame-ce

src_c/window.c
```diff
@@ -124,9 +124,145 @@ window_destroy(pgWindowObject *self, PyObject *_null)
...
+        return 0;
+
+    if (event_window_pg->_is_borrowed) {
+        display_surf = pg_GetDefaultWindowSurface();
+        if (!display_surf) {
+            return 0;
+        }
+        display_surf->surf = SDL_GetWindowSurface(event_window);
+        return 0;
+    }
```
Starbuck5, line 257: If it's borrowed shouldn't the surface be managed by the display.c event watch?

src_c/window.c
```diff
@@ -124,9 +124,145 @@ window_destroy(pgWindowObject *self, PyObject *_null)
         SDL_DestroyWindow(self->_win);
         self->_win = NULL;
     }
+    if (self->surf) {
+        self->surf->surf = NULL;
```
Starbuck5, line 128: Does the surface need SDL_FreeSurface here?

Starbuck5: Actually if you don't NULL it here the natural pygame surface dealloc will take care of it, right?

src_c/window.c
```diff
@@ -553,6 +689,11 @@ window_dealloc(pgWindowObject *self, PyObject *_null)
             SDL_SetWindowData(self->_win, "pg_window", NULL);
         }
     }
+    if (self->surf) {
+        self->surf->surf = NULL;
+        Py_DECREF(self->surf);
+    }
```
Starbuck5, line 695: Same here shouldn't you decref the pygame surface object and let that handle the internal SDL_Surface?

docs/reST/ref/sdl2_video.rst
```diff
@@ -224,6 +224,33 @@
...
+      | :sl:`Get the window surface`
+      | :sg:`get_surface() -> Surface`
+
+      Return a reference to the surface associated with the window.
+      
+      .. seealso:: :func:`update_from_surface`
+
+      .. versionadded:: 2.4.0
+   
+   .. method:: update_from_surface
```
Starbuck5, line 238: There's a real question of what we want to expose here.

Updating with rects is a niche use case, and we moved all the examples to use "flip" instead of "update"-- so should we continue that here and only expose the simpler functionality under the name "flip" ?

We could add both flip and update here.

And on names-- my first thought on "update_from_surface" is that we should just call it "update," but maybe update from surface is better.

Just starting this as a discussion.

zoldalma999: I don't think we should expose `display.update` like partial screen updates. It is hard to actually do for the user, and does not have much speedups.

MyreMylar: I tested this a bit here in #2514

In short, the passing the rect_list adds almost nothing practically performance-wise to CPU rendering. I don't think we should look to duplicate it into the future.

src_c/window.c
yunline: > Does the surface need SDL_FreeSurface here?

According to [SDL wiki](http://wiki.libsdl.org/SDL2/SDL_GetWindowSurface): "This surface will be freed when the window is destroyed. Do not free this surface."

> Actually if you don't NULL it here the natural pygame surface dealloc will take care of it, right?

I set the internal surface to NULL to make the pygame surface invalid, since the internal surface has been freed by SDL. 
If users want to operate on this surface they will get an error.
```py
win = Window()
sf = win.get_surface()
win.destroy()
sf.fill((0,0,0))
```
```
  File "x:\python\pygame_test\window_surface.py", line 11, in <module>
    sf.fill((0,0,0))
pygame.error: Surface is not initialized
```

src_c/window.c
yunline: fixed

src_c/window.c
Starbuck5: Okay, the behavior looks good, thanks for the additional context.

Can some of that be added there as a comment, I feel its hard to understand all the ramifications there.

MyreMylar (changes requested):
Overall this works great on windows OS and lets me use the window class to do familiar CPU blitting and drawing.

I agree that we should remove the list of rects overload from `update_from_surface()` however. The realistic speed ups from this functionality on most modern platforms are too minor, and too niche, to justify continuing it and the confusion it brings when users try to use it to speed up their games. They put a bunch of effort in to maintain an accurate list of sections of the screen that have changed this loop, only to then find that it in many cases it actually makes the program run slower because they have more than 15 or so moving objects in their application and the per-rectangle-cost of updating has exceeded the saving from the reduced area updated. 

See #2514 for more discussion.

test/window_test.py
```diff
@@ -299,6 +299,47 @@ def test_from_display_module(self):
...
 
+    def test_window_surface(self):
+        win = Window(size=(640, 480))
+        surf = win.get_surface()
+
+        self.assertIsInstance(surf, pygame.Surface)
+
+        # test auto resize
+        self.assertTupleEqual(win.size, surf.get_size())
+        win.size = (100, 100)
```
MyreMylar, line 310: do we need to do a `pygame.event.pump()` here to make sure the event watch handler is triggered?

yunline: Auto resize event watch is deleted when `pygame.quit()` is called.
This is the cause of this bug,
So I set the event watch when `pygame.init()` and the bug should be fixed now.

src_c/window.c
```diff
@@ -1113,7 +1085,5 @@ MODINIT_DEFINE(_window)
         return NULL;
     }
 
-    SDL_AddEventWatch(_resize_event_watch, NULL);
```
ankith26: Just commenting because I feel this change could be independent of this PR. I think it could fix another one of those init+quit+init issues.

But I would prefer moving this to the display specific init function instead, instead of messing with `base.c`. The diff would also be much less I reckon

ankith26: My bad, I assumed the changes were happening in `display.c` but I now see they are in `window.c`. My suggestion would still be along similar lines, create an "init-quit hook" (see what `display.c` does) for window, and set/unset this filter from there.

MyreMylar (approved):

MyreMylar (approved):

src_c/window.c
```diff
@@ -959,6 +1067,10 @@ static PyTypeObject pgWindow_Type = {
 static PyMethodDef _window_methods[] = {
     {"get_grabbed_window", (PyCFunction)get_grabbed_window, METH_NOARGS,
      DOC_SDL2_VIDEO_GETGRABBEDWINDOW},
+    {"_internal_mod_init", (PyCFunction)_window_internal_mod_init, METH_NOARGS,
+     "auto initialize for _window module"},
+    {"_internal_mod_quit", (PyCFunction)_window_internal_mod_quit, METH_NOARGS,
+     "auto quit for _window module"},
```
ankith26, line 1073: Yes these hooks look good, but I think they should be called from the display init/quit hook instead of the base pygame init/quit function (like how event and time are handled right now). This is so that if someone only does `pygame.display.init()` instead of the full `pygame.init()` we would still want the window hooks to run, as they are related

yunline: fixed

MyreMylar (approved):

Starbuck5 (approved):
Let's get this in. Thanks yunline!

Note from testing: window resize doesn't work properly when using a borrowed window--
```py
import pygame

pygame.init()

window = Window.from_display_module()

window_screen = window.get_surface()
running = True
while running:
        if event.type == pygame.QUIT:
            running = False

            if event.key == pygame.K_1:
                window.size = (500,500)
            if event.key == pygame.K_2:
                window.size = (700,700)

    window.update_from_surface()
pygame.quit()
```

After this is merged I have several things I'd like to do myself on this. I want to change `update_from_surface()` to `flip()`, I'd like to write some things in the docs about how it's analogous to set_mode and flip(), and I'd like to add a standalone window software rendering example.

The way forward on this is getting this PR merged. @yunline I'm leaving it to you to merge-- I think some of the commits could be squashed down, but it's up to your discretion.

PR #2351, opened 2023-07-26
Rename PyClockObject to pgClockObject, remove rendered field, small cleanups

zoldalma999:
As far as I could tell, nothing uses the rendered field, so no reason to keep it around

EDIT: I found the [commit](https://github.com/pygame/pygame/commit/dddaa9d11c61cdb3717f77b425f61c78efd7cd62#diff-6dd08642f71a022c76a495718c4942ce70262949b2b3ceb7644ee03bd8c37b7a) that added it. It had no use for it back then either. And there was no use added as far as I could tell.

Also changed the name from `PyClockObject` to `pgClockObject` to be more consistent with all the other struct's names.

Removed the `PyClockObject *_clock = (PyClockObject *)self;` lines by making `self`s be already clock objects and casting the functions in the methods table

### DIFF
--- a/src_c/time.c
+++ b/src_c/time.c
@@ -501,7 +501,6 @@ typedef struct {
     PyObject_HEAD Uint64 last_tick, fps_count, fps_tick;
     float fps;
     Uint64 timepassed, rawpassed;
-    PyObject *rendered;
 } PyClockObject;
 
 // to be called by the other tick functions.
@@ -559,7 +558,6 @@ clock_tick_base(PyObject *self, PyObject *arg, int use_accurate_delay)
             _clock->fps_count / ((nowtime - _clock->fps_tick) / 1000.0f);
         _clock->fps_count = 0;
         _clock->fps_tick = nowtime;
-        Py_XDECREF(_clock->rendered);
     }
     return PyLong_FromUnsignedLongLong(_clock->timepassed);
 }
@@ -611,8 +609,6 @@ static struct PyMethodDef clock_methods[] = {
 static void
 clock_dealloc(PyObject *self)
 {
-    PyClockObject *_clock = (PyClockObject *)self;
-    Py_XDECREF(_clock->rendered);
     Py_TYPE(self)->tp_free(self);
 }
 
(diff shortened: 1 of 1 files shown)


src_c/time.c
```diff
@@ -544,85 +542,80 @@ clock_tick_base(PyObject *self, PyObject *arg, int use_accurate_delay)
...
+     DOC_TIME_CLOCK_GETTIME},
+    {"get_rawtime", (PyCFunction)clock_get_rawtime, METH_NOARGS,
+     DOC_TIME_CLOCK_GETRAWTIME},
+    {"tick_busy_loop", (PyCFunction)clock_tick_busy_loop, METH_VARARGS,
      DOC_TIME_CLOCK_TICKBUSYLOOP},
     {NULL, NULL, 0, NULL}};
 
 static void
-clock_dealloc(PyObject *self)
+clock_dealloc(pgClockObject *self)
```
ankith26, line 608: nit: this cast is unneeded

ankith26 (approved):
LGTM, thanks for the PR :apple:

src_c/time.c
```diff
@@ -544,85 +542,80 @@ clock_tick_base(PyObject *self, PyObject *arg, int use_accurate_delay)
...
 
 /* clock object internals */
 
 static struct PyMethodDef clock_methods[] = {
-    {"tick", clock_tick, METH_VARARGS, DOC_TIME_CLOCK_TICK},
-    {"get_fps", clock_get_fps, METH_NOARGS, DOC_TIME_CLOCK_GETFPS},
-    {"get_time", clock_get_time, METH_NOARGS, DOC_TIME_CLOCK_GETTIME},
-    {"get_rawtime", clock_get_rawtime, METH_NOARGS, DOC_TIME_CLOCK_GETRAWTIME},
-    {"tick_busy_loop", clock_tick_busy_loop, METH_VARARGS,
+    {"tick", (PyCFunction)clock_tick, METH_VARARGS, DOC_TIME_CLOCK_TICK},
```
Starbuck5, line 596: This is a good opportunity for METH_O instead of METH_VARARGS, but that's not something you need to deal with in this PR. Same with clock_tick_busy_loop.

Starbuck5 (approved):
Nice simplifications in this PR!

src_c/time.c
Starbuck5: NVM, it can't be METH_O since it's 1 optional parameter, not 1 required parameter.

PR #2352, opened 2023-07-26
Re-add CircleCI for aarch64 builds

ankith26:
Mostly using the old config, but with a few updates. We now make base images directly on CircleCI itself, and use CircleCI DLC (docker layer caching) to speeden it up for subsequent builds

TODO (either in this PR or a future one)
- [x] Integration with the github actions release machinery somehow (can only test it while doing a dev release)

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,6 +1,19 @@
 # Use the latest 2.1 version of CircleCI pipeline process engine.
 # See: https://circleci.com/docs/2.0/configuration-reference
 version: 2.1
+parameters:
+  GHA_Actor:
+    type: string
+    default: ""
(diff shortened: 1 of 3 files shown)


Starbuck5 (approved):
Thanks!

Have you been able to test the release-gh-draft side of this workflow?

Can we get somebody to test the integrity of the generated wheels to gain confidence? Maybe Kyuchumimo? Not sure if they're on aarch64 or not.

ankith26:
However, I tested the `pull_circleci_artifacts.py` by itself locally, and it works. It should hopefully also work on github actions, but if it doesn't we can always handle the aarch64 wheels manually and fix issues as they arise.

PR #2355, opened 2023-07-27
Use OIDC to publish PyPI releases

ankith26:
https://docs.pypi.org/trusted-publishers/using-a-publisher/

TL;DR - moving away from `secrets.PYPI_TOKEN` (which I now removed) to a more secure way to do our automated releases.

After this change, there would be a small addition to our release process that also increases security. The action that does the PyPI release would wait for an approval from a steering council member.


Starbuck5 (approved):
I trust you to make this change.

PR #2356, opened 2023-07-29
Don't use SDL_ name for our compat layer

Starbuck5:
This actually bit me when I tried to compile pygame-ce on a mac running 2.0.21, which has the symbol SDL_IntersectFRectAndLine. The compilation failed because of the duplicate symbol.


ankith26 (approved):
LGTM nice simple PR :tada:

zoldalma999 (approved):
LGTM!

PR #2360, opened 2023-07-29
setup.py: add 'local_ci' and  extend 'test' command for a better developer workflow

dr0id:
### Idea
After some failing PR attempts where the ci builds would fail because I did not run all checks (especially linters!) locally I decided to write a batch to ensure (or at least increase the chance significantly) that my changes would pass the CI builds at first attempt. After a while I realized that I needed something more flexible and since all commands I used were in setup.py I thought it would be cool to have a command in there that would simulate a complete CI build run. So I implemented it as 'local_ci' command in setup.py.

Run **`python setup.py --help-command`** to see all available commands (I would like to sort them alphabetically, but that is another PR!).
Please check **`python setup.py local_ci --help`** to get more detailed help. With that command it is now possible to run all steps or just a slice of them or a few selected or a single one. So now during development I can choose the steps I just need at that moment. And when I think I'm finished just run it once without arguments and the CI build should pass.

The second thing I was bothered with was running the unittests. Setup.py only provided one command to run all tests. But during development I seldom want to run all, normally I just want to run all of the module I'm working on. And I didn't know how to run the test from the repository and not the installed one.
Please check the instructions with **`python setup.py test --help`** and use **`python setup.py test -c "-h"`** to see all the command line options of the pygame test runner.
Now its possible to pass in the command line: e.g. `python setup.py test -c "font ftfont -v -I"` to run all font and ftfont tests in verbose and interactive mode. 

**This has only been tested on windows so far.** Maybe some adaptations are needed for other platforms.

### As an example: working on font related stuff

Strongly recommended to run all font related tests. This can now be done using:
`python setup.py test -c "font ftfont freetype -v"  --copy 1`

And then change code and build using:
`python setup.py local_ci -s 11:14`

Repeat this until you have done your work. Then before pushing run following to ensure ci build will pass:
`python setup.py local_ci`


Starbuck5:
Interesting ideas.

For running tests I do `python -m pygame.tests`, I was unaware you could even run the tests through `setup.py`. For a specific module it's like `python -m pygame.tests.surface_test`

dr0id:
> Interesting ideas.
> 
> For running tests I do `python -m pygame.tests`, I was unaware you could even run the tests through `setup.py`. For a specific module it's like `python -m pygame.tests.surface_test`

Yes, I know that and I have been using this too. But the downside of that is, that is running the 'deployed' tests, e.g. the ones from `lib/sitepackages/pygame/tests`. If I modify those files then I have to copy them back into the repository directory. On the other hand if I modify the ones in the repository then I have to copy them to `lib/sitepackages/pygame/tests` to run.

When doing TDD the loops should be as tight as possible, so copying files is cumbersome.

Running tests from `lib/sitepackages/pygame/tests` and from the repository have both their own benefits.

This gives me an idea: maybe I should add a switch to run the tests from `lib/sitepackages/pygame/tests` through setup.py as well.

dr0id:
I like the general idea, but implementation may be tricky to get right. For one, installing packages with pip from a python program is a bit of a no-no (it's always better to instead, list any missing packages and tell the user to manually install it). I also think a new/improved dev guide that lists commands together at the bottom is an alternative to it

DR0ID_ — 02.08.2023 20:14
Thanks for your feedback.

You don't have to run those steps that install packages via pip programmatically. I find it quite useful.

But sure, it could be changed to a check if the packages is installed (maybe even check the min version required). 
OR 
alternatively just ask for install/update if the user presses yes. (there is such a method in the test/utils package, maybe it could be used or copy from as inspiration)

There is a way to list all commands (e.g. setup.py --help-commands and then you can get help for each command (e.g. for command test setup.py test --help ). I agree, it should be documented in the dev guide.

But I also added a new command: local_ci
This should emulate a ci build locally so you get less cycles pushing -> correcting -> pushing ... (less ci builds, less stress on the infrastructure, less stress). I know it does not guarantee a successful ci build (unless you run it on every OS and version of python)).

dr0id:
@MyreMylar I would like to hear your opinion to this PR.

MyreMylar:
> @MyreMylar I would like to hear your opinion to this PR.

OK.

I like it, it seems pretty handy to just run:
```
python setup.py local_ci
```

In a developer command prompt and come back later to see what's working and what's not - I also like the configurability.

Things I noticed:

1. I had to manually install `mypy` to get past the stubcheck section - since this command was installing modules all over the place it should probably also install that?
2. The docs build failed initially due to the version of sphinx it wanted to install conflicting with another module's sphinx version - I fixed this manually by upgrading the submodule, but it made me wonder - should this whole command be happening inside of a virtual environment/venv?
3. When the steps error they just kind of crash out with along undescriptive error - e.g. one of the tests is currently failing locally for me and the last error I get in the console is:

```
  File "C:\Users\dan\Programming\pygame-ce\setup.py", line 1312, in <module>
    setup(**PACKAGEDATA)
  File "C:\Users\dan\AppData\Local\Programs\Python\Python311\Lib\site-packages\setuptools\_distutils\core.py", line 185, in setup
    return run_commands(dist)
           ^^^^^^^^^^^^^^^^^^
  File "C:\Users\dan\AppData\Local\Programs\Python\Python311\Lib\site-packages\setuptools\_distutils\core.py", line 201, in run_commands
    dist.run_commands()
  File "C:\Users\dan\AppData\Local\Programs\Python\Python311\Lib\site-packages\setuptools\_distutils\dist.py", line 969, in run_commands
    self.run_command(cmd)
  File "C:\Users\dan\AppData\Local\Programs\Python\Python311\Lib\site-packages\setuptools\dist.py", line 1221, in run_command
    super().run_command(command)
  File "C:\Users\dan\AppData\Local\Programs\Python\Python311\Lib\site-packages\setuptools\_distutils\dist.py", line 988, in run_command
    cmd_obj.run()
  File "C:\Users\dan\Programming\pygame-ce\setup.py", line 1188, in run
    self._run_subprocess(step_idx, *cmd_arguments, environ)
  File "C:\Users\dan\Programming\pygame-ce\setup.py", line 1242, in _run_subprocess
    proc = subprocess.run(command_line, capture_output=False, shell=True, check=True, env=environ,
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "C:\Users\dan\AppData\Local\Programs\Python\Python311\Lib\subprocess.py", line 571, in run
    raise CalledProcessError(retcode, process.args,
subprocess.CalledProcessError: Command '['C:\\Users\\dan\\AppData\\Local\\Programs\\Python\\Python311\\python.exe', '-m', 'pygame.tests', '-v']' returned non-zero exit status 1.
```

It would probably be nicer if the command detected a non-success (non-0 error code here) in each step and reported a single line neatly something like like 'Process halted at STEP X due to failures in the step.'

Otherwise 👍

MyreMylar:
Also, I've not reviewed this before now as it is still set as a draft.

zoldalma999:
I'll close this pr as:
 - It seems stale
 - We don't use setup.py anymore
 - There is [act](https://github.com/nektos/act) (thanks Andrew for posting it on discord) which can run github actions locally

PR #2363, opened 2023-07-30
Fixed joystick doc + properly deprecate ``pygame.Joystick.init/get_id``

bilhox:
Hello,

In this PR, I corrected some errors I made while editing the joystick docs last time like : 
- Missing ``*`` for titles
- Wrong mappings for certain buttons

I as well removed ``pygame.Joystick.init`` and ``pygame.Joystick.get_id``, because the doc says that they were supposed to be removed since pygame 2.1, which wasn't the case. (Since it is my first time editing the code directly, I don't know if I removed it correctly)


dr0id (approved):
LGTM

zoldalma999:
I don't think we should remove these functions yet.

I know the docs say these should have been removed in 2.1, but that is because when they were deprecated, 2.1 was supposed to allow breaking changes. Since then, we moved away to a different versioning system (I don't even know what the system is anymore tbh), but we for sure don't plan on breaking compatibility with older pygame versions.

These functions were also not properly deprecated. A lot of people don't check the docs, so they might easily miss deprecations if we only mention it there. Instead, we switched to raising deprecation warnings (again, after this note was added in the docs) in the code itself. Those are a lot harder to miss, so we are more confident in removing things that are marked this way.

So I think this PR should add deprecation warnings to these functions instead of removing them.

dr0id (commented):
add deprecation warnings instead of removing those functions

bilhox:
Okay, I deprecated the methods in the code, shall I change the deprecation starting version from 2.0 to 2.3.1 ? Will this be commited to 2.3.1 branch ?

dr0id (commented):
there is a formatting issue in the c file

bilhox:
Okay, I fixed the problems (I hope).

dr0id (approved):

bilhox:
Changed deprecation version from 2.3.1 to 2.4.0 and added a missing 'f' to a f-string in the joystick example code.
Ready to be merged.

examples/joystick.py
```diff
@@ -65,10 +70,11 @@ def main():
 
         # Get count of joysticks.
         joystick_count = pygame.joystick.get_count()
-        lines.append(indent("Number of joysticks: {joystick_count}", indentation))
+        lines.append(indent(f"Number of joysticks: {joystick_count}", indentation))
         indentation += 1
 
         # For each joystick:
+        j = 0
```
ankith26, line 77: Don't get why this was added, seems like an unused variable. Unless I'm missing something

ankith26 (commented):
Thanks for working on this PR :star: 

Overall nice PR, but left a few minor reviews

examples/joystick.py
bilhox: This was for another test so I could display the two joystick info list at the same time. I forgot to remove it.

bilhox:
So I fixed the minor problems, it should be fine.

bilhox:
Nevermind, a test failed but idk why

ankith26 (approved):
LGTM, but left another minor review, resolve at will

docs/reST/ref/joystick.rst
```diff
@@ -145,7 +145,7 @@ variable. See :ref:`environment variables <environment-variables>` for more deta
       Return True if the Joystick object is currently initialised.
 
       .. ## Joystick.get_init ##
-
+
```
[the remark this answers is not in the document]
bilhox: fixed, it should be fine now

PR #2364, opened 2023-07-30
Added `Window.(min|max)imum_size`

Matiiss:
Ported this SDL2 functionality:
- https://wiki.libsdl.org/SDL2/SDL_SetWindowMinimumSize
- https://wiki.libsdl.org/SDL2/SDL_GetWindowMinimumSize
- https://wiki.libsdl.org/SDL2/SDL_SetWindowMaximumSize
- https://wiki.libsdl.org/SDL2/SDL_GetWindowMaximumSize

Some code to play with:
```py
import pygame
import pygame._sdl2 as pg_sdl2  # noqa


WIDTH, HEIGHT = 640, 360
FPS = 60

pygame.init()
screen = pygame.display.set_mode((WIDTH, HEIGHT), flags=pygame.RESIZABLE)
clock = pygame.Clock()

window = pg_sdl2.Window.from_display_module()
window.minimum_size = (WIDTH / 2, HEIGHT / 2)
window.maximum_size = (WIDTH * 2, HEIGHT * 2)
print(window.minimum_size)
print(window.maximum_size)

running = True
while running:
    dt = clock.tick(FPS) / 1000
    screen.fill("black")
    
    events = pygame.event.get()
    for event in events:
        if event.type == pygame.QUIT:
            running = False
        elif event.type == pygame.VIDEORESIZE:
            print(event.size)
    
    pygame.display.flip()
```

As for the tests, I'm not sure, probably one option is adding those after this gets merged https://github.com/pygame-community/pygame-ce/pull/2230
But also the underlying API would fail silently anyways. (still can check that setting values means they can be retrieved using same API)


src_c/window.c
```diff
@@ -381,6 +381,80 @@ window_get_size(pgWindowObject *self, void *v)
...
+static PyObject *
+window_get_minimum_size(pgWindowObject *self, void *v)
+{
+    int w, h;
+    PyObject *out = NULL;
+
+    SDL_GetWindowMinimumSize(self->_win, &w, &h);
+    out = Py_BuildValue("(ii)", w, h);
+
+    if (!out)
```
yunline, line 415: We don't need to check if out is NULL here. You can directly return out.

src_c/window.c
```diff
@@ -381,6 +381,80 @@ window_get_size(pgWindowObject *self, void *v)
...
+static PyObject *
+window_get_maximum_size(pgWindowObject *self, void *v)
+{
+    int w, h;
+    PyObject *out = NULL;
+
+    SDL_GetWindowMaximumSize(self->_win, &w, &h);
+    out = Py_BuildValue("(ii)", w, h);
+
+    if (!out)
```
yunline, line 452: You can directly return out here.

yunline (changes requested):

yunline:
```py
window.minimum_size = (640,481)
window.maximum_size = (640,480)
print(window.minimum_size) # (640, 481)
print(window.maximum_size) # (0, 0)

window.maximum_size = (640,480)
window.minimum_size = (640,481)
print(window.minimum_size) # (0, 0)
print(window.maximum_size) # (640, 480)
```

Currently, if maximum_size < minimum_size, it returns (0, 0), which is weird.
Let's raise an error if user is trying to set an invalid value.

Matiiss:
> ```python
> window.minimum_size = (640,481)
> window.maximum_size = (640,480)
> print(window.minimum_size) # (640, 481)
> print(window.maximum_size) # (0, 0)
> 
> window.maximum_size = (640,480)
> window.minimum_size = (640,481)
> print(window.minimum_size) # (0, 0)
> print(window.maximum_size) # (640, 480)
> ```
> 
> Currently, if maximum_size < minimum_size, it returns (0, 0), which is weird. Let's raise an error if user is trying to set an invalid value.

Yeah, I suppose raising an exception here should make it easier for users to debug this issue if it ever arises, although I suppose it could also just be documented instead (it is now too, kind of, docs now mention that an error will be raised and why) and the code could raise a warning?

Matiiss:
Also speaking of that `if (!out)` redundancy, `window_set_size` has it too, could that be handled in this PR too or new PR?

yunline:
> although I suppose it could also just be documented instead (it is now too, kind of, docs now mention that an error will be raised and why) and the code could raise a warning?

In SDL3, `SDL_SetWindowMaximumSize()` returns an `int` instead of `void`, and sets an error when maximum is smaller than minimum. So I think it's necessary to raise an exception than a warning.

https://github.com/libsdl-org/SDL/blob/8f00d7856d2ddaf08349c24726471a2d952570f9/src/video/SDL_video.c#L2767-L2771

yunline (approved):
LGTM

buildconfig/stubs/pygame/_sdl2/video.pyi
```diff
@@ -70,6 +70,14 @@ class Window:
...
     id: int
     size: Iterable[int]
+    @property
+    def minimum_size(self) -> Coordinate: ...
+    @minimum_size.setter
+    def minimum_size(self, value: Coordinate) -> None: ...
+    @property
+    def maximum_size(self) -> Coordinate: ...
+    @maximum_size.setter
+    def maximum_size(self, value: Coordinate) -> None: ...
```
Starbuck5, line 80: I feel like the point of making them properties is so that you can type the getter/setter differently.

But the getter is not returning Coordinate, it's returning a Tuple of ints.

Matiiss: Yes, indeed. Also that then makes sense for them to be properties since they're typed differently.

MyreMylar (approved):
LGTM. These sorts of functions are pretty useful for GUI apps with resizable windows.

MyreMylar:
Conflicts need resolving. Merge after 2.3.2.

Starbuck5:
> Merge after 2.3.2.

Main is on 2.4 already, so we can merge as soon as @Matiiss resolves the conflicts. Scanned through it myself and everything looks good.

Starbuck5 (approved):

PR #2369, opened 2023-08-01
Add a `mixer.Channel.id` getter

oddbookworm:


Starbuck5:
@oddbookworm circleci is failing for you but it's passing on main. After scratching my head for a second on this, I believe it's because it's pulling the circleci config from YOUR main branch, rather than the PR branch or pygame-ce's main branch.

@ankith26 tagging for visibility. I haven't thought about mitigation yet, but presumably we'd want circleci to use the config of the PR branches themselves, is there a way to have that?

oddbookworm:
Alright, I pulled in the most recent merges to both my own `main` and the PR branch

docs/reST/ref/mixer.rst
```diff
@@ -457,6 +457,17 @@ change the default buffer by calling :func:`pygame.mixer.pre_init` before
...
       alias.
 
+   .. method:: get_id
+
+      | :sl:`get the channel id for the Channel object`
+      | :sg:`get_id() -> int`
+
+      This simply returns the channel id used to create the ``Channel`` instance.
+
+      ..versionadded:: 2.3.1
```
Starbuck5, line 467: How is this function intended to be used?

The version added is 2.4.0 .

ankith26:
> I haven't thought about mitigation yet, but presumably we'd want circleci to use the config of the PR branches themselves, is there a way to have that?

Yes I'd like that too, but unfortunately there's nothing much we can do about it AFAIK.

docs/reST/ref/mixer.rst
MyreMylar, line 467:
```suggestion
      ..versionadded:: 2.4.0
```

MyreMylar (approved):
LGTM 👍 

It's not earth shattering as users could already keep track of their own channel IDs mapping to channels if they wanted, but might make it slightly easier to find out/debug which channel a sound is playing on with a simple number.

ankith26 (changes requested):
Since this is new API, how about considering making this a read-only property instead? Fewer characters to type and also pythonic

oddbookworm:
> Since this is new API, how about considering making this a read-only property instead? Fewer characters to type and also pythonic

I'll push a commit soon for this

oddbookworm:
force push to fix a half-complete commit

ankith26 (approved):
LGTM, thanks for the PR :+1:

MyreMylar (approved):
LGTM 👍

PR #2370, opened 2023-08-01
Try getting freetype tests running on CI, fix minor test fail issues

ankith26:


test/freetype_test.py
```diff
@@ -547,8 +544,7 @@ def test_freetype_Font_name(self):
         f = self._TEST_FONTS["fixed"]
         self.assertEqual(f.name, "Inconsolata")
 
-        nf = nullfont()
-        self.assertEqual(nf.name, repr(nf))
+        self.assertRaises(AttributeError, lambda: nullfont().name)
```
dr0id, line 547: AttributeError is some what miss leading since the attribute exists.

test/freetype_test.py
```diff
@@ -1693,7 +1689,7 @@ def test_freetype_SysFont_name(self):
         self.assertEqual(font_name_2, font_name)
 
         # Check mixed list of bytes and string.
-        names = [fonts[0], fonts_b[1], fonts[2], fonts_b[3]]
+        names = [fonts_b[i] if i % 2 else str_font for i, str_font in enumerate(fonts)]
```
dr0id, line 1692: could have just asserted for enough fonts?

dr0id (approved):

test/freetype_test.py
dr0id: I see, this is the exception that is thrown right now.

Starbuck5 (approved):
Thanks!

Starbuck5:
This PR is good, but should we have the behavior change in here at all?

It's from #2175 

It changes freetype.Font.name to raise an AttributeError on un-initialized freetype.Fonts. This changes the behavior, is not documented, and breaks the test now that it runs.

dr0id:
> This PR is good, but should we have the behavior change in here at all?
> 
> It's from #2175
> 
> It changes freetype.Font.name to raise an AttributeError on un-initialized freetype.Fonts. This changes the behavior, the change is not documented, and breaks the test now that it runs.

It was a PR-feedback that suggested to change that since 'repr(font)' is confusing when calling `f.name`. The documentation for that is missing, I agree. But honestly, how many times did you encounter an uninitialized font in your coding experience with pygame/pygame-ce?

test/freetype_test.py
ankith26: Well, the goal was to make this test work on cases where there are less than 3 system fonts (adding an assertion would break some platforms like CI). The current code should work regardless of the size (the aim of test is to mix `bytes` and `str` in a list and make sure the function doesn't error for this)

PR #2377, opened 2023-08-04
`BLEND_OVERLAY` blend mode

### LINKED ISSUE #2373: `BLEND_OVERLAY` flag
**Description**

We've had an essential number of blend modes available for a long time, which are:
`MAX` `MIN` `ADD` `SUB` `MULT`
While these flags can be combined together to achieve more effects, it's not possible to achieve all possible blend modes, even some of the most basic ones.

I propose to add one more blend flag called `OVERLAY` (both RGB and RGBA) which is:
- A very common blend mode present in most image editing softwares
- Can help with creating masked special effects for colorization and achieve generally better contrasted results compared to `MULT`.

This is an example screenshot comparing the current blend flags with `OVERLAY` (all use the same "BASE" BW image and the same overlay color):

image

While it's possible to achieve virtually any blend mode with surfarrays/numpy or even GPU shaders i strongly believe expanding on blend flags is a valid path, especially because it would:
- Help achieve a wider range of effects in most CPU-based programs 
- Avoid the incovenience of using external libraries and write more complex programs

**Technicalities**

Implementing such blend flags includes implementing 6 functions internally, 3 for alpha and 3 non-alpha versions, for single pixel, SSE2 and AVX2 versions), not to mention tests to ensure the correct behaviour. This could lead to difficulties during the review phase since 6 functions would need to be checked individually since they're mostly selected at compile time. So for this very reason i was planning to implement these flags in 2 or 3 phases where either:

- Only the single pixel and AVX2 versions are implemented first together and then the SSE2 version
- The single pixel version first, then the SSE2 and then AVX2

itzpr3d4t0r:
Closes #2373, featuring:
`BLEND_OVERLAY` and `BLEND_RGB_OVERLAY` flags and algorithm with:
- 1 pixel at a time implementation that's hardware agnostic
- 8 pixels at a time with AVX2
- 4 pixels at a time with SSE2

For a visual test i'm using the following image and program:
base

```Py
import pygame

pygame.init()
TEXT_COLOR = "black"
screen = pygame.display.set_mode((1000, 500))

img = pygame.image.load("base.png").convert()
img = pygame.transform.scale(img, (250, 250))

background = pygame.Surface(img.get_size())
background.fill((255, 100, 50))

background2 = pygame.Surface(img.get_size())
background2.fill((255, 0, 100))

font = pygame.font.SysFont("Arial", 28, True)
BLEND_OVERLAY = font.render("OVERLAY", True, TEXT_COLOR)
BLEND_ADD = font.render("ADD", True, TEXT_COLOR)
BLEND_MULT = font.render("MULT", True, TEXT_COLOR)
BASE = font.render("BASE", True, TEXT_COLOR)
MIN = font.render("MIN", True, TEXT_COLOR)
MAX = font.render("MAX", True, TEXT_COLOR)
SUB = font.render("SUB", True, TEXT_COLOR)

keep = True
clock = pygame.Clock()

while keep:
    clock.tick(60)

        if event.type == pygame.QUIT:
            keep = False

    screen.blits(
        [
            (img, (0, 250)),
            (img, (250, 250)),
            (img, (500, 250)),
            (img, (750, 250)),
            (img, (0, 0)),
            (img, (250, 0)),
            (img, (500, 0)),
            (img, (750, 0)),
            (background, (0, 0), None, pygame.BLEND_OVERLAY),
            (background, (250, 0), None, pygame.BLEND_ADD),
            (background, (500, 0), None, pygame.BLEND_MULT),
            (background, (250, 250), None, pygame.BLEND_MIN),
            (background, (500, 250), None, pygame.BLEND_MAX),
            (background, (750, 250), None, pygame.BLEND_SUB),
            (BLEND_OVERLAY, BLEND_OVERLAY.get_rect(center=(125, 20))),
            (BLEND_ADD, BLEND_ADD.get_rect(center=(375, 20))),
            (BLEND_MULT, BLEND_MULT.get_rect(center=(625, 20))),
            (BASE, BASE.get_rect(center=(125, 270))),
            (MIN, MIN.get_rect(center=(375, 270))),
            (MAX, MAX.get_rect(center=(625, 270))),
            (SUB, SUB.get_rect(center=(875, 270))),
        ]
    )

    pygame.display.flip()

```


src_c/alphablit.c
```diff
@@ -377,6 +379,25 @@ SoftBlitPyGame(SDL_Surface *src, SDL_Rect *srcrect, SDL_Surface *dst,
...
+#if !defined(__EMSCRIPTEN__)
+#if SDL_BYTEORDER == SDL_LIL_ENDIAN
+                    if (src->format->BytesPerPixel == 4 &&
+                        dst->format->BytesPerPixel == 4 &&
+                        src->format->Rmask == dst->format->Rmask &&
+                        src->format->Gmask == dst->format->Gmask &&
+                        src->format->Bmask == dst->format->Bmask &&
+                        !(src->format->Amask != 0 && dst->format->Amask != 0 &&
+                          src->format->Amask != dst->format->Amask) &&
+                        pg_has_avx2() && (src != dst)) {
```
MyreMylar, line 392: is it time for us (not in this PR, but in general) to MACRO this check up as it is repeated a lot now?

src_c/surface.h
```diff
@@ -202,6 +204,26 @@
...
+    else {                                            \
+        dG = 255 - (((255 - dG) * (255 - sG)) >> 7);  \
+    }                                                 \
+    if (dB < 127) {                                   \
+        dB = ((dB * sB)) >> 7;                        \
+    }                                                 \
+    else {                                            \
+        dB = 255 - (((255 - dB) * (255 - sB)) >> 7);  \
+    }
+
```
MyreMylar, line 226: I did wonder if we should be calculating overall pixel lightness here for the < 127 check? i.e. (r + b + g) / 3 - however every implementation I can find seems to do it on a per channel lightness basis rather than a per pixel lightness basis.

MyreMylar (changes requested):
OK, this all seems to work OK, the tests make sense and I'm convinced that it has value as a blend mode.

Only two things lacking really:

- An SSE2/NEON version for decent performance on macs/arm.
- Documentation on the special flags page - ideally we should include the formula in this one as it is more complicated than the existing set of MUL, ADD & SUB blend modes. Something like the description on the wikipedia page:
image  

Would help I think.

I think we could have the SSE2 version in another PR - because there will at least be a version for all platforms after this one, but we definitely need the docs or nobody will know it exists.

MyreMylar (approved):
Great! 

The new docs look lovely:

image


And the blend mode still works locally, as well as passing the unit tests on all platforms now.

Approved!

docs/reST/ref/special_flags_list.rst
```diff
@@ -62,6 +62,23 @@ Special Flags List
        - ``BLEND_MAX`` / ``BLEND_RGB_MAX``
            Takes the maximum value of each color channel
 
+       .. versionadded:: 2.5.0
```
MyreMylar:
```suggestion
       .. versionadded:: 2.5.6
```

MyreMylar (approved):
OK, this is back working again. Could probably get some more tweaks in to fit the nested macro world that is alphablit.c these days, but I don't think that is essential.

Approved again. 👍

oddbookworm:
After pulling main into this branch and testing your script, this is my result (which matches what I would expect I think)
out

oddbookworm:
@coderabbitai fresh review

PR #2380, opened 2023-08-05
Use Union[FRect, Rect] in places where only Rect was allowed in sprite.pyi (fix #2300)

### LINKED ISSUE #2300: mypy complains about `FRect` being assigned to `Sprite.rect`
**Environment:**

- **Python version**: 3.11.4
- **PyGame version**: 2.3.0
- **mypy version**: 1.4.1

**Expected behavior:**

mypy analysis should succeed when assigning an `FRect` to `Sprite.rect`.

**Steps to reproduce:**

1. Copy the code below and run `mypy` on it.
2. Observe errors.

**Test code**

```
import pygame


class SpriteSubclass(pygame.sprite.Sprite):
    def __init__(self, groups) -> None:
        self.rect = pygame.FRect(0, 0, 100, 100)

```

```
main.py:7: error: Incompatible types in assignment (expression has type "FRect", variable has type "Rect | None")  [assignment]
```

Looking through the [stubs](https://github.com/pygame-community/pygame-ce/blob/main/buildconfig/stubs/pygame/sprite.pyi), there's a handful of things that have type hints for `Optional[Rect]`. I don't know the code base so I don't know if this can be found elsewhere.

devdanzin:
mypy analysis fails when assigning an `FRect` to `Sprite.rect` and some other operations, like in example code below. This PR uses `Union[FRect, Rect]` in places where only `Rect` was allowed in sprite.pyi to fix that.

This fixes #2300.

Code below results in 6 mypy errors unpatched, no errors with PR.
```python
import pygame


class SpriteSubclass(pygame.sprite.Sprite):
    def __init__(self, groups=()) -> None:
        super().__init__(groups)
        self.rect = pygame.FRect(0, 0, 100, 100)


s = SpriteSubclass()
frect = pygame.FRect(0, 0, 1, 1)
s.rect = frect


class DirtySpriteSubclass(pygame.sprite.DirtySprite):
    def __init__(self, groups=()) -> None:
        super().__init__(groups)
        self.source_rect = pygame.FRect(0, 0, 100, 100)


ab: pygame.sprite.AbstractGroup = pygame.sprite.AbstractGroup()
ab.lostsprites = [frect]
ab.spritedict = {s: frect}


class ABSubclass(pygame.sprite.AbstractGroup):
    def draw(self, surface: pygame.Surface) -> list[pygame.FRect | pygame.Rect]:
        return [frect]

```


ankith26 (commented):
Hmmm... Would `FRect` suffice for saying `Union[FRect, Rect]`? I mean, `float` is equivalent to `Union[float, int]` so ideally the former should also hold. If it doesn't with the current implementation of the stubs, maybe that could be fixed too (not necessarily an item for your PR, just thinking out loud here)

ankith26 (approved):

PR #2382, opened 2023-08-05
AVX Surface.fill() setup, AVX BLEND_ADD

itzpr3d4t0r:
Our current implementation of `Surface.fill()` when using blend flags only implements the single-pixel strategy. This is a massive opportunity to speed things up.
This PR tries to start the changes with `BLEND_ADD`.

Results:
```
OLD FILL
2.042152139999234
2.060880370000814
1.977689829999872
2.0481357899989234
1.9969090399994456

NEW FILL
0.01658679000029224
0.01656728999951156
0.01650887000068906
0.016579669998463942
0.016620350000448526

BLIT (avx2 with cached surface)
0.02195337999946787
0.021780080000462478
0.02219769999937853
0.022039489999588113
0.02213338999863481

BLIT (avx2 no cached surface)
0.33580584000083036
0.34673017999957667
0.34221240999904695
0.33031794000053194
0.33174736999790183
```

Test Program:
```Py
from statistics import mean
from timeit import repeat

import pygame

pygame.init()

surf = pygame.Surface((500, 500))
surf.fill((132, 33, 200))

color = pygame.Surface((500, 500))
color.fill((24, 24, 24))

G = globals()

teststr = """
surf.fill((24, 24, 24), None, pygame.BLEND_ADD)
"""
for _ in range(5):
    print(mean(repeat(teststr, globals=G, number=1000, repeat=10)))

print()
teststr = """
surf.blit(color, (0, 0), None, pygame.BLEND_ADD)
"""
for _ in range(5):
    print(mean(repeat(teststr, globals=G, number=1000, repeat=10)))

```


Starbuck5:
Couple preliminary things.

I was interested to see if I could find any examples of people actually using Surface.fill with a blend flag online. I did, I found this: https://github.com/Rabbid76/PyGameExamplesAndAnswers/blob/master/documentation/pygame/pygame_blending_and_transaprency.md#change-the-color-of-a-surface-area-mask

AVX2 only works on x86, our SSE2 code also runs on ARM due to SSE2Neon.h. So SSE2 is more broadly important to us, since it will help x86 computers as well as ARM macs and other ARM devices.

I like that you're taking inspiration from my macro strategy. I see that you're using the SSE and AVX2 registers together, my blit macros use masked stores on the non-aligned edges so everything can be done with AVX2 registers and instructions. Is that something you'd like to do here?

itzpr3d4t0r:
> AVX2 only works on x86, our SSE2 code also runs on ARM due to SSE2Neon.h. So SSE2 is more broadly important to us, since it will help x86 computers as well as ARM macs and other ARM devices.

If what you mean is that i should do the SSE2 version first i guess it's fine to see this PR as a setup for avx and then the blend add SSE2 version and setup will come next instead of expanding AVX2, so after these two PRs we could either implement both the sse and avx versions of the same flags in a single PR or separately.

> I like that you're taking inspiration from my macro strategy.

yep =).

> I see that you're using the SSE and AVX2 registers together, my blit macros use masked stores on the non-aligned edges so everything can be done with AVX2 registers and instructions. Is that something you'd like to do here?

Yeah I've already replicated your work to compare performance and see the benefits. Tbh i thought about the "only avx" stategy myself without knowing your implementation. In practice i didn't see much of a difference, basically the same. I've also switched unrolled loops with a for loop. The main benefit there is that we would just need a single code for filling instead of two which is good. I didn't push that yet but might be wise for simplicity.

Starbuck5:
You’re talking about two codes, in my macro you only need 1, and you only need AVX for the AVX blitter. This is not about performance, this is about code simplicity. And this is the approach I would prefer.

Unrolled vs normal for loop— I don’t care too much. Unrolled has a larger code size, so if there’s no measurable benefit I’d do a normal for loop.

itzpr3d4t0r:
The latest commit split the filler macro in three parts:
- macro for changing sub-8 pixels batches
- macro for changing 8-pixel batches
- macro that implements three different loops to dispatch to avoid evaluating conditions

This brought an approx 33% perf improvement according to my testing:
image
Results:
```
=== Test with dispatch from FILLADD_DISPATCH.json ===
Mean: 0.00035926982913711493
Median: 0.00024040002608671784
Stdev: 0.0003665061626639085
=== Test without dispatch from FILLADD_NODISPATCH.json ===
Mean: 0.0004810019988556403
Median: 0.00026409997371956706
Stdev: 0.0005677493976967148
```

Code for generating the data:
```Py
import pygame
from timeit import timeit

plt.style.use(["dark_background", "no-latex", "notebook"])
pygame.init()

max_size = 1000
color = (233, 12, 34)
total_time = 0
data = []
save_file_name = "FILLADD_DISPATCH.json"

for s in range(max_size + 1):
    base = pygame.Surface((s, s))
    G = globals()
    im_data = timeit("base.fill(color, None, 1)", globals=G, number=10)
    total_time += im_data
    data.append(im_data)
    plt.scatter(s, im_data)


with open(save_file_name, "w") as f:
    f.write(str(data))

print(f"Total time: {total_time}"
      f"\nMean: {mean(data)}"
      f"\nMedian: {median(data)}"
        f"\nStdev: {stdev(data)}")

plt.title("Surface.fill() test")
plt.ylabel("Time (s)")
plt.show()
```
Code for comparing the results:
import json

plt.style.use(["dark_background", "no-latex", "notebook"])

def load_and_plot(file_name, label, color):
    try:
        with open(file_name, "r") as f:
            data = json.load(f)
    except FileNotFoundError:
        print("Run quack.py first!")
        quit()

    print(f"=== Test {label} from {file_name} ===")
    print(
        f"Mean: {mean(data)}"
        f"\nMedian: {median(data)}"
        f"\nStdev: {stdev(data)}"
    )
    print()

    plt.scatter(range(len(data)), data, 0.5, color=color, label=label)

load_and_plot("FILLADD_DISPATCH.json", "with dispatch", "green")
load_and_plot("FILLADD_NODISPATCH.json", "without dispatch", "white")

plt.title("Surface.fill() test")
plt.ylabel("Time (s)")
plt.legend()
plt.show()

src_c/surface_fill.c
```diff
@@ -866,6 +867,16 @@ surface_fill_blend(SDL_Surface *surface, SDL_Rect *rect, Uint32 color,
 
     switch (blendargs) {
         case PYGAME_BLEND_ADD: {
+#if !defined(__EMSCRIPTEN__)
+#if SDL_BYTEORDER == SDL_LIL_ENDIAN
+            if (surface->format->BytesPerPixel == 4 &&
+                (surface->pitch % surface->format->BytesPerPixel == 0) &&
```
Starbuck5, line 873: So I tried to break this using the old weird-pitched surface trick, but it seems like this clever little line stopped me.

This isn't a requirement in the surface blitters, imo the blend_add handler should be able to handle weird pitches.

src_c/simd_surface_fill_avx2.c
```diff
@@ -26,79 +26,67 @@ _pg_has_avx2()
...
+                                                              \
+            {FILL_CODE}                                       \
+                                                              \
+            /* store up to 7 pixels */                        \
+            _mm256_maskstore_epi32(pixels, mask, mm256_dst);  \
+                                                              \
+            pixels += pxl_skip * pxl_excess;                  \
+        }                                                     \
+        mm256_pixels = (__m256i *)pixels;                     \
+        if (n_iters_8 > 0) {                                  \
```
Starbuck5: So you've moved away from this strategy now because you found it slow. That's disappointing, as I would like my surface strategy to be good. 

I see here an unnecessary branch, maybe that contributes to slowness. If the number of iterations is 0 it just won't do the for loop.

Starbuck5:
I saw that you moved to my favored strategy and then moved away from it. I think there are speedups to be had in your implementation of my suggested strategy. 

Another consideration is code size: your current implementation copy-pastes the "add" code into the final binary 14 times by my count, because of the loop unrolling and the macro. Code size could be a bigger impact if it was a bigger routine (like a blit), rather than just a handful of instructions, but it's something to keep in mind.

itzpr3d4t0r:
It looks like your suggestions worked, I made some graphs to get a clearer view but the benefits are there. 
image
This is a zoom up to 1000x1000 pixels, about an 11% perf improvement:
image
And this is a zoom on the 1000-2000 range, perf is a bit worse around 1k but generally it seems faster and more stable past that.
image

Program for the graphs:
```Py
import json

plt.style.use(["dark_background", "no-latex", "notebook"])

def load_and_plot(file_name, label, color):
    try:
        with open(file_name, "r") as f:
            data = json.load(f)
    except FileNotFoundError:
        print("Run quack.py first!")
        quit()

    print(f"=== Test {label} from {file_name} ===")
    print(
        f"Mean: {mean(data)}"
        f"\nMedian: {median(data)}"
        f"\nStdev: {stdev(data)}"
    )
    print()
    plt.scatter(range(len(data)), data, 0.5, color=color, label=label)


#load_and_plot("FILL_ADD_OLD.json", "old (single pixel)", "green")
load_and_plot("FILL_ADD_AVX2.json", "loop unrolled (old)", "white")
load_and_plot("FILL_ADD_AVX2_FOR.json", "for loop (new)", "red")
load_and_plot("FILL_ADD_AVX2_FOR_INVERSE.json", "for loop excess at the end (new)", "yellow")

plt.title("Surface.fill() test")
plt.ylabel("Time (s)")
plt.legend()
plt.show()
```
 **Data files**

----

I had to change them to .txt from .json as github doesn't like .json:
[FILL_ADD_AVX2_FOR_INVERSE.txt](https://github.com/pygame-community/pygame-ce/files/12417666/FILL_ADD_AVX2_FOR_INVERSE.txt)
[FILL_ADD_OLD.txt](https://github.com/pygame-community/pygame-ce/files/12417667/FILL_ADD_OLD.txt)
[FILL_ADD_AVX2.txt](https://github.com/pygame-community/pygame-ce/files/12417668/FILL_ADD_AVX2.txt)
[FILL_ADD_AVX2_FOR.txt](https://github.com/pygame-community/pygame-ce/files/12417669/FILL_ADD_AVX2_FOR.txt)


Program to get the data:
```Py
import pygame
from timeit import timeit

pygame.init()

max_size = 2000
color = (233, 12, 34)
data = []
save_file_name = "FILL_ADD_AVX2_FOR_INVERSE.json"

for s in range(max_size + 1):
    base = pygame.Surface((s, s))
    G = globals()
    im_data = timeit("base.fill(color, None, 1)", globals=G, number=10)
    data.append(im_data)


with open(save_file_name, "w") as f:
    f.write(str(data))
```

src_c/surface_fill.c
```diff
@@ -866,6 +867,16 @@ surface_fill_blend(SDL_Surface *surface, SDL_Rect *rect, Uint32 color,
 
     switch (blendargs) {
         case PYGAME_BLEND_ADD: {
+#if !defined(__EMSCRIPTEN__)
```
dr0id, line 870: What about the other blend modes? In a follow up PR?

itzpr3d4t0r: Yep. I never know how much stuff i can pack in a PR, i could add them all at once but if you want to be very thorough and test everything it's gonna take a long time.

src_c/simd_fill.h
```diff
@@ -1,18 +1,10 @@
...
-#if !defined(PG_ENABLE_ARM_NEON) && defined(__aarch64__)
-// arm64 has neon optimisations enabled by default, even when fpu=neon is not
-// passed
-#define PG_ENABLE_ARM_NEON 1
-#endif
-
 int
 _pg_has_avx2();
 
 // SSE2 functions
```
Temmie3754, line 7: Should this comment be removed too then?

itzpr3d4t0r: Thanks!

MyreMylar:
looks like this needs a merge with main to get past the old CircleCI failure.

MyreMylar (approved):
Alright, LGTM 👍 

SIMD with the add blend is so nice and straightforward! 🎉

Starbuck5 (approved):
This looks good to me.

On to an SSE2 implementation, and roll out to other flags?

Starbuck5:
I'd like to be squashed down a bit before merge, please.

PR #2384, opened 2023-08-05
Not skipping a mixer test that previously was failing often

### LINKED ISSUE #601: SDL2: Fix test_sound_args. Flaky MixerModuleTest. (892)
#  **Issue &numero;892 opened by *[dlon](https://github.com/dlon)* at 2019-03-08 17:42:18**

The test often fails on AppVeyor: https://ci.appveyor.com/project/pygame/pygame/build/job/ja89wd4e8m6hvh76

```
FAIL: test_sound_args (pygame.tests.mixer_test.MixerModuleTest)
  File "C:\Python27-x64\lib\site-packages\pygame\tests\mixer_test.py", line 139, in test_sound_args
    self.assertEqual(get_bytes(mixer.Sound(wave_path)), snd_bytes)
```

**Related Docs**: https://www.pygame.org/docs/ref/mixer.html# pygame.mixer.Sound<hr>

#  Comments

# # #  **[cruxicheiros](https://github.com/cruxicheiros)* commented at 2020-06-26 11:10:06*

Went to check this out. I am pretty sure I am seeing the failure on my machine:

(running test_sound_args)
```
Ran 1 test in 0.343s

FAILED (unexpected successes=1)

Failure
Test should not succeed since it's marked with @unittest.expectedFailure
```

However, the appveyor link in the issue is broken so I can't see the exact problem. Any more details on how it fails on Appveyor and whether this is the same issue? Thanks.

<hr>

# # #  **[MyreMylar](https://github.com/MyreMylar)* commented at 2020-06-26 12:09:26*

I assume this was the way it used to fail on Appveyor:

```
FAIL: test_sound_args (pygame.tests.mixer_test.MixerModuleTest)
  File "C:\Python27-x64\lib\site-packages\pygame\tests\mixer_test.py", line 139, in test_sound_args
    self.assertEqual(get_bytes(mixer.Sound(wave_path)), snd_bytes)
```

Then illume changed the unit test to have the `expectedFailure` decorator so the test would still run but wouldn't block merges if it failed.

It works on stuff that is not Appveyor still, like your computer, so that's why you get:

`FAILED (unexpected successes=1)`

If you look at the relevent parts of the actual test it is a pretty strange failure:

```
snd = mixer.Sound(file=wave_path)
snd_bytes = get_bytes(snd)
self.assertEqual(get_bytes(mixer.Sound(wave_path)), snd_bytes)
```

Seems to imply that sometimes on appveyor the same .wav file loaded twice was producing different bytes. I guess debugging it would involve making a few changes to the test to suss out what might be going on and making more useful test failures (check the sizes of both byte objects, compare the first 20 bytes and the last 20 bytes). 

I notice the failure up there was also reported on python 2.7 so it may just be a python 2 windows only bug that will vanish once we eventually drop python 2 support.  It's also a report from last year so it's possible changes to SDL mixer in that time may also have resolved it.
<hr>

# # #  **[illume](https://github.com/illume)* commented at 2020-12-01 11:08:23*

I guess this test should be skipped, until the issue is diagnosed more (or even fixed!).
(Maybe just skipped on windows, or appveyor... since it seems related to the filesystem there)

Matiiss:
Fixes #601 (hopefully)

### DIFF
--- a/test/mixer_test.py
+++ b/test/mixer_test.py
@@ -142,8 +142,7 @@ class MixerModuleTest(unittest.TestCase):
         mixer.quit()
         self.assertRaises(pygame.error, mixer.get_num_channels)
 
-    # TODO: FIXME: appveyor and pypy (on linux) fails here sometimes.
-    @unittest.skipIf(sys.platform.startswith("win"), "See pygame-ce issue 601.")
+    # TODO: FIXME: pypy (on linux) fails here sometimes.
     @unittest.skipIf(IS_PYPY, "random errors here with pypy")
     def test_sound_args(self):
         def get_bytes(snd):


ankith26 (approved):
Not sure whether the issue is completely gone, but I'm inclined to merge this in and see how things go

MyreMylar (approved):
> Not sure whether the issue is completely gone, but I'm inclined to merge this in and see how things go

Agreed.

PR #2387, opened 2023-08-06
Fix minor issues in joystick and moveit examples

ankith26:
Basically title

- `joystick` - f-string not used properly so the count was not correctly displayed
- `moveit` - `delay(100)` in addition to the clock was killing the performance of the example


MyreMylar (approved):
LGTM 👍

PR #2395, opened 2023-08-14
Bump cython to 3.0.0

yunline:


ankith26 (commented):
I like the idea of getting on cython 3.0.0
I can't remember at this point, what happened to that one PR that was trying to get cython files generated as part of our build step?

MyreMylar:
> I like the idea of getting on cython 3.0.0 I can't remember at this point, what happened to that one PR that was trying to get cython files generated as part of our build step?

It is here #1877

pmp-p (approved):
lgtm, also fixes cython3+emsdk latest wasm multiple compile errors for me

Starbuck5:
Why not add `__reduce__` to the stubs instead of turning off pickling?

yunline:
> Why not add `__reduce__` to the stubs instead of turning off pickling?

Classes like `Renderer` was unable to be picled ever before. And I believe few users need to pickle those objects.
So I think turning off pickling can be a better way than adding something that may mess up the stubs.

dr0id (approved):
if it works...

LGTM

Starbuck5:
> Classes like Renderer was unable to be picled ever before. And I believe few users need to pickle those objects.
So I think turning off pickling can be a better way than adding something that may mess up the stubs.

It's not messing up the stubs to update them with a new method being created.

But it's okay good for now, you're right.

Starbuck5 (approved):
Thanks!

PR #2398, opened 2023-08-14
Add `transform.hsl()`

### LINKED ISSUE #2310: More Surface Adjustment Methods
**Description**

I propose adding three new methods to either the Surface class or the transform module (names aren't final since this is just a functionality suggestion):

- [ ] `Surface.set_hue()` / `transform.hue()`
- [ ] `Surface.set_saturation()` / `tranform.saturation()`
- [ ] `Surface.set_brightness()` / `transform.brightness()`

These methods would enable users to conveniently modify the brightness, hue, and saturation of surfaces, opening up possibilities for a variety of visual effects. They are present in basically any image editing software nowadays.

**Motivation**
While there is a lot of functionality already implemented to fiddle with surface blending/modification, i believe that these new functions could be a very relevant addition mainly for two reasons:
- They would enable a greater variety of visual effects.
- They would avoid extra complexity of using separate libraries/programs/combinations of existing pygame functionality.

Some possible use cases of these new functions are:
- help create lighting effects, simulate day and night transitions, and dynamically adjust brightness in visual environments or objects.
- allowing for different moods and visual atmospheres, so recoloring foliage to from a green summer pallete to a fall one.
- achieve vivid or muted color palettes, for changes in visual impact, say an horror game with hard shifts from color to black and white or close to it.

For those who don't know what hue or saturation mean, 
- `hue` basically defines the color tone, so that a surface that's mainly red can be shifted to mainly green.
- `saturation` defines the color intensity.

This is a simple example of the scales these three factors use:
image

itzpr3d4t0r:
Closes #2310.
This PR implements a general purpose function that lets users modify Surface's hue, saturation and lighness values.
Still a draft as it's missing docs, tests and possibly improved performance / fixes.

### DIFF
--- a/src_c/transform.c
+++ b/src_c/transform.c
@@ -2118,6 +2118,214 @@ surf_grayscale(PyObject *self, PyObject *args, PyObject *kwargs)
     }
 }
 
+#define MIN3(a, b, c) MIN(MIN(a, b), c)
+#define MAX3(a, b, c) MAX(MAX(a, b), c)
+
+static void
+RGB_to_HSL(Uint8 r, Uint8 g, Uint8 b, float *h, float *s, float *l)
+{
+    float min, max, delta;
+    float r_1, g_1, b_1;
+    r_1 = r / 255.0;
+    g_1 = g / 255.0;
+    b_1 = b / 255.0;
+
+    min = MIN3(r_1, g_1, b_1);
+    max = MAX3(r_1, g_1, b_1);
+
+    float chroma = max - min;
+    *l = (max + min) / 2.0;
+
+    if (chroma == 0) {
+        *h = 0;
+    }
+    else if (max == r_1) {
+        *h = 60 * fmod(((g_1 - b_1) / chroma), 6);
+    }
+    else if (max == g_1) {
+        *h = 60 * (((b_1 - r_1) / chroma) + 2);
+    }
+    else if (max == b_1) {
+        *h = 60 * (((r_1 - g_1) / chroma) + 4);
+    }
+
+    if (*h < 0) {
+        *h += 360;
+    }
+
+    if (*l == 0 || *l == 1) {
+        *s = 0;
+    }
+    else {
+        *s = chroma / (1 - fabs(2 * *l - 1));
+    }
+}
+
+static void
+HSL_to_RGB(float h, float s, float l, Uint8 *r, Uint8 *g, Uint8 *b)
+{
+    /* 0 <= h <= 360, 0 <= s <= 1, 0 <= l <= 1 */
+    float chroma = (1 - fabs(2 * l - 1)) * s;
+    float h_1 = h / 60.0;
+    float x = chroma * (1 - fabs(fmod(h_1, 2) - 1));
+    float m = l - chroma / 2.0;
+
+    float r_1, g_1, b_1;
+    if (h_1 < 1) {
+        r_1 = chroma;
+        g_1 = x;
+        b_1 = 0;
+    }
+    else if (h_1 < 2) {
+        r_1 = x;
+        g_1 = chroma;
+        b_1 = 0;
+    }
+    else if (h_1 < 3) {
+        r_1 = 0;
+        g_1 = chroma;
+        b_1 = x;
+    }
+    else if (h_1 < 4) {
+        r_1 = 0;
+        g_1 = x;
+        b_1 = chroma;
+    }
+    else if (h_1 < 5) {
+        r_1 = x;
+        g_1 = 0;
+        b_1 = chroma;
+    }
+    else if (h_1 < 6) {
+        r_1 = chroma;
+        g_1 = 0;
+        b_1 = x;
+    }
+    else {
+        r_1 = 0;
+        g_1 = 0;
+        b_1 = 0;
+    }
+
+    *r = (Uint8)((r_1 + m) * 255);
+    *g = (Uint8)((g_1 + m) * 255);
+    *b = (Uint8)((b_1 + m) * 255);
+}
+
+static SDL_Surface *
+modify_hsl(pgSurfaceObject *src, pgSurfaceObject *dst, float h, float s,
+           float l)
+{
+    SDL_Surface *surf = pgSurface_AsSurface(src);
+    SURF_INIT_CHECK(surf)
+    SDL_Surface *newsurf;
+
+    if (!dst) {
+        newsurf = newsurf_fromsurf(surf, surf->w, surf->h);
+        if (!newsurf)
+            return NULL;
+    }
+    else {
+        newsurf = pgSurface_AsSurface(dst);
+    }
+
+    if (newsurf->w != surf->w || newsurf->h != surf->h) {
+        return (SDL_Surface *)(RAISE(
+            PyExc_ValueError,
+            "Destination surface must be the same size as source surface."));
+    }
+    if (surf->format->BytesPerPixel != newsurf->format->BytesPerPixel) {
+        return (SDL_Surface *)(RAISE(
+            PyExc_ValueError,
+            "Source and destination surfaces need the same format."));
+    }
+
+    int x, y;
+    Uint8 r, g, b, a;
+    Uint8 *pixels = (Uint8 *)surf->pixels, *pix;
+    float s_h, s_s, s_l;
+    Uint32 pixel;
+
+    for (y = 0; y < surf->h; y++) {
+        for (x = 0; x < surf->w; x++) {
+            SURF_GET_AT(pixel, surf, x, y, pixels, surf->format, pix);
+            SDL_GetRGBA(pixel, surf->format, &r, &g, &b, &a);
+            RGB_to_HSL(r, g, b, &s_h, &s_s, &s_l);
+
+            if (h != 0) {
+                s_h += h;
+                s_h = s_h > 360 ? s_h - 360 : s_h < 0 ? s_h + 360 : s_h;
+            }
+            if (s != 0) {
+                s_s = s_s * (1 + s);
+                s_s = s_s > 1 ? 1 : s_s < 0 ? 0 : s_s;
+            }
+            if (l != 0) {
+                if (l < 0) {
+                    s_l = s_l * (1 + l);
+                }
+                else {
+                    s_l = s_l * (1 - l) + l;
+                }
+                s_l = s_l > 1 ? 1 : s_l < 0 ? 0 : s_l;
+            }
+
+            HSL_to_RGB(s_h, s_s, s_l, &r, &g, &b);
+            pixel = SDL_MapRGBA(newsurf->format, r, g, b, a);
+            SURF_SET_AT(pixel, newsurf, x, y, (Uint8 *)newsurf->pixels,
+                        newsurf->format, pix);
+        }
+    }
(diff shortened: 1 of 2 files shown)


PurityLake:
Looks fairly good, only thing really needed now is tests, and it should be good to go.

itzpr3d4t0r:
Bulk of the work is done. I changed the algorithm for a 19% perf improvement, added tests and docs. 
What is still missing is checking that the algorithm always works without strange quirks, possibly adding more tests (for the various bit depths) and fixing the reason why the tests are failing, which is a one-value-off issue that's a combination of factors.

src_c/transform.c
```diff
@@ -2118,6 +2118,225 @@ surf_grayscale(PyObject *self, PyObject *args, PyObject *kwargs)
...
+        }
+    }
+    else if (h < 0) {
+        while (h < -360) {
+            h += 360;
+        }
+    }
+    h /= 360.0;
+
+    newsurf = modify_hsl(surfobj, surfobj2, h, s, l);
```
yunline, line 2325: Shall we release the GIL here?

MyreMylar, line 2325:
```suggestion
    pgSurface_Lock(surfobj);
    Py_BEGIN_ALLOW_THREADS;
    newsurf = modify_hsl(surfobj, surfobj2, h, s, l);
    Py_END_ALLOW_THREADS;
    pgSurface_Unlock(surfobj);
```

Do you mean something like this?

src_c/transform.c
```diff
@@ -2118,6 +2118,225 @@ surf_grayscale(PyObject *self, PyObject *args, PyObject *kwargs)
...
+            }
+
+            HSL_to_RGB(s_h, s_s, s_l, &r, &g, &b);
+            pixel = SDL_MapRGBA(newsurf->format, r, g, b, a);
+            SURF_SET_AT(pixel, newsurf, x, y, dst_pixels, newsurf->format,
+                        pix);
+        }
+    }
+
+    SDL_UnlockSurface(newsurf);
```
MyreMylar, line 2268: Don't we need to lock this earlier to unlock it here?

MyreMylar (changes requested):
Overall this works well, I just have a couple of questions about surface locking and unlocking and as this seems to be done in several different ways all over the transform module. Looking up the python C API it seems to say:

> Note Calling system I/O functions is the most common use case for releasing the GIL, but it can also be useful before calling long-running computations which don’t need access to Python objects, such as compression or cryptographic functions operating over memory buffers. For example, the standard [zlib](https://docs.python.org/3/library/zlib.html#module-zlib) and [hashlib](https://docs.python.org/3/library/hashlib.html#module-hashlib) modules release the GIL when compressing or hashing data.

Which I think all our transforms would qualify for.

src_c/transform.c
MyreMylar: This was meant in reply to @yunline

yunline: > Do you mean something like this?

Yes that is what I mean.

src_c/transform.c
itzpr3d4t0r: I followed what i saw in grayscale, not sure maybe since we create a new surface it's already locked? gotta take a better look since i don't really remember...

src_c/transform.c
itzpr3d4t0r: I added the thread stuff but seems like it led to errors. I don't know wy tbh.

src_c/transform.c
```diff
@@ -2118,6 +2118,227 @@ surf_grayscale(PyObject *self, PyObject *args, PyObject *kwargs)
...
+        newsurf = newsurf_fromsurf(surf, surf->w, surf->h);
+        if (!newsurf)
+            return NULL;
+    }
+    else {
+        newsurf = pgSurface_AsSurface(dst);
+    }
+
+    if (newsurf->w != surf->w || newsurf->h != surf->h) {
+        return (SDL_Surface *)(RAISE(
```
yunline, line 2217: You can't `RAISE` here, since `PyErr_SetString()` needs GIL.
This may be the reason why the tests failed

yunline (changes requested):

src_c/transform.c
yunline: Try move the `RAISE`s out of this function

src_c/transform.c
```diff
@@ -2211,6 +2211,216 @@ surf_grayscale(PyObject *self, PyObject *args, PyObject *kwargs)
...
+    v1 = 2 * l - v2;
+    *r = hue_to_rgb(v1, v2, h + (1.0f / 3.0f));
+    *g = hue_to_rgb(v1, v2, h);
+    *b = hue_to_rgb(v1, v2, h - (1.0f / 3.0f));
+}
+
+static void
+modify_hsl(SDL_Surface *surf, SDL_Surface *dst, float h, float s, float l)
+{
+    int x, y;
```
MyreMylar, line 2295:
```suggestion
    int surf_locked = 0;
    if (SDL_MUSTLOCK(surf)) {
        if (SDL_LockSurface(surf) == 0) {
            surf_locked = 1;
        }
    }
    int x, y;
```

src_c/transform.c
```diff
@@ -2211,6 +2211,216 @@ surf_grayscale(PyObject *self, PyObject *args, PyObject *kwargs)
...
+                s_l = s_l > 1 ? 1 : s_l < 0 ? 0 : s_l;
+            }
+
+            HSL_to_RGB(s_h, s_s, s_l, &r, &g, &b);
+            pixel = SDL_MapRGBA(dst->format, r, g, b, a);
+            SURF_SET_AT(pixel, dst, x, y, dst_pixels, dst->format, pix);
+        }
+    }
+
+    SDL_UnlockSurface(dst);
```
MyreMylar, line 2335:
```suggestion
    if (surf_locked) {
        SDL_UnlockSurface(dst);
    }
```

MyreMylar (commented):
I spent a bit of time looking into surface locking in SDL. Basically it is only needed when the surface has been RLE encoded.

As far as I can tell (though [there is an open bug](https://github.com/libsdl-org/SDL/issues/5594) about it, but I agree with the last comment), `MUST_LOCK(surf)` tells us whether the surface needs locking - which, if true, is something we should do before any direct pixel access.

I've added two suggestions to the code here that will lock the surface before accessing the pixels and then gate the call to unlock only if the surface is already locked.

Untested, but you could probably create a surface with RLE - blitting it to activate the RLE encoding on the surface - and then try running the hsl function post blit to create a good test of locking/unlocking.

src_c/transform.c
```diff
@@ -2211,6 +2211,224 @@ surf_grayscale(PyObject *self, PyObject *args, PyObject *kwargs)
...
+
+static void
+modify_hsl(SDL_Surface *surf, SDL_Surface *dst, float h, float s, float l)
+{
+    int surf_locked = 0;
+    if (SDL_MUSTLOCK(surf)) {
+        if (SDL_LockSurface(surf) == 0) {
+            surf_locked = 1;
+        }
+    }
```
MyreMylar, line 2300:
```suggestion
    int surf_locked = 0;
    if (SDL_MUSTLOCK(surf)) {
        if (SDL_LockSurface(surf) == 0) {
            surf_locked = 1;
        }
    }
    int dest_locked = 0;
    if (SDL_MUSTLOCK(dst)) {
        if (SDL_LockSurface(dst) == 0) {
            dst_locked = 1;
        }
    }
```

yunline (approved):
LGTM👍

src_c/transform.c
```diff
@@ -2211,6 +2211,233 @@ surf_grayscale(PyObject *self, PyObject *args, PyObject *kwargs)
...
+
+static PyObject *
+surf_hsl(PyObject *self, PyObject *args, PyObject *kwargs)
+{
+    pgSurfaceObject *surfobj;
+    pgSurfaceObject *surfobj2 = NULL;
+    SDL_Surface *dst, *src;
+    float h = 0, s = 0, l = 0;
+
+    static char *keywords[] = {"surface",   "hue",          "saturation",
```
Starbuck5, line 2363: The keywords don't match the docs.

src_c/transform.c
```diff
@@ -2211,6 +2211,233 @@ surf_grayscale(PyObject *self, PyObject *args, PyObject *kwargs)
...
+    if (h > 0) {
+        while (h > 360) {
+            h -= 360;
+        }
+    }
+    else if (h < 0) {
+        while (h < -360) {
+            h += 360;
+        }
+    }
```
Starbuck5, line 2403: Isn't this equivalent to `h %= 360` ?

src_c/transform.c
```diff
@@ -2211,6 +2211,233 @@ surf_grayscale(PyObject *self, PyObject *args, PyObject *kwargs)
...
+
+    if (dst->w != src->w || dst->h != src->h) {
+        return RAISE(
+            PyExc_ValueError,
+            "Destination surface must be the same size as source surface.");
+    }
+    if (src->format->BytesPerPixel != dst->format->BytesPerPixel) {
+        return RAISE(PyExc_ValueError,
+                     "Source and destination surfaces need the same format.");
+    }
```
Starbuck5, line 2426: If you want to check format, check src->format->format. Them being the same bytes per pixel does not mean they are compatible at all (yes I know other methods in transform do this, but best not to replicate).

src_c/transform.c
itzpr3d4t0r: Thanks!

src_c/transform.c
itzpr3d4t0r: Thanks!

src_c/transform.c
itzpr3d4t0r: Thanks!

itzpr3d4t0r:
I have an update to share. I’ve managed to achieve a substantial performance improvement of approximately 43% for this Pull Request when dealing with 4bpp surfaces. This enhancement could be significant for real-time usage before we get AVX or SSE in. For instance, in a test case with an 800x800 resolution, the frame rate increased from 51 fps to 74 fps.

The idea is bypassing SDL’s functions for 4bpp surfaces, which are notably slow without any apparent reason. Instead, I’ve replaced the get/set sections with faster versions, ensuring no loss of functionality.
image

Starbuck5:
Itzpr brought up a bug on discord where only changing lightness would lead to unexpected results. I see he's given this PR a "dont merge" label to reflect this, but I'm going to move it to draft just to take it off the review docket while this is resolved.

itzpr3d4t0r:
I have an update to share on this. After some experiments I found out that the current strategy for lightness isn't actually bugged but rather it's considered a naive way of handling lightness change. Take this picture for example:
base2
If we use my algorithm for just increasing its lightness by 60% we get this:
base_pygame
Wich seems weird but it's actually correct.

If we take an image manipulation software such as paint.net (or Krita) and we increase the surface lightness by 60% we get this:
base_paint

Now, fortunately I've found out that Krita offers 2 modes for changing lightness and one is so called "legacy". The lòegagy one actually matches up with my implementation's result:
image

Fortunately Krita is open source so I can look up the implementation and correct this issue now.

Gabryel-lima (approved):
Wow, very good!

PR #2400, opened 2023-08-15
Support int and str input for color values in video module

yunline:
```py
import pygame

win=video.Window()
ren=video.Renderer(win)

sf=pygame.image.load("cobweb.png")
sf=pygame.transform.scale(sf, (128,128))

tex=video.Texture(ren,(128,128))
tex.update(sf)

img = video.Image(tex,(0,0,64,64))

tex.color = "red"
img.color = 0x66ccff
ren.draw_color = "#00ff00"

ren.fill_rect((256,128,128,128))
tex.draw(dstrect=(120,128,128,128))
img.draw(dstrect=(400,128,128,128))
ren.present()

while 1:
    pygame.event.get()
```
image


buildconfig/stubs/pygame/_sdl2/video.pyi
```diff
@@ -131,7 +140,10 @@ class Renderer:
     @classmethod
     def from_window(cls, window: Window) -> Renderer: ...
     draw_blend_mode: int
-    draw_color: Color
+    @property
+    def draw_color(self) -> Color: ...
```
dr0id, line 144: why is it called `draw_color` here and `color` in the other classes?

yunline: > why is it called `draw_color` here and `color` in the other classes?

I don't know.
They have been named like this since 5 years ago.
image

Personally, I think  the name `Texture.color` is a little bit confusing. However, these naming things are out of scope of this PR.

Starbuck5:
This PR seemingly relies on #2349, so that must be merged first.

docs/reST/c_api/color.rst
```diff
@@ -36,10 +36,16 @@ Header file: src_c/include/pygame.h
...
+   Return ``1`` on success, ``0`` otherwise and sets a python exception.
    This in an extension of :c:func:`pg_RGBAFromObj` optimized for pgColor_Type
-   instances.
+   instances and tuples. It can also handle integer and string color inputs based
+   on ``handle_flags``.
+
+.. c:function:: int pg_MappedColorFromObj(PyObject *val, SDL_PixelFormat *format, Uint32 *color, pgColorHandleFlags handle_flags)
+
+   Like above function, but returns mapped color instead. One notable point of difference is
+   the way in which ints are handled (this function directly interprets the int passed as the
```
Starbuck5, line 50: Interpreting as a mapped is less direct imo

ankith26:
The PR this PR depended on has been merged, though this still has extra commits? Fix with a rebase?

buildconfig/stubs/pygame/_sdl2/video.pyi
dr0id: ok, was just curious. Thanks for looking it up.

dr0id (approved):
LGTM

src_c/cython/pygame/_sdl2/video.pxd
```diff
@@ -434,7 +442,6 @@ cdef class Renderer:
 
 cdef class Texture:
     cdef SDL_Texture* _tex
-    cdef Color _color
```
Starbuck5, line 437: If you cdef the uint8[4] array here, it will be part of the Texture struct right? That way you don't need to worry about mallocing and freeing it.

Starbuck5:
Left a comment, there's also a merge conflict.

The ALL color flag is the same one used by the Color constructor, that's good. It's probably good to just have that be the standard and never support mapped color for Textures?

src_c/cython/pygame/_sdl2/video.pyx
```diff
@@ -340,23 +337,30 @@ cdef class Texture:
...
-                                              &self._color.data[2])
+                                              &(rgba[0]),
+                                              &(rgba[1]),
+                                              &(rgba[2]))
+
         if res < 0:
             raise error()
 
-        return self._color
+        return pgColor_NewLength(rgba, 3)
```
Starbuck5, line 351: Should be color length 4, because 4 is standard and the length is being deprecated. See https://pyga.me/docs/ref/color.html#pygame.Color.set_length

You'll need to explicitly set the 4th byte to 255 since it will be unset coming from color mod.

src_c/cython/pygame/_sdl2/video.pyx
```diff
@@ -872,13 +880,17 @@ cdef class Renderer:
...
         cdef int res = SDL_SetRenderDrawColor(self._renderer,
-                                              new_value[0],
-                                              new_value[1],
-                                              new_value[2],
-                                              new_value[3])
+                                              rgba[0],
+                                              rgba[1],
+                                              rgba[2],
+                                              rgba[3])
+
```
Starbuck5, line 893: Instead of storing the state here, is it easy to use https://wiki.libsdl.org/SDL2/SDL_GetRenderDrawColor instead?

That way we don't need to worry about maintaining state that can be mutated by the user.

yunline: Fixed

Starbuck5 (approved):
Thanks!

PR #1842, opened 2023-02-13
Added pg_IntFromObjEx(), pg_TwoIntsFromObjEx() and pg_IntFromSeqIndexEx()

itzpr3d4t0r:
Added three new C API functions:
- pg_IntFromObjEx
- pg_IntFromSeqIndexEx
- pg_TwoIntsFromObjEx

These three C API functions are supposed to be slightly faster than the current ones and have a different behaviour as they raise errors on their own, something that their current couterparts do not do. Errors should be more precise and provide more information, epecifically about what the error is, what it expected and what it got in terms of value/type/numebr of values.

You can either use a custom error message to override the default ones or pass NULL to use the predefined error messages and formatting. I recommend passing NULL most of the time as it will give you more precise and informative error messages. Only use a custom message when you need to add more description.

Modified:
- Rect.collidepoint() to use the new functionality as an example.

I'd like to higlight that in the case of collidepoint, passing a tupleof 2 items is about 83% faster and a list of 2 items about 38% faster. This could entail similar benefits for every function that prefers tuples/lists of 2 items.

**Rect.collidepoint() main:**
```
pygame-ce 2.1.4.dev1 (SDL 2.0.22, Python 3.11.0)

====||SEQUENCES||====
-) tup inside tup: 0.6059378 s
-) coll tup: 0.56530388 s
-) not coll tup: 0.5557564 s
-) coll list: 0.8165761 s
-) not coll list: 0.80312168 s
-) coll 2 args: 0.3694743 s
-) not coll 2 args: 0.36558976 s
-) coll unpack tuple 1->2: 0.63821076 s
-) not coll unpack tuple 1->2: 0.62786246 s
-) coll unpack list 1->2: 1.00910418 s
-) not coll unpack list 1->2: 0.98168378 s
______________
-) Mean time: 0.66714737

====||VECTORS||====
-) coll vector: 0.60935196 s
-) not coll vector: 0.61871136 s
______________
-) Mean time: 0.61403166
```
**This PR:**
```
pygame-ce 2.1.4.dev1 (SDL 2.0.22, Python 3.11.0)

====||SEQUENCES||====
-) tup inside tup: 0.31907396 s
-) coll tup: 0.30785382 s
-) not coll tup: 0.31080586 s
-) coll list: 0.59054886 s
-) not coll list: 0.59381622 s
-) coll 2 args: 0.31044572 s
-) not coll 2 args: 0.3121588 s
-) coll unpack tuple 1->2: 0.56166662 s
[...]

PR #2405, opened 2023-08-18
`pg_IntFromObj` rework

### LINKED ISSUE #2404: `pg_IntFromObj`
```
>>> import pygame 
pygame-ce 2.4.0.dev1 (SDL 2.26.5, Python 3.10.6)
>>> sf = pygame.Surface((100,100)) 
>>> pygame.draw.circle(sf, "red", (0,0), 1<<33) 
TypeError: radius argument must be a number
>>>
```

yunline:
`pg_IntFromObj` is used for converting python objects into c int.
Currently, when encountering an error, this API clears the error indicator and returns 0 representing error occured. Some error like `OverflowError` will be cleared and may be replaced with a different error like `TypeError`, which may make debug harder. 
When handling with a float that way larger than `INT_MAX`, this function just cast the number to int without any error returned. This may make the behavior of code unpredictable.

So I changed the API. The new function first converts the input object into python int object. Then convert the python int object into c int. If any error occurs , the function returns 0 with the error indicator kept. If the type of input object is not able to be converted as a python int object, the function will return 0 with no error set. 
This is compatible with the old API and (hopefully) is able to pass all the tests.
Here is an example of this function: 
```c
if (!pg_IntFromObj(radiusobj, &radius)) {
    if (!PyErr_Occurred()) {
        PyErr_SetString(PyExc_TypeError,
                        "radius argument must be a number");
    }
    return 0;
}
```

In this PR, I fixed #2404 in draw.c. I think there are still some bugs like that (especially in rect_impl.h). I'll fix those in other PRs in the same way if this PR is merged.

### DIFF
--- a/src_c/base.c
+++ b/src_c/base.c
@@ -479,20 +479,29 @@ static int
 pg_IntFromObj(PyObject *obj, int *val)
 {
     int tmp_val;
+    PyObject *tmp_obj;
+    SDL_bool do_decref = SDL_FALSE;
 
-    if (PyFloat_Check(obj)) {
-        /* Python3.8 complains with deprecation warnings if we pass
-         * floats to PyLong_AsLong.
-         */
-        double dv = PyFloat_AsDouble(obj);
-        tmp_val = (int)dv;
+    if (PyLong_Check(obj)) {
+        tmp_obj = obj;
+    }
+    else if (PyNumber_Check(obj)) {
+        tmp_obj = PyNumber_Long(obj);
+        if (!tmp_obj)
+            return 0;
+
+        do_decref = SDL_TRUE;
     }
     else {
-        tmp_val = PyLong_AsLong(obj);
+        return 0;  // Not a number type
     }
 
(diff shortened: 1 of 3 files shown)


itzpr3d4t0r:
There's already an effort towards changing how these functions behave with #1842. Specifically this function should in theory work as well, without converting to PyLongs: https://github.com/pygame-community/pygame-ce/pull/1842/files#diff-1b6175f657b362b42cd3209fbf24664562071e54b22636b8cce5dd4c2e9f87e6R533-R549. Also i don't think this will works out of the bos as you'd have to rewrite the whole codebase since all functions that use this one set an error when it fails.

yunline:
> There's already an effort towards changing how these functions behave with #1842. Specifically this function should in theory work as well, without converting to PyLongs

I read the code in your pr and I think some problems haven't been fixed yet.
I'll submit a review about this.

yunline:
Duplicate of #1842

PR #2410, opened 2023-08-20
wasm: in 32bits mode (mvp) lseek64 is not present

pmp-p:
fix compilation on recent emsdk 3.1.45

### DIFF
--- a/src_c/rwobject.c
+++ b/src_c/rwobject.c
@@ -36,6 +36,9 @@
 #elif defined(__APPLE__)
 /* Mac does not implement lseek64 */
 #define PG_LSEEK lseek
+#elif defined(__EMSCRIPTEN__)
+/* emsdk mvp 1.0 does not implement lseek64  */
+#define PG_LSEEK lseek
 #else
 #define PG_LSEEK lseek64
 #endif


src_c/rwobject.c
```diff
@@ -36,6 +36,9 @@
 #elif defined(__APPLE__)
 /* Mac does not implement lseek64 */
 #define PG_LSEEK lseek
+#elif defined(__EMSCRIPTEN__)
+/* emsdk mvp 1.0 does not implement lseek64  */
+#define PG_LSEEK lseek
 #else
 #define PG_LSEEK lseek64
```
ankith26, line 43: I know it's not really the focus of your PR, but now that you are modifying this part of the code it would be nice if you could also make a related change.
`lseek64` is a glibc only thing, so it must be used only there and on every other case `lseek` itself should be used. Maybe doing this also fixes the emscripten codepath

pmp-p: Like testing specifically glibc or then default to lseek ?
```
#elif defined(__EMSCRIPTEN__)
/* emsdk mvp 1.0 does not implement lseek64  */
#define PG_LSEEK lseek
#elif defined(_LARGEFILE64_SOURCE)
#define PG_LSEEK lseek
#else
#define PG_LSEEK lseek
#endif
```
nb: I'd prefer to keep the emsdk test since emscripten is gaining 64bits support head may define stuff we don't want for maximum backward compatibility

ankith26: Okay, I like the current block of code you have proposed

ankith26 (approved):
LGTM, thanks for the PR :star_struck:

dr0id (approved):
LGTM

PR #2413, opened 2023-08-21
wasm: add pixelarray support

pmp-p:
and also modules order/alias for correctly loading sprite module


dr0id (approved):
LGTM (not tested either)

Issue #2417, opened 2023-08-23
Some colors in dark theme make text invisible


yunline:
![image](https://github.com/pygame-community/pygame-ce/assets/31395137/a8e70295-7d76-4434-b073-6029980c00fb)
![image](https://github.com/pygame-community/pygame-ce/assets/31395137/ece46df0-d2fe-4e55-a112-966ee7edf842)
![image](https://github.com/pygame-community/pygame-ce/assets/31395137/bae09d60-d1f8-4e97-be83-b43944ac60a5)

Notenlish:
fix: https://github.com/pygame-community/pygame-ce/pull/2418

PR #2418, opened 2023-08-23
Fix caution visual bug for dark theme 

Notenlish:
Old:
image
New:
image

reported by @yunline

### DIFF
--- a/docs/reST/themes/classic/static/pygame.css_t
+++ b/docs/reST/themes/classic/static/pygame.css_t
@@ -735,6 +735,23 @@ div.caution {
     border: 1px solid #aabb88;
 }
 
+.dark-theme div.caution {
+    background-color: {{ theme_dark_cautionbgcolor }};
+    border: none;
+    border-radius: 7px;
+}
+.dark-theme dl.definition .admonition-title {
+    font-weight: bold;
+    color: #BA1200;
+    font-size: 1.2rem;
+}
+.dark-theme div.admonition.caution {
+    padding: 10px;
+}
+.dark-theme div.admonition.caution code {
+    background-color: transparent;
+}
+
 div.warning {
     background-color: #ffaec4;
     border: 1px solid #f66;
--- a/docs/reST/themes/classic/static/pygame.css_t
+++ b/docs/reST/themes/classic/static/pygame.css_t
@@ -737,19 +737,8 @@ div.caution {
 
 .dark-theme div.caution {
     background-color: {{ theme_dark_cautionbgcolor }};
-    border: none;
-    border-radius: 7px;
-}
-.dark-theme dl.definition .admonition-title {
-    font-weight: bold;
-    color: #BA1200;
-    font-size: 1.2rem;
-}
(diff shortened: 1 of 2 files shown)


Notenlish:
https://github.com/pygame-community/pygame-ce/issues/2417

yunline (changes requested):
Thanks for your fix.
However, there are 2 bugs mentioned in the issue, and you just fixed one of them.
The another one is in https://pyga.me/docs/tutorials/en/move-it.html
image
image

Starbuck5:
@yunline It's fine for this PR to only fix one of the problems, as long as the original issue doesn't get closed.

docs/reST/themes/classic/theme.conf
```diff
@@ -53,6 +53,7 @@ dark_h1color                = #e1cd62
 dark_h2color                = #00b800
 dark_notebgcolor            = #555349
 dark_notebordercolor        = #6e6b5e
+dark_cautionbgcolor         = rgb(255 25 25 / 30%)
```
Starbuck5, line 56: Opacity in the css? This is single-handedly bringing pygame-ce into the future :) /s

novialriptide: Should be a hex value. `/ 30%` shouldn't be here as well.

Starbuck5: I think it's fine for it to have opacity. It looks good in the output.

rethanon:
Just wanted to point out that we already have a color we use for a warning in the _sdl2 docs, should we try to be consistent and use the same color?
image

rethanon:
Which I guess does lead to the question, rather than just apply changes to the dark mode, could we apply the background color/text color universally like we have for the above warning in sdl2_video?

docs/reST/themes/classic/static/pygame.css_t
```diff
@@ -735,6 +735,23 @@ div.caution {
...
+    color: #BA1200;
+    font-size: 1.2rem;
+}
+.dark-theme div.admonition.caution {
+    padding: 10px;
+}
+.dark-theme div.admonition.caution code {
+    background-color: transparent;
+}
+
```
MyreMylar, line 754:
```suggestion
.dark-theme div.caution {
    background-color: {{ theme_dark_cautionbgcolor }};
    border: 1px solid;
    border-color: {{ theme_dark_cautionbordercolor }};
}

```

docs/reST/themes/classic/theme.conf
MyreMylar, line 56:
```suggestion
dark_cautionbgcolor         = rgb(255 25 25 / 30%)
dark_cautionbordercolor     = rgb(255 60 60)
```

MyreMylar:
I think it is fine for us to have different colours for Warning and Caution boxes.

I applied the suggestions from the earlier comments so we can get this merged.

Caution boxes in the dark theme now look like this:

image

MyreMylar (approved):
LGTM now.

PR #2427, opened 2023-08-29
`display.message_box()`

yunline:
As a replacement of `video.messagebox().`

## Examples:
```py
pygame.display.message_box("Hello")
```
图片
---
```py
pygame.display.message_box(
    "Hello",
    "Hello pygame-ce",
    type = "warn"
)
```
图片
---
```py
result = pygame.display.message_box(
    "Error",
    "V我50",
    type="error",
    buttons=("Wednesday", "Thursday", "Friday"),
    return_button = 1,
    escape_button = None
)

print(result)
```
图片


docs/reST/ref/display.rst
```diff
@@ -794,4 +794,32 @@ required).
...
+   :param str title: A title string.
+   :param str message: A message string. If this parameter is set to ``None``, the message will be the title.
+   :param str type: Set the type of messagebox, could be ``"info"``, ``"warn"`` or ``"error"``.
+   :param tuple buttons: An optional sequence of button name strings to show to the user.
+   :param int return_button: Button index to use if the return key is hit (``None`` for no button linked).
+   :param int escape_button: Button index to use if the escape key is hit (``None`` for no button linked).
+
+..
+   (Uncomment this after the window API is published)
+   :param Winodw parent_window: The parent window of the messagebox
```
MyreMylar, line 811:
```suggestion
   :param Window parent_window: The parent window of the messagebox
```

MyreMylar:
Should we take the opportunity of this function moving modules (from a dead module) to rename it from `messagebox()` to `message_box()` which is more in the PEP8 style?

Matiiss:
Should we have some constants for the modes too?
For example
`pygame.MESSAGE_BOX_INFO` or `pygame.display.MESSAGE_BOX_INFO` or something of the like that would simply represent those strings.

src_c/display.c
```diff
@@ -2676,6 +2682,175 @@ pg_set_allow_screensaver(PyObject *self, PyObject *arg, PyObject *kwargs)
...
+    else if (!strcmp(msgbox_type, "warn")) {
+        msgbox_data.flags |= SDL_MESSAGEBOX_WARNING;
+    }
+    else if (!strcmp(msgbox_type, "error")) {
+        msgbox_data.flags |= SDL_MESSAGEBOX_ERROR;
+    }
+    else {
+        PyErr_Format(PyExc_ValueError,
+                     "type should be 'info', 'warn' or 'error', "
+                     "got '%s'.",
```
Matiiss, line 2742: Mostly the error messages don't end with a period

src_c/display.c
```diff
@@ -2676,6 +2682,175 @@ pg_set_allow_screensaver(PyObject *self, PyObject *arg, PyObject *kwargs)
...
+                         buttons->ob_type->tp_name);
+            return NULL;
+        }
+        Py_ssize_t num_buttons = PySequence_Size(buttons);
+        if (num_buttons < 0) {
+            return NULL;
+        }
+        else if (num_buttons == 0) {
+            return RAISE(PyExc_TypeError,
+                         "'buttons' should contain at least 1 button.");
```
Matiiss, line 2797: Error messages don't end with a period mostly

zoldalma999:
Should this be added in system instead of display? I think it would make sense there, especially since SDL3 will (probably) have support for (save, open file, open folder) file dialogs as well.

Starbuck5:
> Should this be added in system instead of display? I think it would make sense there, especially since SDL3 will (probably) have support for (save, open file, open folder) file dialogs as well.

I've supported having this in display in the past. Good point about SDL file dialogs though (https://github.com/libsdl-org/SDL/issues/7445). They have a full implementation waiting in a PR.

In the SDL implementation, the messagebox and file dialog systems are very separate: different API styles, different modules. Hence it will be okay for us to have messagebox in `pygame.display` and file dialogs in `pygame.system`.

src_c/display.c
```diff
@@ -2676,6 +2682,175 @@ pg_set_allow_screensaver(PyObject *self, PyObject *arg, PyObject *kwargs)
...
+pg_message_box(PyObject *self, PyObject *arg, PyObject *kwargs)
+{
+    const char *title = NULL;
+    PyObject *message = Py_None, *parent_window = Py_None;
+    const char *msgbox_type = "info";
+    PyObject *buttons = NULL;
+    PyObject *return_button_index_obj = Py_None;
+    PyObject *escape_button_index_obj = Py_None;
+
+    static char *keywords[] = {"title",         "message", "type",
```
dr0id, line 2695:
```suggestion
    static char *keywords[] = {"title", "message", "type",
```

docs/reST/ref/display.rst
```diff
@@ -794,4 +794,32 @@ required).
...
+.. function:: message_box
+
+   | :sl:`Create a native GUI message box`
+   | :sg:`message_box(title, message=None, type='info', parent_window=None, buttons=('OK',), return_button=0, escape_button=0)`
+
+   :param str title: A title string.
+   :param str message: A message string. If this parameter is set to ``None``, the message will be the title.
+   :param str type: Set the type of message_box, could be ``"info"``, ``"warn"`` or ``"error"``.
+   :param tuple buttons: An optional sequence of button name strings to show to the user.
+   :param int return_button: Button index to use if the return key is hit (``None`` for no button linked).
```
dr0id, line 806: please mention that -1 is the default return value if 'None' is provided.

src_c/display.c
yunline: This doesn't pass the clang format check.

docs/reST/ref/display.rst
yunline: Index `-1` means "the last value of a sequence" in python. This may make users confused.
In my opinion, if index is less than zero or out of range, raising an error could be better.

src_c/display.c
dr0id: hmm ok, still strange formatting, sorry for the trouble

docs/reST/ref/display.rst
yunline: I updated the code, and `-1` is an invalid input now.

test/message_box_test.py
Starbuck5, line 1: Shouldn't this be in display_test?

docs/reST/ref/display.rst
```diff
@@ -794,4 +794,31 @@ required).
    .. versionadded:: 2.2.0
    .. ## pygame.display.set_allow_screensaver ##
 
+.. function:: message_box
+
+   | :sl:`Create a native GUI message box`
+   | :sg:`message_box(title, message=None, type='info', parent_window=None, buttons=('OK',), return_button=0, escape_button=None)`
```
Starbuck5, line 800: The sg tag doesn't show a return in the line like other functions do.

src_c/display.c
```diff
@@ -2676,6 +2682,192 @@ pg_set_allow_screensaver(PyObject *self, PyObject *arg, PyObject *kwargs)
...
+
+        if (0 > return_button_index || return_button_index >= num_buttons) {
+            PyErr_SetString(PyExc_IndexError,
+                            "return_button index out of range.");
+            return NULL;
+        }
+        if (0 > escape_button_index || escape_button_index >= num_buttons) {
+            PyErr_SetString(PyExc_IndexError,
+                            "escape_button index out of range.");
+            return NULL;
```
Starbuck5, line 2813: Might as well use `RAISE` here like the error case above it.

test/message_box_test.py
yunline: It's easier to run the msgbox interactive test only, since display has it's own interactive tests.

Starbuck5: That's good for your development of it but probably not great to have committed. I don't think any other functions have their own test files outside their module, it could be confusing.

yunline: fixed

Starbuck5 (approved):
Alright, lets pull this in. I couldn't find anything wrong in testing.

Thanks for moving this forward!

Eventually it would be nice to have _sdl2.video.messagebox be a compat shim over this API, but that's for another pull request.

src_c/display.c
```diff
@@ -2675,6 +2681,190 @@ pg_set_allow_screensaver(PyObject *self, PyObject *arg, PyObject *kwargs)
...
+            arg, kwargs, "s|OsO!OiO", keywords, &title, &message, &msgbox_type,
+            &pgWindow_Type, &parent_window, &buttons, &return_button_index,
+            &escape_button_index_obj)) {
+        return NULL;
+    }
+
+    int escape_button_index = 0;
+    SDL_bool escape_button_used = SDL_FALSE;
+    if (escape_button_index_obj != Py_None) {
+        escape_button_index = PyLong_AsLong(escape_button_index_obj);
```
ankith26, line 2709: `escape_button_index` can be negative, it would be convenient to users if this could be interpreted as python-style indexing (like if I want to specify the last button in the list, I could say `-1`)

src_c/display.c
```diff
@@ -2675,6 +2681,190 @@ pg_set_allow_screensaver(PyObject *self, PyObject *arg, PyObject *kwargs)
...
+#if SDL_VERSION_ATLEAST(2, 0, 12)
+    msgbox_data.flags |= SDL_MESSAGEBOX_BUTTONS_LEFT_TO_RIGHT;
+#endif
+
+    if (parent_window == Py_None)
+        msgbox_data.window = NULL;
+    else
+        msgbox_data.window = ((pgWindowObject *)parent_window)->_win;
+
+    msgbox_data.colorScheme = NULL;  // use system color scheme settings
```
ankith26, line 2744: would be nice to gives users the ability to override this, with an extra keyword argument mayhaps? However, if you want to leave that out of this PR for a future PR, that is cool too

Starbuck5: Future PR for sure.

src_c/display.c
yunline: fixed

ankith26 (approved):

Left two reviews that are nice-to-haves

docs/reST/ref/display.rst
```diff
@@ -794,4 +794,31 @@ required).
...
+..
+
+   :return: The index of the button that was pushed.
+
+   This function should be called on the thread that ``set_mode()`` is called.
+   It will block execution of that thread until the user clicks a button or
+   closes the message_box.
+
+   This function may be called at any time, even before ``pygame.init()``.
+
```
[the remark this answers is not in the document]
yunline: fixed

docs/reST/ref/display.rst
```diff
@@ -818,6 +818,9 @@ required).
 
    This function may be called at any time, even before ``pygame.init()``.
 
+   Negative values of ``return_button`` and ``escape_button`` are allowed
+   just like the list index.
```
Starbuck5, line 822: I read this sentence and said "what list?"

`just like standard Python list indexing.` ?

Starbuck5:
@yunline I'm not sure why circleci is failing for you, maybe a rebase against main would fix it?

This looks ready to merge, I left a tiny wording note and this needs to be squashed down to fewer commits before merging.

pmp-p:
That seems like a blocking function, it is really necessary to add things that cannot be ported on some platforms ?

i mean was really video.messagebox used anyway ?

PR #2428, opened 2023-08-29
Change relevant mouse event attributes to floats

Starbuck5:
In preparation for SDL3, where these attributes are floats. See SDL3 include: https://github.com/libsdl-org/SDL/blob/main/include/SDL3/SDL_events.h. 

It would be un-ideal to get these attributes as floats and then cast them to ints before returning them to our users, just throwing away the precision. So we should get ready for the higher precision.

This is not 100% backwards compatible, somebody could be indexing an area with a mousemotion.pos for instance. But I think the vast majority of projects will be unaffected. Especially in Python3, where ints do floating division by default anyways. @ankith26 I'm curious if you think the MOUSEWHEEL change will be too disruptive? I feel like it's more of a change than the `pos` ones.

### DIFF
--- a/docs/reST/ref/event.rst
+++ b/docs/reST/ref/event.rst
@@ -83,9 +83,14 @@ specific attributes.
     VIDEOEXPOSE       none
     USEREVENT         code
 
-.. versionchangedold:: 2.0.0 The ``joy`` attribute was deprecated, ``instance_id`` was added.
+.. versionchangedold:: 2.0.0 The ``joy`` attribute was deprecated,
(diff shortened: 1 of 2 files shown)


ankith26:
An alternative is to add `precise_pos` and `precise_rel` instead. I know adding more attributes is not super elegant, but atleast it's going to be consistent with `precise_x` and `precise_y` that we already added kinda recently (and not deprecate it)?

I'm not pushing for this suggestion, just put up a point for discussion with more people

Starbuck5:
Closing to wait on this

PR #2429, opened 2023-08-29
Change mouse.get_(pos|rel) to output floats

Starbuck5:
In SDL3, the underlying APIs give us floats. So we should get ready to return those numbers.

I attempted to preserve itzprs speedups by making a float variant of the int tuple packer. I have not run performance tests.


ankith26 (commented):
Alternate idea - Make the function take a kwarg and decide based on that?

removes breaks/disruption but slightly less elegant

Starbuck5:
Closing to wait on this.

PR #2407, opened 2023-08-18
Cap the sphinx version under 7.2.0

oddbookworm:
Addresses #2406

PR #2431, opened 2023-08-30
Revert #2407

oddbookworm:
#2407 was to cap the sphinx version due to a breaking change in their behavior. It was fixed and the two most recent releases work again. No need to keep the version capped now
Sphinx issue: https://github.com/sphinx-doc/sphinx/issues/11618


Starbuck5:
I think we should continue with this pinned. Sphinx will break again in the future for sure, as they have active deprecation and removal cycles.

oddbookworm:
Closing because Starbuck5 makes a valid point

PR #2432, opened 2023-09-01
Add SIMD versions of the greyscale transform (attempt #2)

MyreMylar:
Original PR was here: #2214

The last attempt ended up a bit of a mess after it got split into four different PRs, then there was an attempt to rebase and git seemed to get completely bamboozled.

This PR requires #2421 to function, and serves as a demonstration, or proof of concept for that PR. The intent being that we could add SIMD versions of many more of the transform functions making some of them more practical for 'real-time' (i.e. once per frame) full-screen usage in a game.

I'm thinking here of having a live desaturation effect to indicate health loss (or some other game state) across the whole screen by multiply blending a grey scale version of the current full-display frame with a greyscale version of it. Lots of fun post process type effects become possible without having to delve into setting up shaders if you can make the transform functions fast enough for this kind of real-time usage.

So the other obvious question is, does this work? I think last time I gave it a test it was about 12x faster for the AVX2 version on a medium sized image but I may try out exactly the sort of example I talk about above with a test program to see what sort of practical frame-rate difference we end up with.

I think there was some feedback last time on the SIMD code that I still need to respond to before it all got messed up, so I'll try and get to that as well. This will also remain a draft until #2421 gets itself merged.


itzpr3d4t0r:
I've been fiddling with the code for a while and ended up with this. Seems to works and requires far less code.
```C
void
grayscale_avx2(SDL_Surface *src, SDL_Surface *newsurf)
{
    const int bpp = src->format->BytesPerPixel;
    const int w = src->w;
    const int src_pitch = (src->pitch - w * bpp) >> 2;
    const int dst_pitch = (newsurf->pitch - w * bpp) >> 2;
    const int excess = w % 8;
    const int excess_plus_pitch = excess + src_pitch;
    const int num_batches = w / 8;
    int h = src->h, i;

    const __m256i rgb_weights =
        _mm256_set1_epi16((short)(0x4C + (0x96 << 8) + (0x1D << 16)));
    const __m128i rgb_weights128 = _mm256_castsi256_si128(rgb_weights);
    __m256i src_pixels, src_alpha, gray_pixels;
    const __m256i mask = _mm256_set_epi32(
        0, excess > 6 ? -1 : 0, excess > 5 ? -1 : 0, excess > 4 ? -1 : 0,
        excess > 3 ? -1 : 0, excess > 2 ? -1 : 0, excess > 1 ? -1 : 0,
        excess > 0 ? -1 : 0);
    Uint32 *srcp = (Uint32 *)src->pixels;
    Uint32 *dstp = (Uint32 *)newsurf->pixels;

    while (h--) {
        for (i = 0; i < num_batches; i++) {
            src_pixels = _mm256_loadu_si256((__m256i *)srcp);
            src_alpha = _mm256_subs_epu8(src_pixels, rgb_weights);
            gray_pixels = _mm256_mullo_epi16(src_pixels, rgb_weights);
            gray_pixels = _mm256_srli_epi16(gray_pixels, 8);
            gray_pixels = _mm256_packus_epi16(gray_pixels, gray_pixels);
            gray_pixels = _mm256_adds_epu8(gray_pixels, src_alpha);
            _mm256_storeu_si256((__m256i *)dstp, gray_pixels);
            srcp += 8;
            dstp += 8;
        }

        if (excess) {
            src_pixels = _mm256_maskload_epi32((int *)srcp, mask);
            src_alpha = _mm256_subs_epu8(src_pixels, rgb_weights);
            gray_pixels = _mm256_mullo_epi16(src_pixels, rgb_weights);
            gray_pixels = _mm256_srli_epi16(gray_pixels, 8);
            gray_pixels = _mm256_packus_epi16(gray_pixels, gray_pixels);
            gray_pixels = _mm256_adds_epu8(gray_pixels, src_alpha);
            _mm256_maskstore_epi32((int *)dstp, mask, gray_pixels);
            srcp += excess_plus_pitch;
            dstp += excess_plus_pitch;
            continue;
        }
        srcp += src_pitch;
        dstp += dst_pitch;
    }
}
```

MyreMylar:
I created a testbed for 'real world' testing of many small blits a while ago, but it also serves for this.

```python
import random
import pygame
from pygame import Rect, Font, Surface, Color, Vector2
from pygame.sprite import Group, Sprite


class Splash(Sprite):
    def __init__(self, splash_animation, position, splash_group: Group):
        super().__init__(splash_group)
        self.splash_anim = splash_animation
        self.splash_current_frame = 0
        self.image = self.splash_anim[self.splash_current_frame]
        self.splash_frame_time = 0.033
        self.splash_frame_time_acc = 0.0

        self.rect = Rect(0, 0, 16, 16)
        self.rect.midbottom = position

    def update(self, td: float):
        self.splash_frame_time_acc += td
        if self.splash_frame_time_acc >= self.splash_frame_time:
            self.splash_frame_time_acc = 0.0
            self.splash_current_frame += 1
            if self.splash_current_frame >= len(self.splash_anim):
                self.kill()
            else:
                self.image = self.splash_anim[self.splash_current_frame]


class RainDrop(Sprite):
    def __init__(self, drop_surf, position, screen_bottom,
                 rain_drop_group: Group, splash_animation, splash_group: Group):
        super().__init__(rain_drop_group)
        self.image = drop_surf

        self.position = Vector2(position[0], position[1])
        self.screen_bottom = screen_bottom

        self.splash_group = splash_group
        self.splash_anim = splash_animation

        self.fall_vec = Vector2(0.0, 1.0)
        self.fall_speed = random.normalvariate(400.0, 10.0)

        self.rect = Rect(0, 0, 1, 6)
        self.rect.midbottom = self.position

    def update(self, td: float, platforms: Group):

        self.position += self.fall_vec * self.fall_speed * td
        self.rect.midbottom = self.position

        for platform in platforms.sprites():
            if self.rect.colliderect(platform.rect):
                self.rect.bottom = platform.rect.top
                self.position.y = platform.rect.top
                self.kill()

        if self.alive() and self.rect.bottom >= self.screen_bottom:
            self.rect.bottom = self.screen_bottom
            self.position.y = self.screen_bottom
            self.kill()

        if not self.alive():
            Splash(self.splash_anim, self.position, self.splash_group)


class Platform(Sprite):
    def __init__(self, position, size, plat_group: Group):
        super().__init__(plat_group)
        self.image = Surface(size)
        self.image.fill((50, 50, 50))
        self.image.fill((30, 30, 30), Rect((1, 1), (size[0]-2, size[1]-2)))

        self.rect = Rect(position, size)


pygame.init()

desktop_size = pygame.display.get_desktop_sizes()[0]
display_surf = pygame.display.set_mode(desktop_size)
full_screen_surf = pygame.Surface(desktop_size)
greyscale_surf = pygame.Surface(desktop_size)

fifty_percent_surf = pygame.Surface(desktop_size)
fifty_percent_surf.fill((128, 128, 128))

seventy_five_percent_surf = pygame.Surface(desktop_size)
seventy_five_percent_surf.fill((191, 191, 191))

twenty_five_percent_surf = pygame.Surface(desktop_size)
twenty_five_percent_surf.fill((64, 64, 46))

background = pygame.image.load("images/background.png").convert()

splash_sheet = pygame.image.load("images/splash.png").convert_alpha()
splash_anim = [splash_sheet.subsurface(Rect(0, 0, 16, 16)),
               splash_sheet.subsurface(Rect(16, 0, 16, 16)),
               splash_sheet.subsurface(Rect(32, 0, 16, 16)),
               splash_sheet.subsurface(Rect(48, 0, 16, 16)),
               splash_sheet.subsurface(Rect(64, 0, 16, 16))]

splash_horiz_pos = random.uniform(0, 1904)

rain_drop_surf = Surface((1, 6), pygame.SRCALPHA)
rain_drop_surf.fill((40, 150, 170, 200), Rect(0, 5, 1, 1))
rain_drop_surf.fill((40, 150, 170, 175), Rect(0, 4, 1, 1))
rain_drop_surf.fill((40, 150, 170, 150), Rect(0, 3, 1, 1))
rain_drop_surf.fill((40, 150, 170, 100), Rect(0, 2, 1, 1))
rain_drop_surf.fill((40, 150, 170, 50), Rect(0, 1, 1, 1))
rain_drop_surf.fill((40, 100, 120, 50), Rect(0, 1, 1, 1))

splashes = Group()
drop_group = Group()
platform_group = Group()

Platform((700, 500), (150, 32), platform_group)
Platform((300, 700), (50, 32), platform_group)
Platform((100, 250), (100, 32), platform_group)
Platform((1200, 800), (150, 32), platform_group)
Platform((500, 900), (150, 32), platform_group)
Platform((900, 600), (150, 32), platform_group)

Platform((1400, 850), (120, 32), platform_group)
Platform((1600, 950), (175, 32), platform_group)
Platform((1750, 300), (100, 32), platform_group)

font = Font(size=48)

clock = pygame.time.Clock()
running = True

while running:

    time_delta = clock.tick()/1000.0

        if event.type == pygame.QUIT or (event.type == pygame.KEYDOWN and event.key == pygame.K_ESCAPE):
            running = False

    full_screen_surf.blit(background, (0, 0))

    for _ in range(int(random.uniform(0, 15))):
        RainDrop(rain_drop_surf, (int(random.uniform(0, 1920)), int(random.uniform(-100, -10))),
                 full_screen_surf.get_height(), drop_group, splash_anim, splashes)

    splashes.update(time_delta)
    drop_group.update(time_delta, platform_group)

    platform_group.draw(full_screen_surf)
    drop_group.draw(full_screen_surf)
    splashes.draw(full_screen_surf)

    pygame.transform.grayscale(full_screen_surf, greyscale_surf)

    # take 75% of greyscale & 25% of normal surf then add them together for a 75% desaturation effect, we could 
    # make this dynamic by filling the two surfaces based on some sliding variable between 0 and 255 
    greyscale_surf.blit(seventy_five_percent_surf, (0, 0), special_flags=pygame.BLEND_RGBA_MULT)
    full_screen_surf.blit(twenty_five_percent_surf, (0, 0), special_flags=pygame.BLEND_RGBA_MULT)
    display_surf.blit(full_screen_surf, (0, 0))
    display_surf.blit(greyscale_surf, (0, 0), special_flags=pygame.BLEND_RGBA_ADD)

    fps_text = font.render(f"{1.0/(max(time_delta, 0.0000000001)):.2f} FPS", True, Color("white"))
    display_surf.blit(fps_text, (desktop_size[0] - 200, 32))

    pygame.display.flip()


```
background.png:
background
splash.png (it's white so hard to see but should be click and saveable:
splash

If you stick them in an images/ subdirectory you should be able to run this program.

For me, I get 30ish FPS when running it without this PR on main. Then after deleting everything temporary and rebuilding build config and pip installing again (it's one of the pains with these PRs that add new .c files to Setup) I get approximately 60FPS (perhaps slightly below).

Seems like a pretty reasonable speedup in a believe-able use case.

Starbuck5:
@MyreMylar Now that the prereq PR is this ready to leave draft status?

MyreMylar:
I believe I still wanted to address some reviews that went up in the
previous attempt PR.

Will look that over tomorrow and get back to you.

MyreMylar:
OK, I've tried to address everything that was brought up in previous reviews now, so hopefully this is ready for fresh reviews if it passes on the CI tests.

Thanks for the patience with this,

src_c/simd_transform_avx2.c
```diff
@@ -42,3 +42,172 @@ pg_avx2_at_runtime_but_uncompiled()
...
+    // generate number of batches of pixels we need to loop through
+    int pixel_batch_length = src->w * src->h;
+    int num_batches = 1;
+    if (s_row_skip > 0) {
+        pixel_batch_length = src->w;
+        num_batches = src->h;
+    }
+
+    int remaining_pixels = pixel_batch_length % 8;
+    int perfect_8_pixels = (pixel_batch_length - remaining_pixels) / 8;
```
Starbuck5, line 80: This is how its done in other blitters as well, but it would also work to do:
`int perfect_8_pixels = pixel_batch_length / 8`, since it is using C integer division, which truncates.

src_c/simd_transform_sse2.c
```diff
@@ -34,3 +34,116 @@ pg_neon_at_runtime_but_uncompiled()
     }
     return 0;
 }
+
+#if (defined(__SSE2__) || defined(PG_ENABLE_ARM_NEON))
+void
+grayscale_sse2(SDL_Surface *src, SDL_Surface *newsurf)
+{
+    /* For the SSE2 SIMD version of grayscale we do one pixel at a time
```
Starbuck5, line 42: It seems like it could do two at a time?

src_c/simd_transform_avx2.c
```diff
@@ -42,3 +42,172 @@ pg_avx2_at_runtime_but_uncompiled()
...
+            mm256_dst = _mm256_adds_epu8(mm256_dst, mm256_alpha);
+
+            _mm256_maskstore_epi32((int *)dstp, _partial8_mask, mm256_dst);
+
+            srcp += remaining_pixels_batch_counter;
+            dstp += remaining_pixels_batch_counter;
+        }
+        srcp += s_row_skip;
+        srcp256 = (__m256i *)srcp;
+    }
```
Starbuck5, line 204: So you don't modify dstp / dstp256 here because they are assumed (and checked in dispatch) to be more continuous than the source surface?

src_c/simd_transform_avx2.c
```diff
@@ -42,3 +42,172 @@ pg_avx2_at_runtime_but_uncompiled()
...
+     *        for this operation in isolation.
+     *     4. pack pixels back together from A & B while adding with a
+     *        horizontal add (e.g. adds A+R and G+B in a ARGB layout)
+     *     5. shift and add to make final grey pixel colour in 0th
+     *        8Bit channel in each 'pixel'
+     *     6. shuffle again to push the grey from the 0th channel into every
+     *        channel of every pixel.
+     *     7. add the alpha channel back in.
+     */
+    int s_row_skip = (src->pitch - src->w * src->format->BytesPerPixel) >> 2;
```
Starbuck5, line 69: Might as well hardcode "4" instead of src->format->BytesPerPixel, since it will always be 4 in these SIMD things.

And would `/ 4` be clearer than `>> 2` about what it's doing? Compiler would likely optimize the first into the second anways.

src_c/transform.c
```diff
@@ -2059,24 +2089,26 @@ grayscale(pgSurfaceObject *srcobj, pgSurfaceObject *dstobj)
...
-            Uint32 new_pixel =
-                SDL_MapRGBA(newsurf->format, grayscale_pixel, grayscale_pixel,
-                            grayscale_pixel, a);
-            SURF_SET_AT(new_pixel, newsurf, x, y, (Uint8 *)newsurf->pixels,
-                        newsurf->format, pix);
+    if (src->format->BytesPerPixel == 4 &&
+        src->format->Rmask == newsurf->format->Rmask &&
+        src->format->Gmask == newsurf->format->Gmask &&
+        src->format->Bmask == newsurf->format->Bmask &&
+        (src->pitch % src->format->BytesPerPixel == 0) &&
```
Starbuck5, line 2096: Might as well hardcode "4" here for BytesPerPixel since it always be 4.

src_c/transform.c
```diff
@@ -2059,24 +2089,26 @@ grayscale(pgSurfaceObject *srcobj, pgSurfaceObject *dstobj)
...
+        (newsurf->pitch == (newsurf->w * newsurf->format->BytesPerPixel))) {
```
Starbuck5, line 2097: This is what I was talking about with the guarantee of "more continuous" destination Surface.

So you're thinking because it's a new surface it'll almost always be allocated at just the right size instead of passed in as a destination surface?

src_c/simd_transform_avx2.c
MyreMylar: True.

src_c/simd_transform_avx2.c
MyreMylar: Yes, we do this test in the initial dispatch:
```
newsurf->pitch == (newsurf->w * newsurf->format->BytesPerPixel)
```

Since the skip value is calculated as:
```
dst->pitch - info.width * dst->format->BytesPerPixel
```
if pitch is equal to width times bytes per pixel, there will be no skip.

I'm making this test because I believe that:

a) 98% of transformed surfaces are going to be newly created in the transform.
b) from what I've observed newly created 32 bit surfaces are always built with pitch equal to width * bytes per pixel - i.e. one big block of memory.
c) building it this way we can AVX2 SIMD churn at maximum speed until the last 0-7 pixels.

src_c/transform.c
MyreMylar: Exactly. We are testing here for any destination surfaces that would have a non-zero d_skip value. See how this value gets created in alphablit.c for example:

https://github.com/pygame-community/pygame-ce/blob/dfdf8b8b6b42f5705835cc859d8929f61ec3f44c/src_c/alphablit.c#L113

If they do have an unusual pitch then they fallback to the old ways of non-SIMD grayscale. My belief is that this is almost no 32bit surfaces in practice.

src_c/transform.c
MyreMylar: changed.

src_c/simd_transform_sse2.c
MyreMylar: Yes, it probably could, thinking about it - using `_mm_unpackhi_epi16`, `_mm_mulhi_epi16` and `_mm_shufflehi_epi16` for the other pixel, your `LOAD_64_INTO_M128()` & `STORE_M128_INTO_64()` macros and a copy of the current code for handling the last pixel in odd sized surfaces.

I'll have a poke at it.

MyreMylar: 2 pixels at a time turns out to be about 20% faster 👍

MyreMylar:
In my testing:

**SSE2/NEON old one-pixel method:** 6 times faster than non-SIMD greyscale.
**SSE2/NEON new two-pixel method:** 7 times faster than non-SIMD greyscale.
**AVX2 method:** 11 times faster than non-SIMD greyscale.

Used this test program:

<details>

```python
from sys import stdout
from pstats import Stats
from cProfile import Profile


import pygame

pygame.init()


def grayscale_test():
    dimensions = [
        (1, 1),
        (3, 3),
        (65, 65),
        (126, 127),
        (254, 255),
        (1024, 1024),
        (2047, 2048),
    ]
    for iterations in range(0, 1000):
        for w, h in dimensions:
            src_surf = pygame.Surface((w, h), pygame.SRCALPHA, 32)
            src_surf.fill((100, 255, 128, 200))

            result = pygame.transform.grayscale(src_surf)
            # res_color = result.get_at((0, 0))
            # if res_color == pygame.Color(195, 195, 195, 200):
            #     pass


if __name__ == "__main__":
    print("Grayscale")
    profiler = Profile()
    profiler.runcall(grayscale_test)
    stats = Stats(profiler, stream=stdout)
    stats.strip_dirs()
    stats.sort_stats("cumulative")
    stats.print_stats()

```

</details>

src_c/transform.c
Starbuck5: I mean for `src->format->BytesPerPixel`

MyreMylar: Oops, I changed a bunch of these and missed this one.

Starbuck5 (approved):
Pygame-ce w/ the SIMD be like 🚀 

This looks good!

src_c/simd_transform_avx2.c
```diff
@@ -42,3 +42,172 @@ pg_avx2_at_runtime_but_uncompiled()
...
+                _mm256_add_epi16(mm256_dst, _mm256_srli_epi32(mm256_dst, 16));
+            mm256_dst = _mm256_shuffle_epi8(mm256_dst, mm256_shuff_mask_gray);
+
+            mm256_dst = _mm256_subs_epu8(mm256_dst, mm256_alpha_mask);
+            mm256_dst = _mm256_adds_epu8(mm256_dst, mm256_alpha);
+
+            _mm256_storeu_si256(dstp256, mm256_dst);
+
+            srcp256++;
+            dstp256++;
```
itzpr3d4t0r, line 165: Just like in #2534 we could use bitwise logic instructions for the alpha parts. There are some conflicts to solve as well.

MyreMylar: Done.

src_c/transform.c
```diff
@@ -2086,24 +2116,25 @@ grayscale(pgSurfaceObject *srcobj, pgSurfaceObject *dstobj)
...
-            SDL_GetRGBA(pixel, src->format, &r, &g, &b, &a);
-
-            // RGBA to GRAY formula used by OpenCV
-            Uint8 grayscale_pixel = (Uint8)(0.299 * r + 0.587 * g + 0.114 * b);
-            Uint32 new_pixel =
-                SDL_MapRGBA(newsurf->format, grayscale_pixel, grayscale_pixel,
-                            grayscale_pixel, a);
-            SURF_SET_AT(new_pixel, newsurf, x, y, (Uint8 *)newsurf->pixels,
-                        newsurf->format, pix);
+    if (src->format->BytesPerPixel == 4 &&
```
Starbuck5, line 2119: Wait hold up does this need to be EMSCRIPTEN guarded as well? Like the blitters?

MyreMylar: Added 👍 May as well be safe.

src_c/simd_transform_sse2.c
```diff
@@ -43,6 +43,15 @@ pg_neon_at_runtime_but_uncompiled()
 #define _pg_storeu_si32(p, a) (void)(*(int *)(p) = _mm_cvtsi128_si32((a)))
 #define _pg_storeu_si64(p, a) (_mm_storel_epi64((__m128i *)(p), (a)))
 
+#if defined(ENV64BIT)
+#define LOAD_64_INTO_M128(num, reg) *reg = _mm_cvtsi64_si128(*num)
+#define STORE_M128_INTO_64(reg, num) *num = _mm_cvtsi128_si64(*reg)
+#else
+#define LOAD_64_INTO_M128(num, reg) \
+    *reg = _mm_loadl_epi64((const __m128i *)num)
+#define STORE_M128_INTO_64(reg, num) _mm_storel_epi64((__m128i *)num, *reg)
```
itzpr3d4t0r, line 52: Can't you pass the register directly here instead of as a pointer and dereference?

src_c/simd_transform_avx2.c
```diff
@@ -42,3 +42,172 @@ pg_avx2_at_runtime_but_uncompiled()
...
+            mm256_dstA = _mm256_srli_epi16(mm256_dstA, 8);
+
+            mm256_dstB = _mm256_mullo_epi16(mm256_srcB, mm256_dstB);
+            mm256_dstB = _mm256_add_epi16(mm256_dstB, mm256_two_five_fives);
+            mm256_dstB = _mm256_srli_epi16(mm256_dstB, 8);
+
+            mm256_dst = _mm256_hadd_epi16(mm256_dstA, mm256_dstB);
+            mm256_dst =
+                _mm256_add_epi16(mm256_dst, _mm256_srli_epi32(mm256_dst, 16));
+            mm256_dst = _mm256_shuffle_epi8(mm256_dst, mm256_shuff_mask_gray);
```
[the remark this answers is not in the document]
MyreMylar: This just allows us to add all 8 pixels G+B in one instruction and reunite the two 4 pixel registers into one 8 pixel one. So it is doing a fair bit of work.

hadd requires two registers (which we have here because we earlier split the calculation into two 4 pixel registers, then it adds A+R and G+B for all 8 pixels. We don't care about A+R but the G+B is useful and R is effectively left untouched in the 8 pixel data which sets us up for the next two instructions where we shift all 8 pixels down by 16 bits so we can add R to the combined G+B. I feel like a lot of shuffling and adding would probably be slower?

itzpr3d4t0r (approved):
LGTM!

src_c/simd_transform_avx2.c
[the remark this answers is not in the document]
itzpr3d4t0r: Got it. Thanks for the explaination. The wizard never fails!

PR #2437, opened 2023-09-07
Patches so constants.c compiles in SDL3

Starbuck5:
Zeroes out constants that have been removed in SDL3 by using a layer of macro indirection.

This is a very basic small step to start getting us compatible with SDL3. Small PRs are easier to review :)

After this patch, the "base", "color", and "constants" modules can compile.

If you'd like to replicate this, I attached a zipped windows SDL3 build, include, and SDL3-centric Setup file I'm using:
[SDL3.zip](https://github.com/pygame-community/pygame-ce/files/12546541/SDL3.zip)


dr0id (approved):

ankith26 (approved):
LGTM, thanks :tada:

PR #2438, opened 2023-09-07
Skipping mixer test because it randomly fails on Linux

Matiiss:
This test, especially recently, tends to fail randomly on Linux, skipping it for now until a solution is found.
image

### DIFF
--- a/test/mixer_test.py
+++ b/test/mixer_test.py
@@ -143,6 +143,10 @@ class MixerModuleTest(unittest.TestCase):
         self.assertRaises(pygame.error, mixer.get_num_channels)
 
     # TODO: FIXME: pypy (on linux) fails here sometimes.
+    @unittest.skipIf(
+        sys.maxsize <= 2**32,
+        "randomly fails on comparing bytes",
+    )
     @unittest.skipIf(IS_PYPY, "random errors here with pypy")
     def test_sound_args(self):
         def get_bytes(snd):


Starbuck5:
I don't really like skipping tests... I just wish this would consistently fail so it's easy to track down.

MyreMylar:
If it is never failing locally on linux - but it is failing on CI, and it is about the data from files loaded into memory - could it be something to do with multiple tests accessing the same file at the same time (as the CI appears to compile and run tests from multiple modules in parallel).

If so, one idea to try first might be making a second copy of `house_lo.wav` called `house_lo_sound_args_test.wav` and use that only in this test and see if it clears up occurrences of this error? Of course it would increase our file size slightly.

ankith26:
Another observation: it always seems to fail on 32-bit platforms. I think hunting for the root cause and fixing it wouldn't be worth the effort (we have very very few, or practically no real users on 32-bit linux)

I think the test skip can be added but conditional on the platform being 32 bit

test/mixer_test.py
```diff
@@ -143,6 +143,10 @@ def test_quit(self):
         self.assertRaises(pygame.error, mixer.get_num_channels)
 
     # TODO: FIXME: pypy (on linux) fails here sometimes.
+    @unittest.skipIf(
+        sys.platform.startswith("linux") and sys.maxsize <= 2**32,
+        "randomly fails on comparing bytes",
+    )
```
ankith26, line 149: I think we should make this a general 32-bit fail, I don't think it's linux specific (this error has happened on 32-bit windows before IIRC)

ankith26 (approved):

MyreMylar (approved):
LGTM, time will tell if it cracks this test fail.

PR #2441, opened 2023-09-08
Constants typehint show value

MightyJosip:
Really small PR.

Without:
image

With:
image


MightyJosip:
> These constants vary across platforms

Yea you are right, this would indeed be the problem if we don't generate those during the setup process, which we don't

PR #2444, opened 2023-09-10
Use RAISE when appropriate

Starbuck5:
Just a single pass, there are definitely other instances as well. These are just the easy ones.


dr0id (approved):
LGTM

ankith26 (approved):

PR #2459, opened 2023-09-20
Remove experimental notice on system module

Starbuck5:
We're going to treat this module with as much back-compat effort as anything else right now, so this notice is moot.


yunline (approved):
LGTM

ankith26 (commented):
I'd like `get_power_state` to be in experimental state just in case something comes up (it hasn't seen a release yet)

I agree the rest should be leaving experimental status

Starbuck5:
As requested, I added a more specific experimental notice there.

MyreMylar (approved):
LGTM

PR #2460, opened 2023-09-20
Remove conan buildconfig

Starbuck5:
I'm not sure this was ever used, but it was developed several years ago and hasn't been used since.


ankith26 (approved):
I remember having an older PR with deps updates that did these cleanups, but I like that you opened a dedicated PR as it can get merged more easily.

PS: You are deleting a file in the `travis` folder, IMO the whole folder can be deleted. The `appveyor` folder too. We don't really use it, and even if we need to at some point, then we always have git history ;)

Anyways you have my approval. You can take the call on doing the other cleanups in this PR

Starbuck5:
I'll stick with just conan for now, keep it small. :)

pmp-p (approved):

PR #2461, opened 2023-09-20
Fix linter failing due to new clang-format version

Matiiss:
So `clang-format` got updated and now the linter is failing because it wants to format some stuff, well, this PR formats that stuff.


yunline (approved):
LGTM

Starbuck5 (approved):
Thanks!

PR #2464, opened 2023-09-22
Raise minimum deps of SDL and friends

Starbuck5:
To correspond to Debian buster. Because Debian buster is on 3.7, so it's our lowest popular Linux distro to support. And it supports these SDL and SDL_library versions.


MyreMylar (approved):
All looks good to me 👍

ankith26 (approved):

PR #2466, opened 2023-09-24
Fix uint64 underflow in `tick_busy_loop`

oddbookworm:
It was reported on discord that using `tick_busy_loop` could cause a crash in a specific scenario.

Reproducer
```python
import pygame


WIDTH, HEIGHT = 640, 360
FPS = 60

pygame.init()
clock = pygame.Clock()

running = True
while running:
    dt = clock.tick_busy_loop(FPS) / 1000
    screen.fill("black")
    
    events = pygame.event.get()
    for event in events:
        if event.type == pygame.QUIT:
            running = False

    pygame.display.flip()
```
It's an odd one though, I didn't see it on my own machine. After some troubleshooting with @Matiiss and the reporter, I believe it is due to an int underflow when there is a lag spike in cpu processing, causing `delay` to become insanely large (thus blocking the `while` loop from ever finishing).

I don't think a regression test is possible with our current framework

### DIFF
--- a/src_c/time.c
+++ b/src_c/time.c
@@ -27,7 +27,7 @@
 #include "doc/time_doc.h"
 
 #define WORST_CLOCK_ACCURACY 12
-#define MAX_UINT_64 18446744073709551615
+#define MAX_UINT_64 18446744073709551615ULL
 
 /* Enum containing some error codes used by timer related functions */
 typedef enum {
--- a/src_c/time.c
+++ b/src_c/time.c
@@ -27,6 +27,7 @@
 #include "doc/time_doc.h"
 
 #define WORST_CLOCK_ACCURACY 12
+#define MAX_UINT_64 18446744073709551615
 
 /* Enum containing some error codes used by timer related functions */
 typedef enum {
@@ -330,7 +331,7 @@ accurate_delay(Sint64 ticks)
     }
     do {
         delay = ticks - (PG_GetTicks() - funcstart);
-    } while (delay > 0);
+    } while (delay > 0 && delay < (MAX_UINT_64 >> 1));
 
     return PG_GetTicks() - funcstart;
 }
--- a/src_c/time.c
+++ b/src_c/time.c
@@ -27,7 +27,6 @@
 #include "doc/time_doc.h"
 
 #define WORST_CLOCK_ACCURACY 12
-#define MAX_UINT_64 18446744073709551615ULL
 
 /* Enum containing some error codes used by timer related functions */
 typedef enum {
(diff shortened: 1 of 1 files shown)


src_c/time.c
```diff
@@ -27,6 +27,7 @@
 #include "doc/time_doc.h"
 
 #define WORST_CLOCK_ACCURACY 12
+#define MAX_UINT_64 18446744073709551615ULL
```
MyreMylar, line 30: I don't think it makes a difference for this check but it seem like `UInt64` might be an `unsigned long` or an `unsigned long long` depending on the platform.

SDL uses this to define the max value:

```C
#define SDL_MAX_UINT64  ((Uint64)0xFFFFFFFFFFFFFFFFull)     /* 18446744073709551615 */
``` 

So maybe this has caused odd bugs in the past?

src_c/time.c
```diff
@@ -330,7 +331,7 @@ accurate_delay(Sint64 ticks)
     }
     do {
         delay = ticks - (PG_GetTicks() - funcstart);
-    } while (delay > 0);
+    } while (delay > 0 && delay < (MAX_UINT_64 >> 1));
```
MyreMylar, line 334: Perhaps you can remove this bitshift if you define `MAX_UINT_64` in the SDL way?

Matiiss: As far as I understand, the point of the bitshift is to essentially halve the value so that if it underflows and goes back to top it fails to meet the condition that is `delay < bitshifted_value` and exits the loop.

oddbookworm: Yeah, the reason for the bitshift is because I don't want to compare directly to `MAX_UINT_64` for the same reason this PR exists in the first place. I chose a bitshift over a smaller buffer because of 2 reasons:

1) the whole reason this PR exists is an underflow due to a lag spike in CPU processing. I tried buffers of up to 100 milliseconds in testing, but even that was unreliable. Bitshifting down by one bit is a halving of the maximum, but the maximum is going to be hundreds of millions of years anyway, so I doubt there will be a delay that long for any possible reason.

2) bitshifting is a fast operation iirc

src_c/time.c
oddbookworm: I can remove my custom define and use SDL's, but they're equivalent

src_c/time.c
MyreMylar: maybe we should just make this clear with a variable then? So,mething like

```C
unit64 max_delay = (MAX_UINT_64 >> 1); // a large number that should never be reached by delay - to handle underflow subtractions
```

MyreMylar (approved):
LGTM 👍

Matiiss (approved):
LGTM, the window doesn't hang anymore

ankith26 (approved):
LGTM, thanks :tada:

PR #2470, opened 2023-09-26
Bump versions of many manylinux deps

ankith26:
Should do a decent amount bugfixes and enhancements.

Also fixes that manylinux python 3.7 fail happening due to a new `urllib3` version being incompatible with older SSL versions (`urllib3` has been downgraded for now)


.circleci/config.yml
```diff
@@ -34,7 +34,7 @@ jobs:
       - CIBW_SKIP: '*-musllinux_*'
       - CIBW_MANYLINUX_AARCH64_IMAGE: manylinux2014_base_aarch64
       - CIBW_MANYLINUX_PYPY_AARCH64_IMAGE: manylinux2014_base_aarch64
-      - CIBW_BEFORE_BUILD: pip install Sphinx"<7.2.0" && python setup.py docs
+      - CIBW_BEFORE_BUILD: pip install Sphinx"<7.2.0" urllib3"<2.0.0" && python setup.py docs
```
Starbuck5, line 37: Isn't the urllib3 thing only needed on circleci, since they have the outdated image? Why is it also in build-manylinux?

Starbuck5 (approved):
Left one small comment.

Otherwise looks great! Standard stuff.

.circleci/config.yml
ankith26: The problem isn't an outdated image as I initially misunderstood.

The problem is the newer images have an intentional change, which leads to a downgrade of OpenSSL specific to Python 3.7. It only fails on circleci and not github actions is because the github actions script doesn't refresh the cache on base image changes (it only refreshes on changes triggered in our side of things)

Starbuck5: Ah okay, so it'll start failing on ours eventually.

ankith26: Pushed a commit with comments for it

MyreMylar (approved):
LGTM 👍

PR #2471, opened 2023-09-27
Make mpg123, fluidsynth and sndfile not link directly to lower-level audio playback libs

ankith26:
Depends on #2470 (it should be merged before this)

EDITED description to what this PR does now (taking starbucks suggestions):

Previously, `mpg123`, `fluidsynth` and `sndfile` would link directly to lower-level libraries `alsa`, `pulseaudio` and others and all these would get included in the wheels. But these links aren't required for audio playback

As a side effect of these changes, our manylinux wheels should get a bit lighter (maybe 1-2 mb)

To test this PR, we need to make sure that fluidsynth-based music and mp3 music still work as before on linux (which I can test) and mac (for which I need a volunteers help)

### DIFF
--- a/buildconfig/manylinux-build/docker_base/fluidsynth/build-fluidsynth.sh
+++ b/buildconfig/manylinux-build/docker_base/fluidsynth/build-fluidsynth.sh
@@ -14,12 +14,17 @@ cd $FSYNTH
 mkdir build
 cd build
 
-if [[ "$OSTYPE" == "darwin"* ]]; then
+if [[ "$OSTYPE" == "linux-gnu"* ]]; then
+    export FLUIDSYNTH_EXTRA_PLAT_FLAGS="-Denable-alsa=NO -Denable-systemd=NO"
+elif [[ "$OSTYPE" == "darwin"* ]]; then
     # We don't need fluidsynth framework on mac builds
-    export FLUIDSYNTH_EXTRA_MAC_FLAGS="-Denable-framework=NO"
+    export FLUIDSYNTH_EXTRA_PLAT_FLAGS="-Denable-framework=NO"
 fi
 
-cmake .. $PG_BASE_CMAKE_FLAGS -Denable-readline=OFF $FLUIDSYNTH_EXTRA_MAC_FLAGS
+cmake .. $PG_BASE_CMAKE_FLAGS -Denable-readline=OFF $FLUIDSYNTH_EXTRA_PLAT_FLAGS \
+    -Denable-pulseaudio=NO \
+    -Denable-pipewire=NO
+
 make
 make install
 
--- a/buildconfig/manylinux-build/docker_base/mpg123/build-mpg123.sh
+++ b/buildconfig/manylinux-build/docker_base/mpg123/build-mpg123.sh
@@ -12,7 +12,8 @@ bzip2 -d ${MPG123}.tar.bz2
 tar xf ${MPG123}.tar
 cd $MPG123
 
-./configure $ARCHS_CONFIG_FLAG --enable-int-quality --disable-debug
+./configure $ARCHS_CONFIG_FLAG --enable-int-quality --disable-debug \
+    --disable-components --enable-libmpg123 --enable-libsyn123
 make
 make install
 
--- a/buildconfig/manylinux-build/docker_base/sndfile/build-sndfile.sh
+++ b/buildconfig/manylinux-build/docker_base/sndfile/build-sndfile.sh
@@ -13,7 +13,8 @@ tar xf ${SNDFILE}
 cd $SNDNAME
 # autoreconf -fvi
 
-./configure $ARCHS_CONFIG_FLAG --disable-mpeg
+# alsa is only needed for examples here
+./configure $ARCHS_CONFIG_FLAG --disable-mpeg --disable-alsa
 make
 make install
 


ankith26:
```diff
 libasound-e2127b4b.so.2.0.0
-libattr-4f2a9577.so.1.1.0
 libbrotlicommon-c828db13.so.1.1.0
 libbrotlidec-3b276a64.so.1.1.0
-libbz2-a273e504.so.1.0.6
-libcap-79733b59.so.2.22
-libdbus-1-5d678da9.so.3.14.14
-libdw-0-780627ce.176.so
-libelf-0-3d9de92a.176.so
-libFLAC-9c16284b.so.12.1.0
+libFLAC-11b10a77.so.12.1.0
-libfluidsynth-78c3029a.so.3.2.2
+libfluidsynth-13e12a66.so.3.2.2
 libfreetype-ce7af30f.so.6.20.1
-libgcrypt-18957bae.so.11.8.2
 libgomp-a34b3233.so.1.0.0
-libgpg-error-3f09c3c7.so.0.10.0
 libharfbuzz-b5bd39f9.so.0.60821.0
-libjpeg-1944207f.so.62.4.0
-liblz4-af1653fb.so.1.8.3
-liblzma-004595ca.so.5.2.2
+libjpeg-034e96d7.so.62.4.0
 libmodplug-8c1b3b53.so.1.0.0
 libmpg123-7a05540a.so.0.48.0
 libogg-3f4a1572.so.0.8.5
 libopus-4c85b165.so.0.9.0
 libopusfile-df4005ff.so.0.4.5
-libpcre-9513aab5.so.1.2.0
 libpng16-92b73642.so.16.40.0
 libportmidi-bc2a3536.so.2.0.3
-libpulse-5b5d1e90.so.0.23.0
-libpulsecommon-14-9b36c523.2.so
-libpulse-simple-4ba3beb5.so.0.1.1
 libSDL2-2-cdcf49c3.0.so.0.2600.5
 libSDL2_image-2-d19463c5.0.so.0.2.3
 libSDL2_mixer-2-705164cf.0.so.0.600.3
 libSDL2_ttf-2-e58b8489.0.so.0.2000.2
-libselinux-0922c95c.so.1
 libsharpyuv-5ed6246f.so.0.0.1
 libsndfile-bfa200bd.so.1.0.37
-libsystemd-9be61d21.so.0.6.0
 libtiff-61e60d59.so.6.0.2
 libvorbis-ac399090.so.0.4.9
 libvorbisenc-bb396c4a.so.2.0.12
 libvorbisfile-8aced0b7.so.3.3.8
 libwebp-3bee5bfa.so.7.1.8
 ```
```

MyreMylar (approved):
LGTM 👍

oddbookworm:
Tested on mac
flac, ogg, mp3, and tests ran fine
ogx did not load
midi loaded as long as I specified a soundfont with SDL_SOUNDFONT, but crashed when I tried to restart the sound from the beginning
(see my reply to @ankith26 in the discord contributing channel for error logs)

buildconfig/manylinux-build/docker_base/mpg123/build-mpg123.sh
```diff
@@ -12,7 +12,12 @@ bzip2 -d ${MPG123}.tar.bz2
 tar xf ${MPG123}.tar
 cd $MPG123
 
-./configure $ARCHS_CONFIG_FLAG --enable-int-quality --disable-debug
+if [[ "$OSTYPE" == "linux-gnu"* ]]; then
+    # use sdl2, and only sdl2 backend
+    export MPG123_EXTRA_ARGS="--with-audio=sdl --with-default-audio=sdl"
```
Starbuck5, line 17: MPG123 shouldn't be in charge of playing audio at all, and it does not do that in SDL_mixer.

The backend is for the library libout123 in libmpg123: https://www.mpg123.de/features.shtml#libout123. We should turn off libout123 altogether rather than just making it less disruptive.

According to the libmpg123 `configure` file this can be done by passing `--enable-libmpg123` "build (only) libmpg123"

SDL_Mixer's cmake config does this in a Cmake way: https://github.com/libsdl-org/SDL_mixer/blob/f8fabfcfaf4914da2c7ee4696ed39fa9db8f5ea9/CMakeLists.txt#L884

buildconfig/manylinux-build/docker_base/fluidsynth/build-fluidsynth.sh
```diff
@@ -14,12 +14,18 @@ cd $FSYNTH
...
+    export FLUIDSYNTH_EXTRA_PLAT_FLAGS="-Denable-alsa=NO -Denable-systemd=NO"
+elif [[ "$OSTYPE" == "darwin"* ]]; then
     # We don't need fluidsynth framework on mac builds
-    export FLUIDSYNTH_EXTRA_MAC_FLAGS="-Denable-framework=NO"
+    export FLUIDSYNTH_EXTRA_PLAT_FLAGS="-Denable-framework=NO"
 fi
 
-cmake .. $PG_BASE_CMAKE_FLAGS -Denable-readline=OFF $FLUIDSYNTH_EXTRA_MAC_FLAGS
+# pick the sdl2 backend, and only the sdl2 backend
+cmake .. $PG_BASE_CMAKE_FLAGS -Denable-readline=OFF $FLUIDSYNTH_EXTRA_PLAT_FLAGS \
```
Starbuck5, line 25: Like mpg123, SDL_mixer doesn't use this libraries audio playback capabilities, it uses its decoding / synthesizing capabilities and then passes that to SDL to get it onto speakers.

Therefore there is no need to compile fluidsynth with the ability to output sound at all, including through SDL, since SDL_mixer already will handle that. I think `-Denable=sdl2=NO` should also be added here.

Starbuck5:
> ogx did not load

Is this okay?

@ankith26 I believe all the audio backends of mpg123 and fluidsynth should be turned off, because they're not relevant to SDL_mixer. This also means the SDL + libs build order doesn't need to be rearranged, so this PR will get simpler and shave even more off of dependency size?

buildconfig/manylinux-build/docker_base/mpg123/build-mpg123.sh
ankith26: Yeah good catch, I will do the changes

ankith26:
Regarding flac-in-ogg support, I realised that `SDL_mixer` didn't support these on their end, so I have opened a PR there, lets see how it goes. If they do merge that PR, then the fix in this PR will start to work.

I will take starbucks suggestions to simplify the diffs in this PR, and move this flac fix stuff out too in the process

Starbuck5 (approved):
Looks good to me.

I really like how this turned out, better buildconfig -> smaller wheels.

I just checked and it looks like this takes the Linux wheels down 1.4 mb each
Much less significant effect on the Mac wheels, only down about 2 kb each

PR #2474, opened 2023-09-30
Up cibuildwheel

Starbuck5:
Gets us to 3.12b3

(Yes, beta 3, they added another beta)


ankith26 (commented):
Also needs a patch bump in the mac and manylinux files

Starbuck5:
Those are handled by dependabot. If I did those I would hurt its feelings.

ankith26 (approved):
Let's get this in then

Issue #1741, opened 2023-02-12
VideoInfo.pixel_format: remove SDL_ prefix (3526)


GalacticEmperor1:
# **Issue &numero;3526 opened by *[illume](https://github.com/illume)* at 2022-10-29 16:12:55**

For https://github.com/pygame/pygame/pull/2964

Probably would be good to remove the `SDL_` prefix to remain consistent.<hr>

# Comments

PR #2477, opened 2023-09-30
Strip the leading SDL_ from the pixel format name

MyreMylar:
Fixes issue #1741.

I'm not sure this is the best way to strip 4 characters from a C string, so please let me know.

### DIFF
--- a/buildconfig/stubs/pygame/display.pyi
+++ b/buildconfig/stubs/pygame/display.pyi
@@ -29,6 +29,7 @@ class _VidInfo:
     blit_sw_A: int
     current_h: int
     current_w: int
+    pixel_format: str
 
 def init() -> None: ...
 def quit() -> None: ...
--- a/docs/reST/ref/display.rst
+++ b/docs/reST/ref/display.rst
@@ -311,12 +311,12 @@ required).
    mode to verify specific display options were satisfied. The VidInfo object
    has several attributes:
 
-   ::
+   .. code-block:: text
 
      hw:         1 if the display is hardware accelerated
      wm:         1 if windowed display modes can be used
-     video_mem:  The megabytes of video memory on the display. This is 0 if
-                 unknown
+     video_mem:  The megabytes of video memory on the display.
+                 This is 0 if unknown
      bitsize:    Number of bits used to store each pixel
      bytesize:   Number of bytes used to store each pixel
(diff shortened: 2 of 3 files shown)


ankith26 (changes requested):
A one-line change to achieve this could be passing `pixel_format_name + 4` directly instead of the copying to new memory. This of course, assumes that `pixel_format_name` is a non-null pointer and the string has atleast 4 chars (which should be checked, either way)

ankith26:
```c
if (!strncmp(pixel_format_name, "SDL_", 4)) {
    pixel_format_name += 4
}
```
(untested code)
This should do maybe?

zoldalma999:
The vidinfo type also has a getter, defined a bit above the str function. This currently does not even return a value for pixel format, so I think this could be an opportunity to fix that (and of course, strip "SDL_" off there as well).

ankith26:
> ps: [SDL_GetPixelFormatName will always return a string starting with "SDL_"](https://github.com/libsdl-org/SDL/blob/SDL2/src/video/SDL_pixels.c#L88C13-L88C35)

Yeah I realised that too as I made my comment, but then it's safer (more future-proof) this way and does no harm.

MyreMylar:
> The vidinfo type also has a getter, defined a bit above the str function. This currently does not even return a value for pixel format, so I think this could be an opportunity to fix that (and of course, strip "SDL_" off there as well).

Should work now:

```
import pygame

pygame.init()
print(pygame.display.Info().pixel_format)
```

Output:
```
pygame-ce 2.4.0.dev1 (SDL 2.26.5, Python 3.11.1)
PIXELFORMAT_RGB888
```

zoldalma999 (approved):
Looks good to me!

buildconfig/stubs/pygame/display.pyi
```diff
@@ -29,6 +29,7 @@ class _VidInfo:
     blit_sw_A: int
     current_h: int
     current_w: int
+    pixel_format: str
```
ankith26, line 32: This change needs to be documented too

MyreMylar:
Documented the `pixel_format` attribute and fixed that attribute documentation block while I was in there. Before it was formatted as python code and had an unnecessarily messy horizontal scrollbar.

👍

ankith26 (approved):

PR #2479, opened 2023-09-30
Remove old CI files

zoldalma999:
We don't use appveyor or tarvis anymore, so these files should not be needed


ankith26 (approved):
Yay more nice cleanups, thanks! :green_apple:

MyreMylar (approved):
LGTM 👍

PR #2481, opened 2023-09-30
Add PG_CreateSurface(From) for SDL2/3 compat

Starbuck5:
This was the end goal of my previous PRs switching out SDL_CreateRGBSurface for SDL_CreateRGBSurfaceWithFormat. This allows us to write our create surface calls in "SDL3 form" on both on SDL2 and SDL3, because of the macro.

You might be wondering about the depth argument always being 0. This is okay. In SDL2's SDL_CreateRGBSurfaceWithFormat, the depth argument is ignored. (https://github.com/libsdl-org/SDL/blob/f9dc49c21c99cd6eb63d80d60b6e706a51234b46/src/video/SDL_surface.c#L72-L78). This makes sense, as the depth info is in the pixelformat.

I like this patch as well because it actually simplifies the calling code (at least makes it shorter), as you can see with the line count reduction in this PR.

There are still a small few SDL_CreateRGBSurface calls in the wilds of the codebase, they'll get handled eventually.

### DIFF
--- a/src_c/_camera.c
+++ b/src_c/_camera.c
@@ -109,9 +109,7 @@ surf_colorspace(PyObject *self, PyObject *arg)
     surf = pgSurface_AsSurface(surfobj);
 
     if (!surfobj2) {
-        newsurf = SDL_CreateRGBSurfaceWithFormat(0, surf->w, surf->h,
-                                                 surf->format->BitsPerPixel,
-                                                 surf->format->format);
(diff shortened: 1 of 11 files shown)


MyreMylar (approved):
Yes, this looks a lot neater, nice refactoring. 👍

ankith26 (approved):

PR #2483, opened 2023-10-01
Port SDL_ShowCursor to SDL3

Starbuck5:


MyreMylar (approved):
LGTM 👍

src_c/mouse.c
```diff
@@ -193,8 +192,18 @@ mouse_set_visible(PyObject *self, PyObject *args)
         }
     }
 
-    toggle = SDL_ShowCursor(toggle);
-    return PyBool_FromLong(toggle);
+    prevstate = PG_CursorVisible();
+
+    // Cursor visibility API also can indicate errors, but we historically
+    // haven't raised those as pygame.errors. Should that change?
```
ankith26, line 198: Went through both SDL2 and SDL3 source code, and no function actually errors and returns `-1`. Maybe this comment should be along the lines of "according to SDL docs these functions can error with a negative return, but in practice it doesn't happen as of now"?

Starbuck5: Comment updated.

ankith26 (approved):

Issue #493, opened 2023-02-12
surface.convert_alpha(surf) (with an argument) does not do anything differently? (599)


GalacticEmperor1:
#  **Issue &numero;599 opened by *[dlon](https://github.com/dlon)* at 2018-11-04 11:40:14**

It looks like `src` isn't used and it does the same thing as if you don't pass in a surface at all:

```c
    pgSurface_Prep(self);
    if (srcsurf) {
        /*
         * hmm, we have to figure this out, not all depths have good
         * support for alpha
         */
        src = pgSurface_AsSurface(srcsurf);
        newsurf = SDL_DisplayFormatAlpha(surf);
    }
    else
        newsurf = SDL_DisplayFormatAlpha(surf);
    pgSurface_Unprep(self);
```

https://github.com/pygame/pygame/blob/b9f8f97fc06c1285e453099bd186b38c3fab8796/src_c/surface.c# L1780-L1816

**To Do to close:**

- [x] Discuss if there are any good use cases for this doing anything. Either here or discord.
- [x] If it's confirmed that it doesn't do anything, and we don't want it to do anything, we should remove the option from the docs (and probably the .pyi stubs) to move people away from using it -without breaking compatibility.

**Related Docs:** https://www.pygame.org/docs/ref/surface.html# pygame.Surface.convert_alpha<hr>

#  Comments

# # #  **[MyreMylar](https://github.com/MyreMylar)* commented at 2020-05-16 10:51:08*

Moved 'to do' to first post.
<hr>

# # #  **[objarni](https://github.com/objarni)* commented at 2020-10-26 10:57:28*

Noticed convert() does the same thing as convert_alpha() on Mac machines, but does the (expected) removal of alpha-channel on Windows machines (as per documentation here: https://www.pygame.org/docs/ref/surface.html# pygame.Surface.convert ):

    The converted Surface will have no pixel alphas. They will be stripped if the
    original had them. See convert_alpha() for preserving or creating per-pixel
    alphas.

PyGame version: pygame-2.0.0.dev12

<hr>

# # #  **[illume](https://github.com/illume)* commented at 2020-10-26 12:11:43*

@objarni we only just recently noticed that on Mac and fixed it. On some Mac the blending was on by default. I think in pygame==2.0.0.dev22 or so. We did a work-around, but maybe it's actually something we should report to SDL.

[Ed: we do that for surface creation... maybe convert_alpha is still broken]
<hr>

[...]

PR #2486, opened 2023-10-01
Remove the unused Surface param from convert alpha docs

### LINKED ISSUE #1753: docs: document `Surface.convert_alpha()` silently ignoring argument for now (maybe also issue warnings for arguments passed?) (3557)
#  **Issue &numero;3557 opened by *[francis-clairicia](https://github.com/francis-clairicia)* at 2022-11-12 11:12:23**

Currently, `Surface.convert_alpha()` always uses the display surface, ignoring the one as parameter.

https://github.com/pygame/pygame/blob/0164760b2a3b603fab111f2dc95d170d95783db4/src_c/surface.c# L1663

And the documentation does not mention this.

It would be nice to inform the users that giving this setting does not change anything (for now or forever ?), or something else.<hr>

#  Comments

# # #  **[Matiiss](https://github.com/Matiiss)* commented at 2022-11-13 11:03:01*

It looks like they need to figure it, so I guess, it's kind of temporary until it gets fixed I suppose. Also, doesn't that warning already inform users?
<hr>

# # #  **[francis-clairicia](https://github.com/francis-clairicia)* commented at 2022-11-13 12:13:58*

Nope, the argument is just silently ignored at runtime
<hr>

# # #  **[ankith26](https://github.com/ankith26)* commented at 2022-11-14 11:17:47*

This is a known issue (duplicate of # 599).

> And the documentation does not mention this.

Good call, this should be documented for now, because this issue is tricky and may take a while to properly address. I will just re-purpose this issue as a docs issue, since the other issue is already tracking the underlying issue

MyreMylar:
It does nothing at all.

fixes #1753


zoldalma999:
Apart from changing the docs, we could also remove the stubs for it and deprecate it in code. That would also close #493. Unless we want to add this functionality back?

MyreMylar:
> Apart from changing the docs, we could also remove the stubs for it and deprecate it in code. That would also close #493. Unless we want to add this functionality back?

Pretty sure we don't want this back.

I've added the deprecation warnings.

Temmie3754 (approved):
No longer appears in docs, stubs and appropriate depreciation warning displays

docs/reST/ref/surface.rst
```diff
@@ -240,17 +240,15 @@
...
+      with per pixel alpha.
 
       Unlike the :meth:`convert()` method, the pixel format for the new
-      image will not be exactly the same as the requested source, but it will
-      be optimized for fast alpha blitting to the destination.
+      image will not be exactly the same as the display surface, but it will
+      be optimized for fast alpha blitting to it.
 
       As with :meth:`convert()` the returned surface has the same class as
       the converted surface.
```
ankith26, line 254: A note here of the deprecation would be nice

ankith26 (changes requested):
Also a small and simple test that tests the deprecation warning would also be nice

docs/reST/ref/surface.rst
MyreMylar: 👍 added a `versionchanged` note.

MyreMylar:
> Also a small and simple test that tests the deprecation warning would also be nice

Added a test for the deprecation warning, and removed a few places in the test that were now triggering it as they were passing in the surface argument (which has done nothing for years).

test/surface_test.py
```diff
@@ -11,6 +11,7 @@
 except (ImportError, NameError):
     pass
 import pygame
+import pygame
```
zoldalma999, line 14: this seems accidental

zoldalma999 (approved):
Apart from a minor nitpick, this looks good to me

test/surface_test.py
MyreMylar, line 14:
```suggestion
```

MyreMylar: oops, thanks 👍 

Fixed.

ankith26 (approved):
LGTM now, thanks for the PR :tada:

PR #2487, opened 2023-10-02
Remove redundant check in pg_ResizeEventWatch

Starbuck5:
Two locations were checking if `window == pygame_window`, which isn't necessary, because on line 707 (line 694 above this diff for some reason) the function returns early if `window != pygame_window`

Diff looks complex because code got dedented one level after the redundant check was removed.


ankith26 (approved):
LGTM thanks for the PR :tada:

zoldalma999 (approved):
LGTM, nice cleanups

PR #2490, opened 2023-10-02
Deprecate set_mode depth argument

### LINKED ISSUE #635: SDL2: display.set_mode() ignores depth value (991)
#  **Issue &numero;991 opened by *[charlesej](https://github.com/charlesej)* at 2019-04-17 14:58:12**

The depth value is ignored when setting the display mode for pygame SDL2 builds. The depth can be set correctly for pygame SDL1.2 builds.

Code example to show the depth value being ignored.
```python
import pygame

pygame.display.init()

for depth in (8, 16, 24, 32):
    surface = pygame.display.set_mode((200, 200), 0, depth)
    print('set depth = {}, get depth = {}'.format(
        depth, surface.get_bitsize()))

pygame.display.quit()
```

Output from the above code.
```
pygame 2.0.0.dev0 (SDL 2.0.9, python 3.7.2)
set depth = 8, get depth = 32
set depth = 16, get depth = 32
set depth = 24, get depth = 32
set depth = 32, get depth = 32
```

System details:
- os: windows 10 (64bit)
- python: 3.7.2 (64bit)
- pygame: 2.0.0.dev0 (SDL: 2.0.9) at 0f051ad9423350a215a43ac72f1c98581f22579e

**Related Docs:** https://www.pygame.org/docs/ref/display.html# pygame.display.set_mode<hr>

#  Comments

# # #  **[robertpfeiffer](https://github.com/robertpfeiffer)* commented at 2019-10-29 06:31:41*

Old Pygame also ignored some flags and depth parameters. Some were more like *requests*, and you had to check whether you actually got the right kind of display surface.
<hr>

# # #  **[illume](https://github.com/illume)* commented at 2019-10-29 06:49:10*

Yeah, depth is a request. However, SDL2 definitely supports more than this depth.

I can reproduce it on MacOS 10.13.6
```
$ python display_bpp.py 
pygame 2.0.0.dev5 (SDL 2.0.10, python 2.7.16)
set depth = 8, get depth = 32
set depth = 16, get depth = 32
set depth = 24, get depth = 32
set depth = 32, get depth = 32
```
<hr>

# # #  **[robertpfeiffer](https://github.com/robertpfeiffer)* commented at 2019-10-31 10:11:26*

I can implement this to work as expected, but I'd think most of the time you actually don't want this. I know old PGS4A had always used 16bpp depth display surfaces because the of GLES 1.0 it used under the hood, but barring hardware restrictions, there is no good reason to use an 8-bit indexed display surface with SDL2. And there is certainly no good reason to use a 16-bit surface.

The only reason to support this is backward compatibility.
<hr>

# # #  **[ThanhChinhBK](https://github.com/ThanhChinhBK)* commented at 2022-10-20 03:51:40*

I want to investigate on this issue, is it OK?
<hr>

# # #  **[ankith26](https://github.com/ankith26)* commented at 2022-10-21 09:07:08*

Feel free to go ahead, but I suspect it won't be easy (considering SDL API doesn't expose this anymore I think).
If this needs elaborate and complex changes that significantly increases the maintenance burden, it may potentially not be worth it. That being said if the change ends up being small, it would be a very much welcome and appreciated addition!

### LINKED ISSUE #2329: Update the docs of the `depth` argument of `display.set_mode()`
Since we don't really use the `depth` argument in our C code, the docs here seems to be out dated.

image

zoldalma999:
Fixes #635, fixes #2329 

I don't think we want to support setting the depth of the window surface

### DIFF
--- a/docs/reST/ref/display.rst
+++ b/docs/reST/ref/display.rst
@@ -134,10 +134,9 @@ required).
    are set to ``0``, the Surface will have the same width or height as the
    screen resolution.
 
-   It is usually best to not pass the depth argument. It will default to the
-   best and fastest color depth for the system. If your game requires a
-   specific color format you can control the depth with this argument. Pygame
-   will emulate an unavailable color depth which can be slow.
+   Since pygame 2, the depth argument is ignored, in favour of the best
+   and fastest one. It also raises a deprecation warning since pygame-ce
+   2.4.0 if the passed in depth is not 0 or the one pygame selects.
 
    When requesting fullscreen display modes, sometimes an exact match for the
    requested size cannot be made. In these situations pygame will select
@@ -212,6 +211,8 @@ required).
    .. versionchanged:: 2.2.0 explicit request for "adaptive vsync"
 
    .. versionchanged:: 2.2.0 ``vsync=1`` does not require ``SCALED`` or ``OPENGL``
+
+   .. deprecated:: 2.4.0 The depth argument is ignored, and will be set to the optimal value
 
 
    Basic example:
--- a/src_c/display.c
+++ b/src_c/display.c
@@ -999,7 +999,6 @@ pg_set_mode(PyObject *self, PyObject *arg, PyObject *kwds)
                 SDL_GL_SetAttribute(SDL_GL_DOUBLEBUFFER, 0);
         }
 
-#pragma PG_WARN(Not setting bpp ?)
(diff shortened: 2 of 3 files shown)


ankith26 (approved):
I think this is indeed the way forward, thanks for the PR :balloon:

MyreMylar (approved):
LGTM 👍

PR #2492, opened 2023-10-02
Fix issue #2109: Added requested tests

### LINKED ISSUE #2109: Improving `Surface.get_rect` and `Surface.get_frect` tests
Both the tests for `pygame.Surface.get_rect` and `pygame.Surface.get_frect` lack a couple of tests for some of the main features of these methods. Using keyword arguments to set rect values. As well as checking whether appropriate exceptions are raised.

For both methods test:
- [ ] whether using correct keyword argument(s) sets the appropriate rect attribute
- [ ] that when using multiple different kwargs, only the last one is applied
- [ ] that appropriate exceptions are raised when using kwarg names that don't match rect attributes (like `centre` instead of `center`)

For `Rect`:
- [ ] add the difference between current `get_rect` and `get_frect` tests to `get_rect` (test position and instance type)

https://github.com/pygame-community/pygame-ce/blob/aff289d8bec34b6b9d19e0d34676b1d3f21a99b3/test/surface_test.py#L587-L605

JorasOliveira:
Added the requested tests for both pygame.Surface.get_rect and pygame.Surface.get_frect functions.

For both methods test:
 - [X] whether using correct keyword argument(s) sets the appropriate rect attribute
- [X]  that when using multiple different kwargs, only the last one is applied
 - [X] that appropriate exceptions are raised when using kwarg names that don't match rect attributes (like centre instead of center)

For Rect:
- [X]  add the difference between current get_rect and get_frect tests to get_rect (test position and instance type)

Closes #2109


ankith26:
requesting matt's review as they are the OP of the original issue

test/surface_test.py
```diff
@@ -621,20 +621,32 @@ def test_get_rect(self):
...
+        rect_kwargs = surf.get_rect(topleft=(1.0, 1.0), center=(100, 100))
 
+        self.assertEqual(rect.topleft, (0.0, 0.0))
         self.assertEqual(rect.size, size)
-
+        self.assertIsInstance(rect, pygame.Rect)
+        self.assertEqual(rect_kwargs.center, (100, 100))
+        self.assertNotEqual(rect_kwargs.topleft, (1.0, 1.0))
+        with self.assertRaises(AttributeError):
+            surf.get_rect(centre=(100, 100))
```
MyreMylar, line 632:
```suggestion
        if not IS_PYPY:
            with self.assertRaises(AttributeError):
                surf.get_rect(centre=(100, 100))
```

MyreMylar: This test doesn't work on PyPy because pypy doesn't raise an attribute error when passed incorrect attribute names to `PyObject_SetAttr()`.

You could also add a comment as to why we are skipping this particular bit of test for PyPy.

MyreMylar (commented):
Just a couple of spots to resolve the CI test fails on PyPy 👍

test/surface_test.py
JorasOliveira: will implement the changes tomorrow!

Matiiss: And don't forget that `get_frect` has the same issue

test/surface_test.py
```diff
@@ -621,20 +621,34 @@ def test_get_rect(self):
...
         rect = surf.get_rect()
+        rect_kwargs = surf.get_rect(topleft=(1.0, 1.0), center=(100, 100))
 
+        self.assertEqual(rect.topleft, (0.0, 0.0))
         self.assertEqual(rect.size, size)
-
+        self.assertIsInstance(rect, pygame.Rect)
+        self.assertEqual(rect_kwargs.center, (100, 100))
+        self.assertNotEqual(rect_kwargs.topleft, (1.0, 1.0))
+        if not IS_PYPY: #PyPy doesn't raise an AttributeError, so for PyPy we can just skip it.
```
MyreMylar, line 631:
```suggestion
        if (
            not IS_PYPY
        ):  # PyPy doesn't raise an AttributeError, so for PyPy we can just skip it.
```

test/surface_test.py
```diff
@@ -621,20 +621,34 @@ def test_get_rect(self):
...
         surf = pygame.Surface(size)
         frect = surf.get_frect()
+        frect_kwargs = surf.get_frect(topleft=(1.0, 1.0), center=(100, 100))
 
         self.assertEqual(frect.topleft, (0.0, 0.0))
         self.assertEqual(frect.size, size)
         self.assertIsInstance(frect, pygame.FRect)
+        self.assertEqual(frect_kwargs.center, (100, 100))
+        self.assertNotEqual(frect_kwargs.topleft, (1.0, 1.0))
+        if not IS_PYPY: #PyPy doesn't raise an AttributeError, so for PyPy we can just skip it.
```
MyreMylar, line 649:
```suggestion
        if (
            not IS_PYPY
        ):  # PyPy doesn't raise an AttributeError, so for PyPy we can just skip it.
```

MyreMylar (approved):
LGTM 👍

test/surface_test.py
```diff
@@ -621,20 +621,38 @@ def test_get_rect(self):
...
-
+        self.assertIsInstance(rect, pygame.Rect)
+        self.assertEqual(rect_kwargs.center, (100, 100))
+        self.assertNotEqual(rect_kwargs.topleft, (1.0, 1.0))
+        if (
+            not IS_PYPY
+        ):  # PyPy doesn't raise an AttributeError, so for PyPy we can just skip it.
+            with self.assertRaises(AttributeError):
+                surf.get_rect(centre=(100, 100))
+
```
MyreMylar, line 636:
```suggestion

```

test/surface_test.py
```diff
@@ -621,20 +621,37 @@ def test_get_rect(self):
...
         self.assertEqual(rect.size, size)
-
+        self.assertIsInstance(rect, pygame.Rect)
+        self.assertEqual(rect_kwargs.center, (100, 100))
+        self.assertNotEqual(rect_kwargs.topleft, (1.0, 1.0))
+        if (
+            not IS_PYPY
+        ):  # PyPy doesn't raise an AttributeError, so for PyPy we can just skip it.
+            with self.assertRaises(AttributeError):
+                surf.get_rect(centre=(100, 100))
```
MyreMylar, line 635:
```suggestion
                surf.get_rect(centre=(100, 100))

```

MyreMylar, line 635:
```suggestion
                surf.get_rect(centre=(100, 100))

```

Matiiss (changes requested):
> that when using multiple different kwargs, only the last one is applied

That basically means that there should be a test something like
```py
....get_rect(center=(5, 5), center=(10, 10))
```
where you test that center is indeed at (10, 10).

Also should test that if a positional argument is passed, it raises a `TypeError` and that it raises a `TypeError` when the value can't be assigned to the rect, for example
```py
....get_rect(5)
....get_rect(center=5)
....get_rect(center="5")
```

MyreMylar:
Try merging the latest main branch into this PR branch to fix the CI errors.

JorasOliveira:
> > that when using multiple different kwargs, only the last one is applied
> 
> That basically means that there should be a test something like
> 
> ```python
> ....get_rect(center=(5, 5), center=(10, 10))
> ```
> 
> where you test that center is indeed at (10, 10).

I get a syntaxError when I tried do this test, the gef_rect and get_frect functions don't allow repeated keywords. I'll add a test for the syntaxError

test/surface_test.py
```diff
@@ -621,20 +621,46 @@ def test_get_rect(self):
...
+            with self.assertRaises(TypeError):
+                surf.get_rect(5)
+                surf.get_rect(center=5)
+                surf.get_rect(center="5")
```
Matiiss, line 639: I don't think this will test all the functions and it shouldn't be under `if not IS_PYPY:`
you can also do this instead
```py
self.assertRaises(TypeError, surf.get_rect, 5)
self.assertRaises(TypeError, surf.get_rect, center=5)
self.assertRaises(TypeError, surf.get_rect, center="5")
```

Matiiss (changes requested):

test/surface_test.py
Matiiss: Same for `get_frect`

Matiiss (approved):
LGTM, thanks for the PR 🎉

PR #2493, opened 2023-10-02
Drop python 3.7

ankith26:
Python 3.7 has reached EOL.
Now that we support Python 3.12, we have 5 active versions of Python to support (Python 3.8 - 3.12)


zoldalma999 (approved):
LGTM

MyreMylar (approved):
LGTM 👍

PR #2500, opened 2023-10-05
Fix SDL_ConvertSurface(Format) for SDL3

Starbuck5:
SDL3 just removes the unused flags argument, so it's very easy to port.


zoldalma999 (approved):
the _sdl2.video module still uses SDL_ConvertSurface, but that is cython, so I think that can wait (and it will be ported off of it at some point anyway). LGTM

ankith26 (approved):
LGTM, thanks :tada:

PR #2501, opened 2023-10-05
Port last SDL_CreateRGBSurface calls to SDL3-safe PG_CreateSurface

Starbuck5:
The surface_init one could be refactored to push the mask->pixelformat conversion to only when it's actually parsing out masks from the user, and have the format live as a pixelformat most of the time. But that would be a more complex and risky diff than this one, so lets stay the course for now.

I removed `_raise_create_surface_error` because it's no longer useful. It caught a case emitted internally by SDL_CreateRGBSurface when it converts masks to pixelformat, now that we do that ourselves we can directly set ValueError there. As for it's use in subsurface, the format will never be invalid there because it comes from already initialized surfaces-- therefore there's no point with a possible ValueError about masks.

### DIFF
--- a/src_c/image.c
+++ b/src_c/image.c
@@ -1509,7 +1509,7 @@ SaveTGA_RW(SDL_Surface *surface, SDL_RWops *out, int rle)
     Uint8 surf_alpha;
     int have_surf_colorkey = 0;
     Uint32 surf_colorkey;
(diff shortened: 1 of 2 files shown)


ankith26 (approved):

MyreMylar (approved):
LGTM 👍

PR #2502, opened 2023-10-06
Remove RDTSC key because detection removed in SDL3

Starbuck5:
My rationale for straight up removing it is because I doubt anyone is using it.

I'm open to other suggestions of how to do the change management.


ankith26 (approved):
I too think removing it entirely is the way to proceed, thanks :tada:

yunline (approved):
LGTM

PR #2503, opened 2023-10-06
Fix gfxdraw.bezier for a lot of points

Temmie3754:
Resolves the issue by changing the Bezier functions to use integers instead of doubles to prevent equality checks failing unexpectedly for when there is a very small difference. Do note that due to the imprecision of doubles, this issue returns at approximately ~150-200 points in which case it will default to the last point being used rather than (0, 0).


MyreMylar:
Looks like this will need to merge with main to hopefully clear the grub-efi error, and then run:

```
clang-format src_c/SDL_gfx/SDL_gfxPrimitives.c -i
```
(If you are on windows you can download the stuff you need for clang-formant from: https://github.com/llvm/llvm-project/releases/tag/llvmorg-17.0.2 usually [LLVM-17.0.2-win64.exe](https://github.com/llvm/llvm-project/releases/download/llvmorg-17.0.2/LLVM-17.0.2-win64.exe))

Just so only the actual changes show up in the Github diff.

MyreMylar:
This diff still looks odd.

Temmie3754:
> This diff still looks odd.

I've run it through the formatter and that's what it's given, perhaps it's a CRLF vs LF issue where it's still in CRLF but what I'm uploading is LF.

MyreMylar:
It should be failing the auto formatter now if you are merged with current
main?

Temmie3754:
I don't think the formatter accounts for line endings.

MyreMylar:
I've just toggled off 'show whitespace' in Github to see the diff now. 

I guess clang doesn't erase unnecessary spaces, or it is the line endings thing (but only some line endings?). Anyway, we'll just go with your whitespace changes as well since they don't actually change anything important.

MyreMylar (approved):
LGTM. Fixes the issue in #2503. You can now have very long bezier curves with many points (tested up to 150).

👍

yunline:
> Do note that due to the imprecision of doubles, this issue returns at approximately ~150-200 points in which case it will default to the last point being used rather than (0, 0).

This should be mentioned in the docs to tell usesrs don't pass too many points to this function.

PR #2505, opened 2023-10-07
Try compiling deps on win manually

ankith26:
Very WIP (will update details as I go along)


ankith26:
After over 2 years of intermittent work, this PR is FINALLY done from my side and should now be ready for review. I have updated the opening comment to contain almost all info about this PR that I can think of at this point, but if anything is not clear lmk.

cc @Starbuck5 review pls :point_right: :point_left:

Starbuck5:
Comparison of DLLs in artifacts from this PR vs artifacts from main.

| Section      | Name before    | Size before | Name after      | Size after | Size increase percent |
|--------------|----------------|-------------|-----------------|------------|-----------------------|
| Font         | SDL2_ttf       | 1751552     | SDL2_ttf        | 53219      |                       |
|              | freetype       | 654336      | libfreetype-6   | 559455     |                       |
|              |                |             | libbrotlicommon | 141893     |                       |
|              |                |             | libbrotlidec    | 48326      |                       |
|              |                |             | libharfbuzz-0   | 991860     |                       |
| Font totals  |                | 2405888     |                 | 1794753    | -25.40163964          |
|              |                |             |                 |            |                       |
| Image        | SDL2_image     | 157696      | SDL2_image      | 104819     |                       |
|              | libjpeg-62     | 1272353     | libjpeg-62      | 599211     |                       |
|              | libpng16-16    | 279040      | libpng16        | 186651     |                       |
|              |                |             | libzlib         | 145692     |                       |
|              | libtiff-5      | 387584      | libtiff         | 462115     |                       |
|              | libwebp-7      | 444416      | libwebp         | 330081     |                       |
|              | libwebpdemux-2 | 24064       | libwebpdemux    | 21858      |                       |
|              |                |             | libsharpyuv     | 31973      |                       |
| Image totals |                | 2565153     |                 | 1882400    | -26.61646303          |
|              |                |             |                 |            |                       |
| Mixer        | SDL2_mixer     | 421376      | SDL2_mixer      | 144899     |                       |
|              | libogg         | 35328       | ogg             | 31907      |                       |
|              | libopus-0      | 370176      | libopus-0       | 450814     |                       |
|              | libopusfile-0  | 51200       | libopusfile-0   | 54058      |                       |
|              | libwavpack-1   | 175616      | libwavpack-1    | 181307     |                       |
|              | libxmp         | 411648      | libxmp          | 478758     |                       |
|              |                |             | libsndfile      | 483755     |                       |
|              |                |             | libvorbis       | 154483     |                       |
|              |                |             | libvorbisenc    | 574651     |                       |
|              |                |             | libvorbisfile   | 33378      |                       |
|              |                |             | libmpg123-0     | 451556     |                       |
|              |                |             | libFLAC         | 262058     |                       |
|              |                |             | libfluidsynth-3 | 1544556    |                       |
| Mixer totals |                | 1465344     |                 | 4846180    | 230.7196126           |
|              |                |             |                 |            |                       |
| Other        | SDL2           | 1586176     | SDL2            | 1523559    |                       |
|              | portmidi       | 41984       | libportmidi     | 32132      |                       |
|              |                |             | libgcc_s_seh    | 150196     |                       |
|              |                |             | libwinpthread   | 65018      |                       |
|              |                | 1628160     |                 | 1770905    | 8.767258746           |  |                 | 1802878    | 10.73100924           |

Starbuck5:
This is very cool, thanks for working on this Ankith! ARM support would be excellent.

Analysis:
- The size is fine, except for the mixer deps.
     - Maybe no need to switch from performance to size optimized, I think the root of the size gain is in the dependency layout, not how the deps are built.
- The libgcc_s_seh and libwinpthread deps are ugly IMO, since it's showing a separate C runtime than Python uses. GCC vs MSVC.
- The font module had the most potential for size reduction, given the duplication of freetype statically into SDL_ttf and dynamically for our freetype module. The size reduction is close to matching just deleting the existing freetype DLL, which suggests the existing freetype DLL does not have text shaping built in.
- Would it be possible to do phase this in group by group? We know from prior experience that changing this stuff can lead to very hard to predict outcomes, like https://github.com/pygame-community/pygame-ce/issues/2650. Also it seems to me the mixer group might change further, versus the font or image groups.

ankith26:
thanks for the detailed breakdown and grouping of dlls, it is helpful.

> Maybe no need to switch from performance to size optimized, I think the root of the size gain is in the dependency layout, not how the deps are built.

well yeah, the root of size gain is definitely the changes in mixer deps. But I still think that size optimised is better because we are doing decent size savings (across all wheels independent of platform) without sacrificing performance too much (in theory).

> The libgcc_s_seh and libwinpthread deps are ugly IMO, since it's showing a separate C runtime than Python uses. GCC vs MSVC.

I could eliminate these by statically linking them wherever they are used, but that is a bit more work :sweat_smile: and I'm not sure how it would affect size. These libraries are really wrappers over the MSVC c runtime and win threads respectively anyways, so I think having these DLLs is not a big deal.

> Would it be possible to do phase this in group by group?

I already have grouped the changes into separate commits for ease of bisect, I don't think we need separate PRs for this purpose. And I maintain my opinion that for dependency changes, it is best to test out things in integration, hence a common PR.

.github/workflows/release-pypi.yml
```diff
@@ -28,11 +28,16 @@ jobs:
...
+              echo "[DELETED non-python-artifact] $fname"
+              continue
+            fi
             if gh attestation verify $fname -R ${{ github.repository }}; then
               echo "[ALLOWED] $fname"
             else
               rm $fname
-              echo "[DELETED] $fname"
+              echo "[DELETED failed-attestation] $fname"
             fi
```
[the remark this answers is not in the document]
ankith26: fixed

meson.build
```diff
@@ -143,121 +143,182 @@ elif plat == 'emscripten-pyodide'
...
-            sdl_image_lib_dir / 'optional' / (sdl_api == 3 ? 'libtiff-6.dll' : 'libtiff-5.dll'),
-            sdl_image_lib_dir / 'optional' / 'libwebp-7.dll',
-            sdl_image_lib_dir / 'optional' / 'libwebpdemux-2.dll',
+        pg_inc_dirs += fs.relative_to(prebuilt_dir / 'include', base_dir)
+        pg_inc_dirs += fs.relative_to(prebuilt_dir / 'include' / 'SDL2', base_dir)
+        pg_inc_dirs += fs.relative_to(prebuilt_dir / 'include' / 'freetype2', base_dir)
+        pg_lib_dirs += lib_dir
+    elif host_machine.cpu_family().startswith('x86')
+        warning('Missing windows prebuilts. Falling back to legacy prebuilts')
+
```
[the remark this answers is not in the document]
ankith26: fixed

ankith26:
After taking @Starbuck5 s suggestion on discord, I have decided to do this PR in two parts. The goal of this PR is simply ARM windows support now, so that it can be merged ASAP in the upcoming 2.5.7 release. The x86/x64 windows builds will remain untouched and function in the same way as before, using what I now call legacy prebuilts.

All the benefits of this PR I listed in the opening comment will therefore only apply on the ARM builds as of now. With an ARM windows supporting release, we should hopefully get wider testing and feedback based on which we can fix any bugs that come up. After we have reasonable confidence with that we can extend the approach to x86/x64.

Starbuck5:
@ankith26 I considered trying to figure out a way to find someone knowledgeable that happens to have Windows on ARM hardware, and then I got myself an Azure account and fired up a Windows ARM VM.

[314_arm64_testlog.txt.txt](https://github.com/user-attachments/files/25122710/314_arm64_testlog.txt.txt). I'm not concerned about the few fails.

I also ran a few examples, everything seemed to work as expected. The one thing I think would be good to address it the SIMD setup with NEON, since it gives a warning on import. And in hindsight I didn't need a windows VM for the SIMD thing, it's in the logs from CI as well.

Another thing-- this adds wheels for 3.10 and above, I don't think anyone is running 3.10 on ARM64. There's no release for it from python.org. I'd be comfortable dropping 3.11 and 3.12 and only doing 3.13+ for ARM, but it's not a big deal to do them all.

I've also discovered why the lack of windows on arm has been okay. I downloaded the new pymanager and when I tried to install 3.14 it actually installed x64 3.14 everything was emulated seamlessly. It's the default behavior until the "arm64 package ecosystem matures." I was able to download normal pygame-ce and run examples no problem. Could actually be an interesting test to see about getting a speedup with native arm64 (+NEON) vs emulated x64.

ankith26:
Thanks for testing this PR out.

I too had noticed the SIMD thing. The issue is basically that the `sse2neon.h` header doesn't support the MSVC compiler on ARM when included from a C file. I see three possible fixes for this issue:
- Switch to another library for SIMD compat, like [SIMDe](https://github.com/simd-everywhere/simde).
- Use MinGW instead of MSVC for building pygame-ce on ARM (just like what we do for our deps).
- Update the SIMD files to be C++ 

I haven't really worked on any of these ideas yet, its definitely something for a future PR and not this one.

Starbuck5:
> I too had noticed the SIMD thing. The issue is basically that the sse2neon.h header doesn't support the MSVC compiler on ARM when included from a C file. I see three possible fixes for this issue:

Are you sure the problem wasn't on our end? It really looks like that would be supported, looking at the sse2neon GitHub. And if doesn't work it seems like we should raise an issue about that.

meson.build
```diff
@@ -143,120 +143,186 @@ elif plat == 'emscripten-pyodide'
...
+            dlls += lib_dir / 'libgcc_s_dw2-1.dll'
+        elif host_machine.cpu_family() == 'x86_64'
+            dlls += lib_dir / 'libgcc_s_seh-1.dll'
+        endif
+
+        pg_inc_dirs += fs.relative_to(prebuilt_dir / 'include', base_dir)
+        pg_inc_dirs += fs.relative_to(prebuilt_dir / 'include' / 'SDL2', base_dir)
+        pg_inc_dirs += fs.relative_to(prebuilt_dir / 'include' / 'freetype2', base_dir)
+        pg_lib_dirs += lib_dir
+    elif host_machine.cpu_family().startswith('x86')
```
Starbuck5, line 213: I would like a comment or something in the warning to say that this is expected behavior for nowTM. Since this PR does not move over windows deps on x86 or x64.

.github/workflows/build-windows.yml
```diff
@@ -50,3 +52,157 @@ jobs:
...
           compression-level: 0  # wheels are already zip files, no need for more compression
+
+  deps:
+    name: ${{ matrix.winarch }} deps
+    runs-on: ${{ matrix.os }}
+    strategy:
+      fail-fast: false  # if a particular matrix build fails, don't skip the rest
+      matrix:
+        include:
+          - { winarch: x86_64, march: x86_64, msystem: MINGW64, os: windows-2022 }
```
Starbuck5, line 65: I think it could be helpful to have a comment here that the x86 and x86_64 dep bundles aren't used yet.

.github/workflows/release-pypi.yml
Starbuck5, line 1: My alternate patch here would be to only download the wheels in the first place rather than deleting them from the release.

release-gh-draft.yml
```
      - name: Download all artifacts
        uses: actions/download-artifact@v7
        with:
          path: pygame-wheels
          merge-multiple: true
```
\+  pattern: pygame-wheels-*

Even works for sdist, because its artifact is "pygame-wheels-sdist"

Starbuck5:
Now that this is targeted to Windows on ARM only, could you update the PR description to reflect that? It will no longer close the audio issues you've highlighted, and the local contributors solution is different, among other things.

.github/workflows/build-windows.yml
ankith26: Technically they are used, but only for unreleased msys builds. Sure, I could however add a comment saying our release wheels don't use it.

.github/workflows/release-pypi.yml
ankith26: That is exactly the approach I have already implemented in #3625, I still think having an extra layer of checking for sdist or wheel files is better so we should do both IMO.

pyproject.toml
Starbuck5, line 142: I got very confused by this block, I thought when it said skip it meant the wheel, and then I thought it meant skip testing, and now I get it means skip installing numpy for the tests. But I feel like you might as well say why you're not requiring any test deps here... Presumably numpy doesn't have a 3.10 ARM wheel?

ankith26:
> Are you sure the problem wasn't on our end? It really looks like that would be supported, looking at the sse2neon GitHub. And if doesn't work it seems like we should raise an issue about that.

Yes, the error I hit when trying to get SIMD support on this PR basically came from

https://github.com/DLTcollab/sse2neon/blob/f1bc16e4b107f368a6098edd0d423803657837bd/sse2neon.h#L493

.github/workflows/release-pypi.yml
Starbuck5: Lol maybe I got the idea subconsciously from your other PR then. I'm okay with both but I'd like to see them both in this PR and I'd like a comment (ofc :smile:) explaining that it's an extra safety thing.

pyproject.toml
ankith26: Yes, numpy doesn't have 3.10 windows ARM wheels and this block is skipping numpy installation. Currently I'm thinking of going with what you said earlier and skipping 3.10 ARM windows wheels entirely. I would like to retain retain 3.11+ wheels however.

Starbuck5 (approved):

PR #2507, opened 2023-10-07
Remove spurious comment

Julian-O:
Comment referred to print statement requesting user-base support that has since been removed.


MyreMylar (approved):
Yeah, this seems fine to remove.

Temmie3754 (approved):
Checked the history and the contribute link this comment was presumably about has since been removed so it makes sense to remove this too.

Matiiss (approved):
LGTM 👍

PR #2508, opened 2023-10-07
run clang-format on the vendored contents of SDL_gfx/

MyreMylar:
pygame-ce is probably now the most up to date source of this code, the original has not been updated for many years so we are unlikely to re-vendor it.

@Temmie3754 just tried to fix a bug in it and it produced a massive diff as it had not been run through our standard clang-format pass.

I know we have long term plans to be rid of gfxdraw and all it's stuff, but for now we may as well make it easier to maintain.

Related to #2503 and we should merge this first so we can see the diff clearly on that.


Temmie3754 (approved):
Successfully formatted the files and is included in the auto format routine.

ankith26 (approved):
:+1: lets get this in

Issue #2495, opened 2023-10-02
Documentation: Intro to surfarray chart rendered incorrectly on phone


Damus666:
- **Operating system**: Any phone/desktop site in phone view
- **PyGame version**: 2.3.X Documentation

**Current behavior:**

On the desktop site (either on computer or phone in desktop view) the function-surface-format chart is rendered properly as expected.
On a phone or on a computer in phone view, the same chart is rendered in an unexpected way as shown in the screenshots.
Note: The chart is rendered correctly on the [intro-to-surfarray.rst file](https://github.com/pygame-community/pygame-ce/blob/main/docs/reST/tutorials/en/intro-to-surfarray.rst) inside the documentation.

**Expected behavior:**

The chart, which is not big, should be rendered equally on both phone and desktop views.

**Screenshots**

Correct behaviour/desktop view:
![image](https://github.com/pygame-community/pygame-ce/assets/97639432/36b91fe1-3c8a-456d-961a-bba799cd201e)

Unexpected behaviour (phone view):
![image](https://github.com/pygame-community/pygame-ce/assets/97639432/4dc6ffe0-f957-4074-a253-6fdba3fe4a28)

**Steps to reproduce:**

- Go on the following link "https://pyga.me/docs/tutorials/en/intro-to-surfarray.html" on phone (if on computer, press CTRL+SHIFT+I and enable phone view)
- Scroll down until you reach the chart, and look at how it's rendered

PR #2509, opened 2023-10-07
Remove 'responsive' tables from docs CSS

MyreMylar:
Related to #2495

As far as I can tell removing adding `display: block` to all table elements on mobile doesn't gain us very much?

It definitely ruins any actual tables though. Seems to be old CSS rather than recent.

### DIFF
--- a/docs/reST/themes/classic/static/pygame.css_t
+++ b/docs/reST/themes/classic/static/pygame.css_t
@@ -1000,10 +1000,6 @@ pre.commentContent {
 /* -- responsive design --------------------------------------------------------- */
 
 @media only screen and (max-width: 680px) {
-    /* Make tables be responsive-ish. */
-    table, thead, tbody, th, td, tr {
-        display: block;
-    }
 
     div.body img.inlined-right {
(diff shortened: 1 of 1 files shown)


ankith26:
I'm sorry I'm not clear, does it fix the the issue or is it just a cleanup?

MyreMylar:
> I'm sorry I'm not clear, does it fix the the issue or is it just a cleanup?

This resolves the broken tables on mobiles issue. I'm hopeful that somebody with more recent css experience than I can give this a look and explain why this was in there in the first place. It's possible (though I don't know) that a more subtle change could also resolve this, if it turns out there is any good reason for it to be there in the first place. 

The word 'responsive' from the original comment above this block doesn't mean anything to me in this context- which is possibly my lack of modern web dev experience.

ankith26 (approved):
While I'm not that great at CSS to fully understand things, I'm approving because it does seem to fix the issue.

dr0id (approved):
could not find any relevant information about `display: block` and responsive in tables in a short web search.

PR #2511, opened 2023-10-08
Add docs for windows DPI awareness hint

MyreMylar:
After a bit of back and forth in #2081 the best solution seemed to be to document the current behaviour of DPI awareness on windows and revisit this topic in the SDL 3 switchover. Currently in SDL2 DPI awareness is a bit of a mess with different behaviour on different platforms and different APIs in SDL 2.

### DIFF
--- a/docs/reST/ref/pygame.rst
+++ b/docs/reST/ref/pygame.rst
@@ -546,4 +546,19 @@ before calling :func:`pygame.init()` or :func:`pygame.joystick.init()`.
 Useful for testing touch events on desktop platforms (e.g. with a trackpad)
 where this is set to 0 by default.
 
+|
+
+::
+
+ "SDL_WINDOWS_DPI_AWARENESS"
+ Set to "permonitorv2" on windows 10 (and later) to declare the pygame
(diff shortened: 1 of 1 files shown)


MyreMylar:
This is my little test program to play around with DPI in SDL2:

```
import pygame
import os
import platform

from pygame._sdl2.window import Window


if platform.uname().system == "Windows":
    build_num = int(platform.uname().version.split(".")[2])
    if build_num > 14393:
        os.environ["SDL_WINDOWS_DPI_AWARENESS"] = "permonitorv2"
    elif build_num > 9600:
        os.environ["SDL_WINDOWS_DPI_AWARENESS"] = "permonitor"
    else:
        os.environ["SDL_WINDOWS_DPI_AWARENESS"] = "system"


pygame.init()

window = Window(allow_high_dpi=True)

running = True
while running:
        if event.type == pygame.QUIT:
            running = False
```

Temmie3754 (approved):
Suitably covers how to easily implement DPI awareness in pygame on windows and other platforms.

PR #2515, opened 2023-10-12
Moving current scrap API docs to the top to avoid confusion

Matiiss:
There is this notice that "everything below is deprecated" so it sort of made sense to move the working stuff to above that, I don't think that would cause confusion now (but I am not entirely sure what to do with `.. ## pygame.scrap ##` at the end of the docs, should it just stay at the end anyway? (I don't think we are using that feature anyway)).

Regarding the experimental notice, is that still relevant?

### DIFF
--- a/docs/reST/ref/scrap.rst
+++ b/docs/reST/ref/scrap.rst
@@ -21,6 +21,54 @@ in a future release of pygame.
    clipboard as the rest of the current API, but only strings are compatible with the 
    new API as of right now.
 
+.. function:: put_text
+
+   | :sl:`Places text into the clipboard.`
+   | :sg:`put_text(text) -> None`
+
+   Places the input text into the clipboard. The data should be a string.
+   This is the same clipboard as the legacy scrap API when using ``SCRAP_TEXT``.
+
+   :param string text: String to be placed into the clipboard
+   :rtype: None
+
+   :raises pygame.error: if video mode has not been set_mode
+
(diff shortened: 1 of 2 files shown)


MyreMylar (approved):
LGTM 👍 

I mean we could probably actually consider removing the old scrap stuff for 2.4.0 based on the 'two major versions since deprecation' guideline. However, since we've been galloping up major versions quite fast at the minute and this was a fairly large feature set I'd be happy to leave it another version at present. We also haven't officially codified a policy yet on the Wiki.

CPython's policy deprecation-to-removal cycle is two years - which for them also equals two major versions.  Numpy uses a 1 year deprecation-to-removal cycle and two major releases (which for them are once every 6 months).

This modules was deprecated 7 months ago in March so there is a little ways to go yet on either rule.

Starbuck5 (approved):
Good idea!

Starbuck5:
I say keep the experimental notice for now.

> I mean we could probably actually consider removing the old scrap stuff for 2.4.0 based on the 'two major versions since deprecation' guideline. However, since we've been galloping up major versions quite fast at the minute and this was a fairly large feature set I'd be happy to leave it another version at present. We also haven't officially codified a policy yet on the Wiki.

We don't have the market share to make large breaking changes. People can switch over to use pygame-ce with no problem because we have made very few backwards-incompatible changes.

PR #2522, opened 2023-10-16
Improve draw performance

### LINKED ISSUE #2517: Refactor `set_at()` and it's usage in draw.c
Currently the function `set_at()` in draw.c looks like this:

```C
static int
set_at(SDL_Surface *surf, int x, int y, Uint32 color)
{
    SDL_PixelFormat *format = surf->format;
    Uint8 *pixels = (Uint8 *)surf->pixels;
    Uint8 *byte_buf, rgb[4];

        return 0;

    switch (format->BytesPerPixel) {
        case 1:
            *((Uint8 *)pixels + y * surf->pitch + x) = (Uint8)color;
            break;
        case 2:
            *((Uint16 *)(pixels + y * surf->pitch) + x) = (Uint16)color;
            break;
        case 4:
            *((Uint32 *)(pixels + y * surf->pitch) + x) = color;
            break;
        default: /*case 3:*/
            SDL_GetRGB(color, format, rgb, rgb + 1, rgb + 2);
#if (SDL_BYTEORDER == SDL_LIL_ENDIAN)
            *(byte_buf + (format->Rshift >> 3)) = rgb[0];
            *(byte_buf + (format->Gshift >> 3)) = rgb[1];
            *(byte_buf + (format->Bshift >> 3)) = rgb[2];
#else
            *(byte_buf + 2 - (format->Rshift >> 3)) = rgb[0];
            *(byte_buf + 2 - (format->Gshift >> 3)) = rgb[1];
            *(byte_buf + 2 - (format->Bshift >> 3)) = rgb[2];
#endif
            break;
    }
    return 1;
}
```

There has already been one attempt to improve it by making an 'unsafe' version that doesn't do clip checking. 

This does not go far enough. This function is used all over draw.c inside of while loops that draw lines containing hundreds of pixels, and shapes that may contain many lines. 

For each pixel of these lines we are checking what the bit depth of the surface is. **That could be a thousand checks per polygon for a large polygon - but we only need one check.**

I propose, unless `set_at()` is being called directly by a user to change literally a single pixel, we instead make all our tests for surface bit depth (and ideally shape clipping) happen outside of any per pixel while loops and inside of them we call very specific functions that just set the pixel.

e.g.

```C
// This 'safe' version assumes we have determined the surface bit depth but does clipping per pixel,
// ideally we could clip our draw shapes at a higher level
static int set_at_32bit_safe(Uint8 *pixels, int pitch, int x, int y, Uint32 color, SDL_Rect clip_rect)
{
    if (x < clip_rect.x || x >= clip_rect.x + clip_rect.w ||
        y < clip_rect.y || y >= clip_rect.y + clip_rect.h)
        return 0;

    *((Uint32 *)(pixels + y * pitch) + x) = color;
    return 1;
}
//This 'unsafe' version just sets a 32bit surface pixel to a colour and assumes safe postions (i.e. inside the surface)
//and that we have already determined the needed pixel tpye before calling
static void set_at_32bit_unsafe(Uint8 *pixels, int pitch, int x, int y, Uint32 color)
{
    *((Uint32 *)(pixels + y * pitch) + x) = color;
}
```

In all cases we should also prioritise the 32 bit path - the current set_at() has an 8bit surface depth as the first case, before testing for a 16 bit surface, then a 24bit one before finally testing for a 32bit. These switch cases should be in surface commonality order which is 32bit (4byte) then 24bit (3byte), then 8bit (1byte) and finally the rarely seen 16bit (2 byte) surface.

**To Do list:**

- [ ] Identify all the places where set_at() is eventually used inside of a pixel loop. For a refactorings list.
- [ ] Create simple pixel setter functions like the 32bit onese above for all the normal bit depth surfaces (32bit, 24bit, 16bit, 8bit).
- [ ] copy the surface depth switch case that currently sits inside set_at() to sit outside each of the pixel loops (or higher) so we only test the surface bit depth once per user call to draw.
- [ ] use the new simple pixel setter functions inside the pixel loops now separated cleanly by bit depth. 
- [ ] make sure the tests still pass.
- [ ] check the performance versus current main branch across however many draw functions this problem affects.
- [ ] can any of this be macro'd up to reduce repetition?

MightyJosip:
Fix #2517

Still need to check what needs clip version, and what doesn't.

### DIFF
--- a/src_c/draw.c
+++ b/src_c/draw.c
@@ -40,56 +40,174 @@
 /* Declaration of drawing algorithms */
 static void
 draw_line_width(SDL_Surface *surf, Uint32 color, int x1, int y1, int x2,
-                int y2, int width, int *drawn_area);
+                int y2, int width, int *drawn_area, int (*set_func)(SDL_Surface*, int, int, Uint32));
 static void
 draw_line(SDL_Surface *surf, int x1, int y1, int x2, int y2, Uint32 color,
-          int *drawn_area);
+          int *drawn_area, int (*set_func)(SDL_Surface*, int, int, Uint32));
 static void
 draw_aaline(SDL_Surface *surf, Uint32 color, float startx, float starty,
-            float endx, float endy, int *drawn_area);
+            float endx, float endy, int *drawn_area, int (*set_func)(SDL_Surface*, int, int, Uint32));
 static void
 draw_arc(SDL_Surface *surf, int x_center, int y_center, int radius1,
          int radius2, int width, double angle_start, double angle_stop,
          Uint32 color, int *drawn_area);
 static void
 draw_circle_bresenham(SDL_Surface *surf, int x0, int y0, int radius,
-                      int thickness, Uint32 color, int *drawn_area);
+                      int thickness, Uint32 color, int *drawn_area, int (*set_func)(SDL_Surface*, int, int, Uint32));
 static void
 draw_circle_bresenham_thin(SDL_Surface *surf, int x0, int y0, int radius,
-                           Uint32 color, int *drawn_area);
+                           Uint32 color, int *drawn_area, int (*set_func)(SDL_Surface*, int, int, Uint32));
 static void
 draw_circle_filled(SDL_Surface *surf, int x0, int y0, int radius, Uint32 color,
-                   int *drawn_area);
+                   int *drawn_area, int (*set_func)(SDL_Surface*, int, int, Uint32));
 static void
 draw_circle_quadrant(SDL_Surface *surf, int x0, int y0, int radius,
                      int thickness, Uint32 color, int top_right, int top_left,
-                     int bottom_left, int bottom_right, int *drawn_area);
+                     int bottom_left, int bottom_right, int *drawn_area, int (*set_func)(SDL_Surface*, int, int, Uint32));
 static void
 draw_ellipse_filled(SDL_Surface *surf, int x0, int y0, int width, int height,
-                    Uint32 color, int *drawn_area);
+                    Uint32 color, int *drawn_area, int (*set_func)(SDL_Surface*, int, int, Uint32));
 static void
 draw_ellipse_thickness(SDL_Surface *surf, int x0, int y0, int width,
                        int height, int thickness, Uint32 color,
-                       int *drawn_area);
+                       int *drawn_area, int (*set_func)(SDL_Surface*, int, int, Uint32));
 static void
 draw_fillpoly(SDL_Surface *surf, int *vx, int *vy, Py_ssize_t n, Uint32 color,
-              int *drawn_area);
+              int *drawn_area, int (*set_func)(SDL_Surface*, int, int, Uint32));
 static int
 draw_filltri(SDL_Surface *surf, int *xlist, int *ylist, Uint32 color,
-             int *drawn_area);
+             int *drawn_area, int (*set_func)(SDL_Surface*, int, int, Uint32));
 static void
 draw_rect(SDL_Surface *surf, int x1, int y1, int x2, int y2, int width,
           Uint32 color);
 static void
 draw_round_rect(SDL_Surface *surf, int x1, int y1, int x2, int y2, int radius,
                 int width, Uint32 color, int top_left, int top_right,
-                int bottom_left, int bottom_right, int *drawn_area);
+                int bottom_left, int bottom_right, int *drawn_area, int (*set_func)(SDL_Surface*, int, int, Uint32));
 
 // validation of a draw color
 #define CHECK_LOAD_COLOR(colorobj)                               \
     if (!pg_MappedColorFromObj((colorobj), surf->format, &color, \
                                PG_COLOR_HANDLE_ALL)) {           \
         return NULL;                                             \
+    }
+
+#define CHECK_CLIP_RECT                                                        \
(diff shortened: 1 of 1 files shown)


src_c/draw.c
```diff
@@ -108,6 +226,7 @@ aaline(PyObject *self, PyObject *arg, PyObject *kwargs)
     int drawn_area[4] = {INT_MAX, INT_MAX, INT_MIN,
                          INT_MIN}; /* Used to store bounding box values */
     Uint32 color;
+    int (*set_func)(SDL_Surface*, int, int, Uint32);
```
MyreMylar, line 229: Looks like we need to set a default value for `set_func` here to please gcc. I guess something like:

```suggestion
    int (*set_func)(SDL_Surface*, int, int, Uint32) = &set_at_32bit_clip;;
```

MyreMylar (commented):
One thing I'm slightly concerned about using this approach is that C function pointers and performance intensive tight loops are classically not always the best of friends, though this is somewhat compiler dependant.

As a test, I've tried swapping out the function pointer call in `set_and_check_rect()` for a clone of the function it is calling:

```C

static int
set_at_32bit_no_func_pointer(SDL_Surface *surf, int x, int y, Uint32 color)
{
    {
        return 0;
    }

    *((Uint32 *)((Uint8 *)surf->pixels + y * surf->pitch) + x) = color;
    return 1;
}

static void
set_and_check_rect(SDL_Surface *surf, int x, int y, Uint32 color,
                   int *drawn_area, int (*set_func)(SDL_Surface*, int, int, Uint32))
{
    if (set_at_32bit_no_func_pointer(surf, x, y, color))
        add_pixel_to_drawn_list(x, y, drawn_area);
}
```

And on my PC the version without the function pointer was running 15-30% faster in my simple line drawing performance test:

```python
from sys import stdout
from pstats import Stats
from cProfile import Profile


import pygame

pygame.init()


def draw_line_short_inside():
    lines = [
        ((256, 256), (0, 0)),
        ((128, 256), (160, 0)),
        ((50, 256), (128, 0)),
        ((0, 240), (12, 56)),
        ((50, 40), (60, 11)),
    ]
    widths = [1]
    surf = pygame.Surface((256, 256), pygame.SRCALPHA, 32)
    surf.fill((255, 255, 255, 255))
    for iterations in range(0, 10000):
        for line in lines:
            for width in widths:
                pygame.draw.line(
                    surf, pygame.Color("red"), line[0], line[1], width=width
                )


def draw_line_long_inside():
    lines = [
        ((1256, 256), (0, 0)),
        ((1128, 256), (128, 0)),
        ((150, 1256), (128, 0)),
        ((0, 240), (1200, 56)),
        ((50, 40), (1060, 11)),
    ]
    widths = [1]
    surf = pygame.Surface((2048, 2048), pygame.SRCALPHA, 32)
    surf.fill((255, 255, 255, 255))
    for iterations in range(0, 10000):
        for line in lines:
            for width in widths:
                pygame.draw.line(
                    surf, pygame.Color("red"), line[0], line[1], width=width
                )


def draw_line_short_outside():
    lines = [
        ((256, 256), (0, 0)),
        ((128, 256), (160, 0)),
        ((50, 256), (128, 0)),
        ((0, 240), (12, 56)),
        ((50, 40), (80, 11)),
    ]
    widths = [1]
    surf = pygame.Surface((64, 64), pygame.SRCALPHA, 32)
    surf.fill((255, 255, 255, 255))
    for iterations in range(0, 10000):
        for line in lines:
            for width in widths:
                pygame.draw.line(
                    surf, pygame.Color("red"), line[0], line[1], width=width
                )


def draw_line_long_outside():
    lines = [
        ((2256, 256), (0, 0)),
        ((2128, 256), (128, 0)),
        ((150, 4256), (128, 0)),
        ((0, 240), (2200, 56)),
        ((50, 40), (4060, 11)),
    ]
    widths = [1]
    surf = pygame.Surface((1024, 1024), pygame.SRCALPHA, 32)
    surf.fill((255, 255, 255, 255))
    for iterations in range(0, 10000):
        for line in lines:
            for width in widths:
                pygame.draw.line(
                    surf, pygame.Color("red"), line[0], line[1], width=width
                )


if __name__ == "__main__":
    print("Draw Line - short, inside surface")
    profiler = Profile()
    profiler.runcall(draw_line_short_inside)
    stats = Stats(profiler, stream=stdout)
    stats.strip_dirs()
    stats.sort_stats("cumulative")
    stats.print_stats()

    print("\nDraw Line - long, inside surface")
    profiler = Profile()
    profiler.runcall(draw_line_long_inside)
    stats = Stats(profiler, stream=stdout)
    stats.strip_dirs()
    stats.sort_stats("cumulative")
    stats.print_stats()

    print("\nDraw Line - short, edges outside surface")
    profiler = Profile()
    profiler.runcall(draw_line_short_outside)
    stats = Stats(profiler, stream=stdout)
    stats.strip_dirs()
    stats.sort_stats("cumulative")
    stats.print_stats()

    print("\nDraw Line - long, edges a long way outside surface")
    profiler = Profile()
    profiler.runcall(draw_line_long_outside)
    stats = Stats(profiler, stream=stdout)
    stats.strip_dirs()
    stats.sort_stats("cumulative")
    stats.print_stats()

```

I'd be curious if you see similar results.

MightyJosip:
Closed in favor of #2551

PR #2525, opened 2023-10-22
Add docs for "devicename" parameter in 'pygame.mixer.init()'

JiffyRob:
Just noticed that this parameter wasn't documented so I asked around in the discord server to see what it does.  You can use `pygame._sdl2.audio.get_audio_device_names()` to get the valid names for it but since the module is hidden I just mentioned that it changes based on device.  Maybe in a future PR we add a public function for it?  It seems odd to have a parameter just dangle like this.

### DIFF
--- a/docs/reST/ref/mixer.rst
+++ b/docs/reST/ref/mixer.rst
@@ -67,6 +67,10 @@ change the default buffer by calling :func:`pygame.mixer.pre_init` before
    playback. The buffer size must be a power of two (if not it is rounded up to
    the next nearest power of 2).
 
+   The devicename parameter is the name of sound device to open for audio
+   playback.  Allowed device names will vary based on the host system.
+   If left as ``None`` then a sensible default will be chosen for you.
+
    Some platforms require the :mod:`pygame.mixer` module to be initialized
    after the display modules have initialized. The top level ``pygame.init()``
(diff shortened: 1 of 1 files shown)


docs/reST/ref/mixer.rst
```diff
@@ -67,6 +67,10 @@ change the default buffer by calling :func:`pygame.mixer.pre_init` before
    playback. The buffer size must be a power of two (if not it is rounded up to
    the next nearest power of 2).
 
+   The devicename parameter is the name of sound device to open for audio
+   playback.  Allowed device names will vary based on the host system.
```
Matiiss, line 71: Minor nitpick but does this mean that if I use WSL, the device name will be based off Windows instead of Linux as the "host" system would be Windows? Maybe just omit the word "host"?

Matiiss (commented):
Looks fine to me otherwise.

JiffyRob:
Usually I'd say "device" but I figured that would be kind of confusing.  Writing just system with no specifier also seems very ambiguous.  What system?

MyreMylar (approved):
LGTM 👍 

Better to have some words about this rather than none. I agree that it is kind of lacking without a documented, non-experimental way to get the list of system audio device names SDL uses - but that is for the future,

ankith26 (approved):

Issue #1758, opened 2023-02-12
Make pygame.event.custom_type() documentation more clear (3569)


GalacticEmperor1:
#  **Issue &numero;3569 opened by *[Felix3qH4](https://github.com/Felix3qH4)* at 2022-11-16 19:48:48**

**Description**

I find the documentation for [pygame.event.custom_type()](https://www.pygame.org/docs/ref/event.html# pygame.event.custom_type) pretty confusing as it is now.

make custom user event type
custom_type() -> int
Reserves a pygame.USEREVENT for a custom use.

If too many events are made a [pygame.errorstandard pygame exception](https://www.pygame.org/docs/ref/pygame.html# pygame.error) is raised.

I think it would be better this way:

custom_type() -> int
Returns an integer value (pygame.USEREVENT) to which no event has yet been assigned to.
If too many events are made a [pygame.errorstandard pygame exception](https://www.pygame.org/docs/ref/pygame.html# pygame.error) is raised.
<hr>

#  Comments

# # #  **[oddbookworm](https://github.com/oddbookworm)* commented at 2022-11-16 23:43:54*

Yours doesn’t mention that the returned int is now reserved. For all pygame cares, that int now has an event assigned to it. So, the current doc is correct in that the returned int becomes reserved whether you actually use it or not
<hr>

# # #  **[Felix3qH4](https://github.com/Felix3qH4)* commented at 2022-11-17 13:07:15*

That's true so it would maybe be better to just add a sentence to the existing documentation to clarify what it does.

PR #2527, opened 2023-10-23
changed to wording in order to make it more clear

JorasOliveira:
I tried to keep in mind the original suggestion made by @Felix3qH4 and the correction made by @oddbookworm. Hopefully the behaviour of pygame.event.custom_type() is clearer now.

### DIFF
--- a/docs/reST/ref/event.rst
+++ b/docs/reST/ref/event.rst
@@ -455,7 +455,7 @@ On Android, the following events can be generated
    | :sl:`make custom user event type`
    | :sg:`custom_type() -> int`
 
-   Reserves an event slot for use in a custom event ``pygame.USEREVENT`` and returns the integer that the event slot is reserved to.
+   Reserves an event slot for use in a custom event ``userevent`` and returns the integer that the event slot is reserved to.
 
    If too many events are made a :exc:`pygame.error` is raised.
 
--- a/docs/reST/ref/event.rst
+++ b/docs/reST/ref/event.rst
@@ -455,7 +455,8 @@ On Android, the following events can be generated
    | :sl:`make custom user event type`
    | :sg:`custom_type() -> int`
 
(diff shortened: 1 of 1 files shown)


oddbookworm:
The issue this is meant to address for convenience of reviewers: #1758

docs/reST/ref/event.rst
```diff
@@ -455,7 +455,8 @@ On Android, the following events can be generated
    | :sl:`make custom user event type`
    | :sg:`custom_type() -> int`
 
-   Reserves a ``pygame.USEREVENT`` for a custom use.
+   Returns an integer value that represents a reserved ``pygame.USEREVENT`` for a custom use.
```
oddbookworm, line 458: Hmmmmm, this doesn't really seem any clearer IMO. Maybe something alone the lines of

"Reserves an event slot for use in a custom event (``pygame.USEREVENT``) and returns the integer that the event slot is reserved to"

Felix3qH4: I think @oddbookworm solution is easier to understand.

MyreMylar (approved):
LGTM 👍

docs/reST/ref/event.rst
```diff
@@ -455,7 +455,7 @@ On Android, the following events can be generated
    | :sl:`make custom user event type`
    | :sg:`custom_type() -> int`
 
-   Reserves a ``pygame.USEREVENT`` for a custom use.
+   Reserves an event slot for use in a custom event ``pygame.USEREVENT`` and returns the integer that the event slot is reserved to.
```
ankith26, line 458: On the topic of clarifications, I believe we can replace `pygame.USEREVENT` with "userevent". Because `pygame.USEREVENT` is an event type by itself.

MyreMylar (approved):

ankith26 (approved):

Issue #2523, opened 2023-10-18
Segfault when draw on display surface after display window is destroyed


yunline:
Platform: Windows 10
```py
import pygame
from pygame._sdl2 import video

pygame.init()

sf = pygame.display.set_mode((640,480))

win = video.Window.from_display_module()
win.destroy()

sf.fill((0,0,0))
```
```
pygame-ce 2.4.0.dev3 (SDL 2.26.5, Python 3.10.6)
Fatal Python error: pygame_parachute: (pygame parachute) Segmentation Fault
Python runtime state: initialized

Current thread 0x00001a60 (most recent call first):
  File "x:\python\pygame_test\window_surface.py", line 11 in <module>

Extension modules: pygame.base, pygame.constants, pygame.rect, pygame.rwobject, pygame.surflock, pygame.bufferproxy, pygame.math, pygame.surface, pygame._window, pygame.display, pygame.draw, pygame.joystick, pygame.event, pygame.imageext, pygame.image, pygame.key, pygame.mouse, pygame.time, pygame.mask, pygame.pixelcopy, pygame.transform, pygame.font, pygame.mixer_music, pygame.mixer, pygame.scrap, numpy.core._multiarray_umath, numpy.core._multiarray_tests, numpy.linalg._umath_linalg, numpy.fft._pocketfft_internal, numpy.random._common, numpy.random.bit_generator, numpy.random._bounded_integers, numpy.random._mt19937, numpy.random.mtrand, numpy.random._philox, numpy.random._pcg64, numpy.random._sfc64, numpy.random._generator, pygame.system, pygame._sdl2.sdl2, pygame._sdl2.audio, pygame._sdl2.video, pygame._freetype (total: 43)
PS X:\python\pygame_test>
```

novialriptide:
Unreproduceable on Apple Silicon (M2 Max).
`pygame-ce 2.4.0.dev3 (SDL 2.28.4, Python 3.11.5)`

ankith26:
Can reproduce on Ubuntu 23.10
`pygame-ce 2.4.0.dev3 (SDL 2.26.5, Python 3.11.6)`

novialriptide:
I got a virtual machine running Ubuntu 22.04.3, I'll give it my best shot on this segfault.

novialriptide:
`pg_MappedColorFromObj()` is the one that's actually segfaulting, not `surf_fill()`.

PR #2530, opened 2023-10-26
Fix segmentation fault when destroying window

novialriptide:
Fixes: https://github.com/pygame-community/pygame-ce/issues/2523

### DIFF
--- a/src_c/surface.c
+++ b/src_c/surface.c
@@ -2465,7 +2465,6 @@ _PgSurface_SrcAlpha(SDL_Surface *surf)
 static int
 _is_pg_window_active()
 {
-    pgSurfaceObject *default_win_surface = pg_GetDefaultWindowSurface();
     SDL_Surface *win_surface;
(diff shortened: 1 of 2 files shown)


ankith26 (commented):
I'm not really sure this is the way to proceed on fixing this issue. In general, we allow surface operations even when there is no display surface active. Adding these checks might break existing working code

novialriptide:
@ankith26 Yeah... I think it's been improved now

MyreMylar (approved):
LGTM 👍 

Fixes the issue in my testing. You now get a:

```
  File "C:\Users\dan\Programming\scrap\window_destroy_segfault.py", line 15, in <module>
    sf.fill((0, 0, 0))
pygame.error: Surface is not initialized
```

When trying any surface operations on the display surface after destroying the window, which seems fair enough to me.

pmp-p (approved):

PR #2532, opened 2023-10-26
De-emphasise passing a list of rectangles to update

### LINKED ISSUE #2514: Thoroughly test performance of pygame.display.update(list_of_rects) vs pygame.display.flip()
This was a performance optimisation that was relevant twenty plus years ago, but with the passage of time, and the change to modern versions of SDL - my intuition is that it is no longer providing any performance benefit. The idea of it being effective is baked into a lot of pygame-ce functionality (with drawing and rendering functions returning rectangles for use in this performance optimisation). If we could prove it was, or was not, effective in the modern world we could make better decisions on what to optimise in the code, document in the docs and include in tutorials.

However, we should test it before taking any action. A to do list.

- [x] create a base test program with a standard game loop that draws/blits something to about about 25% of the display surface every loop. Rather than a while loop just have it loop something like 100 times.
- [x] split the base test program into two versions - one that flips the display surface to the screen with `pygame.display.flip()`, and one that collects all the rectangles returned from the draws/blits into a list of rectangles and passes that list to `pygame.display.update()`.
- [x] test the performance of the two programs and see which is faster/slower/whether it makes any difference at all.
- [x] present the results and see if anyone else has any ideas for changes to the performance test that might change the results.
- [x] open a PR to update the docs & tutorials regarding the usefulness (or not) of the rect_list - unless anyone else has any better uses for it?

MyreMylar:
fixes #2514

This is just a series of documentation changes - mostly to tutorials to stop emphasising passing a list of rectangles to `pygame.display.update()` as a great way to boost performance. This is no longer nearly universally true, in fact in most normal game cases it will be slower to pass a list of rectangles than just to update the whole screen in one calls. 

So we should stop giving out bad advice.

### DIFF
--- a/docs/reST/ref/display.rst
+++ b/docs/reST/ref/display.rst
@@ -253,9 +253,7 @@ required).
    | :sl:`Update the full display Surface to the screen`
    | :sg:`flip() -> None`
 
-   This will update the contents of the entire display. If your display mode is
-   using the flags ``pygame.HWSURFACE`` and ``pygame.DOUBLEBUF`` on pygame 1,
-   this will wait for a vertical retrace and swap the surfaces.
+   This will update the contents of the entire display.
 
    When using an ``pygame.OPENGL`` display mode this will perform a gl buffer
    swap.
@@ -264,23 +262,31 @@ required).
 
 .. function:: update
 
-   | :sl:`Update portions of the screen for software displays`
+   | :sl:`Update all, or a portion, of the display. For non-OpenGL displays.`
    | :sg:`update(rectangle=None) -> None`
    | :sg:`update(rectangle_list) -> None`
 
-   This function is like an optimized version of ``pygame.display.flip()`` for
-   software displays. It allows only a portion of the screen to be updated,
-   instead of the entire area. If no argument is passed it updates the entire
-   Surface area like ``pygame.display.flip()``.
+   For non OpenGL display Surfaces, this function is very similar to
+   ``pygame.display.flip()`` with an optional parameter that allows only
+   portions of the display surface to be updated, instead of the entire area.
+   If no argument is passed it updates the entire Surface area like
+   ``pygame.display.flip()``.
 
-   Note that calling ``display.update(None)`` means no part of the window is
-   updated. Whereas ``display.update()`` means the whole window is updated.
+   .. note:: calling ``display.update(None)`` means no part of the window is
+             updated. Whereas ``display.update()`` means the whole window is
+             updated.
 
    You can pass the function a single rectangle, or a sequence of rectangles.
-   It is more efficient to pass many rectangles at once than to call update
-   multiple times with single or a partial list of rectangles. If passing a
-   sequence of rectangles it is safe to include None values in the list, which
-   will be skipped.
+   Generally you do not want to pass a sequence of rectangles as there is a
+   performance cost per rectangle passed to the function. On modern hardware,
+   after a very small number of rectangles passed in, the per-rectangle cost
+   will exceed the saving of updating less pixels. In most applications it is
+   simply more efficient to update the entire display surface at once, it also
+   means  you do not need to keep track of a list of rectangles for each call
+   to update.
+
+   If passing a sequence of rectangles it is safe to include None
+   values in the list, which will be skipped.
 
    This call cannot be used on ``pygame.OPENGL`` displays and will generate an
    exception.
--- a/docs/reST/tutorials/en/intro-to-sprites.rst
+++ b/docs/reST/tutorials/en/intro-to-sprites.rst
@@ -220,10 +220,11 @@ sprite module.
 
   :class:`RenderUpdates <pygame.sprite.RenderUpdates>`
 
-    This is the Cadillac of rendering ``Groups``. It is inherited from
-    ``RenderClear``, but changes the ``draw()`` method to also return a list of
-    pygame ``Rects``, which represent all the areas on screen that have been
-    changed.
+    This group is inherited from ``RenderClear``, but changes the
+    ``draw()`` method to also return a list of pygame ``Rects``,
+    which represent all the areas on screen that have been changed.
+    Generally you don't need to use this group, but it is included for
+    completeness.
 
 That is the list of different groups available We'll discuss more about these
 rendering groups in the next section. There's nothing stopping you from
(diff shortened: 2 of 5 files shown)


docs/reST/ref/display.rst
```diff
@@ -253,9 +253,7 @@ required).
    | :sl:`Update the full display Surface to the screen`
    | :sg:`flip() -> None`
 
-   This will update the contents of the entire display. If your display mode is
-   using the flags ``pygame.HWSURFACE`` and ``pygame.DOUBLEBUF`` on pygame 1,
-   this will wait for a vertical retrace and swap the surfaces.
```
MyreMylar, line 258: This stuff about pygame and SDL 1 is no longer relevant as we don't build against SDL 1 any more and haven't for a long time.

docs/reST/ref/display.rst
```diff
@@ -264,23 +262,31 @@ required).
...
+   ``pygame.display.flip()`` with an optional parameter that allows only
+   portions of the display surface to be updated, instead of the entire area.
+   If no argument is passed it updates the entire Surface area like
+   ``pygame.display.flip()``.
 
-   Note that calling ``display.update(None)`` means no part of the window is
-   updated. Whereas ``display.update()`` means the whole window is updated.
+   .. note:: calling ``display.update(None)`` means no part of the window is
+             updated. Whereas ``display.update()`` means the whole window is
+             updated.
```
MyreMylar, line 277: Just updating this to be a proper `.. note::` rather than just saying "Note"

docs/reST/tutorials/en/intro-to-sprites.rst
```diff
@@ -241,33 +242,9 @@ The Rendering Groups
...
-The ``RenderUpdates`` class also does a good job at minimizing overlapping
-areas in the list of updated rectangles. If the previous position and current
-position of an object overlap, it will merge them into a single rectangle.
-Combined with the fact that it properly handles deleted objects, this is
-one powerful ``Group`` class. If you've written a game that manages the changed
-rectangles for the objects in a game, you know this the cause for a lot of
-messy code in your game. Especially once you start to throw in objects that can
-be deleted at any time. All this work is reduced to a ``clear()`` and
-``draw()`` method with this monster class. Plus with the overlap checking, it
-is likely faster than when you did it manually.
```
MyreMylar, line 270: This stuff reminds me I need to get on with my sprite module refactorings as we seem to have a lot of code also dedicated to this old optimisation feature.

docs/reST/tutorials/en/newbie-guide.rst
```diff
@@ -358,35 +359,30 @@ will avoid this.
 Colorkey vs. Alpha.
 -------------------
 
-There's a lot of confusion around these two techniques, and much of it comes
+There's a lot of confusion around these two terms, and much of it comes
```
MyreMylar, line 362: Just because 'Alpha' isn't really a technique on it own, color key transparency is a complete technique.

docs/reST/tutorials/en/newbie-guide.rst
```diff
@@ -358,35 +359,30 @@ will avoid this.
...
-'Per-pixel alpha' is the other flavor of alpha, and it's more complicated.
-Basically, each pixel in the source image has its own alpha value, from 0 to
-255.  Each pixel, therefore, can have a different opacity when blitted onto a
-background.  This type of alpha can't be mixed with colorkey blitting,
-and it overrides per-image alpha.  Per-pixel alpha is rarely used in
-games, and to use it you have to save your source image in a graphic
-editor with a special *alpha channel*.  It's complicated -- don't use it
-yet.
+transparent, and 255 is completely opaque. Note that surface alpha and alpha
+blitting can be combined.
```
MyreMylar, line 385: Flipping the advice completely here because color key sucks in 2024. It has no-SIMD acceleration, it can't do any nice feathering - it is slow and bad. All the common image tools support alpha channels well now so we should stop promoting color key like it is 1999.

docs/reST/tutorials/en/newbie-guide.rst
```diff
@@ -407,7 +403,12 @@ you start due to somewhat nebulous organizational problems.
 
 This brings us to the concept of software architecture and design patterns. You
 may be familiar with pygame's "standard" base template (there are many equivalent
-variations of this, so don't stress about the small details too much)::
+variations of this, so don't stress about the small details too much):
+
+.. code-block:: python
+   :caption: Standard Pygame game loop
+   :name: game_loop.py
+   :linenos:
```
MyreMylar, line 411: I just think this modern style of .rst code block is neat so I updated it while I was in here. I may also have tweaked the standard game loop pattern a bit.

docs/reST/tutorials/en/newbie-guide.rst
```diff
@@ -358,35 +359,30 @@ will avoid this.
...
-transparent instead of black.
-
-'Alpha' is different, and it comes in two flavors. 'Image alpha' applies to the
-whole image, and is probably what you want.  Properly known as 'translucency',
-alpha causes each pixel in the source image to be only *partially* opaque.
-For example, if you set a surface's alpha to 192 and then blitted it onto a
+and so don't obscure the background.  Simply call :meth:`.Surface.set_colorkey()`,
+and pass in an RGB tuple -- say (0,0,0). This would make every pixel in the source
+image transparent instead of black. It used to be significantly cheaper to blit
+than per pixel alpha, but this is no longer the case and in mayn cases it is
```
ankith26, line 376: typo. mayn -> many

docs/reST/tutorials/en/newbie-guide.rst
```diff
@@ -435,6 +437,9 @@ variations of this, so don't stress about the small details too much)::
         pygame.display.flip()  # Refresh on-screen display
         clock.tick(60)         # wait until next frame (at 60 FPS)
 
+    pygame.quit()
+    raise SystemExit
```
ankith26, line 441: This line isn't really needed

docs/reST/tutorials/en/newbie-guide.rst
MyreMylar, line 376:
```suggestion
than per pixel alpha, but this is no longer the case and in many cases it is
```

MyreMylar: Thanks 👍

docs/reST/tutorials/en/newbie-guide.rst
MyreMylar, line 441:
```suggestion
```

MyreMylar: Sure, I never use it either - I just copied it from the original while adjusting this code slightly.

ankith26 (approved):

docs/reST/tutorials/en/intro-to-sprites.rst
dr0id: yes :)

src_c/doc/display_doc.h
```diff
@@ -6,7 +6,7 @@
 #define DOC_DISPLAY_SETMODE "set_mode(size=(0, 0), flags=0, depth=0, display=0, vsync=0) -> Surface\nInitialize a window or screen for display"
 #define DOC_DISPLAY_GETSURFACE "get_surface() -> Surface\nGet a reference to the currently set display surface"
 #define DOC_DISPLAY_FLIP "flip() -> None\nUpdate the full display Surface to the screen"
-#define DOC_DISPLAY_UPDATE "update(rectangle=None) -> None\nupdate(rectangle_list) -> None\nUpdate portions of the screen for software displays"
+#define DOC_DISPLAY_UPDATE "update(rectangle=None) -> None\nupdate(rectangle_list) -> None\nUpdate all, or a portion, of the display. For non-OpenGL displays."
```
dr0id, line 9:
```c
None\nUpdate all, or a portion, of the display. For non-OpenGL displays."
```
The dot between `display. For` looks odd.

dr0id (approved):
LGTM

Issue #2476, opened 2023-09-30
SIMD-ifying the transform submodule


MyreMylar:
We have pretty much reached peak current-gen SIMD-ifying of all the blitting in pygame-ce now. 

The next obvious place to take these efforts is the transform submodule as that also tends to do pixel by pixel changes to large blocks of pixels. SIMD makes everything faster and pretty much every platform we are on has SSE2 equivalent (NEON on Arm) level SIMD capability, often allowing us to do math operations on at least one entire pixel at a time (4 channels) rather than per channel of a pixel - and sometimes even multiple pixels at the same time.

Their are two current-gen tiers of SIMD we are pursuing:

 - Base level - SSE2/NEON - Available almost everywhere pygame-ce is deployed (100% on the steam hardware survey).
 - Enhanced - AVX2 - 91.24% overall availability on the steam hardware survey (majority Windows (92%) & Linux (93%), only at 31% on Mac).

Basic support for adding SIMD versions of transform functions has already been added.

These are the current possibly SIMD-able transforms available in the sub-module as a checklist: 

- [ ] pygame.transform.rotate | rotate an image
- [ ] pygame.transform.rotozoom | filtered scale and rotation
- [ ] pygame.transform.scale2x | specialized image doubler (Is this still relevant? - needs performance metrics, looks like vendored scalar code)
- [ ] pygame.transform.smoothscale | scale a surface to an arbitrary size smoothly - #2473
- [ ] pygame.transform.smoothscale_by | resize to new resolution, using scalar(s) - #2473
- [ ] pygame.transform.chop | gets a copy of an image with an interior area removed
- [ ] pygame.transform.laplacian | find edges in a surface
- [ ] pygame.transform.box_blur | blur a surface using box blur
- [ ] pygame.transform.gaussian_blur | blur a surface using gaussian blur
- [ ] pygame.transform.average_surfaces | find the average surface from many surfaces.
- [ ] pygame.transform.average_color | finds the average color of a surface
- [ ] pygame.transform.invert | inverts the RGB elements of a surface
- [ ] pygame.transform.grayscale | grayscale a surface - #2432
- [ ] pygame.transform.threshold | finds which, and how many pixels in a surface are within a threshold of a 'search_color' or a 

These functions are probably not SIMD-able (please tell me if you think I'm wrong):

- pygame.transform.flip | flip vertically and horizontally - no math, just mem copying.
[...]

PR #2534, opened 2023-10-27
Add SIMD versions of the invert transform

MyreMylar:
Related to #2476.

This adds SSE2/Neon and AVX2 versions of the invert transform.

In quick performance testing the SSE2 version is about 12 times faster at Walt inversions, and AVX2 is about 13 times faster. I'd expect AVX2 performance to do better at larger image sizes. Though; there is only one operation in the invert transform and SSE2 can do four pixels at a time so there isn't as much in it as in some operations.


src_c/simd_transform_avx2.c
```diff
@@ -42,3 +42,98 @@ pg_avx2_at_runtime_but_uncompiled()
...
+        remaining_pixels_batch_counter = remaining_pixels;
+        while (perfect_8_pixels_batch_counter--) {
+            mm256_src = _mm256_loadu_si256(srcp256);
+            mm256_alpha = _mm256_subs_epu8(mm256_src, mm256_rgb_mask);
+
+            /* do the invert */
+            mm256_dst = _mm256_subs_epu8(mm256_two_five_fives, mm256_src);
+
+            mm256_dst = _mm256_subs_epu8(mm256_dst, mm256_alpha_mask);
+            mm256_dst = _mm256_adds_epu8(mm256_dst, mm256_alpha);
```
itzpr3d4t0r, line 104: Instead of this "save the alpha, subtract and readd it" procedure, wouldn't it be simpler to just use the blendv instruction?
```suggestion
            mm256_src = _mm256_loadu_si256(srcp256);

            /* do the invert */
            mm256_dst = _mm256_subs_epu8(mm256_two_five_fives, mm256_src);

            mm256_dst = _mm256_blendv_epi8(mm256_dst, mm256_src, mm256_alpha_mask);
```

MyreMylar: Neat. Thanks for the tip on this instruction, it works out slightly faster and less code. 👍

MyreMylar:
I've been quick visual and performance testing with this test program:

```python
import pygame
import timeit


def invert_walt(walt_image):
    return pygame.transform.invert(walt_image)


pygame.init()


display_surf = pygame.display.set_mode((800, 350))


walt_opaque = walt.convert()
walt_inverse = invert_walt(walt)


running = True

while running:
        if event.type == pygame.QUIT:
            running = False

    display_surf.fill((255, 150, 255))
    display_surf.blit(walt, (0, 0))
    display_surf.blit(walt_inverse, (400, 0))

    pygame.display.flip()

time_taken = timeit.timeit(
    "invert_walt(walt)",
    setup="import pygame\n\n"
    "pygame.init()\n\n"
    "display_surf = pygame.display.set_mode((800, 350))\n\n"
    "walt = pygame.image.load('images/walt.png').convert_alpha()\n\n"
    "def invert_walt(walt_image): "
    "    return pygame.transform.invert(walt_image)",
    number=1000,
)
print("time taken to do 1000 walt inverts:", time_taken)
```

It is not compulsory to test with Walter - but it is traditional.

src_c/simd_transform_avx2.c
```diff
@@ -42,3 +42,94 @@ pg_avx2_at_runtime_but_uncompiled()
...
+            /* do the invert */
+            mm256_dst = _mm256_subs_epu8(mm256_two_five_fives, mm256_src);
+            /* blend the original alpha in */
+            mm256_dst =
+                _mm256_blendv_epi8(mm256_dst, mm256_src, mm256_alpha_mask);
+
+            _mm256_storeu_si256(dstp256, mm256_dst);
+
+            srcp256++;
+            dstp256++;
```
itzpr3d4t0r, line 106: To be honest I think I might have found an even better strategy for both the AVX and SSE versions, basically we could:
1) prepare the `mm256_two_five_fives` vector to have 0 for each place that's the alpha channel
2) `and` the src with the alphamask to get the `alpha` vector (this looks to take 0.33 CPI instead of 0.5 for `subs`)
3) invert the src using the newly transformed `mm256_two_five_fives` vector
4) `or` that inverted vector with the `alpha` vector  (this also takes 0.33 CPI instead of 0.5 for `adds`)

With this we should have 1 + 1 latency instead of up to 3 with blendv and 0.66 CPI instead of 1 with blendv.

Also made a littly program that illustrates the algorithm (for a single pixel)
```Py
pixel = [123, 255, 122, 23]  # ARGB
alphamask = [255, 0, 0, 0]  # A000

# calculate the 255 vector with 0 for the alpha channel
mm_255 = [0, 255, 255, 255]


def invert(pixel, mask, mm_255):
    alpha = [value & m for value, m in zip(pixel, mask)]
    inverted = [max(0, v255 - v) for v255, v in zip(mm_255, pixel)]
    result = [i | a for i, a in zip(inverted, alpha)]

    print(f"{pixel=}")
    print(f"{alpha=}")
    print(f"{inverted=}")
    print(f"{result=}")


invert(pixel, alphamask, mm_255)
```

MyreMylar: OK, I've switched over to using the bitwise operators. 

It seems to perform the same in my little test, but I understand from reading up on it that better throughput would only really shine when the CPU was very busy with stuff to do which it probably isn't inverting just this one image on my otherwise unbusy desktop. Anyway, the thing I like is that we can use basically the same three instruction code on SSE2 and AVX2. 👍

src_c/transform.c
```diff
@@ -3329,19 +3348,25 @@ invert(pgSurfaceObject *srcobj, pgSurfaceObject *dstobj)
...
-            Uint32 pixel;
-            Uint8 *pix;
-            SURF_GET_AT(pixel, src, x, y, (Uint8 *)src->pixels, src->format,
-                        pix);
-            unsigned char r, g, b, a;
-            SDL_GetRGBA(pixel, src->format, &r, &g, &b, &a);
-            Uint32 new_pixel = SDL_MapRGBA(newsurf->format, ~r, ~g, ~b, a);
-            SURF_SET_AT(new_pixel, newsurf, x, y, (Uint8 *)newsurf->pixels,
-                        newsurf->format, pix);
+    if (src->format->BytesPerPixel == 4 &&
```
Starbuck5, line 3351: Brought this up on the greyscale PR more briefly, but I think this needs to be emscripten guarded.

The emscripten guard is a low-tech way of saying "don't compile intrinsics even though the host system supports them," because emscripten is the only active time we compile on one system for a different type of system.

MyreMylar: We should probably get some emscripten CI running at some point. I'm fairly sure I recall reading at some point that most SIMD stuff _does_ run on emscripten - but not 100% of it. Seems like CI tests (& performance tests) would be a good way of teasing that out.

MyreMylar: Oh, and I added the guards.

src_c/simd_transform_avx2.c
```diff
@@ -42,3 +42,101 @@ pg_avx2_at_runtime_but_uncompiled()
...
+
+    int perfect_8_pixels_batch_counter = perfect_8_pixels;
+    int remaining_pixels_batch_counter = remaining_pixels;
+
+    Uint32 *srcp = (Uint32 *)src->pixels;
+    Uint32 *dstp = (Uint32 *)newsurf->pixels;
+
+    Uint32 rgbmask =
+        (src->format->Rmask | src->format->Gmask | src->format->Bmask);
+    Uint32 amask = ~rgbmask;
```
itzpr3d4t0r, line 72:
```suggestion
    Uint32 amask = src->format->Amask;
    Uint32 rgbmask = ~amask;
```

src_c/simd_transform_avx2.c
```diff
@@ -42,3 +42,101 @@ pg_avx2_at_runtime_but_uncompiled()
...
+    mm256_alpha_mask = _mm256_set1_epi32(amask);
+
+    __m256i _partial8_mask =
+        _mm256_set_epi32(0x00, (remaining_pixels > 6) ? 0x80000000 : 0x00,
+                         (remaining_pixels > 5) ? 0x80000000 : 0x00,
+                         (remaining_pixels > 4) ? 0x80000000 : 0x00,
+                         (remaining_pixels > 3) ? 0x80000000 : 0x00,
+                         (remaining_pixels > 2) ? 0x80000000 : 0x00,
+                         (remaining_pixels > 1) ? 0x80000000 : 0x00,
+                         (remaining_pixels > 0) ? 0x80000000 : 0x00);
```
itzpr3d4t0r, line 90: Could just be -1 and zeroes. Not a blocker but would make code cleaner.

src_c/simd_transform_sse2.c
```diff
@@ -413,4 +413,84 @@ filter_expand_Y_SSE2(Uint8 *srcpix, Uint8 *dstpix, int width, int srcpitch,
...
+    int perfect_4_pixels = pixel_batch_length / 4;
+
+    int perfect_4_pixels_batch_counter = perfect_4_pixels;
+    int remaining_pixels_batch_counter = remaining_pixels;
+
+    Uint32 *srcp = (Uint32 *)src->pixels;
+    Uint32 *dstp = (Uint32 *)newsurf->pixels;
+
+    Uint32 rgbmask =
+        (src->format->Rmask | src->format->Gmask | src->format->Bmask);
```
itzpr3d4t0r, line 438: Could just be:
```suggestion
    Uint32 rgbmask = ~src->format->Amask;
```

itzpr3d4t0r (approved):
LGTM

src_c/simd_transform_sse2.c
```diff
@@ -610,4 +610,84 @@ grayscale_sse2(SDL_Surface *src, SDL_Surface *newsurf)
...
+    Uint64 rgbmask64 = ((Uint64)rgbmask << 32) | rgbmask;
+    Uint64 amask64 = ~rgbmask64;
+
+    __m128i mm_src, mm_dst, mm_alpha, mm_rgb_invert_mask, mm_alpha_mask;
+
+    __m128i *srcp128 = (__m128i *)src->pixels;
+    __m128i *dstp128 = (__m128i *)newsurf->pixels;
+
+    mm_rgb_invert_mask = _mm_set1_epi64x(rgbmask64);
+    mm_alpha_mask = _mm_set1_epi64x(amask64);
```
Starbuck5, line 645: Rather than having an rgbmask and a rgbmask64, you could use _mm_set1_epi32 to broadcast the 32 bit out itself. This would make it more similar to the AVX2 code as well.

src_c/simd_transform_sse2.c
```diff
@@ -610,4 +610,84 @@ grayscale_sse2(SDL_Surface *src, SDL_Surface *newsurf)
...
+            mm_dst = _mm_or_si128(mm_dst, mm_alpha);
+
+            _mm_storeu_si128(dstp128, mm_dst);
+            /*dstp = 0xAARRGGBBAARRGGBBAARRGGBBAARRGGBB*/
+            srcp128++;
+            dstp128++;
+        }
+        srcp = (Uint32 *)srcp128;
+        dstp = (Uint32 *)dstp128;
+        if (remaining_pixels_batch_counter > 0) {
```
Starbuck5, line 670: Won't this only try to process the first of potentially 3 remaining pixels?

MyreMylar: Yeah, shit this is left over from when it was two pixels at a time. It  needs to be switched out for a while loop now we upgraded to 4 at a time.

Starbuck5 (approved):
Tested this out locally, gives pixel perfect results for all my test cases before and after this PR.

```py
import random
import hashlib

import pygame

random.seed(36)


expected_hashes = [
    "82c38da441d3325c0ae6fa3fffa74b8b760a26d346a9fc7441c4433efc1210c7",
    "3b05aaa0469f40bfef54207d34f48a868c3a6fa665a76f8a6a8e7f34feb47935",
    "0192eeaf93728e40ab4e11bb5c9d85302ff927b48832fb95fbaceb69e9108813",
    "e8e41e3b0c08e81f29062520b0abee1f3104fc8a90966fbd29a2b46acd097791",
    "340cf63be38ad92cb7a4d836f82d2dcedc550c14ec68c783f203fd127184f6cc",
    "35f3d8ed7268d3f3fb54ebed760e08c85b09ae98960107e3c26f559ba4cba80f",
    "3f4dc52b63a567fd77b431c67be01a8cd0a2e5c5d18961a3f8d7d26faaf46976",
    "266377ffd231505911509d03a75f6ee757375faf5e965da3aa020322209a4486",
    "6d3ec34adca4c1337e000ec339901aef45039009c3355328eca28a5ba5df55e1",
    "edc66664ce5448c742e5848cce2dc2cbaed0383a16a46093551c40c5df37e384",
    "0fecf8c40a3ba2c67c382b2900e29cea152ecd1f74d212464d0f8b883131c57e",
    "f4ad7e15493f8c3862a5aa74ba093c4ab2d6b715939c286c52d61c636d03c1e1",
]

surf_height = 5
offset = (3, 7)


def populate_surf(surf):
            surf.set_at(
                (x, y),
                (
                    random.randint(0, 255),
                    random.randint(0, 255),
                    random.randint(0, 255),
                    random.randint(0, 255),
                ),
            )


hashes = []

for pixel_width in range(4, 16):

    dest = pygame.Surface((pixel_width, surf_height), pygame.SRCALPHA)
    populate_surf(dest)

    print(dest, dest.get_pitch())

    dest = dest.premul_alpha()

    dest = pygame.transform.invert(dest)

    sha256 = hashlib.sha256()
    sha256.update(pygame.image.tobytes(dest, "RGBA"))
    digest = sha256.hexdigest()

    hashes.append(digest)
    # print(digest)

print(hashes)
```

PR #2491, opened 2023-10-02
Fix issue #2446: rotozoom keeps the colorkey flag.

igordsm:
Issue #2446 reported that using rotozoom looses the colorkey information and provided code to reproduce the issue. We modified `rotozoomSurface` to copy colorkey from `src` to `rz_dst` and added a test to `test/transform_test.py`.

Closes #2446 

**Before**:
image

**After**:

image

PR #2535, opened 2023-10-27
Fix RLE usage in the transform module

### LINKED ISSUE #1668: If you scale a surface that has alpha and a colorkey, the result's pixel values will change after it's blitted the first time (3377)
#  **Issue &numero;3377 opened by *[davidpendergast](https://github.com/davidpendergast)* at 2022-08-05 23:52:50**

It seems that calling `set_alpha`, `set_colorkey`, and then `transform.scale` on a surface produces an unstable surface whose pixels will change the first time it's blitted onto another surface. This seems like a bug, as blitting a surface onto another shouldn't normally change the source surface I would think.

Demonstration:
```py
import pygame

#  set up a red 16x16 surface with a colorkey & alpha
surf = pygame.Surface((16, 16))
surf.fill((255, 0, 0))
surf.set_colorkey((0, 0, 0))
surf.set_alpha(64)

#  scale it to the same size (size doesn't matter here)
scaled_surf = pygame.transform.scale(surf, (16, 16))

#  color should be (255, 0, 0, 255)  (and it is)
print(f"color pre-blit: {scaled_surf.get_at((8, 8))}")

#  blit it onto another surface
pygame.Surface((100, 100)).blit(scaled_surf, (0, 0))

#  color should still be (255, 0, 0, 255), but it's actually (63, 0, 0, 255) now.
print(f"color post-blit: {scaled_surf.get_at((8, 8))}")
```
output:
```
pygame 2.1.2 (SDL 2.0.18, Python 3.10.5)
color pre-blit: (255, 0, 0, 255)
color post-blit: (63, 0, 0, 255)
```
I observed the same results on Linux Mint (pygame 2.1.0) & Windows (pygame 2.1.0 and 2.1.2)<hr>

#  Comments

# # #  **[oddbookworm](https://github.com/oddbookworm)* commented at 2022-08-05 23:59:22*

I tested this on Manjaro (current `main` branch) and Windows 11 (pygame 2.1.2 and pygame 2.1.3dev4) and observed the exact same result
<hr>

# # #  **[MyreMylar](https://github.com/MyreMylar)* commented at 2022-08-06 07:52:40*

This gives some insight into what is going on.

Expanded test program:

```
import pygame

#  set up a red 16x16 surface with a colorkey & alpha
surf = pygame.Surface((16, 16))
surf.fill((255, 0, 0))
surf.set_colorkey((0, 0, 0))
surf.set_alpha(64)

print('\nInitial Surface flags:', f'0x{surf.get_flags():08x}', '\n------------------')
print('Color Key flag:', surf.get_flags() & pygame.SRCCOLORKEY != 0)
print('Alpha flag:', surf.get_flags() & pygame.SRCALPHA != 0)
print('RLEACCELOK flag:', surf.get_flags() & pygame.RLEACCELOK != 0)
print('Run Length Encoding flag:', surf.get_flags() & pygame.RLEACCEL != 0, '\n')

#  #  scale it to the same size (size doesn't matter here)
scaled_surf = pygame.transform.flip(surf, False, False)

print('Post scale surface flags:', f'0x{scaled_surf.get_flags():08x}', '\n------------------')
print('Color Key flag:', scaled_surf.get_flags() & pygame.SRCCOLORKEY != 0)
print('Alpha flag:', scaled_surf.get_flags() & pygame.SRCALPHA != 0)
print('RLEACCELOK flag:', scaled_surf.get_flags() & pygame.RLEACCELOK != 0)
print('Run Length Encoding flag:', scaled_surf.get_flags() & pygame.RLEACCEL != 0, '\n')

#  color should be (255, 0, 0, 255)  (and it is)
print(f"color pre-blit: {scaled_surf.get_at((8, 8))} \n")

#  blit it onto another surface
temp_surf = pygame.Surface((100, 100))
temp_surf.blit(scaled_surf, (0, 0))

print('Post blit surface flags:', f'0x{scaled_surf.get_flags():08x}', '\n------------------')
print('Color Key flag:', scaled_surf.get_flags() & pygame.SRCCOLORKEY != 0)
print('Alpha flag:', scaled_surf.get_flags() & pygame.SRCALPHA != 0)
print('RLEACCELOK flag:', scaled_surf.get_flags() & pygame.RLEACCELOK != 0)
print('Run Length Encoding flag:', scaled_surf.get_flags() & pygame.RLEACCEL != 0, '\n')

#  color should still be (255, 0, 0, 255), but it's actually (63, 0, 0, 255) now.
print(f"color post-blit: {scaled_surf.get_at((8, 8))}")

```

Output:

```
Initial Surface flags: 0x00011000 
------------------
Color Key flag: True
Alpha flag: True
RLEACCELOK flag: False
Run Length Encoding flag: False 

Post scale surface flags: 0x00013000 
------------------
Color Key flag: True
Alpha flag: True
RLEACCELOK flag: True
Run Length Encoding flag: False 

color pre-blit: (255, 0, 0, 255) 

Post blit surface flags: 0x00017000 
------------------
Color Key flag: True
Alpha flag: True
RLEACCELOK flag: True
Run Length Encoding flag: True 

color post-blit: (63, 0, 0, 255)
```
<hr>

# # #  **[MyreMylar](https://github.com/MyreMylar)* commented at 2022-08-06 07:56:36*

In short. whatever happens in any transform operation is toggling on the 'RLEACCELOK' flag which gives the go ahead to SDL2 to use run length encoding on the surface next time it is blitted. 

As far as I can tell RLE isn't supposed to work with source alpha, so this causes SDL to have a breakdown and the original  surface data gets corrupted with the source alpha values. 

The likely fix is to make sure that `RLEACCELOK` flag is unset when transforming.

As far as I can see this happens with all the transforms (flip, rotate, scales) but not with a basic `surface.copy()`.
<hr>

# # #  **[MyreMylar](https://github.com/MyreMylar)* commented at 2022-08-06 08:14:03*

I suspect the issue is almost cerrtainly in `newsurf_fromsurf` in `transform.c`
<hr>

# # #  **[MyreMylar](https://github.com/MyreMylar)* commented at 2022-08-06 08:22:39*

OK, my lazy patch would be to remove one line from `transform.c` here:

https://github.com/pygame/pygame/blob/22fc48aea92ac08195dccebabaf48389ed1433b1/src_c/transform.c# L186

If you remove `|| SDL_SetSurfaceRLE(newsurf, SDL_TRUE) != 0)` then it no longer sets the RLEACCELOK flag and this bug disappears.

Perhaps someone could code up a better fix though. At the minute this code seems to be saying 'If we have a colour key on our original surface then add a colour key _and_ RLE to our new surface'. Which seems presumptuous to me. Possibly in the past RLE just always worked fine with colour keys, but now it does not. I think this should probably be changed to _only_ add RLE to the new surface if the old surface also possessed it.
<hr>

# # #  **[gresm](https://github.com/gresm)* commented at 2022-08-16 14:07:52*

Something a bit similar happens when filling a surface and rotating it.
This code:
```py
import pygame as pg
from time import sleep

pg.init()
surface = pg.Surface((100, 100))
surface.fill((255, 255, 255))
pg.draw.rect(surface, (255, 0, 0), (1, 1, 50, 50))
surface = pg.transform.rotate(surface, 45)

window = pg.display.set_mode((200, 200))
window.fill((0, 0, 0))

window.blit(surface, (0, 0))
pg.display.flip()
sleep(5)
```
Will result this image:
image
This seems like an edge case when background is not determined and the surface is not transparent, pygame will try to fill blank space with color of top-left pixel.
<hr>

# # #  **[oddbookworm](https://github.com/oddbookworm)* commented at 2022-08-17 03:35:23*

@gresm I believe that is actually intentional. The problem reported in this issue is that if you're blitting surface A onto surface B, surface A ends up being modified where the expectation is that the surface being blitted should remain unchanged.
<hr>

# # #  **[gresm](https://github.com/gresm)* commented at 2022-08-18 07:55:23*

@oddbookworm I also believe that this is intentional, but doesn't seem to be well-documented behaviour, or at least I wasn't able to find anything about it in the docs.
<hr>

# # #  **[davidpendergast](https://github.com/davidpendergast)* commented at 2022-08-18 08:17:16*

@gresm the docs on pygame.transform.rotate seem pretty clear on this:
> Unless rotating by 90 degree increments, the image will be padded larger to hold the new size. If the image has pixel alphas, the padded area will be transparent. Otherwise pygame will pick a color that matches the Surface colorkey or the topleft pixel value.

In any case though, it's completely unrelated to the problem this issue was made for.

MyreMylar:
Building on work done in #2491. Don't merge this until that has gone in.

This makes sure we don't add the RLE flag to surfaces that didn't already have it before they were transformed. This was breaking output after transforms on color key surfaces - and there is no reason for us to be doing it. 

fixes #1668

### DIFF
--- a/src_c/rotozoom.c
+++ b/src_c/rotozoom.c
@@ -12,22 +12,6 @@
 */
 
 #include "_pygame.h"
-
-#if !SDL_VERSION_ATLEAST(2, 0, 14)
-SDL_bool
-PG_SurfaceHasRLE(SDL_Surface *surface)
-{
-    if (surface == NULL) {
-        return SDL_FALSE;
-    }
-
-    if (!(surface->map->info.flags & SDL_COPY_RLE_DESIRED)) {
-        return SDL_FALSE;
-    }
-
-    return SDL_TRUE;
-}
-#endif
 #define NO_PYGAME_C_API
 #include "pygame.h"
 
@@ -46,6 +30,23 @@ typedef struct tColorRGBA {
 #endif
 #ifndef M_PI
 #define M_PI 3.141592654
+#endif
+
+#if !SDL_VERSION_ATLEAST(2, 0, 14)
+// Remove this when our minimum version is 2.0.14 or larger
+SDL_bool
+PG_SurfaceHasRLE(SDL_Surface *surface)
+{
+    if (surface == NULL) {
+        return SDL_FALSE;
+    }
+
+    if (!(surface->map->info.flags & SDL_COPY_RLE_DESIRED)) {
+        return SDL_FALSE;
+    }
+
+    return SDL_TRUE;
+}
 #endif
 
 /*
--- a/src_c/rotozoom.c
+++ b/src_c/rotozoom.c
@@ -511,6 +511,7 @@ rotozoomSurface(SDL_Surface *src, double angle, double zoom, int smooth)
     int dstwidth, dstheight;
     int is32bit;
     int src_converted;
+    Uint32 colorkey;
 
     /*
      * Sanity check
@@ -556,6 +557,7 @@ rotozoomSurface(SDL_Surface *src, double angle, double zoom, int smooth)
         /*
          * -----------------------
          */
+
         int dstwidthhalf, dstheighthalf;
         double sanglezoom, canglezoom, sanglezoominv, canglezoominv;
 
@@ -583,6 +585,13 @@ rotozoomSurface(SDL_Surface *src, double angle, double zoom, int smooth)
          * Target surface is 32bit with source RGBA/ABGR ordering
          */
         rz_dst = PG_CreateSurface(dstwidth, dstheight, rz_src->format->format);
+        if (SDL_GetColorKey(src, &colorkey) == 0) {
+            if (SDL_SetColorKey(rz_dst, SDL_TRUE, colorkey) != 0 ||
+                SDL_SetSurfaceRLE(rz_dst, SDL_TRUE) != 0) {
+                SDL_FreeSurface(rz_dst);
+                return NULL;
+            }
+        }
 
         /*
          * Lock source surface
@@ -629,6 +638,13 @@ rotozoomSurface(SDL_Surface *src, double angle, double zoom, int smooth)
          * Target surface is 32bit with source RGBA/ABGR ordering
          */
         rz_dst = PG_CreateSurface(dstwidth, dstheight, rz_src->format->format);
+        if (SDL_GetColorKey(src, &colorkey) == 0) {
+            if (SDL_SetColorKey(rz_dst, SDL_TRUE, colorkey) != 0 ||
+                SDL_SetSurfaceRLE(rz_dst, SDL_TRUE) != 0) {
+                SDL_FreeSurface(rz_dst);
+                return NULL;
+            }
+        }
 
         /*
          * Lock source surface
--- a/src_c/rotozoom.c
+++ b/src_c/rotozoom.c
@@ -557,7 +557,6 @@ rotozoomSurface(SDL_Surface *src, double angle, double zoom, int smooth)
         /*
          * -----------------------
          */
-
         int dstwidthhalf, dstheighthalf;
         double sanglezoom, canglezoom, sanglezoominv, canglezoominv;
 
--- a/src_c/rotozoom.c
+++ b/src_c/rotozoom.c
@@ -12,6 +12,22 @@
 */
 
 #include "_pygame.h"
+
+#if !SDL_VERSION_ATLEAST(2, 0, 14)
+SDL_bool
+PG_SurfaceHasRLE(SDL_Surface *surface)
+{
+    if (surface == NULL) {
+        return SDL_FALSE;
+    }
+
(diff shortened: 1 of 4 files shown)


src_c/rotozoom.c
```diff
@@ -48,6 +32,23 @@ typedef struct tColorRGBA {
...
+        return SDL_FALSE;
+    }
+
+    if (!(surface->map->info.flags & SDL_COPY_RLE_DESIRED)) {
+        return SDL_FALSE;
+    }
+
+    return SDL_TRUE;
+}
+#endif
```
ankith26, line 50: Is there a reason this block of code was moved around, or was it unintentional?

ankith26 (commented):
Also a simple test case for this would be nice

src_c/rotozoom.c
MyreMylar: Yes. I moved it intentionally, because I added it in a bad position in the former, related PR for rotozoom.

Previously it was in the middle of the includes, now it is below them.

ankith26 (approved):

(Needs a run of black on the test file to make the linter happy however)

PR #2537, opened 2023-10-28
Update `pygame.version` to not be an autogen file

ankith26:
The idea here is to simplify buildconfig a tiny bit.

Instead of having `pygame.version` auto-generated by `setup.py` on every build (by doing a hacky lookup into C sources), the version information would be exported via `pygame.base` as a regular python string.

Also improved tests a bit.

### DIFF
--- a/src_c/base.c
+++ b/src_c/base.c
@@ -2275,6 +2275,17 @@ MODINIT_DEFINE(base)
         goto error;
     }
 
+    PyObject *version =
+        PyUnicode_FromFormat("%d.%d.%d.%s", PG_MAJOR_VERSION, PG_MINOR_VERSION,
+                             PG_PATCH_VERSION, PG_VERSION_TAG);
+    if (!version) {
+        goto error;
+    }
+    if (PyModule_AddObject(module, "__version__", version)) {
+        Py_DECREF(version);
+        goto error;
+    }
+
     /*some initialization*/
     PyObject *quit = PyObject_GetAttrString(module, "quit");
     PyObject *rval;
--- a/src_c/base.c
+++ b/src_c/base.c
@@ -2276,7 +2276,7 @@ MODINIT_DEFINE(base)
     }
 
     PyObject *version =
-        PyUnicode_FromFormat("%d.%d.%d.%s", PG_MAJOR_VERSION, PG_MINOR_VERSION,
+        PyUnicode_FromFormat("%d.%d.%d%s", PG_MAJOR_VERSION, PG_MINOR_VERSION,
                              PG_PATCH_VERSION, PG_VERSION_TAG);
     if (!version) {
         goto error;
--- a/src_py/__init__.py
+++ b/src_py/__init__.py
@@ -141,7 +141,7 @@ import pygame.math
 Vector2 = pygame.math.Vector2
 Vector3 = pygame.math.Vector3
 
-__version__ = ver
+from pygame.base import __version__
 
 # next, the "standard" modules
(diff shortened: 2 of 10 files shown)


src_py/__init__.py
```diff
@@ -141,7 +141,7 @@ def warn(self):
 Vector2 = pygame.math.Vector2
 Vector3 = pygame.math.Vector3
 
-__version__ = ver
+from pygame.base import __version__
```
MyreMylar, line 144: this line gives me a stub error in pycharm. I think we need to add:
`__version__: str` to `base.pyi`

src_py/version.py
```diff
@@ -53,15 +52,24 @@ def __str__(self):
...
     SDL Version class.
     """
 
+
 _sdl_tuple = get_sdl_version()
 SDL = SDLVersion(_sdl_tuple[0], _sdl_tuple[1], _sdl_tuple[2])
+
+ver = __version__  # pylint: disable=invalid-name
+vernum = PygameVersion(*map(int, ver.split(".")[:3]))
+rev = ""  # pylint: disable=invalid-name
```
MyreMylar, line 73: this variable doesn't seem to do anything, afaict it has always been an empty string. I guess we should probably use it for something one day.

src_py/version.py
```diff
@@ -26,16 +26,15 @@
 The python version information should always compare greater than any previous
 releases. (hmm, until we get to versions > 10)
 """
-from pygame.base import get_sdl_version
 
-###############
-# This file is generated with version.py.in
-##
+from pygame.base import get_sdl_version, __version__
```
MyreMylar, line 30: this line gives me a stub error in pycharm. I think we need to add:
`__version__: str` to `base.pyi`

MyreMylar (approved):
LGTM 👍 

One minor stub thing that could be fixed.

src_py/version.py
ankith26: It has been an empty string since pygame switched to git (so a long time). IIRC it has even been documented now to being an always empty string. Maybe we should use it to store our `devX`, but this is for for future discussions/PRs

ankith26:
Do we need `Circle` in the base namespace? I see Myre is making changes just to fix CI fail, but the actual issue here is that someone previously updated the stubgen file to add `Circle` to the base namespace but didn't do it in code.

ankith26:
The CI is still failing, for now I will take the easier approach of excluding the `Circle` from stubs, and it can be fixed in a future PR

MyreMylar:
> The CI is still failing, for now I will take the easier approach of excluding the `Circle` from stubs, and it can be fixed in a future PR

I think I just needed to add:

`Circle: type` to locals.pyi to finish off there.

But don't mind pulling it out of the base namespace stubgen either.

MyreMylar (approved):

src_c/base.c
```diff
@@ -2275,6 +2275,17 @@ MODINIT_DEFINE(base)
         goto error;
     }
 
+    PyObject *version =
+        PyUnicode_FromFormat("%d.%d.%d.%s", PG_MAJOR_VERSION, PG_MINOR_VERSION,
```
novialriptide, line 2279: This is untested, but what happens if `PG_VERSION_TAG` doesn't exist? Would it have an extra period at the end? (`2.4.0.`)

ankith26: Oh yeah, good point, I will fix it

novialriptide (approved):
LGTM 👍

PR #2538, opened 2023-10-30
Correct casing for macOS

novialriptide:
source: https://www.apple.com/macos/sonoma/


Starbuck5:
Are there other cases this should be changed?

novialriptide:
Yeah, give me one second.

test/font_test.py
```diff
@@ -79,12 +79,12 @@ def test_get_fonts_returns_something(self):
...
     # def XXtest_has_file_osx_10_5_sdk(self):
     #    import os
-    #    f = "/Developer/SDKs/MacOSX10.5.sdk/usr/X11/include/ft2build.h"
+    #    f = "/Developer/SDKs/macOSX10.5.sdk/usr/X11/include/ft2build.h"
     #    self.assertEqual(os.path.exists(f), True)
 
     # def XXtest_has_file_osx_10_4_sdk(self):
     #    import os
-    #    f = "/Developer/SDKs/MacOSX10.4u.sdk/usr/X11R6/include/ft2build.h"
+    #    f = "/Developer/SDKs/macOSX10.4u.sdk/usr/X11R6/include/ft2build.h"
```
ankith26, line 87: paths are case-sensitive, even though this is a comment, this change would be inaccurate

setup.py
```diff
@@ -56,7 +56,7 @@
         "Operating System :: Microsoft :: Windows",
         "Operating System :: POSIX",
         "Operating System :: Unix",
-        "Operating System :: MacOS",
+        "Operating System :: macOS",
```
ankith26, line 59: This is an incorrect tag. This must remain MacOS

MyreMylar (approved):
LGTM 👍

MyreMylar:
Though for some reason the CI doesn't like it. I'm guessing that actual script variable name change is breaking something

ankith26:
CI doesn't like it because the branch of this PR is not on this repo, and this change is triggering a docker image rebuild. I think we can safely ignore this fail, upon merging this PR, the error should not persist.

ankith26 (commented):
:+1:

ankith26 (approved):

PR #2543, opened 2023-11-03
Add the default font to sysfonts

### LINKED ISSUE #2319: Loading freesansbold through sysfont emits warning
**Environment:**

You can get some of this info from the text that pops up in the console when you run a pygame program.

- **Operating system** (e.g. Windows, Linux(Debian), Linux(Ubuntu), Mac): Windows
- **Python version** (e.g. 3.7.9, 3.8.5) : 3.9.5
- **SDL version** (e.g. SDL 2.0.12): 2.26.5
- **PyGame version** (e.g. 2.0.0.dev10, 1.9.6): 2.1.3.dev1

**Current behavior:**

Asking for the font "freesansbold" (the default font) causes the font not found warning to be emitted.

**Expected behavior:**

Since you're asking for freesansbold and getting freesansbold, there should not be a warning.

**Test code**

```py
import pygame
pygame.init()
pygame.font.SysFont("freesansbold", 24)
```

```
C:\Users\charl\AppData\Local\Programs\Python\Python39\lib\site-packages\pygame\sysfont.py:486: UserWarning: The system font 'freesansbold' couldn't be found. Did you mean: 'freesans'? Using the default font instead.
  warnings.warn(
```

MyreMylar:
fixes #2319

### DIFF
--- a/src_py/pkgdata.py
+++ b/src_py/pkgdata.py
@@ -21,25 +21,7 @@ __all__ = ["getResource"]
 import sys
 import os
 
-try:
-    from pkg_resources import resource_stream, resource_exists
-except ImportError:
-
-    def resource_exists(_package_or_requirement, _resource_name):
-        """
-        A stub for when we fail to import this function.
-
-        :return: Always returns False
-        """
-        return False
-
-    def resource_stream(_package_of_requirement, _resource_name):
-        """
-        A stub for when we fail to import this function.
-
-        Always raises a NotImplementedError when called.
-        """
-        raise NotImplementedError
+from pkg_resources import resource_stream, resource_exists
 
 
 def getResource(identifier, pkgname=__name__):
--- a/src_py/pkgdata.py
+++ b/src_py/pkgdata.py
@@ -21,7 +21,25 @@ __all__ = ["getResource"]
 import sys
 import os
 
-from pkg_resources import resource_stream, resource_exists
+try:
+    from pkg_resources import resource_stream, resource_exists
+except ImportError:
+
+    def resource_exists(_package_or_requirement, _resource_name):
+        """
+        A stub for when we fail to import this function.
+
+        :return: Always returns False
+        """
+        return False
+
+    def resource_stream(_package_of_requirement, _resource_name):
+        """
+        A stub for when we fail to import this function.
+
+        Always raises a NotImplementedError when called.
+        """
+        raise NotImplementedError
 
 
 def getResource(identifier, pkgname=__name__):
--- a/src_py/sysfont.py
+++ b/src_py/sysfont.py
@@ -363,12 +363,9 @@ def initsysfonts():
         fonts = initsysfonts_unix()
 
     # Try to add the default font to sys fonts
-    try:
-        pygame_folder = os.path.dirname(os.path.abspath(pygame_main_file))
-        default_font_path = os.path.join(pygame_folder, "freesansbold.ttf")
-        _addfont("freesansbold", True, False, default_font_path, fonts)
-    except (OSError, NotImplementedError):
-        pass
+    pygame_folder = os.path.dirname(os.path.abspath(pygame_main_file))
+    default_font_path = os.path.join(pygame_folder, "freesansbold.ttf")
+    _addfont("freesansbold", True, False, default_font_path, fonts)
 
     Sysfonts.update(fonts)
     create_aliases()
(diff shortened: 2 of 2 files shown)


MightyJosip (approved):
LGTM

dr0id (approved):
Commented on the issue #2319 
If for convenience this should be implemented then you have my approval. Maybe the other two methods should be updated accordingly to add the default font exception to their behavior.

MyreMylar:
> Commented on the issue #2319 If for convenience this should be implemented then you have my approval. Maybe the other two methods should be updated accordingly to add the default font exception to their behavior.

I just checked and after this PR 'freesansbold' appears in both `get_fonts` and `match_fonts`. Tested with:

```python
import pygame

fonts = pygame.sysfont.get_fonts()
fonts.sort()
for font in fonts:
    if font == "freesansbold":
        print(font)

found_font = pygame.sysfont.match_font("freesansbold", bold=True)
print(found_font)
```

Output:
```
pygame-ce 2.4.0.dev3 (SDL 2.26.5, Python 3.11.1)
freesansbold
C:\Users\dan\AppData\Local\Programs\Python\Python311\Lib\site-packages\pygame\freesansbold.ttf
```

src_py/sysfont.py
```diff
@@ -26,6 +26,7 @@
 from os.path import basename, dirname, exists, join, splitext
 
 from pygame.font import Font
+from pygame import __file__ as pygame_main_file
```
ankith26, line 29: Perhaps for consistency we should use `pygame.pkgdata.getResource` (which is what the C code uses to open the file) here? I see that function has a slightly different logic, so it seems like there may be cases where the return value does not match whatever this code is doing here?

MyreMylar: I borrowed this code from the strategy you employed to locate the font in the pyinstaller hook:
https://github.com/pygame-community/pygame-ce/blob/430105e4759b392040a3a83531e29f2cbe0c8f25/src_py/__pyinstaller/hook-pygame.py#L12

MyreMylar: I've changed it to using `getResource()`

src_py/sysfont.py
```diff
@@ -361,6 +362,10 @@ def initsysfonts():
     else:
         fonts = initsysfonts_unix()
 
+    # Add the default font to sys fonts
+    default_font_path = getResource("freesansbold.ttf").name
```
ankith26, line 366: Hmm, we should probably still try to resolve this to an absolute path first? And in case it is not resolvable to an absolute path, maybe skip this?

MyreMylar: isn't that what the getResource function does? It certainly returns an absolute path for me. I could nest this in a try except block if you like.

src_py/pkgdata.py
```diff
@@ -21,25 +21,7 @@ def getResource(identifier, pkgname=__name__):
...
-        return False
-
-    def resource_stream(_package_of_requirement, _resource_name):
-        """
-        A stub for when we fail to import this function.
-
-        Always raises a NotImplementedError when called.
-        """
-        raise NotImplementedError
+from pkg_resources import resource_stream, resource_exists
```
ankith26, line 24: Apparently this happens to be a deprecated module, and is actually distributed with `setuptools` (and is not an stdlib thing). Maybe we should retain the old code here

src_py/sysfont.py
MyreMylar: Actually it looks like the pkg_resources module that getResource() relies on has been removed in python 3.12 so it doesn't run. I'll rollback to the previous version.

dr0id:
> ```python
> import pygame
> 
> fonts = pygame.sysfont.get_fonts()
> fonts.sort()
> for font in fonts:
>     if font == "freesansbold":
>         print(font)
> 
> found_font = pygame.sysfont.match_font("freesansbold", bold=True)
> print(found_font)
> ```

using your code on my machine with win 10 I get:

```console
pygame-ce 2.3.2 (SDL 2.26.5, Python 3.11.2)
None

Process finished with exit code 0
```

dr0id (approved):

MyreMylar:
We should probably also raise an issue to either deprecate or update pkgdata.py to modern python APIs.

ankith26 (approved):
Okay, this current code LGTM. Thanks for the PR :tada:

PR #2547, opened 2023-11-04
Reduce amount of usage of "dummy" in codebase to minimum

### LINKED ISSUE #1218: use SDL_VIDEODRIVER=null (2377)
#  **Issue &numero;2377 opened by *[illume](https://github.com/illume)* at 2020-12-03 11:00:47**

Use SDL_VIDEODRIVER=null instead of dummy, because dummy is offensive. It should still support SDL_VIDEODRIVER=dummy, but support `null` to mean the same thing.

- [x] is `null` the right name?
- [x] In display.c like here: https://github.com/pygame/pygame/pull/2292/files
- [x] update unit tests
- [x] update docs
<hr>

#  Comments

# # #  **[ankith26](https://github.com/ankith26)* commented at 2020-12-03 13:13:28*

Hmm this is another (related) issue.
We are not setting the windib handler in the right place. The place where we are handling it right now, pygame.init has already called SDL_Init(SDL_INIT_VIDEO) 
That’s something I have not tested though, just observed from the code.

Anyways, coming to this issue, null sounds like a good alternative 
<hr>

# # #  **[robertpfeiffer](https://github.com/robertpfeiffer)* commented at 2020-12-07 09:56:43*

In the future, we might consider having a separate, non-envVar API for headless pygame (perhaps a kwarg to pygame.init as in `pygame.init(headless=True)`), perhaps without even polling the keyboard and playing sound, in which the display surface just lives in RAM without bothering the OS. In this case, _all_ modules of pygame would know that they exist in a headless scenario and act accordingly. As of right now, you could theoretically have the dummy _video_ driver paired with the evdev joystick driver and audio. A more comprehensive solution for headless pygame would cover all i/o, not just video.

For now, we might as well recognise SDL_VIDEODRIVER=null. There is no downside apart from introducing a bit of API we need to support into the future. I don't know how many people actually _use_ headless pygame, apart from myself, in teaching or production. My own use case is a headless game server that needs to init the display surface in order to load graphics and convert them. Apart from game servers, headless pygame is mainly used in testing. The currently standing recommendation seems to be to use Pillow for non-interactive command line tools that blit images, so this is a low-priority feature either way.

If we add this, I would propose handling SDL_VIDEODRIVER=headless as well. It sounds like a more descriptive name to me, but either way, I don't have strong feelings about his.
<hr>

# # #  **[Starbuck5](https://github.com/Starbuck5)* commented at 2020-12-19 01:42:07*

Does creating a window with the `pygame.HIDDEN` flag effectively do the same thing as a `SDL_VIDEODRIVER=dummy`? Like allowing pygame to be run on headless machines?
<hr>

# # #  **[robertpfeiffer](https://github.com/robertpfeiffer)* commented at 2021-01-11 16:46:55*

@Starbuck5 sorry for the late reply. It is my understanding that pygame.HIDDEN gives a hint to the display system or window manager (so wayland or gnome shell or xfwm4 or whatever they have on windows) to create the window but not show it. If the game is running on a headless server or a text-only machine without a desktop environment and without a graphics card, that would still fail.
<hr>

# # #  **[robertpfeiffer](https://github.com/robertpfeiffer)* commented at 2021-02-18 10:45:27*

I have seen SDL_VIDEODRIVER=offscreen in some projects, and there is src/video/offscreen in the SDL2 source tree, but it's not documented. If "offscreen" does what we want, we could probably close this by documenting it and updating the test cases, as "offscreen" is more descriptive than "null" or "dummy".

MyreMylar:
partially fixes #1218 - may be about as good code-side as we can do code side due to relying on external libraries which will struggle to change over things due to public API.

**Motivation**
See: https://developers.google.com/style/inclusive-documentation for one example, but there are other similar inclusive language style guides.

This guide recommends 'placeholder variable' where we might currently use 'dummy variable'. [Pandas had a long back and forth on the topic here](https://github.com/pandas-dev/pandas/issues/35724). I'm aware not everyone will agree with this but it seems a fairly easy switch to reduce our exposure on this topic.


MyreMylar:
There are a few places in the documentation where we use the term "dummy" as well which I will leave to a future PR. 

There are also instances in build config stuff I don't think we can remove because they are communicating direct with SDL.

src_c/constants.c
```diff
@@ -632,7 +635,9 @@ MODINIT_DEFINE(constants)
     DEC_CONSTS(FONT_RIGHT, 2);
 
     // https://github.com/pygame-community/pygame-ce/issues/1845
-    DEC_CONSTS(IS_CE, 1)
+    DEC_CONSTS(IS_CE, 1);
+
+    ADD_NAMED_STRING_CONST(NULL_VIDEODRIVER, "dummy");
```
dr0id, line 640: `NULL_VIDEODRIVER` not sure I like that but finding an alternative isn't easy. The reason is that `NULL` is not known in python (`None` is similar but probably not the same). 
Maybe `HEADLESS_VIDEODRIVER` or similar?

test/test_utils/png.py
```diff
@@ -440,7 +440,7 @@ def __init__(
         """
 
         # At the moment the `planes` argument is ignored;
-        # its purpose is to act as a dummy so that
+        # its purpose is to act as a null argument so that
```
dr0id, line 443: use 'placeholder argument' instead?

test/window_test.py
```diff
@@ -165,7 +165,7 @@ def test_maximum_size(self):
             ValueError, lambda: setattr(self.win, "maximum_size", (-1, -1))
         )
         self.assertRaises(
-            TypeError, lambda: setattr(self.win, "maximum_size", "dummy_str")
+            TypeError, lambda: setattr(self.win, "maximum_size", "null_str")
```
dr0id, line 168: `placeholder_str` ?

src_c/_sprite.c
dr0id, line 1: Are those changes just caused by clang formatting?

dr0id (commented):
I agreed with your intent after reading the linked documents to avoid the term 'dummy'.

examples/headless_no_windows_needed.py
```diff
@@ -14,12 +14,12 @@
 
 import os
 import sys
+import pygame
```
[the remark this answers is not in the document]
MyreMylar: before initialising yes, but not before importing

src_c/_sprite.c
MyreMylar: Oh probably. I actually have another PR that is removing this file so it may not matter depending on merge order. 

It probably got caught up in my clang formatting of everything that changed.

MyreMylar: Apparently my editor now won't not format that file so it is probably some line endings shenanigans.

test/window_test.py
MyreMylar, line 168:
```suggestion
            TypeError, lambda: setattr(self.win, "maximum_size", "placeholder_str")
```

test/test_utils/png.py
MyreMylar, line 443:
```suggestion
        # its purpose is to act as a placeholder argument so that
```

test/scrap_test.py
```diff
@@ -1,14 +1,15 @@
...
 
-if os.environ.get("SDL_VIDEODRIVER") == "dummy":
-    __tags__ = ("ignore", "subprocess_ignore")
 import unittest
 from pygame.tests.test_utils import trunk_relative_path
 
 import pygame
 from pygame import scrap
 
+if os.environ.get("SDL_VIDEODRIVER") == pygame.NULL_VIDEODRIVER:
```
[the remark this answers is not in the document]
MyreMylar: Just before `pygame.init()` (or actually display.init() which pygame.init() calls) for this one. Importing is fine, it depends bit on the Environment variable you are using.

src_py/sprite.py
```diff
@@ -372,7 +372,7 @@ class AbstractGroup:
 
     """
 
-    # dummy val to identify sprite groups, and avoid infinite recursion
+    # null val to identify sprite groups, and avoid infinite recursion
```
MyreMylar, line 375:
```suggestion
    # protected identifier value to identify sprite groups, and avoid infinite recursion
```

src_c/_sprite.c
MyreMylar: I'll just assume this file will be deleted soon and I won't worry about it.

src_c/constants.c
MyreMylar: The file it is defined in in SDL is called `SDL_nullvideo.c` if that eases your mind:

https://github.com/libsdl-org/SDL/blob/d2e005ee13a35cb6b11bb8966b491b194251a6c4/src/video/dummy/SDL_nullvideo.c

examples/headless_no_windows_needed.py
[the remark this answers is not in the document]
dr0id: you  may be right, although I think I remember that it did not work before the import either. But that might be different now.

test/scrap_test.py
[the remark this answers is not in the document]
dr0id: right, see other comment

src_c/constants.c
dr0id: hmm, but the actual driver name is defined as:

```c
#define DUMMYVID_DRIVER_NAME       "dummy"
```
but this should be replaced.

`NOOUTPUT_VIEDOEDRIVER` is probably no good.

dr0id (approved):

dr0id (commented):
maybe replace those occurrences of 'dummy' here too (search in the docs):

image

MyreMylar:
> maybe replace those occurrences of 'dummy' here too (search in the docs):
> 

I was initially going to save these for another PR, but there were actually only a few (I realised once I deleted the generated documentation). So I've bundled them in as well.

Now the only remaining instances of dummy are all related to the SDL video driver. I suspect SDL will be unable to simply rename this simply as it is a string used all over the place in other people's code. 

**Other potential ideas:**

 - reduce use in the github workflow yaml files by leveraging reusable workflows. See: https://docs.github.com/en/actions/using-workflows/reusing-workflows
 - Persuade SDL to add an alias string that adds to but doesn't replace "dummy"

dr0id (approved):

Starbuck5 (approved):

Issue #2295, opened 2023-07-08
Segfault when passing an empty StringIO to font.Font()


yunline:
**Environment:**
Windows 10
python 3.10.6
pygame 2.3.1 dev1
SDL 2.26.4
SDL_ttf 2.20.2

**Test code**
```py
import pygame
import io

pygame.init()

f=io.StringIO()

font=pygame.font.Font(f)
print(font.name)
```

```
Fatal Python error: pygame_parachute: (pygame parachute) Segmentation Fault
Python runtime state: initialized

Current thread 0x000011cc (most recent call first):
  File "x:\python\pygame_test\font_path.py", line 10 in <module>

Extension modules: pygame.base, pygame.constants, pygame.rect, pygame.rwobject, pygame.surflock, pygame.bufferproxy, pygame.math, pygame.surface, pygame.display, pygame.draw, pygame.joystick, pygame.event, pygame.imageext, pygame.image, pygame.key, pygame.mouse, pygame.time, pygame.mask, pygame.pixelcopy, pygame.transform, pygame.font, pygame.mixer_music, pygame.mixer, pygame.scrap, numpy.core._multiarray_umath, numpy.core._multiarray_tests, numpy.linalg._umath_linalg, numpy.fft._pocketfft_internal, numpy.random._common, numpy.random.bit_generator, numpy.random._bounded_integers, numpy.random._mt19937, numpy.random.mtrand, numpy.random._philox, numpy.random._pcg64, numpy.random._sfc64, numpy.random._generator, pygame.system, pygame._freetype (total: 39)
```

yunline:
I tested some other file operations in pygame, and here is the result
| code | result | 
|-|-|
|`font = pygame.font.Font(f)`<br>`print(font.name)`|  Segmentation fault |
|`from pygame import freetype as ft`<br>`ft.init()`<br>`ft.Font(f)`|  Segmentation fault |
|`pygame.image.load(f)`| Run forever (no return) |
|`pygame.mixer.Sound(f)`| Raise an exception<br>`pygame.error: Couldn't read first 4 bytes of audio data` |
|`pygame.mixer.music.load(f)`| Raise an exception<br>`pygame.error: Couldn't read first 12 bytes of audio data` |
|`pygame.cursors.load_xbm(f,f)`| Raise an exception<br>`IndexError: list index out of range` |

<details>

<summary>Test Code</summary>

```py
import pygame
import io

pygame.init()

f=io.StringIO()

[...]

PR #2371, opened 2023-08-02
Enhancement/font/check initialization

dr0id:
follow up pr for #2175, related to #1506 and #2255

PR #2548, opened 2023-11-04
Fix segfault when passing empty StringIO/File object to Font

MyreMylar:
Related to #2295

Also fixes freetype submodule's `Font` to track submodule generation on `__new__` instead of on `__init__` for more reliable performance of the unit tests with uninitialized Font objects. Previously I was getting different results running on the CI versus locally.

### DIFF
--- a/test/freetype_test.py
+++ b/test/freetype_test.py
@@ -549,12 +549,16 @@ class FreeTypeFontTest(unittest.TestCase):
         f = self._TEST_FONTS["fixed"]
         self.assertEqual(f.name, "Inconsolata")
 
-        with self.assertRaises(AttributeError or pygame.error):
+        with self.assertRaises(Exception) as cm:
             # this test raises an attribute error locally but
             # a pygame.error on the CI - I assume this is multi-threading
             # related.
             null_font = ft.Font.__new__(ft.Font)
             null_font.name
+        exception_type = type(cm.exception)
+        self.assertTrue(
+            (exception_type == pygame.error) or (exception_type == AttributeError)
+        )
 
     def test_freetype_Font_size(self):
         f = ft.Font(None, size=12)
@@ -1360,12 +1364,16 @@ class FreeTypeFontTest(unittest.TestCase):
     def test_freetype_Font_path(self):
         self.assertEqual(self._TEST_FONTS["sans"].path, self._sans_path)
 
-        with self.assertRaises(AttributeError or pygame.error):
+        with self.assertRaises(Exception) as cm:
             # this test raises an attribute error locally but
             # a pygame.error on the CI - I assume this is multi-threading
             # related.
             nullfont = ft.Font.__new__(ft.Font)
             nullfont.path
+        exception_type = type(cm.exception)
+        self.assertTrue(
+            (exception_type == pygame.error) or (exception_type == AttributeError)
+        )
 
     # This Font cache test is conditional on freetype being built by a debug
     # version of Python or with the C macro PGFT_DEBUG_CACHE defined.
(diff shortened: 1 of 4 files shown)


MyreMylar:
I also fixed two unrelated freetype module tests here that I noticed were failing locally while testing - but were not failing on the CI. 

At some point a fix that was added for a quit() and init() segfault and the way the tests multithread on the CI causes these two particular tests which I changed, which use a class level `__new__` constructor, to give one exception on the CI (that the module has been quit before the 'font' is created) and a different exception locally (the exceptions these tests were originally designed to capture). 

It is not a real functional change but it means we can get these tests to pass both locally and on the CI.

test/freetype_test.py
```diff
@@ -544,7 +549,16 @@ def test_freetype_Font_name(self):
         f = self._TEST_FONTS["fixed"]
         self.assertEqual(f.name, "Inconsolata")
 
-        self.assertRaises(RuntimeError, lambda: nullfont().name)
+        with self.assertRaises(Exception) as cm:
+            # this test raises an attribute error locally but
+            # a pygame.error on the CI - I assume this is multi-threading
```
dr0id, line 554: Are sure about this assumption?

I think I changed that in PR #2371 not sure why it behaves this way.

MyreMylar: I had another idea, about adding the freetype generation to all Font Objects - whether they succeed in loading or not - to see if that will erase the difference between CI and local for these tests.

MyreMylar: OK, with a bit of debugging it looks like the answer is that we should be setting the freetype generation for a font on `__new__` rather than on `__init__` so it lines up properly for uninitialized fonts. It worked OK when everything lined up to be zero anyway (no inits or quits of the module) but whenever that was not true this was breaking because the module generation variable was only set to anything other than 0 for initialised font objects.

I added a bunch of stuff to the error messages to debug which I'm going to unwind to make the diff smaller.

test/freetype_test.py
```diff
@@ -1349,7 +1356,10 @@ def test_freetype_Font_resolution(self):
 
     def test_freetype_Font_path(self):
         self.assertEqual(self._TEST_FONTS["sans"].path, self._sans_path)
-        self.assertRaises(pygame.error, lambda: nullfont().path)
+
+        with self.assertRaises(AttributeError):
+            nullfont = ft.Font.__new__(ft.Font)
+            nullfont.path
```
ankith26, line 1362: Just to confirm, the init generation changes + these test changes are not related to the segfault, right?

MyreMylar: Nope, it may be that my adjustments to freetype Font init exposed the issue on the CI, but it the underlying issue is unrelated. 

Basically the tests in question here are using an 'uninitialized font' which means one where you don't call __init__() and the generation variable was only being set in __init__() so unsurprisingly if anything happened with quit() at the module level the uninitialized font's generation value (which was previously always 0) would be wrong and the tests would start spitting out the module has been quit error instead of relevant errors.

ankith26 (approved):
Thanks! Approving this PR because this LGTM, but I wonder if this is something that should also be fixed upstream on `SDL_ttf` if this is an issue on their end?

ankith26 (commented):
Another improvement idea, we should use `SDL_RWsize(source)` instead of doing `source->size(source)`. Shorter and also looks more robust

dr0id (approved):

MyreMylar:
> Another improvement idea, we should use `SDL_RWsize(source)` instead of doing `source->size(source)`. Shorter and also looks more robust


> This function is available since SDL 2.0.10.
https://wiki.libsdl.org/SDL2/SDL_RWsize

PR #2549, opened 2023-11-06
Install meson+ninja on buildscripts, bump glib

ankith26:
This should serve as a base for future deps updates. After this PR, we should be able to install dependencies that support only meson (like the new glib version).


ankith26:
As a side effect of this PR, we don't need to compile `gettext` ourself anymore, and that saves CI time because gettext is a heavy-ish dependency.

Also, I'd appreciate a mac tester. To save your time I'd suggest directly testing #2494 as that PR uses this PR, then we can merge both in one after the other

MyreMylar (approved):
LGTM 👍

buildconfig/manylinux-build/docker_base/glib/build-glib.sh
```diff
@@ -3,21 +3,21 @@ set -e -x
...
-    # pass a 'cache' file while cross compiling to arm64 for glib. This is
-    # needed for glib to determine some info about the target architecture
-    export GLIB_COMPILE_EXTRA_FLAGS="--cache-file=../macos_arm64.cache"
-fi
+# a minimal glib install will do for us
+meson _build $PG_BASE_MESON_FLAGS -Dtests=false \
+    -Dselinux=disabled -Dlibmount=disabled -Ddtrace=false -Dsystemtap=false -Dnls=disabled
+
+ninja -C _build
+ninja -C _build install
```
Starbuck5, line 22: Rather than invoking ninja directly, you can go through meson?

(Untested)
```
meson setup _build $PG_BASE_MESON_FLAGS -Dtests=false -Dselinux=disabled -Dlibmount=disabled -Ddtrace=false -Dsystemtap=false -Dnls=disabled
meson install -C _build
```

Calling ninja directly seems like a strange workflow for using the build system meson-- ninja is just an implementation detail of meson.

ankith26: I more or less copied this code from the glib build instructions. As of now, ninja is the default build backend for meson. However I believe `meson install -C _build` is still not exactly equivalent to directly invoking ninja, but in our case it might still work

Starbuck5: I looked up the glib build instructions, and the one I found says to do it through meson: https://github.com/GNOME/glib/blob/main/INSTALL.md#simple-install-procedure

```
meson setup _build                      # configure the build
meson compile -C _build                 # build GLib

# Become root if necessary
meson install -C _build                 # install GLib
```

ankith26: This is now resolved

MyreMylar (approved):

Starbuck5 (approved):

PR #2506, opened 2023-10-07
Remove deprecated blend attribute from draw_aaline() & draw_aalines()

MyreMylar:
Should have removed this in 2.2.0 but it got forgotten in the shuffle. It has been deprecated for some time now, this just finally pulls it all the way out

fixes #1233

PR #2550, opened 2023-11-06
Remove blend from draw.aaline

MightyJosip:
Already done in #2506, but there was still some stuff that used blend argument.

### DIFF
--- a/src_c/draw.c
+++ b/src_c/draw.c
@@ -973,36 +973,23 @@ compare_int(const void *a, const void *b)
 
 static Uint32
 get_antialiased_color(SDL_Surface *surf, int x, int y, Uint32 original_color,
-                      float brightness, int blend)
+                      float brightness)
 {
     Uint8 color_part[4], background_color[4];
     Uint32 *pixels = (Uint32 *)surf->pixels;
     SDL_GetRGBA(original_color, surf->format, &color_part[0], &color_part[1],
                 &color_part[2], &color_part[3]);
-    if (blend) {
-        if (x < surf->clip_rect.x ||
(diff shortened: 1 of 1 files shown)


MyreMylar:
Looks like this needs running through the linters.

MyreMylar (approved):
LGTM 👍

dr0id (approved):
LGTM

src_c/draw.c
```diff
@@ -973,36 +973,23 @@ compare_int(const void *a, const void *b)
...
     Uint8 color_part[4], background_color[4];
     Uint32 *pixels = (Uint32 *)surf->pixels;
     SDL_GetRGBA(original_color, surf->format, &color_part[0], &color_part[1],
                 &color_part[2], &color_part[3]);
-    if (blend) {
-        if (x < surf->clip_rect.x ||
-            x >= surf->clip_rect.x + surf->clip_rect.w ||
-            y < surf->clip_rect.y ||
-            y >= surf->clip_rect.y + surf->clip_rect.h)
-            return original_color;
```
ankith26, line 987: I'm not sure if this check should be removed. Because at this point we want this check to have already happened. I can imagine the below line can segfault for unchecked input at this stage

ankith26 (changes requested):

src_c/draw.c
MightyJosip: Tried to reproduce segfault but wasn't able to. But anyways I will add it back just in case

ankith26 (approved):
LGTM, thanks for the PR :crown:

PR #2554, opened 2023-11-07
Attempt to make pygame.Rect() possible - DRAFT

Damus666:
The reason I'm making a draft PR is because I'm having issues with setting up a local environment, in particular with the vsvarsal.bat file. If anyone reading this want to help, I'm on the dc pygame-ce server as Damus666 (@damusss).
Making the PR should technically give me compilation and run the py tests automatically.

There are a few reasons why I think pygame.Rect() should be a thing:
- Less code, faster to type. Typing pygame.Rect(0,0,0,0) is different than pygame.Rect()
- Cleaner code. pygame.Rect() is a lot cleaner than the other version
- Consistency. If pygame.Vector2() exists, not forced to type pygame.Vector2(0,0), why shouldn't pygame.Rect() be a thing?

Where could you need to make an empty rect?
- When you want to make a rect and assign positions using the attributes
- When you need a rect you want to assign the properties of later

I've never edited the C source code in a PR (except for some PyObject *_null arguments, which i couldn't really mess up) so i could mess up a lot of things, reviews are welcome.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -43,3 +43,9 @@ dist
 *.so
 __pycache__
 _headers/*
+
+# Damus666 @damusss
+
+Scripts/
+Lib/
+Include/
--- a/mytest.py
+++ b/mytest.py
@@ -1,0 +1,5 @@
+import pygame
+
+rect = pygame.Rect()
+
+print(rect)
--- a/setup.py
+++ b/setup.py
@@ -8,6 +8,7 @@
 
 import platform
 import sysconfig
+import os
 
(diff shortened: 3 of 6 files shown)


Matiiss:
I suppose you can close this PR now? Since you have opened another one: https://github.com/pygame-community/pygame-ce/pull/2655

Damus666:
> I suppose you can close this PR now? Since you have opened another one: #2655

You are totally right, I forgot about it, Thanks!

PR #2557, opened 2023-11-09
Add meson build system [phase 1]

ankith26:
This is intended to be phase 1 of a 4 phase PR I plan to make over the months.

## Why this change?

Well, distutils is not going to be with us forever. Python 3.12 has officially removed it, though at the moment the `setuptools` package still provides it, though this could change in the future.

Most of the big C-API using python projects (numpy, scipy, etc) seem have gone the same way already, so following their footsteps should be safe.

This is intended to improve a lot of things, some benefits listed below
- hopefully smaller code footprint than old buildconfig
- dependency handling more generic, simple and correct
- give more flexibility on per file flags without much hacks
- in the long term, it should make it easier to support other build configurations

It is also worth mention a few potential and subjective downsides, and some of my own notes on this
- meson does not give as much power as python? This is an intentional meson design choice. In most places we can make do with whatever meson provides, but incase we need to do something not possible within meson, we can always write a python script, and call it from meson.
- The meson code looks very explicit? We are manually mentioning every file/folder that needs to be bundled/built when previously we just used globs? Again, intentional meson design choice. Kinda inline with pythons own "explicit is better than implicit".
- This adds a build time dependency? Can it make builds slower or harder? Yes it does add a build-time dependency. But the nice part of using `pyproject.toml` is that a single `pip wheel .`/`pip install .` will just do the right thing and handle the meson build dependency. This does add a bit of build overhead before the actual compilation, but then the actual compilation should be faster (another neat thing, the builds would now default to using all cores on a machine by default, no need for extra flags)

## Roadmap

Here is a phase-wise tentative plan, it is intentionally "slow" (it is aligned to when I get free time lol)

### Phase 1 (December 2023)
That would be this PR. The goal of this PR is to just introduce the new buildconfig, and keep its logic as identical to the old buildconfig as possible.

This PR adds support for all common platforms, that is, binary distributions on windows, mac and manylinux.

At this point the old buildconfig would be left as-is, so people could still use it if their specific workflow does not work. However, right from its introduction, the new buildconfig is going to take "precedence" over the old (by this I mean, all pip commands will stop running `setup.py`, if you need to use legacy buildconfig you would have to invoke it directly)

### Phase 2 (May 2024 or earlier)

The goal of this phase is to remove the need for legacy buildconfig for anything build-related. In this phase, the old buildconfig should start to throw deprecation warnings. Other changes here could be
- Implement sdist wheel generation
- Add support for platforms not covered in phase 1 (most notably, android and wasm) and I might need help on this.
- Implement and improve windows dependency handling in a way independent from legacy buildconfig
- cython files to be generated at build time by this point

### Phase 3 (July 2024 or earlier)

The goal of this phase is to remove the need for legacy buildconfig entirely.
- Implement the remaining legacy buildconfig/`setup.py` features like `python setup.py docs|lint|format|stubcheck`

### Phase 4 (November 2024 or earlier)

Removal of the legacy buildconfig.


Starbuck5:
Gave this a quick test on my system.

I needed to run `pip install .` from the x64 native tools command prompt, so it would pick up MSVC (cl) before my local installation of Mingw. It didn't work to just add MSVC to the path, I had to build in that entire environment.

I also noticed the addition of .lib files after the build process that makes the new installation slightly larger:
extralibs

ankith26:
> I needed to run `pip install .` from the x64 native tools command prompt, so it would pick up MSVC (cl) before my local installation of Mingw. It didn't work to just add MSVC to the path, I had to build in that entire environment.

Hmmm, perhaps in your global environment you have mingw "higher up" than msvc on PATH? Anyways I'm curious what that resulted in, I'm assuming the build failed for some reason or the other (like not finding dependencies), but I guess we could actually make it work somehow. If not in this PR maybe in a future one

> I also noticed the addition of .lib files after the build process that makes the new installation slightly larger:

Yeah I don't know what's up with those. But those should only add a few kb atmost. Any larger difference you are seeing is probably coming in due to the fact that the new buildconfig bundles the tutorials that are under the docs (the old buildconfig has a bug/typo due to which it doesn't over there)

MyreMylar (approved):
LGTM on windows. Simplifies the build process a lot, I'll be able to cut the 'Compiling on windows' wiki page down a lot.

I'm just doing:

```
pip install .
```

In the normal Pycharm terminal now and it builds fine. No need for a special 64 bit developer command prompt or all the windows specific command line stuff.

Tests all pass locally and all my GUI examples stuff also seem to work fine when locally built. 👍

ankith26:
Thanks for the approval. I could also simplify the CI config a bit, but again, as starbuck also pointed out, the 64 bit developer command prompt is required if mingw is also installed on the system. But when its not, then meson can find msvc even when not in the developer command prompt for whatever reasons

MyreMylar:
I do have mingw installed myself but I guess it is not first on the windows
PATH environment variable or something? Either that or the mingw build now
works perfectly and I can't tell the difference.

ankith26:
Newest commit implements a workaround that makes sure the `.lib` files are not installed in the wheel

ankith26:
Force pushed with a minor change that fixes a merge conflict

buildconfig/pip_config.ini
Starbuck5, line 1: So this is deleted because pyproject.toml is replacing it as a source for static information? Or is it just obsolete?

.github/workflows/build-on-msys2.yml
Starbuck5, line 1: Why?

Starbuck5:
So it still tries to use mingw on my system, and still fails.

```
      FAILED: src_c/_camera.cp39-win_amd64.pyd
      "gcc"  -o src_c/_camera.cp39-win_amd64.pyd src_c/_camera.cp39-win_amd64.pyd.p/_camera.c.obj src_c/_camera.cp39-win_amd64.pyd.p/camera_windows.c.obj "-Wl,--allow-shlib-undefined" "-Wl,-O1" "-shared" "-Wl,--start-group" "-Wl,--out-implib=src_c\_camera.cp39-win_amd64.dll.a" "-lMfplat" "-lMf" "-lMfuuid" "-lMfreadwrite" "-lOle32" "C:/Users/charl/Desktop/pygame-ce/prebuilt-x64/SDL2-2.30.0/lib/x64/SDL2.lib" "c:\users\charl\appdata\local\programs\python\python39\python39.dll" "-lkernel32" "-luser32" "-lgdi32" "-lwinspool" "-lshell32" "-lole32" "-loleaut32" "-luuid" "-lcomdlg32" "-ladvapi32" "-Wl,--end-group"
      C:/Program Files/mingw64/bin/../lib/gcc/x86_64-w64-mingw32/8.1.0/../../../../x86_64-w64-mingw32/bin/ld.exe: cannot find -lMfreadwrite

      collect2.exe: error: ld returned 1 exit status
```

Really I think the problem is that msvc's `cl` is not on my path at all. But for pygame development it never needed to be, I had never noticed in years that it wasn't.

I'm still able to do it from the x64 native tools prompt, and I could see about installing `cl` into my path, but I don't think it happens by default when you install VS, so we'd need to add that step to the build instructions.

I have yet to go deep into the build products and scrutinize the output.

But I do have another question, how can one use the old setup process after this PR? My workflow has always been `pip install .`, this PR diverts that from setup.py to meson-python.

meson.build
```diff
@@ -0,0 +1,351 @@
...
+
+pg = 'pygame' # python base package name
+
+if host_machine.system() == 'windows'
+    plat = 'win'
+elif host_machine.system() == 'darwin'
+    plat = 'mac'
+elif host_machine.system() == 'linux'
+    plat = 'linux'
+elif host_machine.system() == 'android'
```
robertpfeiffer, line 22: Host means building *on* android here? Does this support an android or emscripten *target* then?

meson.build
```diff
@@ -0,0 +1,351 @@
...
+        'SDL2_ttf': sdl_ttf_dep.found(),
+        'freetype2': freetype_dep.found(),
+        'portmidi': portmidi_dep.found(),
+        'porttime': portmidi_dep.found() ? porttime_dep.found() : false,
+    },
+    section: 'Dependencies',
+)
+
+simd_avx2_flags = []
+if host_machine.cpu_family().startswith('x86')
```
robertpfeiffer, line 283: again, why host? Why not target? And isn't this supposed to be based on CPU dispatch?

meson.build
```diff
@@ -0,0 +1,351 @@
...
+        'However it may be added in the future',
+    )
+elif host_machine.system() == 'emscripten'
+    plat = 'emscripten'
+    error(
+        'The meson buildconfig of pygame-ce does not support emscripten for now. ',
+        'However it may be added in the future',
+    )
+else
+    # here it one of: cygwin, dragonfly, freebsd, gnu, haiku, netbsd, openbsd, sunos
```
robertpfeiffer, line 35: Is there a way to produce a warning here instead of an error?

buildconfig/pip_config.ini
ankith26: It is obsolete. The reason it existed was to pass parallel build flag, but meson just does that by default

.github/workflows/build-on-msys2.yml
ankith26: The old buildconfig tried to work with msys2, the new one does, and this action has been added to verify that.

I don't think we will "officially" support it or distribute wheels for it, but no harm in having extra checks anyways

ankith26:
> So it still tries to use mingw on my system, and still fails.

Hmm, that's probably happening due to an older mingw version AFAICT. I can add guards to prevent the error you are seeing.

> After this PR, does the CI generated meson-built wheels?

Yes, it does.

meson.build
ankith26: Meson uses the confusing GNU terminology
Build: Machine the code is built on.
Host: Machine the code is built for.
Target: Machine the code built, generates output for.

So yes, using `host_machine` really means *target* in the sense that you are thinking about.

meson.build
ankith26: We still have runtime detection and guarding. This only exists to pass compile time flags. Our strategy is to pass as "high/new" compile time flags as possible, but at runtime downgrade based on the best available option.

Though, I don't have much CPU dispatch experience so I may be missing out something here.

robertpfeiffer: I looked some things up. The upshot: We could massively simplify dispatch _if only we didn't have to support MSVC._

robertpfeiffer:
None of the things I complained about are the fault of the PR, I just didn't understand that meson uses GCC bootstrap terminology.

meson.build
ankith26: Yes, though I'm not sure if we should be doing it. I'd like to evaluate it in case-wise basis in future PRs

.github/workflows/build-on-msys2.yml
Starbuck5: It seems a bit wasteful to of CI resources to do it. Especially 5 separate runs on most PRs.

Starbuck5:
> After this PR, does the CI generated meson-built wheels?
> Yes, it does.

That ups the amount of scrutiny a lot, because we need to be confident in these wheels being ready for release.

I built two wheels, one before this PR, one after, and went through them with a fine-toothed comb.

- The pyd files in _sdl2 seem to be slightly larger on meson. Not sure what's up with that, if anything.
- examples/ now has editorconfig copied in, normally doesn't
- tests/fixtures/fonts now has PyGameMono.sfd copied in, normally doesn't
- docs folder now has pyi files for all top level python files for some reason
- docs/generated/_sources now has a c_api folder and a tutorial folder instead of just a ref folder
- docs/generated now comes with the tutorials folder, it normally doesn't.

These differences end up as the meson build having 73 more files and taking up 1.2 mb more disk space than the standard build on my test.

.github/workflows/build-on-msys2.yml
ankith26: We get 20 concurrent windows builds, of which we are currently using only 13 ATM. Adding the 5 extra builds won't drastically increase wait time for CI on a PR

ankith26:
> The pyd files in _sdl2 seem to be slightly larger on meson. Not sure what's up with that, if anything.

I would attribute that to different compiler flags used by meson implicitly.

> examples/ now has editorconfig copied in, normally doesn't

This file can probably be removed from the source code itself, what do you think?

> tests/fixtures/fonts now has PyGameMono.sfd copied in, normally doesn't

This file is 6.5 KB so I don't think adding it should be a problem.

> docs folder now has pyi files for all top level python files for some reason

This is where those pyi files should be. Is the legacy buildconfig ignoring these files entirely?

> docs/generated/_sources now has a c_api folder and a tutorial folder instead of just a ref folder
docs/generated now comes with the tutorials folder, it normally doesn't.

I consider these changes a "fix" of this PR. At some point these files were infact bundled in our wheels, but then "tut" folder was renamed to "tutorials" while not changing the old buildconfig, so these files were accidentally removed.

Starbuck5:
> This file can probably be removed from the source code itself, what do you think?

Probably

> This is where those pyi files should be. Is the legacy buildconfig ignoring these files entirely?

image

I've never seen these pyi files before in my life. They're not in source control. https://github.com/pygame-community/pygame-ce/tree/main/docs

> I consider these changes a "fix" of this PR. At some point these files were infact bundled in our wheels, but then "tut" folder was renamed to "tutorials" while not changing the old buildconfig, so these files were accidentally removed.

You're right, but I'm very file size sensitive right now given pypi support's lack of raising our limit, so it pains me to see it increase.

ankith26:
> I've never seen these pyi files before in my life. They're not in source control.

They are in source control, [but here](https://github.com/pygame-community/pygame-ce/tree/main/buildconfig/stubs/pygame/docs). They were recently added, and the author forgot to update buildconfig to handle the new subtree.

> You're right, but I'm very file size sensitive right now given pypi support's lack of raising our limit, so it pains me to see it increase.

The real fix of course, is to move docs/tests/examples/stubs into a separate package as we've spoken about. I would like to work on that post this PR. In the mean time, it's probably fine to be leaving out this subtree in this PR.

I will make a new push with the changes.

ankith26:
After the latest changes:
11739702 2.5dev/pygame_ce-2.5.0.dev1-cp310-cp310-win32.whl
11650529 meson-pr/pygame_ce-2.5.0.dev1-cp310-cp310-win32.whl

23390231 total

Starbuck5:
> They are in source control, [but here](https://github.com/pygame-community/pygame-ce/tree/main/buildconfig/stubs/pygame/docs). They were recently added, and the author forgot to update buildconfig to handle the new subtree.

These are bizarre, this is not a public API, why would it get pyi files?

They cause the editor to know about these things:
```py
from pygame import docs
docs.has_local_docs()
```

Which then crash:
```
pygame-ce 2.5.0.dev1 (SDL 2.28.5, Python 3.12.0)
  File "C:\Users\charl\Desktop\testr2.py", line 8, in <module>
    docs.has_local_docs()
    ^^^^^^^^^^^^^^^^^^^
AttributeError: module 'pygame.docs' has no attribute 'has_local_docs'
```

> The real fix of course, is to move docs/tests/examples/stubs into a separate package as we've spoken about. I would like to work on that post this PR. In the mean time, it's probably fine to be leaving out this subtree in this PR.

Thank you. Yes I've been thinking about that as well. I'd like to keep stubs and tests in the main package. Docs work best standalone and provide the best benefit, I could see examples going with the docs. `pygame-ce-docs` on pypi?

ankith26:
> Which then crash:

I tested locally and I don't get the crash. How are you building the wheel?
I have by building off this PR and also the 2.4.1 release on PyPI

> I'd like to keep stubs and tests in the main package

I mean, if we are going to create a new package with the aim of saving PyPI space, we might as well go all out and move as many things as possible. The reasoning for moving docs/examples/tests into a separate package is that the average user definitely doesn't have to care if these packages are installed or not. The stubs are in the gray area (useful for people to have installed), and I don't really have a strong opinion on where they should go.

Starbuck5:
Really, this doesn't crash for you?

I got pygame-ce from pip.

```
>>> from pygame import docs
pygame-ce 2.4.1 (SDL 2.28.5, Python 3.9.5)
>>> docs.has_local_path()
AttributeError: module 'pygame.docs' has no attribute 'has_local_path'
>>>
```

I have a strong opinion that the stubs should stay in this package.

Starbuck5:
Found this bit of docs on meson build about MSVC: https://meson-python.readthedocs.io/en/latest/how-to-guides/meson-args.html#force-the-use-of-the-msvc-compilers-on-windows

Tested this and confirmed it worked without CL needing to be in my path.

Putting cl in my path led to compile errors I wasn't able to resolve. This error: https://stackoverflow.com/questions/77029786/meson-error-compiler-cl-cannot-compile-programs-windows-10

ankith26:
> Really, this doesn't crash for you?
> 
> I got pygame-ce from pip.
> 
> ```
> >>> from pygame import docs
> pygame-ce 2.4.1 (SDL 2.28.5, Python 3.9.5)
> >>> docs.has_local_path()
> Traceback (most recent call last):
>   File "<stdin>", line 1, in <module>
> AttributeError: module 'pygame.docs' has no attribute 'has_local_path'
> >>>
> ```
> 
> I have a strong opinion that the stubs should stay in this package.

Well, you mistyped `docs.has_local_path` for `docs.has_local_docs`

Starbuck5:
> Well, you mistyped docs.has_local_path for docs.has_local_docs

Well, that would do it.

My typo aside, I don't see why a non public API like this would have type stubs.

meson.build
ankith26: Actually, on a rethink I think your suggestion is better, as we are probably never going to be able to test all these platforms on CI. We could probably group all these under a "unix-ey" label, though I'm not actually sure how many of these are actually "unix-ey"

Starbuck5 (approved):
The last test I wanted to do locally was make sure the SIMD dispatch still works, 

Tested grayscale and alpha blitting with printfs internally, made sure it was still reaching those places properly.

Let's get this out! I think any teething problems will need to come up in usage and in time, rather than holding this PR in wait longer.

PR #2559, opened 2023-11-11
Fix setDaemon() deprecation error

MyreMylar:
Deprecated in Python 3.10.

`.daemon` is not a new attribute replacing `.setDaemon()` it is just now being put forth as the **only** way to set this attribute.


Starbuck5 (approved):
I wonder if we should remove this module? (Not in this PR, but in general)

Anyways, this is an easy fix to approve, the `daemon` attribute was added in Python 2.6.

ankith26 (approved):

PR #2565, opened 2023-11-12
Added missing AVX2 fillers

itzpr3d4t0r:
This PR is a continuation of #2382 and adds the `SUB`, `MIN`, `MAX`, and `MULT` blend flags. this should be a 70X to 110X perf improvement over the current fillers with blend flags and is 35% faster than caching a surface with a color and blitting (with AVX2).

Results and test program:
**ON MAIN**
```
Flag: BLEND_SUB
fill: 1.8767017999998643
blit: 0.021530719999827853
--------------------
Flag: BLEND_MULT
fill: 1.7614726199999495
blit: 0.03417129999997996
--------------------
Flag: BLEND_MIN
fill: 1.82431628000013
blit: 0.02125239999986661
--------------------
Flag: BLEND_MAX
fill: 1.872223340000346
blit: 0.021590140000080284
--------------------
```

**WITH THIS PR**
```
Flag: BLEND_SUB
fill: 0.01598090000006778
blit: 0.021626579999974638
--------------------
Flag: BLEND_MULT
fill: 0.024974719999954688
blit: 0.03429605999972409
--------------------
Flag: BLEND_MIN
fill: 0.01568092000015895
blit: 0.021365460000197345
--------------------
Flag: BLEND_MAX
fill: 0.015678200000002106
blit: 0.021505839999917953
--------------------
```

**Test Program**
```Py
from timeit import repeat

import pygame

pygame.init()

surf = pygame.Surface((500, 500))
surf.fill((132, 33, 200))

color = pygame.Surface((500, 500))
color.fill((24, 24, 24))

flags = [
    "BLEND_SUB",
    "BLEND_MULT",
    "BLEND_MIN",
    "BLEND_MAX",
]

G = globals()

for flag in flags:
    print(f"Flag: {flag}")
    teststr = "surf.fill((24, 24, 24), None, pygame." + flag + ")"
    l = [min(repeat(teststr, globals=G, number=1000, repeat=10)) for _ in range(5)]
    print(f"fill: {sum(l) / len(l)}")

    teststr = "surf.blit(color, (0, 0), None, pygame." + flag + ")"
    l = [min(repeat(teststr, globals=G, number=1000, repeat=10)) for _ in range(5)]
    print(f"blit: {sum(l) / len(l)}")
    print("-" * 20)
```

PR #2566, opened 2023-11-12
Add SSE2 fillers

itzpr3d4t0r:
This PR is a continuation of #2382 and it adds all blend modes for fillers implemented with SSE2. This should perform better on ARM and on old x86 CPUs that don't support the newest AVX2 instructions (PR for this: #2565).

Results:
From my testing SSE2 is about 2X slower than AVX2 (makes sense since we work on 1/2 the pixels at a time) but is still a lot faster than the current single pixel implementation and is about as fast as the AVX2 blit with cached color surface, so I'd say that's a win either way.
**ON MAIN**
```
Flag: BLEND_ADD
fill: 1.8567257939238243
blit: 0.022120719999827853
--------------------
Flag: BLEND_SUB
fill: 1.8767017999998643
blit: 0.021530719999827853
--------------------
Flag: BLEND_MULT
fill: 1.7614726199999495
blit: 0.03417129999997996
--------------------
Flag: BLEND_MIN
fill: 1.82431628000013
blit: 0.02125239999986661
--------------------
Flag: BLEND_MAX
fill: 1.872223340000346
blit: 0.021590140000080284
--------------------
```
**WITH THIS PR**
```
Flag: BLEND_ADD
fill: 0.03238646000099834
blit: 0.021517279999534365
--------------------
Flag: BLEND_SUB
fill: 0.022462600001017564
blit: 0.021596119999594522
--------------------
Flag: BLEND_MULT
fill: 0.05370256000023801
blit: 0.034474519999639595
--------------------
Flag: BLEND_MIN
fill: 0.02288747999991756
blit: 0.02128549999979441
--------------------
Flag: BLEND_MAX
fill: 0.022593719998258165
blit: 0.021513199999753853
--------------------
```

Test Program:
```Py
from timeit import repeat

import pygame

pygame.init()

surf = pygame.Surface((500, 500))
surf.fill((132, 33, 200))

color = pygame.Surface((500, 500))
color.fill((24, 24, 24))

flags = [
    "BLEND_ADD",
    "BLEND_SUB",
    "BLEND_MULT",
    "BLEND_MIN",
    "BLEND_MAX",
]

G = globals()

for flag in flags:
    print(f"Flag: {flag}")
    teststr = "surf.fill((24, 24, 24), None, pygame." + flag + ")"
    l = [min(repeat(teststr, globals=G, number=1000, repeat=10)) for _ in range(5)]
    print(f"fill: {sum(l) / len(l)}")

    teststr = "surf.blit(color, (0, 0), None, pygame." + flag + ")"
    l = [min(repeat(teststr, globals=G, number=1000, repeat=10)) for _ in range(5)]
    print(f"blit: {sum(l) / len(l)}")
    print("-" * 20)
```


src_c/simd_surface_fill_sse2.c
```diff
@@ -0,0 +1,168 @@
...
+            pixels += 4;                                    \
+        }                                                   \
+                                                            \
+        if (pxl_excess) {                                   \
+            for (j = 0; j < pxl_excess; j++, pixels++) {    \
+                mm128_dst = _mm_cvtsi32_si128(*pixels);     \
+                                                            \
+                {FILL_CODE}                                 \
+                                                            \
+                    *pixels = _mm_cvtsi128_si32(mm128_dst); \
```
MyreMylar, line 67:
```suggestion
                mm128_dst = _mm_cvtsi32_si128(*pixels);     \
                                                            \
                {FILL_CODE}                                 \
                                                            \
                *pixels = _mm_cvtsi128_si32(mm128_dst);     \
```

MyreMylar (approved):
LGTM 👍 

I can follow all the SIMD logic and I see the same speedup locally with this PR over main as described in the PR comments. Makes a big difference! Only noticed one small indentation thing that seemed off. I guess the auto-formatter didn't catch it because it was inside a pre-processor macro.

src_c/simd_surface_fill_sse2.c
itzpr3d4t0r: That's actually the clang formatter that's doing that, so i guess it's not an issue.

Starbuck5 (approved):
I like the comparisons vs a single color surface in the PR description. Although I think it's using SDL blit, not AVX blit, since the surfaces you make in the test script don't have alpha.

I was just talking to somebody about how to implement a "fade to white" effect, and I think with this merged the MULT fillers would officially be a more efficient way to do that then a white surface and set_alpha.

Anyways, the code looks good to me.

PR #2567, opened 2023-11-13
Fix compiler warnings in event, camera and display

ankith26:
These warnings were caught on my meson buildconfig branch. For some reason, the current buildconfig code does not actually `-Werror` on linux builds (where these warnings are reported)


MyreMylar (approved):
LGTM 👍

yunline (approved):
LGTM

PR #2570, opened 2023-11-16
Rect.contains() optimization

itzpr3d4t0r:
Another fastcall port, this time for the Rect.contains() function.

### Results
The best results come from the `Rect`/`2 args`/`4 args` cases since they go through minimal overhead thanks to the `RectFromFastcallArgs` function and the recent `RectFromObject` optimization. the `2 args` case and `1 arg 2 tuples` could get faster with an optimized `twoPrimitivesFromObj` implementation (which i plan to implement in the future and should see a 15% perf improvement for fast sequences).

**This PR**
```
Rect: 0.00031379999927594326    - 2.33X faster
2 args: 0.0009735999992699362   - 1.52X faster
4 args: 0.0007076000001688953   - 1.69X faster
1 arg: 0.0009988999991037417     - 1.33X faster
1 arg 2 tuples: 0.0014074000027903821  - 1.18X faster
```

**Main**
```
Rect: 0.0007312000016099773
2 args: 0.0014795999995840248
4 args: 0.0012012999977741856
1 arg: 0.0013316000004124362
1 arg 2 tuples: 0.0016660000001138542
```

**Test Program**
```Py
from timeit import repeat
from pygame import Rect

r = Rect(0, 0, 100, 100)
r1 = Rect(10, 10, 50, 50)

G = globals()
N = 10000
R = 1000

tests = [
    ("Rect", "r.contains(r1)"),
    ("2 args", "r.contains((10, 10), (50, 50))"),
    ("4 args", "r.contains(10, 10, 50, 50)"),
    ("1 arg", "r.contains((10, 10, 50, 50))"),
    ("1 arg 2 tuples", "r.contains(((10, 10), (50, 50)))"),
]

for name, test in tests:
    print(name + ": ", end="")
    print(min(repeat(test, globals=G, number=N, repeat=R)))
```


MyreMylar (approved):
LGTM 👍  I see the same speedup locally with a variety of inputs.

src_c/rect_impl.h
```diff
@@ -1796,13 +1796,19 @@ RectExport_contains_internal(RectObject *self, PyObject *arg)
...
         return RAISE(PyExc_TypeError, "Argument must be rect style object");
     }
-    return PyBool_FromLong(ret);
+
+    return PyBool_FromLong(
+        (self->r.x <= argrect->x) && (self->r.y <= argrect->y) &&
+        (self->r.x + self->r.w >= argrect->x + argrect->w) &&
+        (self->r.y + self->r.h >= argrect->y + argrect->h) &&
+        (self->r.x + self->r.w > argrect->x) &&
+        (self->r.y + self->r.h > argrect->y));
```
ankith26, line 1811: I thinks this can be done in a better way. This is repeated code from `RectExport_contains_internal`. How about making `RectExport_contains_internal` fastcall-ey instead, so that this function can directly call it, and any other caller of `RectExport_contains_internal` can that needs to call it with one object can use `nargs = 1`

ankith26 (changes requested):

src_c/rect_impl.h
itzpr3d4t0r: Tried to patch it up.

ankith26 (approved):

PR #2575, opened 2023-11-20
Set default convert format when `Window.get_surface()` is called

yunline:
Test code (should not raise exception):
```py
import pygame

win = Window()

sf1 = win.get_surface()

sf2 = pygame.Surface((10,10)).convert()

while 1:
    event = pygame.event.wait()
    if event.type == pygame.QUIT:
        break
```

If `convert()` is called but no convert format have been set, it should raise:
```
pygame.error: No convert format has been set, try display.set_mode() 
```
After the Window API published, it should raise:
```
pygame.error: No convert format has been set, try display.set_mode() or Window.get_surface()
```


MyreMylar (approved):
LGTM 👍

src_c/surface.c
```diff
@@ -1623,32 +1623,34 @@ surf_convert(pgSurfaceObject *self, PyObject *args)
 static SDL_Surface *
 pg_DisplayFormat(SDL_Surface *surface)
 {
-    SDL_Surface *displaysurf;
-    if (!pg_GetDefaultWindowSurface()) {
-        SDL_SetError("No video mode has been set");
+    if (!pg_GetDefaultConvertFormat()) {
```
Starbuck5, line 1626: Would it be better to call the function once, test the value, then use the value -- instead of calling the function twice?

I see it was previously calling the function twice.

src_c/surface.c
```diff
@@ -1623,32 +1623,34 @@ surf_convert(pgSurfaceObject *self, PyObject *args)
...
+        SDL_SetError("No convert format has been set, try display.set_mode()"
+                     // " or Window.get_surface()"
+                     // (uncomment when Window API is published)
```
Starbuck5, line 1629: You might as well uncomment this, lets say Window will be published next release.

src_c/base.c
```diff
@@ -2057,6 +2057,23 @@ pg_SetDefaultWindowSurface(pgSurfaceObject *screen)
     pg_default_screen = screen;
 }
 
+SDL_PixelFormat *pg_default_convert_format = NULL;
```
Starbuck5, line 2060: Thoughts-- I see that we can store this struct or we can store an SDL_PixelFormatEnum. I was initially thinking maybe it would be better to store as an enum value rather than a dynamically allocated struct.

Is it at risk of being deallocated by SDL if we just have a pointer to it (right now)?

In the pygame source we have PG_ConvertSurface, but there's also PG_ConvertSurfaceFormat that takes a pixelformat enum instead of a pixelformat struct. I looked into the SDL source, and SDL_ConvertSurfaceFormat allocates a format using the enum, calls SDL_ConvertSurface with it, and then throws it away. So that would be less efficient.

Idea to make this code even more safe: pg_SetDefaultConvertFormat takes a SDL_PixelFormatEnum and internally uses SDL_AllocFormat to create a format to store internally in base.c. When called again, it uses SDL_FreeFormat to free the existing format before replacing it with another.

Does that make sense at all? What do you think?

src_c/surface.c
yunline: fixed

Starbuck5:
Hey @yunline, can you rebase this against main?

It will make it easy to test with a full Window.flip() example, and it'll fix the circleci change.

I see you didn't respond to my long rambling paragraph feedback, I'd like to try and add that as a commit to this branch myself, and then if it looks good to you we can move forward.

"Window will be published next release" shouldn't go in the error message! That was information for you, not for the user :)

yunline:
> "Window will be published next release" shouldn't go in the error message! That was information for you, not for the user :)

Oops, I'll fix it soon.

Starbuck5:
@yunline I added a new commit to this branch, which makes the code more but also safer.

I was worried about what would happen if the SDL_PixelFormat struct we hold was deallocated by SDL, making it allocate a SDL_PixelFormat specifically to hold on to will prevent that from happening.

Please let me know if this looks okay to you

src_c/surface.c
```diff
@@ -1590,32 +1590,33 @@ surf_convert(pgSurfaceObject *self, PyObject *args)
...
-        SDL_SetError("No video mode has been set");
+    if (!pg_GetDefaultConvertFormat()) {
+        SDL_SetError(
+            "No convert format has been set, try display.set_mode()"
+            " or Window.get_surface().");
         return NULL;
     }
-    displaysurf = pgSurface_AsSurface(pg_GetDefaultWindowSurface());
-    dformat = displaysurf->format;
+    dformat = pg_GetDefaultConvertFormat();
```
ankith26, line 1619: Perhaps we could move this line above and check the variable instead of calling the function twice, to make the code a bit more cleaner
I see this is a pretty cheap function, so I'm not worried about any performance impact.

src_c/window.c
```diff
@@ -163,6 +163,15 @@ window_get_surface(pgWindowObject *self)
...
     }
     self->surf->surf = _surf;
+
+    if (pg_GetDefaultConvertFormat() == NULL) {
+        if (pg_SetDefaultConvertFormat(_surf->format->format) == NULL) {
+            /* This is very unlikely, I think only would happen if SDL runs
+             * out of memory when allocating the format. */
+            return RAISE(pgExc_SDLError, SDL_GetError());
+        }
+    }
```
ankith26, line 173: Should this happen a few lines above? Like just after `SDL_GetWindowSurface`. So that `pgSurface_New2` and other logic does not run in the error case

src_c/surface.c
yunline: fixed

src_c/window.c
yunline: fixed

Starbuck5 (approved):
Looks good to me.

Issue #2518, opened 2023-10-15
Deprecation warning in chimp.py example


Starbuck5:
When I run the chimp example, I get a sprite deprecation warning. This should be fixed.

Are there other examples using deprecated functionality?

oddbookworm:
aliens

PR #2581, opened 2023-11-20
General example changes

bilhox:
Hi, I took a look at all examples, and made some fixes.

``aacircle.py`` : 
- Removed the pretty weird try block and replaced it by a beautiful while loop

``aliens.py`` : 
- renamed ``all`` to ``all_sprite``, because ``all`` is a native python function.
- removed ``depth`` argument
- added ``pygame.SCALED`` because on Windows, there is a problem with FULLSCREEN mode

``chimp.py`` : 
- removed ``RenderPlain``because it's deprecated
- renamed some variables so it follows python naming convention

``dropevent.py`` : 
- renamed some variables because either it doesn't follow the python naming convention, or it is not intuitive

``playmus.py`` : 
- This is a proposition from Matiiss, switching from ``pygame.freetype`` to ``pygame.Font``
- Revamped the code, as its structure isn't understandable for a beginner.

~~resizing_new.py~~:
- ~~renamed a variable because it's not intuitive~~
Removal plan is being discussed

``scaletest.py`` : 
- removed useless alias
- renamed some variables/functions because either it doesn't follow the python naming convention, or it is not intuitive

``liquid.py`` : 
- renamed some variables/functions because either it doesn't follow the python naming convention, or it is not intuitive
- remove ``time.sleep`` so it uses ``pygame.Clock``


bilhox:
This should fix #2518

examples/aacircle.py
```diff
@@ -24,17 +24,18 @@ def main():
...
-            event = pygame.event.wait()
-            if event.type == pygame.QUIT:
-                break
-            if event.type == pygame.KEYDOWN:
-                if event.key == pygame.K_ESCAPE or event.unicode == "q":
-                    break
-            pygame.display.flip()
-    finally:
-        pygame.quit()
+
```
Starbuck5, line 27: The loop used to have pygame.display.flip in it. It still should, in my opinion. Even though stuff isn't being redrawn it still tells the OS things are happening.

examples/chimp.py
Starbuck5, line 1: I believe chimp.py has a tutorial based on it, so any code changes here should also be changed there.

examples/dropevent.py
```diff
@@ -16,49 +16,49 @@
...
-                    ev.text, 1, (255, 255, 255), wraplength=screen_size[0] - 10
+                    event.text, 1, (255, 255, 255), wraplength=screen_size[0] - 10
                 )
                 spr_file_text_rect = spr_file_text.get_rect()
                 spr_file_text_rect.center = surf.get_rect().center
-            elif ev.type == pygame.DROPFILE:
-                print(ev)
+            elif event.type == pygame.DROPFILE:
+                print(event)
                 spr_file_text = font.render(
```
Starbuck5, line 52: You changed 1->True in the other font.render but not here

examples/aacircle.py
bilhox: I found it in the example, I replaced the code by a standard event loop

examples/playmus.py
```diff
@@ -51,8 +50,7 @@ def __init__(self, title):
         self.screen.fill(self.background_color)
         pygame.display.flip()
 
-        pygame.freetype.init()
-        self.font = pygame.freetype.Font(None, 20)
+        self.font = pygame.sysfont.SysFont(None, 20)
```
Starbuck5, line 53: You can also do `pygame.Font(None, 20)`, which seems more normal to me than using sysfont to get the default font. Although it might get the font at a slightly different due to a long suffering incongruity in the font loader.

examples/playmus.py
Starbuck5, line 1: If you're messing with font rendering here, you could port it to use multiline text rendering from pygame.Font for #2206

bilhox: Okay so I changed a lot of things in ``playmus.py``, mainly how we write lines. The structure of the program was quite terrible, so I changed the structure so beginners can understand the code. The other task in the issue is linked to another PR, so I can't update it.

examples/aacircle.py
```diff
@@ -24,17 +24,20 @@ def main():
...
 
     pygame.display.flip()
-    try:
-        while True:
-            event = pygame.event.wait()
+
+    running = True
+
+    while running:
+        for event in pygame.event.get():
```
dr0id, line 31: this un-capped loop is a waste of resources... at least limit the fps. Or use Wait as before.

examples/aliens.py
```diff
@@ -382,7 +383,7 @@ def main(winstyle=0):
             player.kill()
 
         # draw the scene
-        dirty = all.draw(screen)
+        dirty = all_sprites.draw(screen)
         pygame.display.update(dirty)
```
dr0id, line 387: recommend to use flip over dirty update. Prefer to use flip instead of `pygame.display.update(dirty)` .

dr0id, line 387: recommend to use flip over dirty update. Prefer to use flip instead of `pygame.display.update(dirty)` .

examples/aliens.py
```diff
@@ -383,8 +383,8 @@ def main(winstyle=0):
             player.kill()
 
         # draw the scene
-        dirty = all_sprites.draw(screen)
-        pygame.display.update(dirty)
+        all_sprites.draw(screen)
```
Starbuck5, line 386: Now that you've changed this to not use dirty rects, there is no point in using a RenderUpdates (the group class that all_sprites is), because is tracking the dirty rects. If not using dirty rects, you should use the normal sprite group class I believe.

bilhox: Fixed

examples/chimp.py
Starbuck5: This has not been resolved.

examples/playmus.py
```diff
@@ -20,142 +20,105 @@
...
 
-class Window:
-    """The application's Pygame window
 
-    A Window instance manages the creation of and drawing to a
-    window. It is a singleton class. Only one instance can exist.
+def write_lines(
+    dest: pygame.Surface,
+    text: str,
+    font=pygame.font.Font(None, 30),
```
Starbuck5, line 30: `pygame.Font` rather than `pygame.font.Font`

examples/chimp.py
bilhox: Hi, sorry for this waiting time, I'll update the tutorial when I woke up tomorrow !
Happy new year :)

bilhox: fixed

examples/playmus.py
bilhox: fixed

MyreMylar (approved):
LGTM 👍

examples/aliens.py
```diff
@@ -314,7 +320,7 @@ def main(winstyle=0):
                         print("Changing to FULLSCREEN")
                         screen_backup = screen.copy()
                         screen = pygame.display.set_mode(
-                            SCREENRECT.size, winstyle | pygame.FULLSCREEN
+                            SCREENRECT.size, winstyle | pygame.FULLSCREEN | pygame.SCALED,
```
MyreMylar, line 323:
```suggestion
                            SCREENRECT.size, 
                            winstyle | pygame.FULLSCREEN | pygame.SCALED,
```

ankith26 (commented):
This PR needs a lint run to make CI happy

examples/resizing_new.py
```diff
@@ -23,10 +22,6 @@
             done = True
         if event.type == pygame.QUIT:
             done = True
-        # if event.type==pygame.WINDOWRESIZED:
-        #    screen=pygame.display.get_surface()
-        if event.type == pygame.VIDEORESIZE:
-            screen = pygame.display.get_surface()
```
ankith26, line 29: Why is this block being removed? It is needed because during resizing, the `screen` object needs to be kept updated, otherwise the output looks like garbage.

examples/scaletest.py
ankith26, line 1: This example is missing a clock. I played around a bit with this and I could feel my laptop start to heat up lol. Now that you are making changes here it would be a good time to fix it :smile:

examples/scaletest.py
```diff
@@ -30,68 +30,74 @@ def main(imagefile, convert_alpha=False, run_speed_test=False):
...
-        if bDown:
+        if down:
             cursize[1] += 2
-        if bLeft:
+        if left:
             cursize[0] -= 2
             if cursize[0] < 1:
                 cursize[0] = 1
-        if bRight:
+        if right:
```
ankith26, line 95: might be a good idea to refactor this to use `get_pressed` instead of manually doing it with the event loop here

examples/resizing_new.py
bilhox: With or without, I don't see the difference. What were you thinking about exactly ?

ankith26: [Screencast from 2024-05-26 18-39-54.webm](https://github.com/pygame-community/pygame-ce/assets/46915066/87d41d60-b2f3-4a7c-b06c-c794c545bbfa)

Here is a screencast that demonstrates the difference between your updated version (the first command I ran) and the old version (the second command that I ran)

bilhox: AH, this problem is on linux system then.

bilhox: I had like to discuss how useful is this example. This example looks old after a small talk with bluberquark on discord. He also told me that ``pygame.display._set_autoresize`` was supposed to be deprecated, which doesn't seem to be the case.
This joins the idea of blublerquark to reduce the number of examples, and making new ones.

bilhox: Reverting the changes made in this file, as there is a plan for a removal.

ankith26 (approved):
I had already tested all these examples in the last reviewing round that I did. The code changes LGTM.
I will go ahead and merge this in now, though there is always scope for further improvements in future (in hopefully smaller and more focussed PRs).

Thanks for contributing! :tada:

PR #2544, opened 2023-11-04
Compile SSE2 smoothscale intrinsics on NEON

Starbuck5:
Follow on to #2473 

On x86, these routines represented a ~20-25% speedup over the already accelerated SSE instructions. However, on ARM there is no acceleration at all (besides any compiler autovectorization). I don't have an ARM device to test, but I expect it to be a ~60% speedup, given the difference between GENERIC (no acceleration) and SSE2 on x86.

I split out the legacy "SCALE_MMX_SUPPORT" define to only govern the old MMX and SSE routines, and used the same defines as the blitters for the new SSE2 and faux-NEON routines.

Myre's perf test program for ARM testing: https://github.com/pygame-community/pygame-ce/pull/2473#pullrequestreview-1651874211

PR #2473, opened 2023-09-29
Add SSE2 intrinsics smoothscale backend

Starbuck5:
Tackles the bulk of what I wanted in #1767 

This was more difficult than I thought, this was quite difficult. I read through the existing smoothscale assembly (written assembly, not intrinsics) and the C implementations and combined what I saw into an SSE2 intrinsic version of the same algorithm.

I then had lots of ideas for optimizations on top of that, so for three of the filters it processes more pixels at a time than the existing SSE/MMX implementations.

This PR "just" implements the SSE2 backend, there are several follow on steps that would be quite nice:
- Get it compiling and up at runtime on NEON, so ARM folks can get fast smoothscale
- Remove legacy MMX/SSE backends eventually (we've been linking the compiled code for this in, at least on Windows, this is a part of the pygame codebase we don't know how to compile). Wanting to remove this uncompileable, unportable and ancient bit of our codebase was my chief motivation for opening this PR.
- AVX2 backend? Now that we've got it in intrinsics, it's much easier to scale up the parallelism with other intrinsics.
- Fix so if a malloc fails in the filters it will accurately report an error

Performance relative to SSE backend:
```
X-shrinking: 25% faster
Y-shrinking: 28% faster
X-expanding: neglible
Y-expanding: 25% faster
```

These are pixel perfect accurate relative to the SSE backend. Verified by saving out to pngs and checking the file hashes.

PR #2583, opened 2023-11-22
Smoothscale: deprecate old backends, update docs

Starbuck5:
Follows on from story started in #2473 and #2544.

The old backends are marked deprecated. I don't want to remove them yet in case bugs are found in the new implementations. Having the old ones around allows users to work around by setting it to use the old ones explicitly.


ankith26 (commented):
Would be also nice to have a unit test testing the deprecation

Starbuck5:
> Would be also nice to have a unit test testing the deprecation

I thought of this but I didn't include it initially since it seems very minor, I've included it now.

MyreMylar (approved):
LGTM 👍

ankith26 (approved):

PR #2584, opened 2023-11-22
Raise minimum SDL version to 2.0.10

Starbuck5:
We recently raised our minimum Python to 3.8. In Linux-land, Python 3.8 corresponds to Ubuntu focal, which packages SDL2 2.0.10. Therefore it is reasonable to raise our minimum version to 2.0.10.


ankith26 (approved):

MyreMylar (approved):
LGTM 👍

PR #2586, opened 2023-11-22
C loader targets a python alias

pmp-p:
which is not yet available when using static loader. https://github.com/pygame-community/pygame-ce/blob/9d69a535b2b79d115e2a7db55d35fd541b140578/src_py/__init__.py#L135


MyreMylar (approved):
LGTM 👍

Starbuck5:
We still commit generated Cython, so I ran it with the changes in the PR and added a commit.

Starbuck5 (approved):

PR #2588, opened 2023-11-23
Multiline more examples

gresm:
Resolves one sub-task from  #2206

This PR:
 - examples/textinput.py - now uses multi-line font rendering
 - examples/textinput.py - fix visual errors with text looking weird
 - examples/textinput.py - mypy no longer upset
 - examples/vgrade.py - PEP 8 exists

### DIFF
--- a/examples/textinput.py
+++ b/examples/textinput.py
@@ -11,7 +11,6 @@ import sys
 import os
 
 import pygame
-import pygame.freetype as freetype
 
 # This environment variable is important
 # If not added the candidate list will not show
@@ -62,11 +61,17 @@ class TextInput:
         # The font name can be a comma separated list
         # of font names to search for.
         self.font_names = ",".join(self.FONT_NAMES)
-        self.font = freetype.SysFont(self.font_names, 24)
-        self.font.origin = True
-        self.font_height = self.font.get_sized_height(self.font.size)
-        self.font_small = freetype.SysFont(self.font_names, 16)
+        self.font = pygame.font.SysFont(self.font_names, 24)
+        self.font_height = self.font.get_height()
+        self.font_small = pygame.font.SysFont(self.font_names, 16)
         self.text_color = text_color
+
+        self.prompt_surf = self.font.render(
+            self.prompt, True, self.text_color
+        )
+        self.prompt_rect = self.prompt_surf.get_rect(
+            topleft=self.CHAT_BOX_POS.topleft
+        )
 
         self.fps = fps
         self.second_counter = 0
@@ -84,6 +89,15 @@ class TextInput:
 
         if self.second_counter >= self.fps:
             self.second_counter = 0
+
+        # Check if input fits in chat box
+        input_size = self.font.size(self._get_ime_text())
+        while input_size[0] > self.CHAT_BOX_POS.w - self.prompt_rect.w:
+            if self._ime_editing_text:
+                # Don't block.
+                break
+            self._ime_text = self._ime_text[:-1]
+            input_size = self.font.size(self._get_ime_text())
 
     def _clamp_to_text_range(self, num: int):
         return min(len(self._ime_text), max(0, num))
@@ -155,11 +169,14 @@ class TextInput:
                 if len(self._ime_text) == 0:
                     return
 
-                # Append chat list
-                self.chat_list.append(self._ime_text)
-                if len(self.chat_list) > self.CHAT_LIST_MAXSIZE:
-                    self.chat_list.pop(0)
+                # Add to chat log
                 self.chat += self._ime_text + "\n"
+
+                chat_lines = self.chat.split("\n")
+                if len(chat_lines) > self.CHAT_LIST_MAXSIZE:
+                    chat_lines.pop(0)
+                    self.chat = "\n".join(chat_lines)
+
                 self._ime_text = ""
                 self._ime_text_pos = 0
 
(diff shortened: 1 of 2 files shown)


examples/textinput.py
```diff
@@ -37,7 +37,13 @@ class TextInput:
...
 
     def __init__(
-        self, prompt: str, pos, screen_dimensions, print_event: bool, text_color="white"
+        self,
+        prompt: str,
+        pos,
+        screen_dimensions,
+        print_event: bool,
+        text_color="white",
+        fps: int = 50,
```
oddbookworm, line 46: Looks like you're missing a couple of type hints here

oddbookworm (changes requested):
Looks like reasonable changes, just one small nitpick about type hinting

examples/textinput.py
gresm: Thanks, didn't catch this because this is actually the code from before my changes (it appeared in diff because I've black on it...).

examples/textinput.py
```diff
@@ -37,7 +37,13 @@ class TextInput:
     ]
 
     def __init__(
-        self, prompt: str, pos, screen_dimensions, print_event: bool, text_color="white"
+        self,
+        prompt: str,
+        pos: tuple[int, int],
+        screen_dimensions: tuple[int, int],
+        print_event: bool,
+        text_color="white",
```
oddbookworm, line 45: another minor nitpick:
spaces around "="

MyreMylar, line 45:
```suggestion
        text_color = "white",
```

MyreMylar (approved):
seems fine 👍

examples/textinput.py
gresm: @MyreMylar https://pep8.org/#other-recommendations
image

gresm: Here I disagree, text_color should be a valid pygame color-like object, but the annotation for it is a bit verbose (in ._common.pyi is defined like ``Union[int, str, Sequence[int]]``), so I don't see much value in adding annotations here, as it doesn't improve code readability and when annotations are missing, according to PEP 8, there shouldn't be any spaces (also running black on the code should also remove them, so the checks will fail).

Matiiss: The way to deal with verbose type annotations is to alias them, so something like `Color: TypeAlias = Union[int, str, Sequence[int], pygame.Color]` and use that for the type hint, then you'd need to add the spaces around which is what Andrew probably meant, if you add the type hint, you need spaces.

gresm: I prefer to leave it like this until a proper API is exposed.

examples/textinput.py
```diff
@@ -53,124 +59,198 @@ def __init__(
         self._ime_text_pos = 0
         self._ime_editing_text = ""
         self._ime_editing_pos = 0
-        self.chat_list = []
+        self.chat = ""
 
         # Freetype
```
Starbuck5, line 64: But it's not using freetype anymore?

examples/textinput.py
```diff
@@ -5,12 +5,12 @@
 
 Shows how to use the TEXTEDITING and TEXTINPUT events.
 """
+from __future__ import annotations
```
Starbuck5, line 8: Why?

examples/textinput.py
```diff
@@ -37,7 +37,13 @@ class TextInput:
     ]
 
     def __init__(
-        self, prompt: str, pos, screen_dimensions, print_event: bool, text_color="white"
+        self,
+        prompt: str,
+        pos: tuple[int, int],
+        screen_dimensions: tuple[int, int],
```
Starbuck5, line 43: The modern type hint style doesn't work on 3.8, which we still support.

```
Python 3.8.8 (tags/v3.8.8:024d805, Feb 19 2021, 13:18:16) [MSC v.1928 64 bit (AMD64)] on win32
Type "help", "copyright", "credits" or "license" for more information.
>>> def a(x: int, y: tuple[int, int]):
...     print(x, y)
...
TypeError: 'type' object is not subscriptable
```

```
Python 3.9.5 (tags/v3.9.5:0a7dcbd, May  3 2021, 17:27:52) [MSC v.1928 64 bit (AMD64)] on win32
Type "help", "copyright", "credits" or "license" for more information.
>>> def a(x: int, y: tuple[int, int]):
...     print(x, y)
...
>>> a
<function a at 0x000001921B3AF040>
```

gresm: Invalid concern: see [``__future__`` imports](https://docs.python.org/3/library/__future__.html).
```py
>>> from __future__ import annotations
>>> import sys
>>> def a(lst: list[str]):
...     print(*lst)
>>> print(sys.version)
>>> a(["a", "b", "c"])
3.8.0b4 (default, Sep  2 2019, 08:08:37) [GCC 8.3.1 20190223 (Red Hat 8.3.1-2)]
a b c
```

examples/textinput.py
gresm: @Starbuck5 done

examples/textinput.py
Starbuck5: I would rather use the old style than use a future import that has been stalled with no definite direction for several years now.

gresm: Well, examples need to have the code refactored, as they use different style conventions. I personally use the ``__future__`` import, but I would use the chosen one for this project, if such style existed. I understand why you dislike ``__future__.annotations``, but I have nothing against it. We are missing a consistent code-styling guide ( it doesn't have to be a formal document of any kind, coherent code base would be enough) and because of that, we are having this discussion about what convention is correct.

Starbuck5: I agree that lots of the examples could use improvement.

I'm not sure what styling guide would go in depth enough to prescribe whether a future import is allowed.

I would like you to make your type annotations compatible with Python 3.8 without this future import.

gresm: Understandable, I can do that.

gresm: @Starbuck5 done

Starbuck5 (approved):
Looks good, thanks!

PR #2590, opened 2023-11-24
Subclassing Sounds + Channels.

### LINKED ISSUE #2589: pygame.mixer.Channel.queue does not support pygame.mixer.Sound derivative class instances
If you attempt using `pygame.mixer.Channel.queue` with a Sound derived class, pygame will raise an error as if you didn't pass a Sound-based class at all. Here is the class I use, which netted me the error:

```py
# _SoundBase is an alias for pygame.mixer.Sound
class Sound(_SoundBase):
    def __init__(self, file: str|ndarray|_SoundBase):
        if isinstance(file,str):
            super().__init__(resource_path(file)) # Keep the source directory in case it's needed, and use an absolute path instead of a relative one
            self.dir = file
        elif isinstance(file, _SoundBase):
            super().__init__(file) # Clone the original Sound object
        else:
            super().__init__(array=file)

    volume = property(
        _SoundBase.get_volume, _SoundBase.set_volume, doc="The current sound's volume."
    ) # Access the volume as a property for convenience.

    num_channels = property(_SoundBase.get_num_channels)

    # Convenience conversion methods
    def __bytes__(self):
        return self.get_raw()

    def __len__(self):
        return self.get_length()

    def __copy__(self):
        return self.__class__(self.dir)

    def play(self, loops: int = 0, maxtime: int = 0, fade_ms: int = 0):
        if is_played(self): # is_played is a function checking if a sound is played by at least one channel.
            super().stop() # Stop the previous playback to avoid sound overlapping
        return super().play(loops, maxtime, fade_ms)

    copy = __copy__
    __deepcopy__ = __copy__
```

And here is the error I got (pygame-ce 2.3.2 (SDL 2.26.5, Python 3.12.0)):

```
  File "C:\Users\XXXXXXX\Documents\Python\Oracle-Project\testEngine.py", line 1733, in <module>
    get_channel(wooshpart1).queue(wooshpart2) # wooshpart1 and wooshpart2 are both instances of Sound (the derived class)
TypeError: The argument must be an instance of Sound
```

Line 1070 in src_c/mixer.c reveals `Channel.queue`'s code uses an exact type check instead of the usual instance verification to block non-sound classes from being queued, like this : 

```c
#define pgSound_Check(x) (Py_TYPE(x) == &pgSound_Type) //i.e. type(x) == pygame.mixer.Sound
...
if (!pgSound_Check(sound)) { //Line 1070
        return RAISE(PyExc_TypeError,
                     "The argument must be an instance of Sound");
}
```

gresm:
Fixes #2589
(Or at least it should)

### DIFF
--- a/test/mixer_test.py
+++ b/test/mixer_test.py
@@ -935,13 +935,8 @@ class ChannelTypeTest(unittest.TestCase):
         self.assertEqual(ch2.id, 2)
 
     def test_subclass(self):
-        try:
-
-            class MyChannel(mixer.Channel):
-                pass
-
-        except Exception:
-            self.fail("This should not raise an exception.")
+        class MyChannel(mixer.Channel):
+            pass
 
 
 class ChannelInteractiveTest(unittest.TestCase):
--- a/test/mixer_test.py
+++ b/test/mixer_test.py
@@ -934,6 +934,15 @@ class ChannelTypeTest(unittest.TestCase):
         self.assertEqual(ch1.id, 1)
         self.assertEqual(ch2.id, 2)
 
+    def test_subclass(self):
+        try:
+
+            class MyChannel(mixer.Channel):
+                pass
+
+        except Exception:
+            self.fail("This should not raise an exception.")
+
 
 class ChannelInteractiveTest(unittest.TestCase):
     __tags__ = ["interactive"]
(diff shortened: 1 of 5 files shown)


MyreMylar (approved):
LGTM 👍

ankith26 (approved):

ankith26 (changes requested):
Okay on a relook, maybe this PR also needs a little docs addition in the `Channel` class saying that it is now subclassable?

MyreMylar (approved):
LGTM 👍

test/mixer_test.py
```diff
@@ -934,6 +934,15 @@ def test_id_getter(self):
         self.assertEqual(ch1.id, 1)
         self.assertEqual(ch2.id, 2)
 
+    def test_subclass(self):
```
Starbuck5, line 937: Nitpick on this testing, won't unittest mark it as failed if an unexpected Exception occurs without needing to explicitly try/except it?

gresm: I was sure that it was implemented this way, but when searching through code-base, I see I was wrong, I'll fix it when I'll be able to.

gresm: @Starbuck5 done

PR #2592, opened 2023-11-29
Add font filtering example

gottadiveintopython:
This example demonstrates how to filter the pre-installed fonts on the system, by a language.

screenshot

### DIFF
--- a/examples/font_filter_by_language.py
+++ b/examples/font_filter_by_language.py
@@ -1,0 +1,85 @@
+"""
+This example demonstrates how to find fonts for each language.
+"""
+
+from typing import Iterator
+
+from pygame.font import get_fonts
+from pygame import freetype
+import pygame
+
+
+def enum_font_from_lang(lang) -> Iterator[str]:
+    '''
+    Enumerates fonts that are suitable for a specified language.
+    '''
+    text = DISCRIMINANTS[lang]
+    for font_name in get_fonts():
+        if can_render_text(font_name, text):
+            yield font_name
+
+
+def can_render_text(font_name, text, *, font_size=14) -> bool:
+    '''
+    Whether a specified font is capable of rendering a specified text.
+    '''
+    try:
+        font = freetype.SysFont(font_name, font_size)
+    except AttributeError as e:
+        # AttributeError occurs on certain fonts, which may be a PyGame bug.
+        if e.args[0] == r"this style is unsupported for a bitmap font":
+            return False
+        else:
+            raise
+    rendered = []
+    for c in text:
+        pixels = font.render_raw(c)
+        if pixels in rendered:
+            return False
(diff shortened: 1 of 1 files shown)


yunline:
You need to run `python setup.py format` to format the code.

MyreMylar:
Line 885 should have:

```
        if ((TTF_GlyphIsProvided(font, ch) != 0) &&
```

gottadiveintopython:
@MyreMylar @yunline  Thank you but I'm not familiar enough with PyGame nor SDL to modify its code, so I shouldn't do that.

MyreMylar:
> @yunline @MyreMylar Thank you but I'm not familiar enough with PyGame nor SDL to modify its code, so I shouldn't do that.

No worries, I wasn't expecting you to solve this. I have along standing plan to gradually deprecate the freetype submodule so I will go to great lengths to keep it from expanding further throughout the codebase.

Anyway I made a PR to look at this issue in font further here:

https://github.com/pygame-community/pygame-ce/pull/2593

gottadiveintopython:
> I have along standing plan to gradually deprecate the freetype submodule

@MyreMylar I thought it's the opposite because the freetype doc says

> The pygame.freetype module is a replacement for pygame.font.

MyreMylar:
Just to be clear, there is a long running confusion over pygame/pygame-ce having multiple font handling submodules (`font`,`freetype`,`sysfont`, `ftfont` that all, have the 'freetype' library underlying them. There isn't a great reason for this, other than the fact that the pygame library has a long history and there were periods of time when one path or another made sense and then workarounds were added and so on.

What I would like is for pygame to have one clear user and developer story on handling fonts while wrapping the freetype library. In the present time, the `font` submodule looks like the best place to centralise because it leverages the work of `SDL_ttf` as well as our own efforts here. They recently solved the problem of rendering arabic alphabets correctly by integrating `harfbuzz` with `freetype` and the font module was able to quickly adopt those improvements.

Centralising will be a slow process of first making sure we can support all the use cases of the other submodules with code in `font`. This example here just triggered me to investigate why it couldn't be done entirely with font.

The documentation for the `freetype` module saying it is a replacement for `font` is both of it's time and also a true thing to say in that you can do most of the same things wither either submodule (but not everything in either direction).

gottadiveintopython:
@MyreMylar  Thank you for clarification :) I'll rewrite the example without using the `pygame.freetype` module then. Maybe after #2593 is merged.

MyreMylar (commented):
Example itself seems fine but is awaiting another PR. Should we merge it as is anyway?

gottadiveintopython:
I might need to add more comments to the example if we merge it as is, such as 

> The result of can_render_text is not 100% accurate.

PR #2596, opened 2023-12-10
SDL_image 2.8.0

Starbuck5:
It's a long time coming, since we've decided we need SDL_image built with libpng, and SDL_image prebuilts after 2.0.5 don't come with that.

This PR hopes to get around that using CI on a pygame-community fork of SDL_image (https://github.com/pygame-community/SDL_image/releases/tag/2.8.0-pgce) to generate release bundles with our parameters.

This allows our Windows CI and local development workflow to remain almost exactly the same through this change.


MyreMylar:
There was a change made in CPython 3.13.0 & 3.12.1, the latter of which came out this week, removing skipped tests from the total number of completed tests in the unit test module. This caused our check that the number of started tests equalled the number of completed tests to fail. See:

https://github.com/python/cpython/issues/110890

I initially assumed it was threading related due to the print output - but actually none of the threading code in our test running code is actually being used, and it doesn't actually work anymore.

In conclusion; we should probably be using pytest.

MyreMylar (approved):
All looks good to me. I see 2.8.0 as the version locally - the tests all pass and it works with all my GUI library examples.

Though please squash and merge to hide my messy experimentation :)

Starbuck5:
Thanks so much for investigating and fixing this!

> Though please squash and merge to hide my messy experimentation :)

I squashed all your commits in the branch together, since I don't want to have all this end up as one commit. This also unintentionally adds me as a coauthor, which I don't think I can get around.

oddbookworm (approved):
LGTM!

src_c/imageext.c
```diff
@@ -385,19 +191,14 @@ image_save_ext(PyObject *self, PyObject *arg, PyObject *kwarg)
...
         else if (!strcasecmp(ext, "png")) {
-#ifdef PNG_H
             /*Py_BEGIN_ALLOW_THREADS; */
             if (rw != NULL) {
-                result = pg_SavePNG_RW(surf, rw, 0);
+                result = IMG_SavePNG_RW(surf, rw, 0);
             }
             else {
-                result = pg_SavePNG(surf, name);
+                result = IMG_SavePNG(surf, name);
```
ankith26, line 199: Might be worth adding a python warning here that fires if an old `SDL_image` is used

Starbuck5: Interesting idea, but I don't think the old SDL_image behavior will be that disruptive, I don't think it's worth it. And there's nothing people will be able to do about the warning either, since they presumably installed pygame-ce the only way they know how.

test/image_test.py
```diff
@@ -196,6 +196,10 @@ def testSavePNG32(self):
             del reader
             os.remove(f_path)
 
+    @unittest.skipIf(
+        pygame.image.get_sdl_image_version() == (2, 0, 5),
```
ankith26, line 200: Maybe do a `< (2, 8, 0)` instead?

docs/reST/ref/image.rst
```diff
@@ -35,6 +35,8 @@ formats.
    
    * ``PNM``
 
+   * ``QOI``
+
    * ``SVG`` (limited support, using Nano SVG)
 
    * ``TGA`` (uncompressed)
```
ankith26, line 42: related to this PR but not important, we should also document supporting the `XCF` format for completeness.

src_c/imageext.c
ankith26: okay, fair enough

test/image_test.py
Starbuck5: The PNG saving code this PR deletes was copied from SDL_image by you, and it seems to be the code from SDL_image 2.6. Therefore I'm only excluding 2.0.5.

docs/reST/ref/image.rst
Starbuck5: I would like that to be a separate PR, because I don't know anything about XCF and once it's documented it should also be tested.

test/image_test.py
ankith26: We still support 2.0.4 so this skip should likely apply there as well, and in case of 2.6.x this issue will pop up with the default stbimage backend but not the libpng backend which this code was taken from. In 2.8.0 this issue should not appear on either backend so we should run this test then.

Starbuck5: None of this has been running on the stbimage backend, and if it does, more than just this test will fail.

I don't think it's worth the effort trying to make it show passing in all instances. The whole point of the tests failing is that  we consider some of these behaviors not passing.

Also we should raise to 2.0.5, it's the default in Ubuntu focal (https://packages.ubuntu.com/focal/libsdl2-image-2.0-0) (Python 3.8)

ankith26: Okay I was definitely confusing the 24bit png issue and palette png issue so this is fine

ankith26 (approved):
Gave this PR a good spin locally on linux and all is working well!

I tested both, the CI wheel and built-locally wheel (that in my case, would build against SDL_image 2.6.3) and both pass all tests as well as my little image loader test script. 

This is a good simplification of things and will also help us get a lot of new features and maybe bugfixes :partying_face:

PR #2597, opened 2023-12-10
Fix linked/non linked SDL image version getter

Starbuck5:
It was reversed from what it should be.

Checked all the other version getters, this is the only one with the problem.

Noticed this when I built a pre release SDL_image 2.9 with SDL_image 2.8 headers and it reported itself as SDL_image 2.8.


oddbookworm (approved):
LGTM! Thanks for fixing this mistake I accidentally made lol

ankith26 (approved):
LGTM, thanks! :tada:

PR #2598, opened 2023-12-11
Allow for clamping the zero vector when possible

oddbookworm:
I was playing around with `clamp_magnitude` and I noticed that `pygame.Vector2().clamp_magnitude(1)` would always result in a `ValueError` being raised, even though a length of 0 is a perfectly fine output here. I've just modified it to follow these examples:
```
pygame.Vector2().clamp_magnitude(5) # should be allowed
pygame.Vector2().clamp_magnitude(0, 5) # should be allowed
pygame.Vector2().clamp_magnitude(0.1, 5) # should not be allowed
```
The first two should be fine to return the zero vector, the last one is impossible to do because it's ambiguous in what direction it should go in (just like normalizing the zero vector is impossible)

### DIFF
--- a/docs/reST/ref/math.rst
+++ b/docs/reST/ref/math.rst
@@ -471,7 +471,7 @@ Multiple coordinates can be set using slices or swizzling
       | :sg:`clamp_magnitude(min_length, max_length, /) -> Vector2`
 
       **Experimental:** feature still in development available for testing and feedback. It may change.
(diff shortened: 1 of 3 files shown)


ankith26 (commented):
This seems like a reasonable change of behaviour, thanks for the PR :tada:

ankith26 (approved):

MyreMylar (approved):
LGTM 👍

PR #2601, opened 2023-12-13
Improve performance of SSE2 no_surf_alpha_opaque_dst blitter

Starbuck5:
Previously, this blitter used a 2x path if the blit had an even width, and a 1x path if the blit had an odd width. Now it uses a 4x path in all cases, with stride switching to handle excess pixels. ("x" = "pixels at a time" in this parlance, 2x = 2 pixels at a time, etc.)

The 4x strategy was inspired by the AVX blitters. There are operations it can do on 4 pixels at a time (load, store, alpha masking), then it goes into sub batches to process the first 2 pixels and the second 2 pixels.

**I measure a 17% time reduction for an 800x100 blit, and a 56% time reduction for a 799x100 blit.**

Why are the SSE blitters important? They are our cross platform accelerators. ARM uses them, using sse2neon. If we ever get WASM SIMD working it will use them. AVX2 is still way faster and shinier for x86, but SSE is important. In fact, it might become more important in coming years, such that we might want to raise our target to SSE3/SSE4 to take advantage of more advanced instructions.

**Testing**
- Add `#undef __AVX2__` to simd_blitters_avx2.c to make sure you're testing SSE2 blitters

<details>

<summary>Performance test script</summary>

```py
import pygame
import time

pygame.init()

width = 799

screen = pygame.Surface((width,100), depth=32)
screen.fill((60, 70, 80))

surf = pygame.Surface((width,100), pygame.SRCALPHA)
surf.fill((5,216,77,120))

print(screen, screen.get_shifts(), screen.get_alpha(), screen.get_colorkey())
print(surf, surf.get_shifts(), surf.get_alpha(), surf.get_colorkey())

# blit once before to get rid of any spurious warnings
screen.blit(surf, (0, 0))  

start = time.time()
for _ in range(100000):
    screen.blit(surf, (0, 0))    
print(time.time() - start)
```

</details>

**Future things**
- It's very scary to me that this function is hardcoded to support ARGB and it seems like it could be called by any 4 byte alpha blit. This PR doesn't change the behavior at all, but I'd like this blitter to support varied alpha channel locations like the modern AVX2 ones.
- This 4x strategy should be relevant to other things in simd_blitters_sse2, like the blend modes.
- In the future this could be more macro-ed up, like our modern AVX2 blitters and the filler codes. It's not ready for that yet. This gets it closer. The other alpha blitters need significant work.

### DIFF
--- a/src_c/simd_blitters_sse2.c
+++ b/src_c/simd_blitters_sse2.c
@@ -479,164 +479,135 @@ alphablit_alpha_sse2_argb_no_surf_alpha_opaque_dst(SDL_BlitInfo *info)
     int srcskip = info->s_skip >> 2;
     int dstskip = info->d_skip >> 2;
 
-    Uint64 *srcp64 = (Uint64 *)info->s_pixels;
-    Uint64 *dstp64 = (Uint64 *)info->d_pixels;
-
-    Uint64 rgb_mask64 = 0x00FFFFFF00FFFFFF;
-    Uint32 rgb_mask32 = 0x00FFFFFF;
-
     Uint32 *srcp32 = (Uint32 *)info->s_pixels;
     Uint32 *dstp32 = (Uint32 *)info->d_pixels;
 
-    __m128i src1, dst1, sub_dst, mm_src_alpha, mm_zero, mm_rgb_mask;
+    int pre_2_width = width % 2;
+    int post_2_width = width / 2;
 
-    /* There are two paths through this blitter:
-           1. Two pixels at once.
-           2. One pixel at a time.
-    */
-    if (((width % 2) == 0) && ((srcskip % 2) == 0) && ((dstskip % 2) == 0)) {
-        width = width / 2;
-        srcskip = srcskip / 2;
-        dstskip = dstskip / 2;
+    __m128i src1, dst1, sub_dst, mm_src_alpha;
+    __m128i mm_rgb_mask = _mm_set1_epi32(0x00FFFFFF);
+    __m128i mm_zero = _mm_setzero_si128();
 
-        mm_zero = _mm_setzero_si128();
+    while (height--) {
+        LOOP_UNROLLED4(
+            {
+                /* src(ARGB) -> src1 (00000000ARGBARGB) */
+                LOAD_64_INTO_M128((Uint64 *)srcp32, &src1);
 
-        /* two pixels at a time */
-        LOAD_64_INTO_M128(&rgb_mask64, &mm_rgb_mask);
-        while (height--) {
-            LOOP_UNROLLED4(
-                {
-                    /* src(ARGB) -> src1 (00000000ARGBARGB) */
-                    LOAD_64_INTO_M128(srcp64, &src1);
+                /* isolate alpha channels
+                 * 00000000A1000A2000 -> mm_src_alpha */
+                mm_src_alpha = _mm_andnot_si128(mm_rgb_mask, src1);
 
-                    /* isolate alpha channels
-                     * 00000000A1000A2000 -> mm_src_alpha */
-                    mm_src_alpha = _mm_andnot_si128(mm_rgb_mask, src1);
+                /* shift right to position alpha channels for manipulation
+                 * 000000000A1000A200 -> mm_src_alpha*/
+                mm_src_alpha = _mm_srli_si128(mm_src_alpha, 1);
 
-                    /* shift right to position alpha channels for manipulation
-                     * 000000000A1000A200 -> mm_src_alpha*/
-                    mm_src_alpha = _mm_srli_si128(mm_src_alpha, 1);
+                /* shuffle alpha channels to duplicate 16 bit pairs
+                 * shuffle (3, 3, 1, 1) (backed 2 bit numbers)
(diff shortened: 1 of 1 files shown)


MyreMylar (approved):
LGTM 👍 

My test results using the test program:

```
Before PR Test Results (multiple runs) - 799x100
--------------------------------------
9.54190182685852
9.535496711730957
9.523637771606445
9.511376142501831

After PR Test Results (multiple runs) - 799x100
--------------------------------------
4.210539102554321
4.189740419387817
4.187772512435913
4.1871418952941895



Before PR Test Results (multiple runs) - 800x600
--------------------------------------
5.045648813247681
5.021234035491943
5.04343843460083
5.013594388961792

After PR Test Results (multiple runs) - 800x600
--------------------------------------
4.173273324966431
4.166048526763916
4.1809751987457275
4.171604871749878
```

A very good speedup on the odd/large number of trailing pixels per row case.

novialriptide (approved):
Major performance boosts on the Apple M2 Max

## Before

```
<799x100>
8.609078884124756
8.535655975341797
8.516473054885864
```

```
<800x100>
4.360736131668091
4.322415113449097
4.34481406211853
```

## After

```
<799x100>
3.417734146118164
3.400517702102661
3.3988189697265625

<800x100>
3.435307025909424
3.3835179805755615
3.361617088317871
```

Starbuck5:
I have the reviews necessary to merge, but I plan on leaving this open a few more days in case @itzpr3d4t0r wants to take a look.

src_c/simd_blitters_sse2.c
```diff
@@ -479,164 +500,129 @@ alphablit_alpha_sse2_argb_no_surf_alpha_opaque_dst(SDL_BlitInfo *info)
...
-                    sub_dst = _mm_srli_epi16(sub_dst, 8);
+            /*[00][00][00][00][00][00][0A][0A] -> mm_src_alpha*/
+            mm_src_alpha = _mm_unpacklo_epi16(mm_src_alpha, mm_src_alpha);
 
-                    /* pack everything back into a pixel */
-                    sub_dst = _mm_packus_epi16(sub_dst, mm_zero);
-                    sub_dst = _mm_and_si128(sub_dst, mm_rgb_mask);
-                    /* reset alpha to 0 */
-                    *dstp32 = _mm_cvtsi128_si32(sub_dst);
+            /*[00][00][00][00][0A][0A][0A][0A] -> mm_src_alpha*/
```
itzpr3d4t0r, line 600: Just curious, could there be a way to reduce the amount of unpacks we do here? like with a shuffle?, maybe like get the alpha in a register, shuffle that value in all other positions, and then rihght shift that? Could that save an instruction?

src_c/simd_blitters_sse2.c
```diff
@@ -479,164 +500,129 @@ alphablit_alpha_sse2_argb_no_surf_alpha_opaque_dst(SDL_BlitInfo *info)
...
-                    mm_src_alpha = _mm_andnot_si128(mm_rgb_mask, src1);
+                /*[AR][GB][AR][GB][AR][GB][AR][GB] -> pixels_src*/
+                pixels_src = _mm_loadu_si128(srcp128);
 
-                    /* shift right to position alpha channels for manipulation
-                     * 000000000A1000A200 -> mm_src_alpha*/
-                    mm_src_alpha = _mm_srli_si128(mm_src_alpha, 1);
+                /* isolate alpha channels
+                 * [A10][00 ][A20][00 ][A30][00 ][A40][00 ] -> mm_src_alpha*/
+                mm_src_alpha = _mm_andnot_si128(mm_rgb_mask, pixels_src);
```
itzpr3d4t0r, line 528: Couldn't we just use `_mm_and_si128` with the alpha mask instead of also inverting the rgb mask?

itzpr3d4t0r (approved):
Aside from minor a minor nitpick and a question, this looks good to me, good job!!

src_c/simd_blitters_sse2.c
itzpr3d4t0r: Yeah actually when experimenting with premul_alpha() i found out a shuffle is straight up better, replaces multiple instructions with one.

src_c/simd_blitters_sse2.c
Starbuck5: This function does not have an alpha mask.

src_c/simd_blitters_sse2.c
Starbuck5: Of course a shuffle would be better, the problem is that SSE2 does not have a byte-level shuffle instruction. SSSE3 does, AVX or AVX2 does.

Starbuck5: Also the alpha fetching here is not dynamic to the actual location of the alpha channel. (Same as it's always been in this function).

I'd be willing to trade performance for making this work correctly on different pixelformats.

Starbuck5: The question is "what amount of ARM systems justify a 3rd optimization level," where we put in an SSE4.2 level of SIMD for translation of advanced features like byte-level shuffles to NEON.

PR #2606, opened 2023-12-16
Fix file-like `close` not called in freetype (Removed `pgRWops_ReleaseObject` and replaced usage with `SDL_RWclose`)

ankith26:
When I first saw `pgRWops_ReleaseObject` I was pretty confused, then came to the conclusion that its implementation is unnecessary and incorrect. I forgot about it for a while but today I stumbled across it again, and went *WHY IS IT NOT REMOVED YET*.

- `pgRWops_ReleaseObject` was only called in freetype objects dealloc, every other dealloc uses the regular `SDL_RWclose` which can handle all cleanups correctly.
- The exact issue with `pgRWops_ReleaseObject` is that it does not call `close` on the file-like object (unless its refcount is exactly 6)

Here is a script that I wrote to test whether any pygame resource loading API leaks memory (testing against a few common kinds of input). Running this program for some time while observing the memory usage, I see a constant memory usage as expected (yay no leaks!). Now before this PR, the output clearly shows `freetype.Font` loads consistently missing `close` calls on file-like objects, after this PR none of the loads have this issue.

<details>
<summary>Click for test script</summary>

```py
from io import BytesIO
from pathlib import Path

import pygame
import pygame.freetype

close_miss_counter = {}


class TrackedBytesIO(BytesIO):
    def __init__(self, *args, caller, **kwargs):
        self._closed = False
        self.caller = caller
        if caller not in close_miss_counter:
            close_miss_counter[self.caller] = 0

        return super().__init__(*args, **kwargs)

    def close(self, *args, **kwargs):
        self._closed = True
        return super().close(*args, **kwargs)

    def __del__(self, *args, **kwargs):
        if not self._closed:
            close_miss_counter[self.caller] += 1

        return super().__del__(*args, **kwargs)


pygame.init()


def test_formats(file: str, caller: str):
    file_obj = Path(file)
    yield file
    yield file_obj
    yield TrackedBytesIO(file_obj.read_bytes(), caller=caller)


loops = 1
try:
    cur = None
    while True:
        cur = pygame.freetype.Font(None)
        del cur

        cur = pygame.font.Font(None)
        del cur

        for i in test_formats(
            "test/fixtures/fonts/PlayfairDisplaySemibold.ttf", "freetype"
        ):
            cur = pygame.freetype.Font(i)
            del cur

        for i in test_formats(
            "test/fixtures/fonts/PlayfairDisplaySemibold.ttf", "font"
        ):
            cur = pygame.font.Font(i)
            del cur

        for i in test_formats("test/fixtures/fonts/A_PyGameMono-8.png", "image"):
            cur = pygame.image.load(i)
            del cur

        for i in test_formats("examples/data/boom.wav", "mixer"):
            cur = pygame.mixer.Sound(i)
            del cur

        for i in test_formats("examples/data/boom.wav", "music"):
            cur = pygame.mixer.music.load(i)
            del cur

        loops += 1

except KeyboardInterrupt:
    pass

finally:
    print(
        f"\nTest results (after {loops} iterations):\n"
        f"Misseded close calls: {close_miss_counter}",
    )
```

</details>


MyreMylar (approved):
Sounds like this was frustrating! 👍

zoldalma999 (approved):
Looks good to me!

PR #2608, opened 2023-12-17
Test PBM and LBM images, test and document XCF

ankith26:
Simple PR to update docs and tests on supported images, we now test every file format that is documented.

EDIT: Nice, the added test in this PR already caught a new `SDL_image` regression (which I patched upstream and now it should hopefully make its way to the next release)

EDIT2: Some more info, `PBM` and `LBM` are not really widely used formats, but `XCF` is the "native image format" used by GIMP so I imagine documented and tested support for it could be useful.


Starbuck5 (approved):
Nice work getting a fix upstreamed! I feel like we've been increasingly more communicative with the SDL peeps, with good results.

This PR seems simple enough, ran the tests locally just to be sure it was all good there. Approved.

MyreMylar (approved):
LGTM, all works locally for me too. 👍

PR #2610, opened 2023-12-17
Border radius for pygame.draw.polygon (resolves #1527)

### LINKED ISSUE #1527: Border radius for polygons (3011)
#  **Issue &numero;3011 opened by *[zenthm](https://github.com/zenthm)* at 2022-02-02 02:09:50**

I want pygame's ```pygame.draw.polygon``` to have a border-radius too, like in ```pygame.draw.rect``` where you can set the border radius of the rectangle<hr>

#  Comments

# # #  **[MalinSvenberg](https://github.com/MalinSvenberg)* commented at 2022-03-03 13:48:20*

Hello, me and my project group in a University course would like to claim this issue, would that be possible? :)
<hr>

# # #  **[MyreMylar](https://github.com/MyreMylar)* commented at 2022-03-03 18:50:16*

I've assigned the issue to you.

ezzakri-anas:
### This pull request resolves the enhancement issue #1527 (previously #3011)  about implementing the border radius for pygame.draw.polygon

# Overview
- Added a new parameter _border_radius_ to the **_pygame.draw.polygon_** function to specify the border radius of the rounded polygon corners.
- Updated the list of keywords in _**pygame.draw.polygon**_ to include _"border_radius"_ in the function signature.
- Added a new function **_draw_round_polygon_** to handle the drawing of rounded polygons. This function takes care of the logic for calculating rounded corners.
- Integrated this function as a part of _**polygon**_ in draw.c.

## Geometry Calculations:
- Implemented various geometric calculation functions, such as **_find_parallel_line_**, **_project_point_onto_segment_**, and **_intersection_**(to calculate intersections between two lines), _**angle**_, _**side**_ ….

# draw_round_polygon Function:

**static void draw_round_polygon(SDL_Surface *surf, int *pts_x, int *pts_y, int radius, int num_points, Uint32 color, int *drawn_area)**

## Purpose:
This function is responsible for drawing a rounded polygon with a specified border radius. It calculates and draws circular segments at each corner of the polygon.
Parameters:
- _surf:_ The target surface to draw on.
- _pts_x:_ Array of x-coordinates of polygon vertices.
- _pts_y:_ Array of y-coordinates of polygon vertices.
- _border_radius:_ The border radius of the rounded corners.
- _num_points_: Number of vertices in the polygon.
- _color_: Color of the polygon.
- _drawn_area_: Array to store the bounding box of the drawn area.

### Path calculation loop:
- Coordinate Initialization: Initializes arrays and structures to store calculated points and circles.
- Parallel Line Calculation: The **_find_parallel_line_** function calculates two parallel lines given three points: the current vertex and its adjacent vertices. The function ensures that the parallel lines are on the correct side of the vertex.
- Circle and Path Calculation: The center of the circular segment is found by calculating the intersection of two parallel lines (line1 and line2) using the **_intersection_** function.
- The circular arc's starting and ending points are obtained by projecting the center onto adjacent line segments using the **_project_point_onto_segment_** function.
- Stores these points in the path array.

### Drawing loop:
- For each segment in the path array, it calculates start and end angles and draws the circular arc using the **_draw_arc_** function.
- Draws a straight line between consecutive circular arcs.

## Debugging Output:
- During the calculation loop, if three consecutive points are aligned the loop is interrupted with a return statement to abort the function.
- A condition on the border_radius is verified on each vertex, at each iteration, to avoid incorrect drawing. if the condition is not satisfied the function is aborted.
- As the draw_round_polygon function returns void (a choice made to align with the same logic of other drawing functions), it is not possible to raise an error directly from draw_round_polygon. Instead, we settle for a return statement whenever an error should be raised to abort the function, and a printf of the error message.

### DIFF
--- a/src_c/doc/draw_doc.h
+++ b/src_c/doc/draw_doc.h
@@ -1,7 +1,7 @@
 /* Auto generated file: with makeref.py .  Docs go in docs/reST/ref/ . */
 #define DOC_DRAW "pygame module for drawing shapes"
 #define DOC_DRAW_RECT "rect(surface, color, rect) -> Rect\nrect(surface, color, rect, width=0, border_radius=0, border_top_left_radius=-1, border_top_right_radius=-1, border_bottom_left_radius=-1, border_bottom_right_radius=-1) -> Rect\ndraw a rectangle"
-#define DOC_DRAW_POLYGON "polygon(surface, color, points) -> Rect\npolygon(surface, color, points, width=0) -> Rect\ndraw a polygon"
+#define DOC_DRAW_POLYGON "polygon(surface, color, points) -> Rect\npolygon(surface, color, points, width=0, border_radius=0) -> Rect\ndraw a polygon"
 #define DOC_DRAW_CIRCLE "circle(surface, color, center, radius) -> Rect\ncircle(surface, color, center, radius, width=0, draw_top_right=None, draw_top_left=None, draw_bottom_left=None, draw_bottom_right=None) -> Rect\ndraw a circle"
 #define DOC_DRAW_ELLIPSE "ellipse(surface, color, rect) -> Rect\nellipse(surface, color, rect, width=0) -> Rect\ndraw an ellipse"
 #define DOC_DRAW_ARC "arc(surface, color, rect, start_angle, stop_angle) -> Rect\narc(surface, color, rect, start_angle, stop_angle, width=1) -> Rect\ndraw an elliptical arc"
--- a/src_c/draw.c
+++ b/src_c/draw.c
@@ -30,7 +30,7 @@
 #include "doc/draw_doc.h"
 
 #include <math.h>
-#define sametype(A, B) _Generic(*((A *)0), B: true, default: false)
+
 #include <float.h>
 
 #ifndef M_PI
--- a/src_c/draw.c
+++ b/src_c/draw.c
@@ -84,10 +84,10 @@ static void
 draw_round_rect(SDL_Surface *surf, int x1, int y1, int x2, int y2, int radius,
                 int width, Uint32 color, int top_left, int top_right,
                 int bottom_left, int bottom_right, int *drawn_area);
-static void
-draw_round_polygon(SDL_Surface *surf, int *pts_x, int *pts_y, int width, int border_radius, 
-                    int num_points, Uint32 color, int *drawn_area);
-
+static int
+draw_round_polygon(SDL_Surface *surf, int *pts_x, int *pts_y, int width,
+                   int border_radius, int num_points, Uint32 color,
+                   int *drawn_area);
 
 // validation of a draw color
 #define CHECK_LOAD_COLOR(colorobj)                               \
@@ -724,13 +724,14 @@ polygon(PyObject *self, PyObject *arg, PyObject *kwargs)
     SDL_Surface *surf = NULL;
     Uint32 color;
     int *xlist = NULL, *ylist = NULL;
-    int width = 0; /* Default width. */
+    int width = 0;         /* Default width. */
     int border_radius = 0; /* Default border_radius. */
     int x, y, result, l, t;
     int drawn_area[4] = {INT_MAX, INT_MAX, INT_MIN,
                          INT_MIN}; /* Used to store bounding box values */
     Py_ssize_t loop, length;
-    static char *keywords[] = {"surface", "color", "points", "width", "border_radius", NULL};
+    static char *keywords[] = {"surface", "color",         "points",
+                               "width",   "border_radius", NULL};
 
     if (!PyArg_ParseTupleAndKeywords(arg, kwargs, "O!OO|ii", keywords,
                                      &pgSurface_Type, &surfobj, &colorobj,
@@ -814,7 +815,7 @@ polygon(PyObject *self, PyObject *arg, PyObject *kwargs)
         return RAISE(PyExc_RuntimeError, "error locking surface");
     }
 
-        if (length != 3) {
+    if (length != 3) {
         if ((border_radius > 0) && (width > 0)) {
             int err;
             err = draw_round_polygon(surf, xlist, ylist, width, border_radius,
@@ -2856,58 +2857,75 @@ side(Point a, Point b, Point c)
 {
     double det = (a.x * b.y + b.x * c.y + c.x * a.y) -
                  (a.y * b.x + b.y * c.x + c.y * a.x);
-        if (det > 0) {
+    if (det > 0) {
         return 1;
-    // If the determinant is positive, point 'c' is on the left side of the line
-    } else if (det < 0) {
-        // If the determinant is negative, point 'c' is on the right side of the line
+        // If the determinant is positive, point 'c' is on the left side of the
+        // line
+    }
+    else if (det < 0) {
+        // If the determinant is negative, point 'c' is on the right side of
+        // the line
         return -1;
-    } else {
+    }
+    else {
         // If the determinant is zero, points 'a', 'b', and 'c' are collinear
         return 0;
     }
 }
 
-// Function to find a parallel line to the line formed by 'pt1' and 'pt2' at a given distance
-// 'pt3' is used to determin in which side of the line, the parallel line should be drawn
-// Returns a Line struct representing the parallel line
-Line find_parallel_line(Point pt1, Point pt2, Point pt3, int distance) {
-    
-    // Calculate direction vector, normalize it, and find the perpendicular vector
-    
+// Function to find a parallel line to the line formed by 'pt1' and 'pt2' at a
+// given distance 'pt3' is used to determin in which side of the line, the
+// parallel line should be drawn Returns a Line struct representing the
+// parallel line
+Line
+find_parallel_line(Point pt1, Point pt2, Point pt3, int distance)
+{
+    // Calculate direction vector, normalize it, and find the perpendicular
+    // vector
+
     // Calculate direction vector from 'pt1' to 'pt2'
     Point direction_vector = {pt2.x - pt1.x, pt2.y - pt1.y};
 
     // Calculate the magnitude of the direction vector
-    float magnitude = sqrt(direction_vector.x * direction_vector.x + direction_vector.y * direction_vector.y);
-    
+    float magnitude = sqrt(direction_vector.x * direction_vector.x +
+                           direction_vector.y * direction_vector.y);
+
     // Normalize the direction vector to get a unit vector
-    Point normalized_direction = {direction_vector.x / magnitude, direction_vector.y / magnitude};
-    
-    // Calculate the perpendicular vector by swapping x and y components and negating one of them
-    Point perpendicular_vector = {-normalized_direction.y, normalized_direction.x};
-    
-    // Adjust the perpendicular vector based on the side of 'pt3' relative to 'pt1' and 'pt2'
+    Point normalized_direction = {direction_vector.x / magnitude,
+                                  direction_vector.y / magnitude};
+
+    // Calculate the perpendicular vector by swapping x and y components and
+    // negating one of them
+    Point perpendicular_vector = {-normalized_direction.y,
+                                  normalized_direction.x};
+
+    // Adjust the perpendicular vector based on the side of 'pt3' relative to
+    // 'pt1' and 'pt2'
     if (side(pt1, pt2, pt3) == -1) {
         perpendicular_vector.x *= -1;
         perpendicular_vector.y *= -1;
     }
     // Create an offset vector and generate the parallel line.
-    Point offset_vector = {perpendicular_vector.x * distance, perpendicular_vector.y * distance};
-    
+    Point offset_vector = {perpendicular_vector.x * distance,
+                           perpendicular_vector.y * distance};
+
     // Generate the points for the parallel line based on the offset
-    Line parallel_line = {{pt1.x + offset_vector.x, pt1.y + offset_vector.y}, {pt2.x + offset_vector.x, pt2.y + offset_vector.y}};
-    
+    Line parallel_line = {{pt1.x + offset_vector.x, pt1.y + offset_vector.y},
+                          {pt2.x + offset_vector.x, pt2.y + offset_vector.y}};
+
     // Return the Line struct representing the parallel line
     return parallel_line;
 }
 
-
-// Function to project a point onto a line segment defined by 'segment_start' and 'segment_end'
-Point project_point_onto_segment(Point point, Point segment_start,
-                                 Point segment_end) {
-    // t is a parameter that represents the position of the projected point along the line segment defined by segment_start and segment_end.                               
-    // Calculate vectors, find the parameter 't' for projection, and calculate the projection.
+// Function to project a point onto a line segment defined by 'segment_start'
+// and 'segment_end'
+Point
+project_point_onto_segment(Point point, Point segment_start, Point segment_end)
+{
+    // t is a parameter that represents the position of the projected point
+    // along the line segment defined by segment_start and segment_end.
+    // Calculate vectors, find the parameter 't' for projection, and calculate
+    // the projection.
     Point segment_vector;
     segment_vector.x = segment_end.x - segment_start.x;
     segment_vector.y = segment_end.y - segment_start.y;
@@ -2917,13 +2935,14 @@ Point project_point_onto_segment(Point point, Point segment_start,
     point_vector.y = point.y - segment_start.y;
 
     // Calculate the parameter 't' for the projection using the dot product
-    double t =
-        (point_vector.x * segment_vector.x + point_vector.y * segment_vector.y) /
-        (segment_vector.x * segment_vector.x +
-        segment_vector.y * segment_vector.y);
+    double t = (point_vector.x * segment_vector.x +
+                point_vector.y * segment_vector.y) /
+               (segment_vector.x * segment_vector.x +
+                segment_vector.y * segment_vector.y);
 
-    // Ensure 't' is within the valid range [0, 1] to make sure that the projection of the point onto the line segment lies specifically within the boundaries of the segment
-    // See line 2906
+    // Ensure 't' is within the valid range [0, 1] to make sure that the
+    // projection of the point onto the line segment lies specifically within
+    // the boundaries of the segment See line 2906
     t = fmax(0, fmin(1, t));
 
     // Calculate the coordinates of the projected point using the parameter 't'
@@ -2935,11 +2954,12 @@ Point project_point_onto_segment(Point point, Point segment_start,
     return projection;
 }
 
-
 // Function to find the intersection point of two line segments
 // Returns the intersection point as a Point struct
-Point intersection(Point line1_start, Point line1_end, Point line2_start,
-                   Point line2_end) {
+Point
+intersection(Point line1_start, Point line1_end, Point line2_start,
+             Point line2_end)
+{
     // Calculate coefficients for each line equation
     double A1 = line1_end.y - line1_start.y;
     double B1 = line1_start.x - line1_end.x;
@@ -2951,26 +2971,33 @@ Point intersection(Point line1_start, Point line1_end, Point line2_start,
 
     // Check if the lines are parallel (det == 0) and handle the case
     double det = A1 * B2 - A2 * B1;
-    // If the lines are parallel (det == 0), they do not intersect, return a default Point
+    // If the lines are parallel (det == 0), they do not intersect, return a
+    // default Point
     if (det == 0) {
         return (Point){0, 0};
-    } else {
+    }
+    else {
         // Calculate the intersection point using Cramer's rule
-        // The lines never intersect out of the range given because of the condition limiting the value of boarder radius
+        // The lines never intersect out of the range given because of the
+        // condition limiting the value of boarder radius
         double x = (B2 * C1 - B1 * C2) / det;
         double y = (A1 * C2 - A2 * C1) / det;
         return (Point){x, y};
     }
 }
 
-// Function to calculate the angle (in radians) between a center point and another point
-double angle(Point center, Point point) {
+// Function to calculate the angle (in radians) between a center point and
+// another point
+double
+angle(Point center, Point point)
+{
     // Calculate the differences in x and y coordinates between the two points
     double x = point.x - center.x;
     double y = point.y - center.y;
-    // Use atan2 to calculate the angle formed by the vector from the center to the point
-    // The angle is measured counterclockwise from the positive x-axis
-    // Note: The negative sign is used to ensure the angle is measured clockwise, which is more common in Cartesian coordinates.
+    // Use atan2 to calculate the angle formed by the vector from the center to
+    // the point The angle is measured counterclockwise from the positive
+    // x-axis Note: The negative sign is used to ensure the angle is measured
+    // clockwise, which is more common in Cartesian coordinates.
     return -atan2(y, x);
 }
 
@@ -3010,13 +3037,13 @@ draw_round_polygon(SDL_Surface *surf, int *pts_x, int *pts_y, int width,
         points[i].y = pts_y[i];
     }
 
-        for (int i = 0; i < num_points; i++) {
+    for (int i = 0; i < num_points; i++) {
         int a = (i - 1 + num_points) %
                 (num_points);            // index of the point before index i
         int b = (i + 1) % (num_points);  // index of the point after index i
 
         // Check if the border-radius can be applied to the current angle
-// it can not be applied to a flat angle(if the three points are
+        // it can not be applied to a flat angle(if the three points are
         // aligned)
         if (side(points[a], points[i], points[b]) == 0) {
             PyErr_SetString(
@@ -3042,14 +3069,14 @@ draw_round_polygon(SDL_Surface *surf, int *pts_x, int *pts_y, int width,
 
         // Check if the border-radius size is valid
         if ((sqrt(pow(proj1.x - points[i].x, 2) +
-            pow(proj1.y - points[i].y, 2)) >=
+                  pow(proj1.y - points[i].y, 2)) >=
              sqrt(pow(points[a].x - points[i].x, 2) +
-                                                    pow(points[a].y - points[i].y, 2)) /
-                 2) ||  
+                  pow(points[a].y - points[i].y, 2)) /
+                 2) ||
             (sqrt(pow(proj2.x - points[i].x, 2) +
-                pow(proj2.y - points[i].y, 2)) >=
+                  pow(proj2.y - points[i].y, 2)) >=
              sqrt(pow(points[b].x - points[i].x, 2) +
-                                                        pow(points[b].y - points[i].y, 2)) /
+                  pow(points[b].y - points[i].y, 2)) /
                  2)) {
             PyErr_SetString(PyExc_ValueError,
                             "Border-radius size must be smaller\n");
@@ -3070,7 +3097,7 @@ draw_round_polygon(SDL_Surface *surf, int *pts_x, int *pts_y, int width,
         Point pt3 = path[(i + 2) % (2 * num_points)];
         Circle circle = circles[i / 2];
 
-                double start_angle = angle(circle.center, pt1);
+        double start_angle = angle(circle.center, pt1);
         double end_angle = angle(circle.center, pt2);
 
         // Adjust angles based on the side of the polygon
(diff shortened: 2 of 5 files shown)


src_c/doc/draw_doc.h
```diff
@@ -1,7 +1,7 @@
 /* Auto generated file: with makeref.py .  Docs go in docs/reST/ref/ . */
 #define DOC_DRAW "pygame module for drawing shapes"
 #define DOC_DRAW_RECT "rect(surface, color, rect) -> Rect\nrect(surface, color, rect, width=0, border_radius=0, border_top_left_radius=-1, border_top_right_radius=-1, border_bottom_left_radius=-1, border_bottom_right_radius=-1) -> Rect\ndraw a rectangle"
-#define DOC_DRAW_POLYGON "polygon(surface, color, points) -> Rect\npolygon(surface, color, points, width=0) -> Rect\ndraw a polygon"
+#define DOC_DRAW_POLYGON "polygon(surface, color, points) -> Rect\npolygon(surface, color, points, width=0, border_radius=0) -> Rect\ndraw a polygon"
```
oddbookworm, line 4: Was this edited manually? I don't see where you generated this output from a modified doc file. If you read the first line of this file, it's an autogen file created from `docs/reST/ref/draw.rst`. You should make your change there in the appropriate method and then run `python setup.py docs --fullgeneration`, then commit the change to the ReStructured Text file and any changes to this header made by the autogen.

src_c/draw.c
```diff
@@ -84,6 +84,10 @@ static void
 draw_round_rect(SDL_Surface *surf, int x1, int y1, int x2, int y2, int radius,
                 int width, Uint32 color, int top_left, int top_right,
                 int bottom_left, int bottom_right, int *drawn_area);
+                static void
```
oddbookworm, line 87: Indentation is weird here, make sure you have `clang-format` and `black` installed in your python environment and run `python setup.py format` to apply our code formatting standards

src_c/draw.c
```diff
@@ -721,19 +725,20 @@ polygon(PyObject *self, PyObject *arg, PyObject *kwargs)
     Uint32 color;
     int *xlist = NULL, *ylist = NULL;
     int width = 0; /* Default width. */
+    int border_radius = 0; /* Default border_radius. */
```
oddbookworm, line 728: This isn't a super important nitpick, but is there a reason this should be a signed integer vs unsigned? What do you expect to happen if a negative `border_radius` is fed in?

src_c/draw.c
```diff
@@ -810,7 +815,11 @@ polygon(PyObject *self, PyObject *arg, PyObject *kwargs)
     }
 
     if (length != 3) {
+        if(border_radius != 0){ 
+            draw_round_polygon(surf, xlist, ylist, width, border_radius, length, color, drawn_area);
+        } else {
         draw_fillpoly(surf, xlist, ylist, length, color, drawn_area);
+        }
```
oddbookworm, line 822: Another wacky indentation. I know you didn't change the indentation of the `draw_fillpoly` line, but with your added level of nesting, it just looks weird. `python setup.py format` should fix that up

src_c/draw.c
```diff
@@ -2818,6 +2827,233 @@ draw_round_rect(SDL_Surface *surf, int x1, int y1, int x2, int y2, int radius,
...
+typedef struct {
+    Point start;
+    Point end;
+} Line;
+
+// Function to determine the side of point 'c' relative to the line formed by points 'a' and 'b'
+// Returns 1 if on the left side, -1 if on the right side, and 0 if collinear
+int side(Point a, Point b, Point c) {
+    double det =
+        (a.x * b.y + b.x * c.y + c.x * a.y) - (a.y * b.x + b.y * c.x + c.y * a.x);
```
oddbookworm, line 2852: I'd like to see a comment explaining how this does what it claims to do

src_c/draw.c
```diff
@@ -2818,6 +2827,233 @@ draw_round_rect(SDL_Surface *surf, int x1, int y1, int x2, int y2, int radius,
...
+    } else if (det < 0) {
+        return -1;
+    } else {
+        return 0;
+    }
+}
+
+// Function to find a parallel line to the line formed by 'pt1' and 'pt2' at a given distance
+// Returns a Line struct representing the parallel line
+Line find_parallel_line(Point pt1, Point pt2, Point pt3, int distance) {
```
oddbookworm, line 2864: Your comment doesn't explain what `pt3` is for. Based on this comment, I would expect for it to calculate the slope of a line between `pt1` and `pt2` and shift it by `distance` (where a negative distance means one direction and a positive distance means the other)

src_c/draw.c
```diff
@@ -2818,6 +2827,233 @@ draw_round_rect(SDL_Surface *surf, int x1, int y1, int x2, int y2, int radius,
...
+    Point offset_vector = {perpendicular_vector.x * distance, perpendicular_vector.y * distance};
+    Line parallel_line = {{pt1.x + offset_vector.x, pt1.y + offset_vector.y}, {pt2.x + offset_vector.x, pt2.y + offset_vector.y}};
+    return parallel_line;
+}
+
+// Function to project a point onto a line segment defined by 'segment_start' and 'segment_end'
+Point project_point_onto_segment(Point point, Point segment_start,
+                                 Point segment_end) {
+    // Calculate vectors, find the parameter 't' for projection, and calculate the projection.
+    // Ensure 't' is within the valid range [0, 1].
```
oddbookworm, line 2885: What does this range mean? Why is it restricted to the closed interval [0, 1]? I'd also like to see comments within the code explaining what it's doing.

src_c/draw.c
```diff
@@ -2818,6 +2827,233 @@ draw_round_rect(SDL_Surface *surf, int x1, int y1, int x2, int y2, int radius,
...
+    Point projection;
+    projection.x = segment_start.x + t * segment_vector.x;
+    projection.y = segment_start.y + t * segment_vector.y;
+
+    return projection;
+}
+
+// Function to find the intersection point of two line segments
+// Returns the intersection point as a Point struct
+Point intersection(Point line1_start, Point line1_end, Point line2_start,
```
oddbookworm, line 2909: What happens if the lines are parallel and intersecting or parallel and nonintersecting? What about if the lines are not parallel but do not intersect within the range given?

src_c/draw.c
```diff
@@ -2818,6 +2827,233 @@ draw_round_rect(SDL_Surface *surf, int x1, int y1, int x2, int y2, int radius,
...
+    return -atan2(y, x);
+}
+
+// Define a function to draw a rounded polygon on an SDL surface
+static void
+draw_round_polygon(SDL_Surface *surf, int *pts_x, int *pts_y, int width, int border_radius, int num_points, Uint32 color, int *drawn_area) {
+
+    // Define arrays to store path points and circle information
+    Point path[2 * num_points];
+    Circle circles[num_points];
```
oddbookworm, line 2945: I'm pretty sure this won't even compile because `num_points` is a dynamic value. If there's no better way, you'll probably have to use `malloc` to create your array and then make sure you `free` the memory when you're done with it
Edit: just checked your CI runs. You indeed have an issue here

src_c/draw.c
```diff
@@ -2818,6 +2827,233 @@ draw_round_rect(SDL_Surface *surf, int x1, int y1, int x2, int y2, int radius,
...
+        } else if (i == (num_points - 1)) {
+            a = num_points - 2;
+            b = 0;
+        } else if (i == (num_points - 2)) {
+            a = num_points - 3;
+            b = num_points - 1;
+        } else {
+            a = i - 1;
+            b = i + 1;
+        }
```
oddbookworm, line 2972: IMO this would be better suited as a `switch/case`

src_c/draw.c
```diff
@@ -2818,6 +2827,233 @@ draw_round_rect(SDL_Surface *surf, int x1, int y1, int x2, int y2, int radius,
...
+
+        // Check if the border-radius can be applied to the current angle
+        if (side(points[a], points[i], points[b]) == 0) {
+            printf("ValueError: Border-radius cannot be applied to a flat angle.\n"
+                    "Please ensure that the specified angle or curvature is within a valid range for border-radius drawing.\n");
```
oddbookworm, line 2977: We don't use `printf` for error reporting. Please use the python C API to raise exceptions

src_c/draw.c
```diff
@@ -2818,6 +2827,233 @@ draw_round_rect(SDL_Surface *surf, int x1, int y1, int x2, int y2, int radius,
...
+        Point proj2 =
+            project_point_onto_segment(circles[i].center, points[i], points[b]);
+
+        // Check if the border-radius size is valid
+        if ((sqrt(pow(proj1.x - points[i].x, 2) +
+            pow(proj1.y - points[i].y, 2)) >= sqrt(pow(points[a].x - points[i].x, 2) +
+                                                    pow(points[a].y - points[i].y, 2)) / 2) ||  
+            (sqrt(pow(proj2.x - points[i].x, 2) +
+                pow(proj2.y - points[i].y, 2)) >= sqrt(pow(points[b].x - points[i].x, 2) +
+                                                        pow(points[b].y - points[i].y, 2)) / 2)) {
```
oddbookworm, line 2998: More math I'd like to see an explanation of

src_c/draw.c
```diff
@@ -2818,6 +2827,233 @@ draw_round_rect(SDL_Surface *surf, int x1, int y1, int x2, int y2, int radius,
...
+            printf("ValueError: Border-radius size must be smaller\n");
```
oddbookworm, line 2999: Again, don't use `printf` to report exceptions

oddbookworm (changes requested):
I see lots of undocumented math that isn't immediately obvious. Please make sure you're leaving enough comments that someone else can come in later and make fixes/changes without having to reinvent your entire thought process.

I also don't see any tests written. You should write tests to make sure that your code is behaving as expected. They live in the `test` directory.

Please update documentation (as I outline in one of my comments)

Formatting of the code can be wacky in a few places, please run `python setup.py format`

No stub updates. Since you're changing the signature of a function, you should update the appropriate stub in `buildconfig/stubs/pygame`

src_c/draw.c
ezzakri-anas: Thank you for the feedback! I'll make sure to take your suggestions into consideration. Quick question: for dynamic arrays, should I stick with malloc or use Pymem_New? Want to make sure it fits pygame conventions.

itzpr3d4t0r: https://docs.python.org/3/c-api/memory.html#:~:text=To%20avoid%20memory,the%20following%20example%3A

itzpr3d4t0r: So you can use malloc/free as long as you're not interacting with PyObjects or memory blocks allocated with Pymem_New.

ezzakri-anas: I'm currently utilizing xlist and ylist from pygame.draw.polygon in the draw.c file. These lists are created using PyMem_New(int, length). I'm trying to use them in creating an array of a struct variable called "Point" within my new function draw_round_polygon. Is this considered as an interaction with PyObjects?

src_c/draw.c
ezzakri-anas: In aligning with the pattern in draw.c, I opted for a void return type in draw_round_polygon. Would you recommend using perror for error reporting, or should I consider making the function return Raise()? Open to your suggestion.

Matiiss:
If you need to raise exceptions from your helper function, I would suggest making it return `-1` on failure and `0` on success (as Python C API functions do for the most part).
Before returning `-1`, you can call [`PyErr_SetString`](https://docs.python.org/3/c-api/exceptions.html#c.PyErr_SetString) with the appropriate exception type (seemingly `PyExc_ValueError`) and the message and then `return -1;`
Then in the "main" polygon function, check for `== -1` and `return NULL;`

itzpr3d4t0r:
You have to add a bunch of explicit casts:
image

src_c/draw.c
oddbookworm: I unresolved this because you didn't actually respond to it. If you're going to resolve a comment, there better be some kind of followup. If it doesn't warrant a change, then add a comment to reply. Since I was asking a question, I would expect an answer. Marking this as resolved without doing that signifies to me that you've done ***something*** to address it. In this case, you have done nothing, and my questions are still valid.

src_c/draw.c
```diff
@@ -720,20 +724,22 @@ polygon(PyObject *self, PyObject *arg, PyObject *kwargs)
...
-    int width = 0; /* Default width. */
+    int width = 0;         /* Default width. */
+    int border_radius = 0; /* Default border_radius. */
     int x, y, result, l, t;
     int drawn_area[4] = {INT_MAX, INT_MAX, INT_MIN,
                          INT_MIN}; /* Used to store bounding box values */
     Py_ssize_t loop, length;
-    static char *keywords[] = {"surface", "color", "points", "width", NULL};
+    static char *keywords[] = {"surface", "color",         "points",
+                               "width",   "border_radius", NULL};
```
oddbookworm, line 734: Spacing here is weird. Was this after running `python setup.py format`?

src_c/draw.c
```diff
@@ -2818,6 +2834,299 @@ draw_round_rect(SDL_Surface *surf, int x1, int y1, int x2, int y2, int radius,
...
+    else {
+        // If the determinant is zero, points 'a', 'b', and 'c' are collinear
+        return 0;
+    }
+}
+
+// Function to find a parallel line to the line formed by 'pt1' and 'pt2' at a
+// given distance 'pt3' is used to determin in which side of the line, the
+// parallel line should be drawn Returns a Line struct representing the
+// parallel line
```
oddbookworm, line 2879: Looks like maybe the autoformatting messed with your comment. Inserting some punctuation would help readability

src_c/draw.c
```diff
@@ -2818,6 +2834,299 @@ draw_round_rect(SDL_Surface *surf, int x1, int y1, int x2, int y2, int radius,
...
+    // Calculate direction vector from 'pt1' to 'pt2'
+    Point direction_vector = {pt2.x - pt1.x, pt2.y - pt1.y};
+
+    // Calculate the magnitude of the direction vector
+    double magnitude = sqrt(direction_vector.x * direction_vector.x +
+                            direction_vector.y * direction_vector.y);
+
+    // Normalize the direction vector to get a unit vector
+    Point normalized_direction = {direction_vector.x / magnitude,
+                                  direction_vector.y / magnitude};
```
oddbookworm, line 2895: You should do a check that `magnitude` is not zero before doing this

src_c/draw.c
```diff
@@ -2818,6 +2834,299 @@ draw_round_rect(SDL_Surface *surf, int x1, int y1, int x2, int y2, int radius,
...
+    double B2 = line2_start.x - line2_end.x;
+    double C2 = A2 * line2_start.x + B2 * line2_start.y;
+
+    // Check if the lines are parallel (det == 0) and handle the case
+    double det = A1 * B2 - A2 * B1;
+    // If the lines are parallel (det == 0), they do not intersect, return a
+    // default Point
+    if (det == 0) {
+        return (Point){0, 0};
+    }
```
oddbookworm, line 2978: I don't like that "return a default Point". There's no obvious way to know from the caller's perspective that something went wrong and the return value is garbage. There's no outward difference between "parallel, nonintersecting lines" and "lines that only intersect at (0, 0)". Not to mention, parallel lines ***can*** intersect. If they are the same line, they'll intersect at every point of their intersection.

src_c/draw.c
```diff
@@ -2818,6 +2834,299 @@ draw_round_rect(SDL_Surface *surf, int x1, int y1, int x2, int y2, int radius,
...
+             Point line2_end)
+{
+    // Calculate coefficients for each line equation
+    double A1 = line1_end.y - line1_start.y;
+    double B1 = line1_start.x - line1_end.x;
+    double C1 = A1 * line1_start.x + B1 * line1_start.y;
+
+    double A2 = line2_end.y - line2_start.y;
+    double B2 = line2_start.x - line2_end.x;
+    double C2 = A2 * line2_start.x + B2 * line2_start.y;
```
oddbookworm, line 2970: A1, B1, C1, A2, B2, C2 are very nondescriptive. Please be more descriptive with your variable names, or leave a comment explaining what they are if there's no good name for them

src_c/draw.c
```diff
@@ -2818,6 +2834,299 @@ draw_round_rect(SDL_Surface *surf, int x1, int y1, int x2, int y2, int radius,
...
+        double x = (B2 * C1 - B1 * C2) / det;
+        double y = (A1 * C2 - A2 * C1) / det;
+        return (Point){x, y};
+    }
+}
+
+// Function to calculate the angle (in radians) between a center point and
+// another point
+double
+angle(Point center, Point point)
```
oddbookworm, line 2992: A more descriptive name than `angle` is probably possible here. There are many, many lines of code in this codebase, and quite a few people that contribute to it. Simple names that aren't descriptive of the actual purpose generally make things harder for people to work with several years down the road

src_c/draw.c
```diff
@@ -2818,6 +2834,299 @@ draw_round_rect(SDL_Surface *surf, int x1, int y1, int x2, int y2, int radius,
...
+        // Check if the border-radius can be applied to the current angle
+        // it can not be applied to a flat angle(if the three points are
+        // aligned)
+        if (side(points[a], points[i], points[b]) == 0) {
+            PyErr_SetString(
+                PyExc_ValueError,
+                "Border-radius cannot be applied to a flat angle.\nPlease "
+                "ensure that the specified angle or curvature is within a "
+                "valid range for border-radius drawing.\n");
+            return -1;
```
oddbookworm, line 3054: Make sure you free all of your manually allocated memory on every error exit. Otherwise that memory leaks

src_c/draw.c
```diff
@@ -2818,6 +2834,299 @@ draw_round_rect(SDL_Surface *surf, int x1, int y1, int x2, int y2, int radius,
...
+        if ((sqrt(pow(proj1.x - points[i].x, 2) +
+                  pow(proj1.y - points[i].y, 2)) >=
+             sqrt(pow(points[a].x - points[i].x, 2) +
+                  pow(points[a].y - points[i].y, 2)) /
+                 2) ||
+            (sqrt(pow(proj2.x - points[i].x, 2) +
+                  pow(proj2.y - points[i].y, 2)) >=
+             sqrt(pow(points[b].x - points[i].x, 2) +
+                  pow(points[b].y - points[i].y, 2)) /
+                 2)) {
```
oddbookworm, line 3080: Woah. Calm down he-who-must-do-it-all-at-once. This is awful to try to parse. Can this be broken up into more bite-sized chunks that make sense?

src_c/draw.c
```diff
@@ -2818,6 +2834,299 @@ draw_round_rect(SDL_Surface *surf, int x1, int y1, int x2, int y2, int radius,
...
+            PyErr_SetString(PyExc_ValueError,
+                            "Border-radius size must be smaller\n");
+            return -1;
```
oddbookworm, line 3083: Again, free your manually allocated memory

src_c/draw.c
```diff
@@ -2818,6 +2834,299 @@ draw_round_rect(SDL_Surface *surf, int x1, int y1, int x2, int y2, int radius,
...
+             sqrt(pow(points[a].x - points[i].x, 2) +
+                  pow(points[a].y - points[i].y, 2)) /
+                 2) ||
+            (sqrt(pow(proj2.x - points[i].x, 2) +
+                  pow(proj2.y - points[i].y, 2)) >=
+             sqrt(pow(points[b].x - points[i].x, 2) +
+                  pow(points[b].y - points[i].y, 2)) /
+                 2)) {
+            PyErr_SetString(PyExc_ValueError,
+                            "Border-radius size must be smaller\n");
```
oddbookworm, line 3082: You don't need newlines in `PyErr_SetString`

test/draw_test.py
```diff
@@ -3645,12 +3645,20 @@ def setUp(self):
...
-    def test_polygon__args_without_width(self):
+    def test_polygon__args_without_radius(self):
+        """Ensures draw polygon accepts the args without a width."""
+        bounds_rect = self.draw_polygon(
+            pygame.Surface((2, 2)), (0, 0, 0, 50), ((0, 0), (1, 1), (2, 2)), 1
+        )
+
+        self.assertIsInstance(bounds_rect, pygame.Rect)
+
+    def test_polygon__args_without_width_nor_radius(self):
```
oddbookworm, line 3661: This is more of a comment on all of your changes to this file:
I'd leave all of the old tests alone and add a few new ones for your changes specifically. We need to ensure that the old functionality still works as it used to. Unless you're expecting old behavior to change (and that would be a big thing), you should not have to really touch the existing tests

oddbookworm (changes requested):
Still got quite a few things I'd like to see changed. You also still have a bunch of implicit conversion warnings which get treated as errors in some of our CI.
```
    src_c/draw.c(3110): error C2220: the following warning is treated as an error
    src_c/draw.c(3110): warning C4244: 'initializing': conversion from 'double' to 'int', possible loss of data
    src_c/draw.c(3111): warning C4244: 'initializing': conversion from 'double' to 'int', possible loss of data
    src_c/draw.c(3112): warning C4244: 'initializing': conversion from 'double' to 'int', possible loss of data
    src_c/draw.c(3113): warning C4244: 'initializing': conversion from 'double' to 'int', possible loss of data
    src_c/draw.c(3114): warning C4244: 'initializing': conversion from 'double' to 'int', possible loss of data
    src_c/draw.c(3115): warning C4244: 'initializing': conversion from 'double' to 'int', possible loss of data
```

src_c/draw.c
ezzakri-anas: I apologize for any confusion. I provided an explanation in the draw.rst file regarding the behavior of the function when border_radius is negative. I assumed this information would be visible and that it would be enough to resolve the comment. I'll make sure to be clearer in my responses.
Regarding the border_radius parameter, similar to the approach with the width parameter, negative values are not considered, and the function does not apply any radius while still drawing the polygon. The border_radius parameter becomes active only when it is positive.

src_c/draw.c
ezzakri-anas: yes, i made sure to run ` python setup.py format`. I'll try to run it again after doing the  changes and see.

src_c/draw.c
ezzakri-anas: Actually, before reaching `find_parallel_line` in `draw_round_polygon`, we check whether the three consecutive points `points[a]`, `points[i]` and `points[b]` are aligned using` side(points[a], points[i], points[b])`. If they are aligned, the function raises an exception since we cannot draw a rounded vertex when the angle is 180 degrees. Now, if the same point is given twice, the three points will be considered aligned, and the function raises the exception before reaching `find_parallel_line`. Thus, `magnitude` will never be zero.

src_c/draw.c
ezzakri-anas: Lines that are passed to this function will never be parallel as we check for parallel lines using `side(points[a], points[i], points[b])`before reaching this point. But, i undertsand your concern. I'll make `intersection` return `NULL` if two lines are parallel and check for NULL in draw_round_polygon.

src_c/draw.c
ezzakri-anas: The variable names A1, B1, C1, A2, B2, and C2 represent the coefficients of the line equations in the form Ax + By = C.
Specifically, A1 is the coefficient of y, B1 is the coefficient of x, and C1 is the constant term for the first line. Same thing for the second line. I'll add a comment to make it clearer

src_c/draw.c
ezzakri-anas: This is because of the autoformatting. I'll make sure to fix it.

test/draw_test.py
ezzakri-anas: Should i add a class for round corners polygon or should i put tests inside the already existing class. Also, i saw that there is no tests for `draw_arc` to see whether the arcs are drawn correctly. `draw_round_polygon` uses `arc` to draw the rounded corners. Do you have any suggestions to test if the corners are actually rounded and/or with the correct radius?

ezzakri-anas:
regarding the implicit conversion warnings, i made sure to round the `double` values and assign them to an `int` variable so that the variables i pass to the functions are actually of type int but i still get the warning. Any thoughts on how to resolve this?

ezzakri-anas:
> Still got quite a few things I'd like to see changed. You also still have a bunch of implicit conversion warnings which get treated as errors in some of our CI.
> 
> ```
>     src_c/draw.c(3110): error C2220: the following warning is treated as an error
>     src_c/draw.c(3110): warning C4244: 'initializing': conversion from 'double' to 'int', possible loss of data
>     src_c/draw.c(3111): warning C4244: 'initializing': conversion from 'double' to 'int', possible loss of data
>     src_c/draw.c(3112): warning C4244: 'initializing': conversion from 'double' to 'int', possible loss of data
>     src_c/draw.c(3113): warning C4244: 'initializing': conversion from 'double' to 'int', possible loss of data
>     src_c/draw.c(3114): warning C4244: 'initializing': conversion from 'double' to 'int', possible loss of data
>     src_c/draw.c(3115): warning C4244: 'initializing': conversion from 'double' to 'int', possible loss of data
> ```

Regarding the implicit conversion warnings, i made sure to round the `double` values and assign them to an `int` variable so that the variables i pass to the functions are actually of type int but i still get the warning. Any thoughts on how to resolve this?

src_c/draw.c
ezzakri-anas: I retried to run `python setup.py format`. It, indeed, messed with the formatting. It keeps adding these unnecessary newlines everytime i run it.

src_c/draw.c
ezzakri-anas: I changed this part and broke it into smaller chunks with clear explanation.

src_c/draw.c
ezzakri-anas: I renamed it to `computeVectorAngle `as it computes the angle of the vector from `center ` to `point`.

src_c/draw.c
ezzakri-anas: I made sure to free it.

src_c/draw.c
Matiiss: I have seen it auto-format like this before, also the lint check passes, so yeah, it sometimes formats it weirdly. Maybe could raise an issue with clang-format though.

src_c/draw.c
Matiiss: Pretty sure oddbookworm is referring to the `\n` character in the string.

Matiiss:
> I've implemented the test cases for rounded polygons within separate functions inside the `DrawPolygonMixin` class. I'm not sure if this aligns with your preferences or if you would prefer them in a different class. Also,`polygon.draw.arc` function lacks tests to ensure accurate arc rendering. Given this, I'm unsure if testing the rounded polygons for correct rounding is coherent, as the function uses `polygon.draw.arc` for drawing the roundness. Any suggestions or guidance on this matter would be appreciated.

I think it's fine if they are in that class. As for arc tests, that will have to be added in a different PR. You can still check for rounded corners with your polygon though, because it also does a bit of implicit arc testing which just adds extra tests as a safety measure pretty much. Testing is one of those places where redundancy is fine.

ezzakri-anas:
> > I've implemented the test cases for rounded polygons within separate functions inside the `DrawPolygonMixin` class. I'm not sure if this aligns with your preferences or if you would prefer them in a different class. Also,`polygon.draw.arc` function lacks tests to ensure accurate arc rendering. Given this, I'm unsure if testing the rounded polygons for correct rounding is coherent, as the function uses `polygon.draw.arc` for drawing the roundness. Any suggestions or guidance on this matter would be appreciated.
> 
> I think it's fine if they are in that class. As for arc tests, that will have to be added in a different PR. You can still check for rounded corners with your polygon though, because it also does a bit of implicit arc testing which just adds extra tests as a safety measure pretty much. Testing is one of those places where redundancy is fine.

In this context, is it necessary to test for the precise radius of roundness, or would it suffice to confirm that the corners are indeed rounded?

ezzakri-anas:
### I don't get why the circleci failed while all i did is just add some tests for basic shapes.

ankith26:
That is not your PRs fault, don't worry. It was a separate issue that has been fixed now

MyreMylar (changes requested):
It looks like the rounded corners are being calculated in the wrong place:

image

The outside edge of the corner is lining up with the middle of the sides rather than the outside edge of them and that is throwing it off. I'm guessing there is a wrong parameter somewhere. Otherwise things look alright.

I'm going to put this on 'request changes'. If the original author doesn't want to work on it anymore (this has been stuck since December I think) then somebody else feel free to take it over and fix that bug.

test/draw_test.py
```diff
@@ -4264,6 +4283,251 @@ def test_polygon_filled_shape(self):
...
+
+    def test_draw_round_square(self):
+        """Ensure square is drawn with rounded corner"""
+        surfw = surfh = 1000
+        polygon_color = pygame.Color("red")
+        surface_color = pygame.Color("black")
+        surface = pygame.Surface((surfw, surfh))
+        surface.fill(surface_color)
+
+        self.draw_polygon(surface, polygon_color, LARGE_SQUARE, 1, 5)
```
MyreMylar, line 4442: I'm guessing we need some tests like this with a width other than "1"

bilhox:
Hi @ezzakri-anas , do you want to revive this PR ?

Issue #2603, opened 2023-12-14
API proposal: publicizing Window


Starbuck5:
I'd like to turn `_sdl2.video.Window` into a public API in 2.4.0

Firstly, where should it live? I see two contenders.
- `pygame.display.Window`     
- `pygame.window.Window`       (like pygame.surface.Surface, primarily advertised as pygame.Window)

Yunline has successfully convinced me it should live at pygame.window.Window. If you disagree, this is the issue to speak out on.

Steps to making this thing public:
- [x] Make the class public at its final address
- [x] Extricate Window docs from _sdl2.video docs and publicize them in its final namespace
- [x] Keep _sdl2.video.Window / _sdl2.Window as an alias for old code
- [ ] Make sure tests are good
- [x] Make sure nothing will bite us with it in SDL3. (Like fullscreen vs fullscreen desktop is going away in SDL3, so we probably shouldn't expose that...)
- [ ] Window class example that shows off cool new functionality while only using software rendering
- [ ] Can we make private or remove things that are iffy?
    - Sharing a surface with pygame.display can be suspect
    - Every SDL display flag is exposed, but not all of them will work. We don't support OPENGL on these for example. And what would HIGHDPI actually do? (also will change in SDL3?)
- [ ] Polish up documentation

MyreMylar:
I agree with `pygame.window.Window`. Keep window in it's own submodule as it is a new and slightly less confusing paradigm that will likely become the dominant way to create windows for desktop pygame usages going forward.

We could probably make good use of 'undocumented' code for dubious flag features that are currently in SDL2 but may or may not be in SDL3, but if it is straight up not working at all yet then let's pop up a `NotImplementedError` or something similar.

There is already a Window Example, but I agree it should be revamped once the code has stabilised to it's final location.

> Like fullscreen vs fullscreen desktop is going away in SDL3

In what sense? I see this in the SDL3 docs: [https://wiki.libsdl.org/SDL3/SDL_SetWindowFullscreenMode](https://wiki.libsdl.org/SDL3/SDL_SetWindowFullscreenMode) which implies you can still set a 'native fullscreen' or a 'desktop fullscreen' mode for a window.

I'm not sure there is a good reason for a programmer to interop between windows created by `display.set_mode()` and windows created by `window.Window()`. I'd be in favour of warning or erroring if we detect this happening and telling people to stick to one lane or the other, especially if it will reduce bugs.

Starbuck5:
[...]

PR #2617, opened 2023-12-20
Remove Window API that will be questionable in SDL3

Starbuck5:
For #2603, see https://github.com/pygame-community/pygame-ce/issues/2603#issuecomment-1862312329 for details


yunline (approved):
LGTM👍

src_c/window.c
```diff
@@ -712,16 +712,6 @@ window_get_opacity(pgWindowObject *self, void *v)
...
 
-static PyObject *
-window_get_display_index(pgWindowObject *self, PyObject *_null)
-{
-    int index = SDL_GetWindowDisplayIndex(self->_win);
-    if (index < 0) {
-        return RAISE(pgExc_SDLError, SDL_GetError());
-    }
-    return PyLong_FromLong(index);
-}
```
ankith26, line 723: I'm not clear on why this is being removed, from my understanding SDL3 will continue to provide this info?

Starbuck5: It won’t. A display index and a display id are not the same.

ankith26: Oh, interesting. Thoughts on renaming this to `display_id`, and exporting display id on SDL3 and display index on SDL2? Will this cause confusion or create problems?

Starbuck5: Display id is not numerical. It’s a type. You can use it to look up things about displays.

It’s competently different from an index.

ankith26: Okay, I agree we can just remove it for now and later decide if we want to re-add it in some form

ankith26 (approved):
LGTM, lets get this in! :100:

PR #2618, opened 2023-12-20
Update always_on_top doc (it's not x11 only)

Starbuck5:
The SDL2 docs say it is, but the SDL2 source has support for Mac/Windows/Linux. I tested it on Windows and it worked fine.

Relevant issue: #2603

### DIFF
--- a/docs/reST/ref/window.rst
+++ b/docs/reST/ref/window.rst
@@ -42,8 +42,8 @@
    :param bool allow_highdpi: Create a window in high-DPI mode if supported.
    :param bool mouse_capture: Create a window that has the mouse captured
                               (unrelated to INPUT_GRABBED).
-   :param bool always_on_top: Create a window that is always on top
-                              (X11 only).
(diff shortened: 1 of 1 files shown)


yunline (approved):
LGTM👍

ankith26 (commented):
I think the reasoning behind marking it `(X11 only)` is that it does not work on something like direct wayland (and I just tested it to confirm)

Starbuck5:
> I think the reasoning behind marking it (X11 only) is that it does not work on something like direct wayland (and I just tested it to confirm)

That is not the reasoning. The reasoning is that SDL says it is X11 only. The SDL docs are wrong. I've submitted a PR to their docs repo, but I expect it will be ignored (my last PR from 2021 has seen no action).

Starbuck5:
SDL merged my docs change! So this is updated on their end now as well.

ankith26 (approved):
LGTM :tada:

PR #2620, opened 2023-12-20
Implement `pygame.image.load_sized_svg`

### LINKED ISSUE #2600: API proposal for scaled (svg) image loading
**Description**

I propose the addition of a new kwarg `scale` (default: `None`) to `pygame.image.load` that takes a size, and makes sure the output surface is in that size (if it is not None).

The idea is that if the source file is something like an SVG, we use the new SDL_image 2.6.0+ functionality that rasterizes the SVG at the given size, so the output looks clean even at arbitrarily large resolutions. Additionally, if the source is not an SVG, I'd like something like `pygame.transform.smoothscale` applied behind the scenes in the image loading function, so this scale argument is generic and not constrained by the format of the source file passed.

I'm gonna work on this proposal if it is accepted, I'm also open to discussing alternatives in this issue

ankith26:
closes #2600

### DIFF
--- a/docs/reST/ref/image.rst
+++ b/docs/reST/ref/image.rst
@@ -37,7 +37,7 @@ formats.
 
    * ``QOI``
 
-   * ``SVG``
+   * ``SVG`` (limited support, using Nano SVG)
 
    * ``TGA`` (uncompressed)
 
--- a/docs/reST/ref/image.rst
+++ b/docs/reST/ref/image.rst
@@ -127,6 +127,10 @@ following formats.
    aspect ratio, so the returned surface could be smaller along at most one
    dimension.
 
+   This function requires SDL_image 2.6.0 or above. If pygame was compiled with
+   an older version, ``pygame.error`` will be raised when this function is
+   called.
+
    .. versionadded:: 2.4.0
 
    .. ## pygame.image.load_sized_svg ##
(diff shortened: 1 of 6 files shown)


docs/reST/ref/image.rst
```diff
@@ -37,7 +37,7 @@ formats.
 
    * ``QOI``
 
-   * ``SVG`` (limited support, using Nano SVG)
+   * ``SVG``
```
Starbuck5, line 40: It’s limited support because it can’t rasterize text, for example.

ankith26: Oh, I see, will re-add that info then

docs/reST/ref/image.rst
```diff
@@ -109,6 +109,28 @@ following formats.
...
+   argument. The ``file`` argument can be either a filename, a Python file-like
+   object, or a pathlib.Path.
+
+   The usage of this function for handling SVGs is recommended, as calling the
+   regular ``load`` function and then scaling the returned surface would not
+   preserve the quality that an SVG can provide.
+
+   It is to be noted that this function does not return a surface whose
+   dimensions exactly match the ``size`` argument. This function preserves
+   aspect ratio, so the returned surface could be smaller along atmost one
```
oddbookworm, line 127: Typo? `atmost` -> `at most`

oddbookworm (approved):
Just one small nitpick that won’t prevent me from approving

Starbuck5 (approved):
I haven't tested this out locally, but everything seems straightforward enough.

Should the docs say "if not on SDL_image 2.6.0 or above, a pygame.error will be raised" ?

PR #2624, opened 2023-12-21
Deprecate from_display_module

Starbuck5:
For #2603 

Making this work in an actually fool proof way would be a nightmare. Now that windows can have their own display surface management, there is no need for this function.

Why not just remove it? It's too widely used. I dislike this fact, but this "experimental api" (for this function) has enough users that I think it's worth giving it an actual deprecation and not just removing it.


ankith26 (approved):

zoldalma999 (approved):
Agreed on not just removing it, all current apps that use Window and surfaces together use this. But keeping this up must also be a nightmare, so it should go.

Issue #2625, opened 2023-12-23
Window should be subclassable


Starbuck5:
Currently it is not, and it is marked as such with types, which is good.

```
>>> class ManagedWindow(pygame.Window):
...     pass
...
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
TypeError: type 'pygame.window.Window' is not an acceptable base type
```

It's hard to sell this as an OOP interface into windowing functions when it can't be used to define your own objects :)

PR #2626, opened 2023-12-23
Add window example

Starbuck5:
Task for #2603

### DIFF
--- a/docs/reST/ref/window.rst
+++ b/docs/reST/ref/window.rst
@@ -18,6 +18,9 @@
    window class will continue to be developed, and we're excited to share
    the new functionality this class offers.
 
+   Example code showing the Window API in action: 
+   https://github.com/pygame-community/pygame-ce/blob/main/examples/window.py
+
    :param str title: The title of the window.
    :param (int, int) size: The size of the window, in screen coordinates.
    :param (int, int) or int position: A tuple specifying the window position, or
--- a/examples/window.py
+++ b/examples/window.py
@@ -1,0 +1,146 @@
+#!/usr/bin/env python
+""" pygame.examples.window
+
+Demonstrates the new Window API, which can be used in place of
+pygame.display.set_mode, providing more control and an object oriented
+interface.
+"""
+import pygame
+
+WIN_MOVE_SPEED = 25
+WIN_GROW_SPEED = 15
+COLOR_PROGRESSION = ["cadetblue2", "darkorange2", "lightslateblue", "seagreen"]
+UNHOVERED_OPACITY = 0.8
+SHOW_WINDOW = pygame.event.custom_type()
+
+pygame.init()
+pygame.key.set_repeat(500, 100)
+
+main_window = pygame.Window("demo window", (500, 500), resizable=True)
+main_surface = main_window.get_surface()
+windows_and_surfaces = [(main_window, main_surface)]
+
+instructions = """Welcome to the window demo!
+Controls:
+m.) maximize main window
+n.) minimize main window
+r.) restore main window 
+    (from being minimized or maximized)
+h.) hide main window for 1 second
+s.) spawn a new window
+arrow keys.) move window(s) around screen
+escape.) destroy the most recently created window
+1.) shrink all windows
+2.) grow all windows
+"""
+
+font = pygame.font.SysFont("Arial", 24)
+rendered_instructions = font.render(instructions, True, "black")
+
+# Make sure instructions can be shown
+main_window.minimum_size = rendered_instructions.get_size()
+
+clock = pygame.Clock()
+running = True
+
+while running:
+    for event in pygame.event.get():
+        # If there are multiple windows, QUIT fires when the last one is destroyed
+        if event.type == pygame.QUIT:
+            running = False
+
+        if event.type == pygame.WINDOWCLOSE:
+            index = [win_surface[0] for win_surface in windows_and_surfaces].index(
+                event.window
+            )
+            del windows_and_surfaces[index]
+            event.window.destroy()
+
+        if event.type == pygame.WINDOWENTER:
+            if event.window != main_window:
+                try:
+                    event.window.opacity = 1.0
+                except pygame.error:
+                    pass
+
+        if event.type == pygame.WINDOWLEAVE:
+            # Test not None because WINDOWLEAVE will trigger when a window
+            # closes, can't set opacity of now-nonexistent Window.
+            if event.window != main_window and event.window is not None:
+                try:
+                    event.window.opacity = UNHOVERED_OPACITY
+                except pygame.error:
+                    pass
+
+        if event.type == SHOW_WINDOW:
+            main_window.show()
+
+        if event.type == pygame.KEYDOWN:
+            if event.key == pygame.K_ESCAPE:
+                win, _ = windows_and_surfaces.pop()
+                win.destroy()
+
+                # Destroying the windows will not automatically do a QUIT on
+                # last window, unlike closing the windows manually.
+                if len(windows_and_surfaces) == 0:
+                    pygame.event.post(pygame.Event(pygame.QUIT))
+
+            if event.key == pygame.K_m:
+                main_window.maximize()
+
+            if event.key == pygame.K_n:
+                main_window.minimize()
+
+            if event.key == pygame.K_r:
+                main_window.restore()
+
+            if event.key == pygame.K_h:
+                main_window.hide()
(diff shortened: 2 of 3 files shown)


examples/window.py
```diff
@@ -0,0 +1,146 @@
...
+WIN_MOVE_SPEED = 25
+WIN_GROW_SPEED = 15
+COLOR_PROGRESSION = ["cadetblue2", "darkorange2", "lightslateblue", "seagreen"]
+UNHOVERED_OPACITY = 0.8
+SHOW_WINDOW = pygame.event.custom_type()
+
+pygame.init()
+pygame.key.set_repeat(500, 100)
+
+main_window = pygame.Window("demo window", (500, 500), resizable=True)
```
ankith26, line 19: could the main window also be `always_on_top`?

Another trivial note, how about setting an icon with `set_icon`?

examples/window.py
```diff
@@ -0,0 +1,146 @@
...
+                for win, _ in windows_and_surfaces:
+                    win.position += pygame.Vector2(WIN_MOVE_SPEED, 0)
+
+            if event.key == pygame.K_1:
+                for win, _ in windows_and_surfaces:
+                    win.size -= pygame.Vector2(WIN_GROW_SPEED, WIN_GROW_SPEED)
+
+            if event.key == pygame.K_2:
+                for win, _ in windows_and_surfaces:
+                    win.size += pygame.Vector2(WIN_GROW_SPEED, WIN_GROW_SPEED)
```
ankith26, line 131: might also be a good idea to cap it to some upper bound? Perhaps using the `maximum_size` property?

examples/window.py
```diff
@@ -0,0 +1,146 @@
...
+            )
+            del windows_and_surfaces[index]
+            event.window.destroy()
+
+        if event.type == pygame.WINDOWENTER:
+            if event.window != main_window:
+                try:
+                    event.window.opacity = 1.0
+                except pygame.error:
+                    pass
```
ankith26, line 64: How about adding a `event.window.focus()` here?

examples/window.py
```diff
@@ -0,0 +1,146 @@
...
+pygame.key.set_repeat(500, 100)
+
+main_window = pygame.Window("demo window", (500, 500), resizable=True)
+main_surface = main_window.get_surface()
+windows_and_surfaces = [(main_window, main_surface)]
+
+instructions = """Welcome to the window demo!
+Controls:
+m.) maximize main window
+n.) minimize main window
```
ankith26, line 26: In addition to maximise/minimise, could also two keys for add/remove full screen?

examples/window.py
```diff
@@ -0,0 +1,146 @@
...
+                main_window.hide()
+                pygame.time.set_timer(SHOW_WINDOW, 1000, 1)
+
+            if event.key == pygame.K_s:
+                win = pygame.Window("spawned window", (300, 300))
+                try:
+                    win.opacity = UNHOVERED_OPACITY
+                except pygame.error:
+                    pass
+                windows_and_surfaces.append((win, win.get_surface()))
```
ankith26, line 107: ## API comment
This has gotten me thinking about the API. As a user, I think it would be much neater if we could do something like `win.surface` and not have to explicitly track the output of `win.get_surface`. Even helps conceptually to show that "this surface instance is a special window-surface tied to the window"

ankith26 (changes requested):
Nice work on the example! :sunglasses: 
Seeing this example in action has gotten me more interested in `Window` :smile: as it demonstrates new cool stuff and also serves as a good way to test things in action

I added a couple of reviews, some reviews for the example, and also a few API reviews that I noticed while playing around with the example.

examples/window.py
```diff
@@ -0,0 +1,146 @@
...
+                for win, _ in windows_and_surfaces:
+                    win.position += pygame.Vector2(0, WIN_MOVE_SPEED)
+
+            if event.key == pygame.K_LEFT:
+                for win, _ in windows_and_surfaces:
+                    win.position += pygame.Vector2(-WIN_MOVE_SPEED, 0)
+
+            if event.key == pygame.K_RIGHT:
+                for win, _ in windows_and_surfaces:
+                    win.position += pygame.Vector2(WIN_MOVE_SPEED, 0)
```
ankith26, line 123: ## API comment

SMH my head, `window.position` *is also broken under direct wayland*

Nothing moves when I press the arrow keys. :cry: 

All these "direct wayland" issues are fine/ignorable for now because these issues don't happen under XWayland which SDL2 picks by default, but this could be a problem in the SDL3 era if SDL3 devs decide to do wayland-default.

ankith26 (commented):
This example also reveals memory leaks :sweat: 

Open task manager while this example is running and play around with this example. Every time a new window is spawned the memory usage goes up as expected, but closing the window does not reduce the memory consumption

Similarly everytime the window is resized memory usage increases, even when downsizing.

These memory leaks are not insignificant, I believe entire window surfaces are being leaked, so depending on the window size and display resolutions we may be leaking a lot.

examples/window.py
robertpfeiffer: This is by design. Wayland does not want applications to know this for API design and security reasons. There is nothing SDL3 can do about it short of sending patches to KDE and GNOME and sway

examples/window.py
Starbuck5: Well I think I could be calling get_surface() every time I need a surface. I'm just not sure how that works SDL-side.

examples/window.py
```diff
@@ -0,0 +1,146 @@
...
+
+            if event.key == pygame.K_1:
+                for win, _ in windows_and_surfaces:
+                    win.size -= pygame.Vector2(WIN_GROW_SPEED, WIN_GROW_SPEED)
```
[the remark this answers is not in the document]
Starbuck5: It's unclear reading your comment whether you see the `minimum_size` of the window is explicitly set to contain the instruction text.

It would happen to the main window if it was ever told to do a negative size there. Which it can't because the minimum size dimensions are higher than WIN_GROW_SPEED.

I agree it shouldn't be possible to crash the example by messing around with it, but I'm not sure what the best solution would be.

examples/window.py
Starbuck5: Do I need to add try/except pygame.error blocks around them then? I did it for opacity given your previous report.

I'm also unsure if we really want to raise an exception if the operation is not supported. We could return True/False about success?

examples/window.py
Starbuck5: This is also why I thought of subclassing window, issue raised as #2625. In this demo it would be nice to store the color of the window with the window. (Store arbitrary things in instance of class, subclass)

examples/window.py
Starbuck5: At least on Windows, the spawned windows spawn in the center of the main window (I'm not controlling the spawn point). So if the main window was always_on_top it would seem like the spawn window command does nothing, because the other windows would not be visible unless you move the main window with the mouse.

I was looking around the examples data folder for art to use for this, (icons, backgrounds), and none of it spoke to me. I don't feel the need to set an icon in this example.

docs/reST/ref/window.rst
```diff
@@ -18,6 +18,9 @@
    window class will continue to be developed, and we're excited to share
    the new functionality this class offers.
 
+   Example code showing the Window API in action: 
+   https://github.com/pygame-community/pygame-ce/blob/main/examples/window.py
```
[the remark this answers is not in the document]
Starbuck5: I'll update the import. I'll leave the example in the docs, I think we should have more "snippet" style examples in our documentation in general.

examples/window.py
ankith26: In this case the position is silently ignored by SDL, like `always_on_top`. Only opacity explicitly errors.

I think for this example, we should not have this position-moving thing as a highlight, and make it more low-key somehow?

examples/window.py
ankith26: So, SDL stores the surface in the window struct (both in SDL2 and SDL3), and if the surface needs updating it updates the surface, so for most of the calls to this function it is just a cheap struct member access.

On the python side, we could basically the change this `get_surface` to a property.

ankith26: Here is a commit on a branch I made on top of yours, implementing my idea: https://github.com/pygame-community/pygame-ce/commit/6b620f18b3a37eab5a0147508aa5b2da00ea384b

I also updated this example to see/test the property in action

Starbuck5:
This PR is too old, I'm going to close.

PR #2627, opened 2023-12-25
Push window out of 2.4.0

Starbuck5:
It seems like too much of a push to get in 2.4.0 while keeping 2.4.0 this year. Let's push it.

I'm happy that we've made significant progress on it in the meantime.


ankith26 (approved):
LGTM :+1:

ankith26 (commented):
Force merging because this is a docs change that shouldn't break anything, and we should not miss our release schedule.

PR #2631, opened 2023-12-25
Start 2.5.0

Starbuck5:
Since this is before the release of 2.4.0 final, regression and bug fixes will need to be backported to the 2.4.x branch, and the release can happen from there.

This PR brings main up to 2.5.0 so development can continue without a merge freeze in the ~week predicted until 2.4.0 release (maybe longer if something comes up)


ankith26 (approved):
LGTM

Issue #2605, opened 2023-12-15
Maximizing a window invalidates the window surface


Starbuck5:
OS: Windows 10, Python: 3.9

```py
import pygame
from pygame._sdl2 import Window

window = Window(resizable=True)
screen = window.get_surface()

clock = pygame.Clock()

while True:
    for event in pygame.event.get():
        if event.type == pygame.QUIT:
            pygame.quit()
            raise SystemExit
        
        if event.type == pygame.KEYDOWN and event.key == pygame.K_1:
            window.maximize()
        
    screen.fill("white")
    window.flip()
```

When I press 1, get `pygame.error: Window surface is invalid, please call SDL_GetWindowSurface() to get a new surface`

Starbuck5:
Relevant to #2603

yunline:
The bug can be fixed if you add a `pygame.init()`, because the init hook of window module is called when display module init, then the init hook sets the event watch.

Starbuck5:
Ah, I didn't notice it was just about the resizability!

Using pygame.display.set_mode with pygame.RESIZABLE this problem never surfaces, so it probably shouldn't here?

yunline:
> Using pygame.display.set_mode with pygame.RESIZABLE this problem never surfaces

The resize event watch of display module is set in `set_mode()`, so it's not influenced by this bug.

Starbuck5:
Sure. I’m saying that since this error can’t occur with pygame.display, it shouldn’t be able to occur with pygame.window.

ankith26:
I see two ways to resolve this
- `Window` constructor calls `pygame.display.init()`
- `Window` constructor requires `pygame.display.init()` to be called already
I prefer the second way. `pygame.display.set_mode` actually goes the first way however

Starbuck5:
The way I see it, the modern way is to make initializations more automatic. And it's precedented in display.set_mode.

But either option would be better than the current setup.

PR #2633, opened 2023-12-25
Improved `Window.flip` error message when no surface is associated with the `Window`

Matiiss:
Previously the error message would have been `"Window surface is invalid, please call SDL_GetWindowSurface() to get a new surface"`

This also adds a video init check, which is related to this issue: https://github.com/pygame-community/pygame-ce/issues/2605
But I don't know if it should stay here, because if, for instance, we implicitly init video on Window creation, this check shouldn't be needed here?

### DIFF
--- a/src_c/window.c
+++ b/src_c/window.c
@@ -183,17 +183,16 @@ window_flip(pgWindowObject *self)
 {
     int result;
 
-    VIDEO_INIT_CHECK();
+    if (!self->surf) {
+        return RAISE(pgExc_SDLError,
+                     "the Window has no surface associated with it, did "
+                     "you forget to call Window.get_surface()");
+    }
 
     Py_BEGIN_ALLOW_THREADS;
     result = SDL_UpdateWindowSurface(self->_win);
     Py_END_ALLOW_THREADS;
     if (result) {
-        if (!self->surf) {
-            return RAISE(pgExc_SDLError,
-                         "the Window has no surface associated with it, did "
-                         "you forget to call Window.get_surface()");
-        }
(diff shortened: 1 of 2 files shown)


src_c/window.c
```diff
@@ -183,11 +183,19 @@ window_flip(pgWindowObject *self)
 {
     int result;
 
+    VIDEO_INIT_CHECK();
```
Starbuck5, line 186: This is not the time for a video init check. It will let you create a window, put the window onto the screen, then when you try to start software rendering it will yell at you for not being initialized? Not a great experience.

src_c/window.c
```diff
@@ -183,11 +183,19 @@ window_flip(pgWindowObject *self)
...
     Py_BEGIN_ALLOW_THREADS;
     result = SDL_UpdateWindowSurface(self->_win);
     Py_END_ALLOW_THREADS;
-    if (result)
+    if (result) {
+        if (!self->surf) {
+            return RAISE(pgExc_SDLError,
+                         "the Window has no surface associated with it, did "
+                         "you forget to call Window.get_surface()");
+        }
```
Starbuck5, line 196: Why not raise this before the UpdateWindowSurface?

Performance? it's a single C if statement in a function that gets called only once per frame.

Matiiss: Fair enough, it would reduce the indentation level

ankith26 (approved):

(yes this is going to conflict with my `window.surface` PR but I can handle that later if/when the PR gets accepted, so let's get this in first)

Starbuck5 (approved):
Thanks! Looks good to me.

PR #2635, opened 2023-12-26
Remove array struct support

### LINKED ISSUE #405: pypy, array interface and ctypes.pythonapi (425)
#  **Issue &numero;425 opened by *[illume](https://github.com/illume)* at 2018-03-19 14:18:32**

pythonapi isn't on pypy, and they don't want to implement it.

- [x] find what uses it (see below).
- [x] add skips for tests that use pythonapi. We don't have to skip all of it.
- [x] ask pypy devs about alternatives.
- [ ] use use memoryview for arritter.ArrayInterface with ctypes.

# #  What is pythonapi used for?

Array testing. pygame uses the python array interface, and has a ctypes based testing module for checking this stuff.

[pygame/test/test_utils/arrinter.py](https://github.com/pygame/pygame/blob/master/test/test_utils/arrinter.py)

These are the functions from pythonapi it uses, with links to the CPython API docs.
- pythonapi.[PyCObject_AsVoidPtr](https://docs.python.org/2/c-api/cobject.html# c.PyCObject_AsVoidPtr) (arrinter) (py2)
- pythonapi.[PyCObject_GetDesc](https://docs.python.org/2/c-api/cobject.html?highlight=pycobject_getdesc# c.PyCObject_GetDesc) (arrinter)
- pythonapi.[PyCapsule_IsValid](https://docs.python.org/2/c-api/capsule.html# c.PyCapsule_IsValid) (optional) (arrinter)
- pythonapi.[PyCapsule_GetPointer](https://docs.python.org/3.6/c-api/capsule.html?highlight=pycapsule_getpointer# c.PyCapsule_GetPointer) (arrinter)
- pythonapi.[PyCapsule_GetContext](https://docs.python.org/3.6/c-api/capsule.html?highlight=pycapsule_getcontext# c.PyCapsule_GetContext) (arrinter)
- pythonapi.[PyCapsule_New](https://docs.python.org/3.6/c-api/capsule.html?highlight=pycapsule_new# c.PyCapsule_New) (py3) (arrinter)
- pythonapi.[PyCObject_FromVoidPtr](https://docs.python.org/2.7/c-api/cobject.html?highlight=pycobject_fromvoidptr# c.PyCObject_FromVoidPtr) (py2) (arrinter)
- pythonapi.[PyBytes_FromString](https://docs.python.org/3/c-api/bytes.html# c.PyBytes_FromString) (mixer_test.py) 

Tests which use pythonapi (and skip such tests on pypy):
- [x] base_test.py
- [x] color_test.py
- [x] freetype_test.py
- [x] mixer_test.py
- [x] pixelarray_test.py
- [x] pixelcopy_test.py
- [x] surface_test.py

<hr>

#  Comments

# # #  **[illume](https://github.com/illume)* commented at 2018-03-19 17:16:54*

This could be used to get strides and such:
```
memoryview(array)
```
<hr>

# # #  **[illume](https://github.com/illume)* commented at 2022-01-22 20:01:51*

The pypy issue tracker is here now: https://foss.heptapod.net/pypy/pypy/-/issues

Haven’t looked into if we need this stuff still. At least the python stuff could be removed. Maybe we should rip out the old buffer stuff first and then reassess.
<hr>

# # #  **[illume](https://github.com/illume)* commented at 2022-01-22 20:02:52*

Here is the old buffer removal issue https://github.com/pygame/pygame/issues/2949

### LINKED ISSUE #1497: Remove support for NumPy __array_struct__ (2949)
#  **Issue &numero;2949 opened by *[mattip](https://github.com/mattip)* at 2021-12-24 08:09:26**

**Description**

NumPy's `__array_struct__` is a legacy interface which pre-dates the buffer interface (memoryview). I think pygame is one of the only projects to support it (at least thats what [searching the top 4000 PyPI packages](https://github.com/hpyproject/top4000-pypi-packages/search?q=__array_struct__) turns up. What do you think about deprecating it?<hr>

#  Comments

# # #  **[Starbuck5](https://github.com/Starbuck5)* commented at 2021-12-25 09:14:13*

Hello, thanks for warning us of a possible upcoming deprecation. My understanding after researching this a tiny bit is that `__array_struct__` has been mostly replaced with the Python 3+ buffer protocol, so it seems like it makes sense to deprecate.

For future reference, `__array_struct__` is present in these files:
```
src_c/base.c
src_c/bufferproxy.c
src_c/color.c
src_c/mixer.c
src_c/pixelarray.c
test/base_test.py
test/bufferproxy_test.py
test/test_utils/arrinter.py
```

Snooping around the C code it looks pretty easy to remove. It seems that we support the protocol but don't use it internally. And it looks like we're good about supporting the new buffer protocol too.
<hr>

# # #  **[illume](https://github.com/illume)* commented at 2021-12-26 00:15:50*

Starbuck5 covered most of it. The other parts are that we now only support python 3.6 and up, and what is left of the old buffer code is planned to be removed. So I don’t see any problem removing it in numpy from our side. Probably we should remove it first to make sure that’s true :)

Maybe this will help with pygame on pypy too… since the buffer tests are one of the remaining buggy bits that are skipped on pypy. I don’t know if pypy was ever using the old buffer style or not though.

<hr>

# # #  **[PurityLake](https://github.com/PurityLake)* commented at 2022-07-28 19:38:09*

So are we looking to remove it and all its uses, so in `src_c/base.c` there is a function called `pgGetArrayStruct` that uses `__array_struct__`. This function is used a few times in that file alone, should I get rid of it altogether or compromise with returning an error value or a "deprecation warning"?

zoldalma999:
Fixes #1497, fixes #405


src_c/_pygame.h
```diff
@@ -485,7 +485,7 @@ typedef enum {
 #define PYGAMEAPI_PIXELARRAY_NUMSLOTS 2
 #define PYGAMEAPI_COLOR_NUMSLOTS 5
 #define PYGAMEAPI_MATH_NUMSLOTS 2
-#define PYGAMEAPI_BASE_NUMSLOTS 29
+#define PYGAMEAPI_BASE_NUMSLOTS 28
```
robertpfeiffer, line 488: This seems like a bad breaking change?

src_c/base.c
```diff
@@ -2281,7 +1753,7 @@ MODINIT_DEFINE(base)
     c_api[11] = pg_mod_autoquit;
     c_api[12] = pg_RGBAFromObj;
     c_api[13] = pgBuffer_AsArrayInterface;
-    c_api[14] = pgBuffer_AsArrayStruct;
+    c_api[14] = pg_SetDefaultConvertFormat;
```
robertpfeiffer, line 1756: Maybe I don't understand the C API, but this looks wrong?

buildconfig/stubs/pygame/bufferproxy.pyi
```diff
@@ -7,7 +7,6 @@ class BufferProxy:
     # possibly going to be deprecated/removed soon, in which case these
     # typestubs must be removed too
```
ankith26, line 8: this comment is outdated now?

buildconfig/stubs/pygame/mixer.pyi
```diff
@@ -61,7 +61,6 @@ class Sound:
     # possibly going to be deprecated/removed soon, in which case these
     # typestubs must be removed too
```
ankith26, line 62: same as above

src_c/bufferproxy.c
```diff
@@ -23,8 +23,8 @@
   data through the Python buffer protocol or the array interface.
   The new buffer protocol is available for Python 3.x. For Python 2.x
   only the old protocol is implemented (for PyPy compatibility).
-  Both the C level array structure - __array_struct__ - interface and
-  Python level - __array_interface__ - are exposed.
+  Python level - __array_interface__ - are exposed. (TODO(zoldalma): Reprashe
+  this to make sense and also be factually correct)
```
ankith26, line 27: Is this a TODO to self that you intended to resolve in this PR and forgot, or is this intentional?

zoldalma999: It was meant for me, and I just forgot it. Good catch, thanks for pointing it out!

src_c/base.c
zoldalma999: The reason this is done this way is so that I don't have to renumber all the slots. I also changed the `include/_pygame.h` file, so it still works. And as far as I know, we are more lenient towards C API changes than python API ones and only code that use `pgBuffer_AsArrayStruct` would break (which should be very few, as most if not all code should have migrated to the buffer protocol) so I removed it.

test/test_utils/arrinter.py
```diff
@@ -1,438 +0,0 @@
...
-
-PAI_CONTIGUOUS = 0x01
-PAI_FORTRAN = 0x02
-PAI_ALIGNED = 0x100
-PAI_NOTSWAPPED = 0x200
-PAI_WRITEABLE = 0x400
-PAI_ARR_HAS_DESCR = 0x800
-
-
-class ArrayInterface:
```
ankith26, line 103: This PR is removing a lot of tests, some removals are justified, but I wonder if this class can be, say, rewritten to use the array interface, or the new buffer interface to test that?

MyreMylar: I think there is already a class like this for the buffer interface called `buftools.py`

test/base_test.py
```diff
@@ -378,105 +330,6 @@ def test_PgDict_AsBuffer_PyBUF_flags(self):
...
-        self.assertEqual(b.format, frev + "h")
-        self.assertEqual(b.len, e.len)
-        self.assertEqual(b.itemsize, e.itemsize)
-        self.assertEqual(b.shape, e.shape)
-        self.assertEqual(b.strides, e.strides)
-        self.assertTrue(b.suboffsets is None)
-        self.assertTrue(b.readonly)
-        self.assertEqual(b.buf, e.data)
-        self.assertRaises(BufferError, Importer, a, buftools.PyBUF_FULL)
-
```
MyreMylar, line 479: Why is this test gone? It looks to be related to the buffer interface not the array struct.

MyreMylar (approved):
LGTM overall. 

I don't think anyone will miss any of this array stuff as it looks like it never gets used anymore.

zoldalma999:
I'll close this because this seems like too much to push through in one PR, and there's a lot of conflicts now too. I'll split this up into smaller PRs if/when I have time.

Issue #2458, opened 2023-09-20
Possible divide by zero in camera


Starbuck5:
Reported: https://stackoverflow.com/questions/77107362/float-division-by-zero-python-webcam-pygame-camera

Source: https://github.com/pygame-community/pygame-ce/blob/681f5b5e4e91f813f806c13b1183a7a1e18450d6/src_py/_camera_opencv.py#L74

It makes sense that this could return 0 if unknown, and then crash. The solution is likely to rewrite the OpenCV camera backend to use threading so it can convert the blocking OpenCV API into our non blocking API.

novialriptide:
Stack Overflow page redirects to a 404.
<img width="937" alt="image" src="https://github.com/pygame-community/pygame-ce/assets/35881688/a0de564a-8399-4107-bcb6-927925f41af0">

Starbuck5:
Ok, stack overflow question got removed. The possibility of OpenCV returning 0 as webcam FPS and leading to a divide by zero error is still there.

PR #2637, opened 2023-12-27
Camera: OpenCV Backend Fixed

maghedo243:
Fixes the issue in #2458 and #2050 where the OpenCV backend can return 0 FPS, resulting in an error.

All I've done is added exception handling to let the programmer know that there's an issue with the camera FPS. There's still an error, but the error message is more descriptive than a "ZeroDivisionError: Divide by Zero".

To test, you can use the "OpenCV" backend in any camera code, I used the "examples/camera.py" to test.

### DIFF
--- a/src_py/_camera_opencv.py
+++ b/src_py/_camera_opencv.py
@@ -71,7 +71,11 @@ class Camera:
         self._flipy = False
         self._brightness = 1
 
+        if self._cam.get(cv2.CAP_PROP_FPS) == 0:
+            raise ZeroDivisionError("Camera returning 0 FPS. Device ID:",self._device_index)
+
         self._frametime = 1 / self._cam.get(cv2.CAP_PROP_FPS)
+
         self._last_frame_time = 0
 
         self._open = True


oddbookworm:
You should make sure that you're creating your pull to merge into `main`, not the `2.4.x` branch

maghedo243:
> You should make sure that you're creating your pull to merge into `main`, not the `2.4.x` branch

Fixed!

src_py/_camera_opencv.py
```diff
@@ -71,7 +71,11 @@ def start(self):
         self._flipy = False
         self._brightness = 1
 
+        if self._cam.get(cv2.CAP_PROP_FPS) == 0:
+            raise ZeroDivisionError("Camera returning 0 FPS. Device ID:",self._device_index)
```
oddbookworm, line 75: I don't know that raising an exception here is the right thing to do. If `self._cam.get(cv2.CAP_PROP_FPS)` is returning 0, that doesn't mean a problem occurred. It means that the camera isn't reporting an FPS. That's a hardware limitation of the camera, but something we should be able to handle. Maybe default to some frametime that makes sense or something. Maybe @Starbuck5 will have a different opinion though

Starbuck5:
I think this backend should be rewritten to not require knowing the FPS at all.

src_py/_camera_opencv.py
maghedo243: If it's a hardware issue, I could change the error message. The issue is I think it starts trying to return a camera feed immediately afterwards. We could have some kind of "status" attribute to tell the user that it's having a hardware issue or something. How does that sound?

maghedo243:
> I think this backend should be rewritten to not require knowing the FPS at all.

I mean that works too. I didn't really have any prior guidance on how we wanted it so I could do that

Starbuck5:
> I mean that works too. I didn't really have any prior guidance on how we wanted it so I could do that

In the #2458 description my take was "The solution is likely to rewrite the OpenCV camera backend to use threading so it can convert the blocking OpenCV API into our non blocking API."

maghedo243:
> > I mean that works too. I didn't really have any prior guidance on how we wanted it so I could do that
> 
> In the #2458 description my take was "The solution is likely to rewrite the OpenCV camera backend to use threading so it can convert the blocking OpenCV API into our non blocking API."

I'm struggling a little to see how threading would help the 0 FPS issue... and on top of that if I have to rewrite an entire backend for a system I don't fully understand yet I'm not sure if that's a "good first issue" for me 😅

Starbuck5:
Well, the current API wants to know the FPS so it can guess when the next frame will be. It does this because OpenCV is blocking- if it doesn’t have a frame it will wait until it has one.

If this was threaded it wouldn’t need to guess at when the frames will be generated. The thread will just busy wait on the frames and pass them back to be dispensed by pygame.camera’s non blocking API.

Starbuck5:
@maghedo243 Do you plan on working on this further? No worries if not.

I'm marking this PR a draft for now because I don't believe this is a good way to resolve the issue.

Starbuck5:
Since there's been no action on this and I don't think this actually resolves the issue, I'm going to close this PR.

PR #2638, opened 2023-12-28
Test function for pygame.draw.arc (resolves #2636)

### LINKED ISSUE #2636: Missing Test for pygame.draw.arc in draw_test.py
The class `DrawArcMixin` lacks test coverage to ensure that the function `pygame.draw.arc` draws the arc correctly.

**TODO**:
- [ ] Add a comprehensive test for `pygame.draw.arc`. Initial test placeholder:
```python
def todo_test_arc(self):
    """Ensure draw arc works correctly."""
    # Your test implementation goes here.

```

Houda222:
### This pull request resolves issue #2636
I added a test function `test_arc__correct_drawing` that replaced the `todo_test_arc` function in the `DrawArcMixin` class. It ensures that `pygame.draw.arc` works correctly, by checking whether the drawn arc matches the ellipse defined by the bounding rectangle. The comparison is done by measuring the difference between the ellipse's points and the arc's point and making sure it's less than a certain threshold. 
The test is done on different bounding rectangles.


ankith26:
The circleci fail on this PR is not this PRs fault, don't worry. It's being resolved in another PR

test/draw_test.py
```diff
@@ -6160,9 +6160,55 @@ def test_arc__invalid_color_formats(self):
...
+                        number_of_valid_arc_points += 1
+                    elif surface.get_at((x, y)) == arc_color:
+                        number_of_invalid_arc_points += 1
+
+            surface.fill(surface_color)  # Clear for the next test
+
+            self.assertEqual(
+                number_of_valid_arc_points,
+                number_of_invalid_arc_points + number_of_valid_arc_points,
+            )
```
ankith26, line 6211: this check can more simply be written as `self.assertEqual(number_of_invalid_arc_points, 0)`.

Additionally, we should probably test for `number_of_valid_arc_points` being correct somehow?

test/draw_test.py
```diff
@@ -6160,9 +6160,55 @@ def test_arc__invalid_color_formats(self):
...
+        surfw, surfh = 500, 500
+        surface = pygame.Surface((surfw, surfh))
+        surface_color = pygame.Color("black")
+        surface.fill(surface_color)
+        arc_color = pygame.Color("white")
+        rectangles = [
+            pygame.Rect((200, 200), (300 + i, 200 + i)) for i in range(-100, 50, 30)
+        ]  # Test on different bounding rectangles
+        start = 0
+        stop = 2
```
MyreMylar, line 6174: I feel like we should probably test a broader range of arcs than just 0 to 2 radians, probably stuff like: pi, pi/2, pi/4, 2 * pi.

test/draw_test.py
```diff
@@ -6160,9 +6160,55 @@ def test_arc__invalid_color_formats(self):
...
+
+        for rect in rectangles:
+            kwargs = {
+                "surface": surface,
+                "color": arc_color,
+                "rect": rect,
+                "start_angle": start,
+                "stop_angle": stop,
+                "width": 1,
```
MyreMylar, line 6183: Probably should also test on wider range of widths than just 1.

test/draw_test.py
MyreMylar, line 6211:
```suggestion
            self.assertEqual(number_of_invalid_arc_points, 0)
```

MyreMylar (changes requested):
What is here is nice, but I think this test should test more angles and more widths to fully test out draw arc.

MyreMylar:
Hi @Houda222, do you think you will make any of the changes asked for? Or should we pass this PR over to someone else to finish off?

PR #2639, opened 2023-12-28
Call `PyObject_GetBuffer` directly, fix pypy fail

ankith26:
So, with the latest pypy version, we have new test fails. What's happening now on pypy is that `PyObject_CheckBuffer` passes but `PyObject_GetBuffer` fails, and the old code is forwarding the error in this case, but it's better to instead go down the array interface/struct fallback and then error if all those fail.

I have intentionally kept this PR minimal so that it is easy to review, despite my itch to refactor the whole function to make it neater :)

### DIFF
--- a/src_c/base.c
+++ b/src_c/base.c
@@ -1148,12 +1148,9 @@ pgObject_GetBuffer(PyObject *obj, pg_buffer *pg_view_p, int flags)
     flags |= PyBUF_PYGAME;
 #endif
 
-    if (PyObject_CheckBuffer(obj)) {
+    if (PyObject_GetBuffer(obj, view_p, flags) == 0) {
         char *fchar_p;
 
-        if (PyObject_GetBuffer(obj, view_p, flags)) {
-            return -1;
-        }
         pg_view_p->release_buffer = PyBuffer_Release;
 
         /* Check the format is a numeric type or pad bytes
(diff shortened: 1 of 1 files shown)


ankith26:
If this works and is merged, this will need backporting to the 2.4.x branch before the release, because otherwise circleci is going to fail on it, when it makes the release wheels

Starbuck5:
Is this a bug in pypy-- should this be reported to them?

ankith26:
According to the python docs for [`PyObject_CheckBuffer`](https://docs.python.org/3/c-api/buffer.html#c.PyObject_GetBuffer)
> Return 1 if obj supports the buffer interface otherwise 0. When 1 is returned, it doesn’t guarantee that [`PyObject_GetBuffer()`](https://docs.python.org/3/c-api/buffer.html#c.PyObject_GetBuffer) will succeed. This function always succeeds.

So maybe it could be considered a PyPy regression, but the new behaviour is still inline with what the docs say, so instead doing a fix on our end is better IMO.

Starbuck5 (approved):
Ok I guess? The change is small and looks like it will function the same, except for the different error.

I don't feel confident about how the buffers work in pygame, or what is going on with pypy there. Is failing a different way making the tests pass or does this cause it to not fail at all?

I still feel that this might be a good thing to report to pypy. If it's an intentional change they can say so, but it seems like it could be unintentional.

ankith26:
> Is failing a different way making the tests pass or does this cause it to not fail at all?

The tests pass because now there is no fail at all. 

> I still feel that this might be a good thing to report to pypy. If it's an intentional change they can say so, but it seems like it could be unintentional.

Yeah, maybe. Though I'd still like this patch in because it would work regardless of their decision, and as of now it is not failing on any config we support

MyreMylar (approved):
LGTM 👍

PR #2640, opened 2023-12-28
Remove `HAVE_NEWBUF` checks in tests

ankith26:
Just a minor code cleanup: `HAVE_NEWBUF` is a constant that is always defined to `1` in the source. So there's no need to check for it anymore


MyreMylar (approved):
LGTM 👍 simple change that makes the source tidier.

zoldalma999 (approved):

PR #2641, opened 2023-12-30
Add mp3 file-like object tests

### LINKED ISSUE #2154: Loading MP3 music from file like object fails on Linux
Discovered this while I was working on #2152.

Apparently `test_load_object` in mixer_music_test.py will fail on Linux if it gets an mp3 file like object. (Originally `house_lo.mp3`)
```
  File "/home/runner/.local/lib/python3.8/site-packages/pygame/tests/mixer_music_test.py", line 70, in test_load_object
    pygame.mixer.music.load(musf)
pygame.error: ModPlug_Load failed
```

People have seen this out in the world:
https://stackoverflow.com/questions/65614352/pygame-error-modplug-load-failed-when-trying-to-play-sound-from-bytes

I would like someone to investigate why this fails. It might be expected without a namehint, but then why does it only fail on Linux?

ankith26:
fixes #2154

Apparently, the issue here is that mp3 file-like objects are not handled properly in SDL_mixer < 2.6.0 (which is what both our ubuntu builds are on ATM). So include these tests, but only if SDL_mixer is newer than 2.6.0

### DIFF
--- a/test/mixer_music_test.py
+++ b/test/mixer_music_test.py
@@ -55,6 +55,9 @@ class MixerMusicModuleTest(unittest.TestCase):
     def test_load_object(self):
         """test loading music from file-like objects."""
         filenames = ["house_lo.ogg", "house_lo.wav", "surfonasinewave.xm"]
+        if pygame.mixer.get_sdl_mixer_version() >= (2, 6, 0):
+            filenames.append("house_lo.mp3")
+
         data_fname = example_path("data")
         for file in filenames:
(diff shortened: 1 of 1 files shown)


MyreMylar (approved):
OK, seems odd that these only fail on linux CI, but if they've fixed it in SDL in the future then fine.

Starbuck5 (approved):
Thanks for looking into this! I'm happy to get things like this resolved.

PR #2644, opened 2023-12-31
removed use of depth param in examples

oddbookworm:


examples/aliens.py
```diff
@@ -239,8 +239,7 @@ def main(winstyle=0):
     fullscreen = False
     # Set the display mode
     winstyle = 0  # |FULLSCREEN
-    bestdepth = pygame.display.mode_ok(SCREENRECT.size, winstyle, 32)
-    screen = pygame.display.set_mode(SCREENRECT.size, winstyle, bestdepth)
+    screen = pygame.display.set_mode(SCREENRECT.size, winstyle)
```
ankith26, line 242: can `winstyle` be removed entirely as it's just 0?

examples/arraydemo.py
```diff
@@ -24,7 +24,7 @@
 
 def surfdemo_show(array_img, name):
     "displays a surface, waits for user to continue"
-    screen = pygame.display.set_mode(array_img.shape[:2], 0, 32)
+    screen = pygame.display.set_mode(array_img.shape[:2], 0)
```
ankith26, line 27:
```suggestion
    screen = pygame.display.set_mode(array_img.shape[:2])
```

examples/aliens.py
oddbookworm: I left that one in because it seems like it was meant to be bitwise-OR'ed with FULLSCREEN in some cases (given from the comment beside it) and by removing the `#`, it would just work

ankith26 (approved):

MyreMylar (approved):
LGTM 👍

PR #2647, opened 2024-01-02
Don't skip tests if "optional" modules are absent

### LINKED ISSUE #1569: Freetype test doesn't fail if freetype wasn't built (3099)
# **Issue &numero;3099 opened by *[Starbuck5](https://github.com/Starbuck5)* at 2022-03-30 04:18:13**

You can test this by building pygame, commenting out the `_freetype` line in SETUP, then building and installing again.

Although pygame.freetype isn't required, it is expected, so the tests should fail if it isn't present.

Does this happen with other "optional" modules?<hr>

# Comments

ankith26:
We have many "optional" modules. However these modules are widely used and most users would expect these to be present. So, we should run these tests so that the tests explicitly error when these modules are absent, to ensure that the person running the tests knows the modules are not compiled in (and not left out accidentally)

fixes #1569


Starbuck5 (approved):
Nice! This has bothered me in the past.

MyreMylar (approved):
LGTM 👍 

I agree that these optional modules are very marginally relevant for 99% of users.

PR #2653, opened 2024-01-04
support emsdk >3.1.51

pmp-p:
since https://github.com/pygame-community/pygame-ce/pull/2596 static.c build was silently ignored (commented by setuptools) because PNG/JPEG not set anymore in Setup as emsdk is only an additional_platform_setup to config_unix

patch tested against 3.1.52 with pygbag : pygame-ce 2.5.0.dev1 (SDL 2.24.2, Python 3.11.7)

### DIFF
--- a/buildconfig/Setup.Emscripten.SDL2.in
+++ b/buildconfig/Setup.Emscripten.SDL2.in
@@ -6,7 +6,8 @@
 #MIXER = -lSDL2_mixer
 #SCRAP =
 #FREETYPE = -lfreetype -lharfbuzz
-
+PNG = -lpng
+JPEG = -ljpeg
 DEBUG =
 
 # these can build alone and object files merged with ar
--- a/buildconfig/Setup.Emscripten.SDL2.in
+++ b/buildconfig/Setup.Emscripten.SDL2.in
@@ -6,6 +6,9 @@
 #MIXER = -lSDL2_mixer
 #SCRAP =
 #FREETYPE = -lfreetype -lharfbuzz
+
+# png and jpeg are needed for decoders in SDL_image, when not using emscripten preload api because its async
+# which is the case for both pygbag or wasi (no browser, no emscripten js api emulation)
 PNG = -lpng
 JPEG = -ljpeg
 DEBUG =


Starbuck5:
I don't understand this change.

This points to my PR which removed two commented out lines in the emscripten setup. This PR does not revert that change, it makes a different change by reinserting the lines uncommented.

You mention this is important for config_unix, but my PR also removes config_unix's link to libpng and libjpeg.

We don't depend on libjpeg or libpng now, just on SDL_image, which on turn depends on those libraries.

pmp-p:
emscripten inherits from config_unix, SDL2_image recent port cannot link to browser jpeg/png support ( not synchronous == (more) troubles ). So static linking require -lpng -ljpeg explicitely linked in to hide native support. 

But the real culprit is not your change to unix cfg : it is more setuptools that decided to put a # in front of static.c build command in Setup when either PNG or JPEG keys are undefined

Instead of just letting linking fail which would have been obvious even without testsuite ...

Starbuck5:
Extra info from pmp-p on discord:

> @Starbuck regarding https://github.com/pygame-community/pygame-ce/pull/2653 ,  emsdk config inherits values from default unix SDL2.in, so when you removed JPEG/PNG since i had them commented in the extra config they were just not set anymore but static linking require them because sdl2_image (which is a custom build in pygbag)  expect to find them preloaded 
sdl2_image from emsdk port is  too old, not suitable for pygame-ce ( gray image bug, no svg etc ... )
took me hours to understand that setuptools silently sneaked a # in front of the line in Setup, it allows empty value but does not like undefined key at all

Starbuck5:
Could this be fixed instead by removing PNG and JPEG link into static.c in this line? https://github.com/pygame-community/pygame-ce/blob/main/buildconfig/Setup.Emscripten.SDL2.in#L32. It seems like that's something I missed when taking out the libpng and libjpeg direct dependencies.

I'm curious about what you're doing with a custom SDL_image build in pygbag. The main SDL2_image branch supports emscripten natively from what I see, it shouldn't be out of date.

pmp-p:
SDL2_image from emscripten ports repo is usually older than current release and may not come from original.

i use prerelease from https://github.com/libsdl-org/SDL_image/releases when possible and SDL2_image static links in libjpeg/libpng/libwebp as wasm ( no exploit of outdated browser/native lib eg: webp )

png/jpeg come from emscripten ports.
libwebp wasm prebuilt is provided by pygame-web/python-wasm-sdk

Starbuck5 (approved):
Thanks for explaining what this is needed for, it's a surprisingly complex story (at least to me).

Seems good!

ankith26 (approved):

Issue #2649, opened 2024-01-03
Add invlerp() and remap() functions to pygame.math


bigwhoopgames:
Now that the `lerp` function has been implemented into `pygame.math` I thought it would also be a good idea to add the `invlerp` and `remap` functions.

```py
def inv_lerp(a: float, b: float, v: float) -> float:
    return (v - a) / (b - a)

def remap(i_min: float, i_max: float, o_min: float, o_max: float, v: float) -> float:
    t = inv_lerp(i_min, i_max, v)
    return lerp(o_min, o_max, t)
```
    
`inv_lerp` is opposite of `lerp` and `remap` utilizes both to map a value from an input range to an output range. Very useful if you want to remap things like colors or percents etc.

PR #2654, opened 2024-01-04
``invlerp`` and ``remap`` implementation

bilhox:
Implementation of #2649
Docs will come in a short time.


Matiiss (changes requested):
Thanks for the PR, I left a couple of notes:

I don't think you should change anything about `math.lerp` in this PR, you can submit a separate PR for that (that includes both the implementation and the docs).

Also this needs tests too.

buildconfig/stubs/pygame/math.pyi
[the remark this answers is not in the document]
bilhox: Ah, I thought I added it

buildconfig/stubs/pygame/math.pyi
```diff
@@ -334,6 +334,7 @@ class Vector3(_GenericVector):
 
 
 def lerp(a: float, b: float, weight: float, do_clamp: bool = True, /) -> float: ...
+def invlerp(a: float, b: float, weight: float /) -> float: ...
```
Matiiss, line 337: CI is failing because you forgot a comma here, after `weight: float`

bilhox:
> I don't think you should change anything about math.lerp in this PR, you can submit a separate PR for that (that includes both the implementation and the docs).

It's because I had to use ``lerp`` and ``invlerp`` in ``remap``, So I updated it. Should I point it out btw ?

src_c/math.c
```diff
@@ -4233,6 +4245,82 @@ math_clamp(PyObject *self, PyObject *const *args, Py_ssize_t nargs)
...
+
+    PyObject *min = args[0];
+    PyObject *max = args[1];
+    PyObject *value = args[2];
+
+    if (PyNumber_Check(args[2]) != 1)
+        return RAISE(PyExc_TypeError,
+                     "invlerp requires the interpolation amount to be number");
+
+    double t = PyFloat_AsDouble(value);
```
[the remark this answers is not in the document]
bilhox: It's also present in ``pygame.math.lerp``, I'll fix it in another PR.

buildconfig/stubs/pygame/math.pyi
```diff
@@ -334,6 +334,8 @@ class Vector3(_GenericVector):
 
 
 def lerp(a: float, b: float, weight: float, do_clamp: bool = True, /) -> float: ...
+def invlerp(a: float, b: float, weight: float, /) -> float: ...
+def remap(i_min : float, i_max : float, o_min : float, o_max : float, value : float , /) -> float: ...
```
Matiiss, line 338: Was it formatted with those spaces before `:`s using `python setup.py format`?

src_c/math.c
```diff
@@ -4192,6 +4192,18 @@ vector_elementwise(pgVector *vec, PyObject *_null)
...
+{
+    return a + (b - a) * v;
+}
+
+double
+invlerp(double a, double b, double v)
+{
+    return (v - a) / (b - a);
+}
+
```
Matiiss, line 4206: Could maybe `inline` both of these functions

src_c/math.c
Matiiss, line 1: For consistency all of the error messages should begin with a lowercase letter (well, unless it's reasonable not to, like abbreviations or something) and end without a period

src_c/math.c
```diff
@@ -4233,6 +4245,79 @@ math_clamp(PyObject *self, PyObject *const *args, Py_ssize_t nargs)
...
+        t = 0.0;
+    else if (t > 1)
+        t = 1.0;
+
+    double a = PyFloat_AsDouble(min);
+    double b = PyFloat_AsDouble(max);
+
+    if (PyErr_Occurred())
+        return RAISE(PyExc_ValueError,
+                     "Invalid argument values passed to invlerp.");
```
Matiiss, line 4275: I think it's reasonable to at least mention to the user here that their input value might be too small or too large

src_c/math.c
```diff
@@ -4233,6 +4245,79 @@ math_clamp(PyObject *self, PyObject *const *args, Py_ssize_t nargs)
...
+
+    double v = PyFloat_AsDouble(value);
+    double a = PyFloat_AsDouble(i_min);
+    double b = PyFloat_AsDouble(i_max);
+    double c = PyFloat_AsDouble(o_min);
+    double d = PyFloat_AsDouble(o_max);
+
+    if (PyErr_Occurred())
+        return RAISE(PyExc_ValueError,
+                     "Invalid argument values passed to remap.");
```
Matiiss, line 4311: I think it's reasonable to at least mention to the user here that their input value might be too small or too large

Matiiss (commented):
This also needs unit tests

buildconfig/stubs/pygame/math.pyi
bilhox: I think it's just me, let me fix it

buildconfig/stubs/pygame/math.pyi
```diff
@@ -334,6 +334,8 @@ class Vector3(_GenericVector):
 
 
 def lerp(a: float, b: float, weight: float, do_clamp: bool = True, /) -> float: ...
+def invlerp(a: float, b: float, weight: float, /) -> float: ...
+def remap(i_min: float, i_max: float, o_min: float, o_max: float, value: float , /) -> float: ...
```
Matiiss, line 338: There's space before that last comma

bilhox: fixed

test/math_test.py
```diff
@@ -86,6 +86,107 @@ def test_lerp(self):
...
+        a = -10.0
+        b = 10.0
+        c = 10.0
+        d = 110.0
+        self.assertEqual(pygame.math.remap(a, b, c, d, -8.0), 20.0)
+
+        with self.assertRaises(TypeError):
+            a = Vector2(0, 0)
+            b = "fish"
+            c = "durk"
```
durkisneer1, line 158: I approve.

src_c/math.c
```diff
@@ -4233,6 +4245,81 @@ math_clamp(PyObject *self, PyObject *const *args, Py_ssize_t nargs)
...
+        return RAISE(PyExc_TypeError,
+                     "invlerp requires exactly 3 numeric arguments");
+
+    PyObject *min = args[0];
+    PyObject *max = args[1];
+    PyObject *value = args[2];
+
+    if (!PyNumber_Check(min) || !PyNumber_Check(max) || !PyNumber_Check(value))
+        return RAISE(PyExc_TypeError,
+                     "invlerp requires all the arguments to be numbers");
```
ankith26, line 4261: improvement suggestion: instead of checking this explicitly, we can instead call `PyFloat_AsDouble` directly, and then handle the error on a per-argument basis? That we can also tell which of the arguments is incorrect

src_c/math.c
```diff
@@ -4233,6 +4245,81 @@ math_clamp(PyObject *self, PyObject *const *args, Py_ssize_t nargs)
...
+    double b = PyFloat_AsDouble(max);
+
+    if (PyErr_Occurred())
+        return RAISE(PyExc_ValueError,
+                     "invalid argument values passed to invlerp, numbers "
+                     "might be too small or too big");
+
+    if (b - a == 0)
+        return RAISE(PyExc_ZeroDivisionError,
+                     "the result of b - a needs to be different from zero");
```
ankith26, line 4280: Instead of a zero division error, how about raising something else, like maybe a `ValueError`

src_c/math.c
```diff
@@ -4192,6 +4192,18 @@ vector_elementwise(pgVector *vec, PyObject *_null)
     return (PyObject *)proxy;
 }
 
+inline double
+lerp(double a, double b, double v)
+{
+    return a + (b - a) * v;
+}
```
ankith26, line 4199: Nice addition! To take things further, this could also be called from the implementation of `math_lerp`

bilhox: yup, @Matiiss asked me to make a different PR so I can change ``pygame.math.lerp`` code, just to difference new api and enhancement.

src_c/math.c
bilhox: So If I understood, I should write something like this ? :
```c
double v = PyFloat_AsDouble(value);
if (PyErr_Occurred())
    return RAISE(PyExc_ValueError,
                 "invalid `arg` values passed to remap, the value might "
                 "be too small or too big");
double a = PyFloat_AsDouble(i_min);
// same here
double b = PyFloat_AsDouble(i_max);
// same here
double c = PyFloat_AsDouble(o_min);
// same here
double d = PyFloat_AsDouble(o_max);
// same here
```

ankith26: Well, we could do a simple `return NULL` as the error is already set. But the error won't be really specific

src_c/math.c
```diff
@@ -4233,6 +4245,81 @@ math_clamp(PyObject *self, PyObject *const *args, Py_ssize_t nargs)
...
+    if (!PyNumber_Check(min) || !PyNumber_Check(max) || !PyNumber_Check(value))
+        return RAISE(PyExc_TypeError,
+                     "invlerp requires all the arguments to be numbers");
+
+    double t = PyFloat_AsDouble(value);
+
+    if (t < 0)
+        t = 0.0;
+    else if (t > 1)
+        t = 1.0;
```
MyreMylar, line 4268: This check breaks the implementation of inverse lerp. You need to be able to reverse the interploatation of results larger than 1. e.g:

```
inv_lerp_result = pygame.math.invlerp(0, 10, 5)
```

Currently this will return 0.1, It should return 0.5. You need to expand your test cases with third parameters with a magnitude larger than 1. See this page on the inverse lerp function:

https://www.gamedev.net/articles/programming/general-and-gameplay-programming/inverse-lerp-a-super-useful-yet-often-overlooked-function-r5230/

MyreMylar (changes requested):
At the minute the float version of the inverse lerp implementation doesn't work how I would expect. It does not reverse a `lerp()` operation - except if the result is between 0 and 1.

the `remap()` seems ok 👍

test/math_test.py
```diff
@@ -86,6 +86,107 @@ def test_lerp(self):
             b = 2
             pygame.math.lerp(a, b, Vector2(0, 0))
 
+    def test_invlerp(self):
+        a = 0.0
+        b = 10.0
+        self.assertEqual(pygame.math.invlerp(a, b, 0.5), 0.05)
+
```
MyreMylar, line 93:
```suggestion
        a = 0.0
        b = 10.0
        self.assertEqual(pygame.math.invlerp(a, b, 5.0), 0.5)

```

MyreMylar: Changing a test to show how the implementation is not working.

test/math_test.py
```diff
@@ -86,6 +86,107 @@ def test_lerp(self):
...
+        b = 10.0
+        self.assertEqual(pygame.math.invlerp(a, b, 0.1), 0.01)
+
+        a = -10.0
+        b = 10.0
+        self.assertEqual(pygame.math.invlerp(a, b, 0.5), 0.525)
+
+        a = -10.0
+        b = 10.0
+        self.assertEqual(pygame.math.invlerp(a, b, 1.5), 0.55)
```
MyreMylar, line 104:
```suggestion
        a = -10.0
        b = 10.0
        self.assertEqual(pygame.math.invlerp(a, b, 1.0), 0.55)
```

MyreMylar: Correcting test.

test/math_test.py
```diff
@@ -86,6 +86,107 @@ def test_lerp(self):
...
+        b = 10.0
+        self.assertEqual(pygame.math.invlerp(a, b, 0.1), 0.01)
+
+        a = -10.0
+        b = 10.0
+        self.assertEqual(pygame.math.invlerp(a, b, 0.5), 0.525)
+
+        a = -10.0
+        b = 10.0
+        self.assertEqual(pygame.math.invlerp(a, b, 1.0), 0.55)
```
MyreMylar, line 104:
```suggestion
        self.assertEqual(pygame.math.invlerp(a, b, 1.5), 0.575)
```
Better correction.

src_c/math.c
bilhox: technically, I just have to remove the value limit and it's okay

MyreMylar (approved):
LGTM 👍 works as described now.

Matiiss (approved):
LGTM 🎉
If you want to, you can implement Ankith's suggestions, but otherwise it seems to be functionally fine. It is failing one of the CI runs though and as far as I can tell, you'd just need to rebase this branch on top of main and that should fix that I hope.

Matiiss:
> There is one thing I didn't understand from what ankith asked me to do, and with my exams I didn't have time sorry. Anyways, if you can explain me how to specificly now which variable is incorrect, rather than copy/pasting the error condition.

Good question, one approach would of course be a macro that does that, just add it after every double conversion and give it the argument number or name or whatever and make it just add the error message with that name in that place using the macro (make sure you undef it after usage). I guess another approach is a function, but that seems somewhat unnecessary for this. So yeah, probably a macro if you want to do that.

src_c/math.c
```diff
@@ -4233,6 +4245,87 @@ math_clamp(PyObject *self, PyObject *const *args, Py_ssize_t nargs)
     return value;
 }
 
+#define RAISE_ARG_TYPE_ERROR(py_num_obj, var)                         \
+    if (!PyNumber_Check(py_num_obj)) {                                \
+        return RAISE(PyExc_TypeError,                                 \
+                     "The argument '" var "' must be a real number"); \
+    }
```
ankith26, line 4252:
```suggestion
#define RAISE_ARG_TYPE_ERROR(var)                                     \
    if (PyErr_Occurred()) {                                           \
        return RAISE(PyExc_TypeError,                                 \
                     "The argument '" var "' must be a real number"); \
    }
```

src_c/math.c
```diff
@@ -4233,6 +4245,87 @@ math_clamp(PyObject *self, PyObject *const *args, Py_ssize_t nargs)
...
+    RAISE_ARG_TYPE_ERROR(value, "value")
+
+    // if (!PyNumber_Check(min) || !PyNumber_Check(max) ||
+    // !PyNumber_Check(value))
+    //     return RAISE(PyExc_TypeError,
+    //                  "invlerp requires all the arguments to be numbers");
+
+    double t = PyFloat_AsDouble(value);
+    double a = PyFloat_AsDouble(min);
+    double b = PyFloat_AsDouble(max);
```
ankith26, line 4275:
```suggestion
    double a = PyFloat_AsDouble(args[0]);
    RAISE_ARG_TYPE_ERROR("min")
    double b = PyFloat_AsDouble(args[1])
    RAISE_ARG_TYPE_ERROR("max")
    double t = PyFloat_AsDouble(args[2]);
    RAISE_ARG_TYPE_ERROR("value")
```
We check if `PyFloat_AsDouble` raises an error, and if it does, it means that the input is incorrect and therefore raise our custom error that tells which argument exactly is incorrect.

Sorry for the back and forth about the reviews, but if this is resolved I don't think I will have any other issues with this PR. You can also apply a similar patch for the other function being added

src_c/math.c
```diff
@@ -4233,6 +4245,78 @@ math_clamp(PyObject *self, PyObject *const *args, Py_ssize_t nargs)
...
+    RAISE_ARG_TYPE_ERROR("min")
+    double b = PyFloat_AsDouble(args[1]);
+    RAISE_ARG_TYPE_ERROR("max")
+    double t = PyFloat_AsDouble(args[2]);
+    RAISE_ARG_TYPE_ERROR("value")
+
+    if (PyErr_Occurred())
+        return RAISE(PyExc_ValueError,
+                     "invalid argument values passed to invlerp, numbers "
+                     "might be too small or too big");
```
ankith26, line 4271: this check is redundant now

src_c/math.c
```diff
@@ -4233,6 +4245,78 @@ math_clamp(PyObject *self, PyObject *const *args, Py_ssize_t nargs)
...
+        !PyNumber_Check(i_max) || !PyNumber_Check(o_min) ||
+        !PyNumber_Check(o_max))
+        return RAISE(PyExc_TypeError,
+                     "remap requires all the arguments to be numbers");
+
+    double v = PyFloat_AsDouble(value);
+    double a = PyFloat_AsDouble(i_min);
+    double b = PyFloat_AsDouble(i_max);
+    double c = PyFloat_AsDouble(o_min);
+    double d = PyFloat_AsDouble(o_max);
```
ankith26, line 4305: same macro and changes as above have to be applied here too

Starbuck5:
Why is this implemented in C?

MyreMylar:
> Why is this implemented in C?

Why not? I think for this current PR it would be a bit awkward & somewhat confusing to have 90% of the current math library in C and just two functions not written in C. It is a worthwhile question to ask though and I would like more of pygame's code to be written in python.

If you are pondering re-implementing the _whole_ of `pygame.math` in python then that would be interesting to see how the current speed of python 3.12 stacks up against C implementations.

Starbuck5:
> Why not?

Because bigwhoop's feature request had the functions written in Python. It would be trivial to just monkey patch them in in `__init__.py`.

Less code on our end, might even be more performant.

MyreMylar:
> > Why not?
> 
> Because bigwhoop's feature request had the functions written in Python. It would be trivial to just monkey patch them in in `__init__.py`.
> 
> Less code on our end, might even be more performant.

I'd like to see a pull request like this - both to see how it looks as an example case, and to see some performance testing. I think performance is definitely relevant for a math function like this that could be called repeatedly in tight loop drawing code.

ankith26:
Here is a simple performance test script for testing the `invlerp` implemented here
```py
import pygame
import random
import time


use_python = False

objs = [
    (
        random.uniform(-100, 100),
        random.uniform(-100, 100),
        random.uniform(0, 1),
    )
    for _ in range(5000)
]

def py_invlerp(a, b, v, /):
    return (v - a) / (b - a)


start = time.time()

if use_python:
    for arg in objs:
        py_invlerp(*arg)


else:
    for arg in objs:
        pygame.math.invlerp(*arg)


print(time.time() - start)
```
I'm on python 3.12 and the C version implemented in this PR is more performant than the python version. However, at the rate at which the language is evolving (with the jit compiler and stuff), the difference is only going to lessen over the years, seems like.

On this specific case, I still would support the C implementation because it is simple enough, and if we ever decide to do a python re-implementation: it's a one-liner.

bilhox:
I highly believe that at this stage of implementation, and because the rest of math lib is written in C, that we keep this implementation.

docs/reST/ref/math.rst
```diff
@@ -79,6 +79,24 @@ Multiple coordinates can be set using slices or swizzling
 
    .. ## math.lerp ##
 
+.. function:: invlerp
+
+   | :sl:`returns value inverse interpolated between a and b`
+   | :sg:`invlerp(a, b, value, /) -> float`
```
Starbuck5, line 85: Why is a, b, value here and a, b, weight in the stubs?

docs/reST/ref/math.rst
```diff
@@ -79,6 +79,24 @@ Multiple coordinates can be set using slices or swizzling
...
+
+   Returns a number which is an inverse interpolation between ``a``
+   and ``b``. The third parameter determines how far between ``a`` and
+   ``b`` the result is going to be.
+   If ``b - a`` is equal to 0, it raises a ``ZeroDivisionError``.
```
Starbuck5, line 90: This was very confusing to me when I was trying to figure out this function.

It takes 2 endpoints and a value, it returns the lerp amount of the value relative to the endpoints.

I think an example in this doc could help.

```py
>>> a = 10
>>> b = 20
>>> pygame.math.invlerp(10, 20, 11.5)
0.15
>>> pygame.math.lerp(10, 20, 0.15)
11.5
```

Starbuck5: I tried to look up "inverse interpolation" and got completely unhelpful results like this: https://en.wikipedia.org/wiki/Inverse_quadratic_interpolation.

docs/reST/ref/math.rst
bilhox: small mistake

docs/reST/ref/math.rst
bilhox: Added an example and re explained the purpose of the function in the docs.

src_c/math.c
```diff
@@ -4233,6 +4245,72 @@ math_clamp(PyObject *self, PyObject *const *args, Py_ssize_t nargs)
...
+
+static PyObject *
+math_invlerp(PyObject *self, PyObject *const *args, Py_ssize_t nargs)
+{
+    if (nargs != 3)
+        return RAISE(PyExc_TypeError,
+                     "invlerp requires exactly 3 numeric arguments");
+
+    double a = PyFloat_AsDouble(args[0]);
+    RAISE_ARG_TYPE_ERROR("min")
```
MyreMylar, line 4262:
```suggestion
    RAISE_ARG_TYPE_ERROR("a")
```

src_c/math.c
```diff
@@ -4233,6 +4245,72 @@ math_clamp(PyObject *self, PyObject *const *args, Py_ssize_t nargs)
...
+    double b = PyFloat_AsDouble(args[1]);
+    RAISE_ARG_TYPE_ERROR("max")
```
MyreMylar, line 4264:
```suggestion
    RAISE_ARG_TYPE_ERROR("b")
```

PR #2655, opened 2024-01-04
Rect() and FRect() initialization with no arguments

Damus666:
I've added 3 lines of code to the C src to allow `Rect` and `FRect` to be made by passing no arguments to them, which will result in a zero rect (x, y, w, h = 0).
The check is so little it should not impact performance at all.

Why did i do this?
- Code speed and quantity: this is an insignificant point but it takes less time/you type less to do `pygame.Rect()` rather than `pygame.Rect(0, 0, 0, 0)`
- It feels like it should be possible + prettier to see
- Consistency: the main point: why can i do `pygame.Vector2()` and i can't do `pygame.Rect()`? Well, now you do
- Same performace. I am very confident a little sequence length check on the args won't impact performance at all

Other than the C code i modified:
- The `_GenericRect` class in the `rect.pyi` stub to include parameterless `__init__`
- Added some lines of documentation to explain the new behaviour (i'm not the best at documenting, if you'd prefer something different tell me)
- Added a test in `rect_test.py` to verify the correct initialization of the rect


src_c/rect_impl.h
```diff
@@ -850,6 +850,10 @@ RectExport_dealloc(RectObject *self)
 static int
 RectExport_init(RectObject *self, PyObject *args, PyObject *kwds)
 {
+    if (PySequence_Fast_GET_SIZE(args) == 0) {
+        return 0;
+    }
```
itzpr3d4t0r, line 855: This could just be `PyTuple_GET_SIZE` since `args` is guaranteed to be a tuple.

Damus666: You're right. I'll correct this in a commit, thanks

Matiiss: Should also maybe move this check to after that InnerRect declaration line

novialriptide:
The code looks fine, but is there any use case for this?

Damus666:
> The code looks fine, but is there any use case for this? (@novialriptide)

There are some, i think. Maybe you want to create a rectangle and assign it's values later, just like you'd do with a `Vector2`.
In my UI library you could benefit from this alot. Each element requires a `Rect` for positioning and sizing, but very often you just want for the element to resize automatically to fill the parent (specified in the style) so you pass an empty rectangle. I use this so much i even made the `guiscript.ZeroR()` function. Having it built in would be nice, and it would be also consistent as i can do that with `pygame.Vector2` as mentioned above. temporary rects, dummy rects, rects made in the `__init__`/setup and then changed later etc could all be created without parameters. And since the check i'm adding is a simple tuple size check, i'm confident it won't impact performance so it also raises the question "why not?".

src_c/rect_impl.h
```diff
@@ -850,6 +850,10 @@ RectExport_dealloc(RectObject *self)
 static int
 RectExport_init(RectObject *self, PyObject *args, PyObject *kwds)
 {
+    if (PyTuple_GET_SIZE(args) == 0) {
+        return 0;
```
ankith26, line 854: we should probably explicitly set all the fields of `self->r` to 0 here. I dunno the internals of python memory management but relying on it to be 0 initialised sounds like a bad idea anyways

Damus666: That's true, thanks. We don't want our rect to have random values from existing memory (it happened at the previous attempt at this). I'll be updating it now.

itzpr3d4t0r: This is weird, since we already do that in `RectExport_new`, in theory we shouldn't do this again...

Damus666: When I said it happened at the previous attempt, I meant it happened when I tried to achieve the same result a few months ago when I was editing another function. Tell me if you think the change is still needed, and if yes, I posted a weird failure on the discord channel #contributing, maybe you know why

src_c/rect_impl.h
```diff
@@ -850,6 +850,15 @@ RectExport_dealloc(RectObject *self)
 static int
 RectExport_init(RectObject *self, PyObject *args, PyObject *kwds)
 {
+    if (PyTuple_GET_SIZE(args) == 0) {
+        // Setting fields explicitly to 0 to eliminate any chance of bugs
+        // happening
+        self->r.x = self->r.y = (PrimitiveType)0;
+        self->r.w = self->r.h = (PrimitiveType)0;
+        self->weakreflist = NULL;
```
itzpr3d4t0r, line 858: Yeah I don't think this is necessary. rect_new already sets all fields to zero. To check this is true, just run this program for an arbitrary number of times:
```Py
from pygame import Rect

r_zero = Rect(0, 0, 0, 0)

l = [r_zero == Rect.__new__(Rect) for _ in range(10_000_000)]
print(all(l))
```

itzpr3d4t0r (changes requested):

src_c/rect_impl.h
Damus666: Yeah, exactly. I removed those in the newest commit, thanks

test/rect_test.py
```diff
@@ -2847,6 +2847,13 @@ def test_collection_abc(self):
         self.assertTrue(isinstance(r, Collection))
         self.assertFalse(isinstance(r, Sequence))
 
+    def test_construction_no_args(self):
+        r = Rect()
+        self.assertEqual(r.x, 0)
+        self.assertEqual(r.y, 0)
+        self.assertEqual(r.w, 0)
+        self.assertEqual(r.h, 0)
+
```
itzpr3d4t0r, line 2856: A test for FRect is still missing.

docs/reST/ref/rect.rst
```diff
@@ -26,6 +28,10 @@
    values to construct a Rect. This makes it easier to create Rects on the fly
    as arguments to functions.
 
+   If no arguments are given, a zero Rect will be created (x, y, w, h = 0).
```
itzpr3d4t0r, line 31: I don't really like the format of `(x, y, w, h = 0)`, it looks like only the h is set to zero that way. Maybe something like `(x=0, y=0, w=0, h=0)` or `(x, y, w, h) = (0, 0, 0, 0)` would be more clear?

buildconfig/stubs/pygame/rect.pyi
```diff
@@ -128,6 +128,8 @@ class _GenericRect(Collection[_N]):
     def __init__(self, left_top: Coordinate, width_height: Coordinate) -> None: ...
     @overload
     def __init__(self, single_arg: RectValue) -> None: ...
+    @overload
+    def __init__(self) -> None: ...
```
itzpr3d4t0r, line 132: This should return a `Rect`/`Frect`, not `None` no?

test/rect_test.py
itzpr3d4t0r: You should also add tests for Rect/Frect subclasses.

Damus666: The FRect test unit inherits from the normal rect test, so it should have that already. Please correct me if i'm wrong

docs/reST/ref/rect.rst
Damus666: Good idea, I'll change that for the next commit (i'll commit once all the code change requests are resolved)

ankith26 (approved):

I agree with the reasoning to add this API, might not be a huge deal, but does no harm so why not.

I also see that `__new__` inits the fields to 0 already, so there's no issue of uninitialised memory at any point

MyreMylar (approved):
LGTM, and the API makes sense. 👍

itzpr3d4t0r (approved):

PR #2656, opened 2024-01-04
Partially refactor SSE2 blitters with macros

itzpr3d4t0r:
Similar to #2055, this PR tries to streamline and condense our blitter SSE2 code to make it easier to work with and understand/fix. This PR only implements the respective RGB and RGBA variants of the `ADD`, `SUB`,  `MULT`, `MIN` and `MAX`. Makes the file 514 lines shorter.

These changes don't affect performance negatively in any case, instead they improve performance for the SSE2 MULT/RGBA_MULT cases by about 15-21% due to the old version using a worse strategy.

I used two separate programs to test, one for generating the data and another for plotting and printing results.
**And a warning, the generation program has some option at the top of the files which are to ensure the testing is fair and precise. This current setup generates .json files that could be very big so watch out or reduce the `REP` value.**

Data Program:
```Py
import pygame
import json
from timeit import repeat
from datetime import datetime

pygame.init()

tests = []
for mode in [
    "ADD",
    "RGBA_ADD",
    "SUB",
    "RGBA_SUB",
    "MULT",
    "RGBA_MULT",
    "MIN",
    "RGBA_MIN",
    "MAX",
    "RGBA_MAX",
]:
    for old in ["_old"]:
        stmt = f"base.blit(surf, (0, 0), special_flags=pygame.BLEND_{mode})"
        tests.append(("blit_sse2_" + mode + old, stmt))

MAX_SIZE = 1000
REP = 250
NUM = 1
PRINT_EVERY = 100


def run_tests(tests: list):
    # print hours and minutes and seconds of start time
    start_time = datetime.now()
    print(f"Started at {start_time.strftime('%H:%M:%S')}")

    for filename, stmt in tests:
        data = {"title": filename, "data": []}
        test_start_time = datetime.now()
        print(f"=====| {filename} |=====")

        print("<", end="")
            if size % PRINT_EVERY == 0:
                g = 3
                amt = g * size // PRINT_EVERY
                print('\r[' + '*' * (amt) + ' ' * (g * 10 - amt) + ']', end="")

            surf = pygame.Surface((size, size)).convert_alpha()
            base = pygame.Surface((size, size)).convert_alpha()

            g = globals()
            g["surf"] = surf
            g["base"] = base
            times = repeat(stmt, number=NUM, repeat=REP, globals=g)
            data["data"].append(times)

        test_end_time = datetime.now()
        print(f"\nFinished {filename} in {test_end_time - test_start_time}")
        print("====================================\n")

        with open(f"{filename}.json", "w") as f:
            json.dump(data, f)

    # print hours and minutes and seconds of end time and delta time
    end_time = datetime.now()
    print(f"Finished at {end_time.strftime('%H:%M:%S')}")
    print(f"Total time it took: {end_time - start_time}")



run_tests(tests)
pygame.quit()
```

Visualization Program:
```Py

import json

plt.style.use(["dark_background", "no-latex", "notebook"])

TEST_NAME = "Blit Test"
MODE = "MIN"
LIMIT_TO_RANGE = 500
files = []

comparative_tests = [
    # (("blit_sse2_ADD_old", "red", MODE), ("blit_sse2_ADD_new", "green", MODE)),
    # (("blit_sse2_SUB_old", "red", MODE), ("blit_sse2_SUB_new", "green", MODE)),
    # (("blit_sse2_MULT_old", "red", MODE), ("blit_sse2_MULT_new", "green", MODE)),
    # (("blit_sse2_MIN_old", "red", MODE), ("blit_sse2_MIN_new", "green", MODE)),
    # (("blit_sse2_MAX_old", "white", MODE), ("blit_sse2_MAX_new", "blue", MODE)),
    # (("blit_sse2_ADD_old", "white", MODE), ("blit_sse2_ADD_new", "blue", MODE)),
]


def plot_tests(tests: list):
    for file_name, color, mode in tests:
        try:
            with open(f"{file_name}.json", "r") as f:
                data = json.load(f)
        except FileNotFoundError:
            print(f"File {file_name}.json not found!")
            quit()

        f = mean
        match mode:
            case "MEAN":
                f = mean
            case "MIN":
                f = min
            case "MAX":
                f = max
            case "MEDIAN":
                f = median

        timings = [f(data_point) for data_point in data["data"]][:LIMIT_TO_RANGE]

        print(f"=== {file_name} ===")
        print(
            f"Total: {sum([sum(data_point) for data_point in data['data']])}\n"
            f"Mean: {mean(timings)}\n"
            f"Median: {median(timings)}\n"
            f"Stdev: {stdev(timings)}"
        )
        print()
        plt.plot(timings, color=color, label=file_name, linewidth=1)


def plot_and_compare_tests(c_tests: list):
    for test1, test2 in c_tests:
        filename_1, color_1, mode_1 = test1
        filename_2, color_2, mode_2 = test2

        try:
            with open(f"{filename_1}.json", "r") as f:
                data_1 = json.load(f)
        except FileNotFoundError:
            print(f"File {filename_1}.json not found!")
            quit()

        try:
            with open(f"{filename_2}.json", "r") as f:
                data_2 = json.load(f)
        except FileNotFoundError:
            print(f"File {filename_2}.json not found!")
            quit()

        f = mean
        match mode_1:
            case "MEAN":
                f = mean
            case "MIN":
                f = min
            case "MAX":
                f = max
            case "MEDIAN":
                f = median

        timings_1 = [f(data_point) for data_point in data_1["data"]][:LIMIT_TO_RANGE]
        timings_2 = [f(data_point) for data_point in data_2["data"]][:LIMIT_TO_RANGE]

        print(f"=== {filename_1} vs {filename_2} ===")
        print(
            f"Total 1: {sum([sum(data_point) for data_point in data_1['data']])}\n"
            f"Mean 1: {mean(timings_1)}\n"
            f"Median 1: {median(timings_1)}\n"
            f"Stdev 1: {stdev(timings_1)}"
        )
        print(
            f"Total 2: {sum([sum(data_point) for data_point in data_2['data']])}\n"
            f"Mean 2: {mean(timings_2)}\n"
            f"Median 2: {median(timings_2)}\n"
            f"Stdev 2: {stdev(timings_2)}"
        )

        print()
        plt.plot(timings_1, color=color_1, label=filename_1, linewidth=1)
        plt.plot(timings_2, color=color_2, label=filename_2, linewidth=1)


plot_and_compare_tests(comparative_tests)

plt.title(TEST_NAME)
plt.ylabel("Time (s)")
plt.legend()
plt.show()
```


itzpr3d4t0r:
Apparently the latest commits (inlining some registers) made the code 11% faster!

itzpr3d4t0r:
> I don't like having the DUFF loop in the non-4-at-a-time pixel segments. It seems wasteful to me to duplicate the code a lot just so it can be run through once per height. The 4-at-a-time pixel segments could go through 800 pixels every row, the non-4-at-a-time will only ever go through 3 at most.

I'm fine with this, I'll test the perf difference out of curiosity but I'll change that no matter what.

> I'd like to keep the functions themselves (like the function definitions, the names) handwritten in the code, rather than having them defined by the macro. This way you can jump around the code in an IDE to those functions or search for them directly. This is also a style preference. Thoughts?

> I understand you're using the same strategy you used in the SSE2 fill and I didn't complain then, but this has a better argument for wanting to be more generic.

I was trying to replicate the filler strategy. My idea was to keep the macros generic for now. This way, if we want to modify a function like ADD in the future, we can simply remove one line and apply your strategy of using macros within a function definition. This means we won’t have to change all the macros, just replace individual instances with custom code. I understand your point and partially agree. However, one can simply look at the SSE blitters file to see that the blitters are defined with macros, which should give a clear picture of what’s happening.

> Also, tell me about the const. Any performance impact on that? We haven't been using const on our __mm128is before.

The only reason I did that was to help the people working on this stuff when reading to make them know what's constant and what's not. So basically to help them read the code.

Starbuck5 (approved):
Nice code quality improvement! (And also perf improvement for mult)

MyreMylar (approved):
LGTM. 

I can confirm the small performance enhancement on my machine, and the reduction in code size speaks for itself. Nice work! 🎉

Issue #2390, opened 2023-08-10
`Surface.fill()` performance improvements


itzpr3d4t0r:
While investigating for #2388, I've concluded that the main issue with inconsistent or sub-optimal performance is caused by SDL's custom implementantion for memcpy, but also for memset, as it's used inside the `SDL_FilLRect` function. In this Issue i'm focused on the 32-4 byte surface case. Long story short there looks to be a 6X performance improvement still on the table.

And I discovered a similar issue with fill, having very inconsistent performance across surface sizes. At first it looked like just a inconsistency issue, but i later discovered that the SDL_memset4 function was causing massive performance deficits.

This is SDL's implementation of the memset function used to fill a solid rect in the 4-byte case:
```C
SDL_FORCE_INLINE void SDL_memset4(void *dst, int val, size_t len)
{
#if defined(__GNUC__) && defined(i386)
    int u0, u1, u2;
    __asm__ __volatile__ (
        "cld \n\t"
        "rep ; stosl \n\t"
        : "=&D" (u0), "=&a" (u1), "=&c" (u2)
        : "0" (dst), "1" (val), "2" (SDL_static_cast(Uint32, len))
        : "memory"
    );
#else
    size_t _n = (len + 3) / 4;
    Uint32 *_p = SDL_static_cast(Uint32 *, dst);
    Uint32 _val = (val);
    if (len == 0)
        return;
    switch (len % 4)
    {
        case 0: do {    *_p++ = _val;
        case 3:         *_p++ = _val;
        case 2:         *_p++ = _val;
        case 1:         *_p++ = _val;
        } while ( --_n );
    }
#endif
}
```

This is a graph made with the program below. It shows how the performance levels aren't on a smooth curve, but rather on fairly separate levels that cause this inconsistency in performance.

![main sdl memses](https://github.com/pygame-community/pygame-ce/assets/103119829/8cc23fa1-0815-45da-9825-47adf1166816)

Text results:
```
Total time: 2.479284100008954
Mean: 0.004948670858301305
Median: 0.0048034999999799766
Stdev: 0.0032880930364022334
```
Program:
```Py
import pygame
from matplotlib import pyplot as plt
from timeit import timeit
from statistics import mean, stdev, median

pygame.init()

max_size = 500
color = (233, 12, 34)
total_time = 0
data = []

for s in range(max_size + 1):
    base = pygame.Surface((s, s))
    G = globals()
[...]

PR #2657, opened 2024-01-05
SIMD surface.fill

MightyJosip:
This should improve the performance of simple surface.fill

Test code:
```py
from timeit import repeat
import pygame

pygame.init()
surf = pygame.Surface((800, 600))
G = globals()

teststr = "surf.fill((24, 24, 24))"
l = [min(repeat(teststr, globals=G, number=1000, repeat=10)) for _ in range(5)]
print(f"fill: {sum(l) / len(l)}")
```

My results
```py
pygame-ce 2.5.0.dev1 (SDL 2.28.5, Python 3.11.0)
fill: 0.03824264000868425

pygame-ce 2.4.0 (SDL 2.28.5, Python 3.11.0)
fill: 0.08401216000784188
```

### DIFF
--- a/src_c/simd_fill.h
+++ b/src_c/simd_fill.h
@@ -45,6 +45,9 @@ _pg_HasSSE_NEON();
 
 // AVX2 functions
 int
+surface_fill_blend_rgba_none_avx2(SDL_Surface *surface, SDL_Rect *rect,
+                             Uint32 color);
+int
 surface_fill_blend_add_avx2(SDL_Surface *surface, SDL_Rect *rect,
                             Uint32 color);
 int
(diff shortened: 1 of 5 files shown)


itzpr3d4t0r:
Cool to see you venture into SIMD! Unfortunately I think you just found out something I've discovered a while ago, that is going for an AVX/SSE naive strategy for optimizing the fill algorithm (see some of my findings here: https://github.com/pygame-community/pygame-ce/issues/2390). That strategy is fine for small blits but it doesn't scale whell with surface size due to cache pollution. In the graph below the current best SDL version is the `SDL (SSE)` one that uses stream instructions to minimize cache pollutions and what you are going to is the AVX2 naive one. This should be investigated further but I'm pretty sure this is the case.

all_together

MightyJosip:
oh yea you are right

PR #2666, opened 2024-01-14
Further deprecation of deprecated sprite groups

Matiiss:
Follow through of https://github.com/pygame-community/pygame-ce/pull/2036

### DIFF
--- a/buildconfig/stubs/pygame/sprite.pyi
+++ b/buildconfig/stubs/pygame/sprite.pyi
@@ -183,7 +183,12 @@ class Group(AbstractGroup[_TSprite]):
         self, *sprites: Union[_TSprite, AbstractGroup[_TSprite], Iterable[_TSprite]]
     ) -> None: ...
 
+# these are aliased in the code too
+RenderPlain = Group
+RenderClear = Group
+
 class RenderUpdates(Group[_TSprite]): ...
+class OrderedUpdates(RenderUpdates[_TSprite]): ...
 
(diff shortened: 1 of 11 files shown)


Starbuck5:
These things were always minor, but this is still a compatibility break. I don't think they should be removed.

Updating docs and tutorials to not mention them is fine, seems like a good idea.

Matiiss:
> These things were always minor, but this is still a compatibility break. I don't think they should be removed.
> 
> Updating docs and tutorials to not mention them is fine, seems like a good idea.

I was just following the contents of the deprecation warnings, which were put there by MyreMylar. Also, I don't see that much harm in removing these particular groups, since they're just aliases, those few few people who actually use them and at some later point move over to pygame-ce can just do a find and replace and change the names to `Group` or that other group.

Starbuck5:
> I was just following the contents of the deprecation warnings, which were put there by MyreMylar

I didn't notice the deprecation warnings said that. That should be changed.

> those few few people who actually use them and at some later point move over to pygame-ce can just do a find and replace and change the names to Group or that other group

When you recommend someone switches over to pygame-ce, don't you want to be able say "it won't break your code" with a clean conscience?

Matiiss:
Alright, added back the functional bits to maintain compatibility and changed deprecation warnings to just inform users that the features will be removed in a future version.

MyreMylar (approved):
LGTM 👍

Starbuck5 (approved):
Changes make sense, thanks for the PR.

Issue #661, opened 2023-02-12
Mask.to_surface (1070)


GalacticEmperor1:
#  **Issue &numero;1070 opened by *[illume](https://github.com/illume)* at 2019-05-26 11:20:04**

Reading the https://github.com/pygame/pygame/issues/1029 issue again... helpful would be some sort of Mask drawing functions.

Some ideas about a possible API...

```
Mask.to_surface() -> Surface with the set ones as white, the unset ones as black.

Mask.to_surface(surface, setcolor=(255,255,255,255), unsetcolor=None)
-> draws setcolor onto the surface where the mask is set to 1.

Mask.to_surface(surface, setcolor=(255,255,255,255), unsetcolor=(0,0,0,127))
Mask.to_surface(surface, (255,255,255,255), (0,0,0,127))
-> draws setcolor onto the surface where the mask is set to 1, 
    and unsetcolor where the mask is set to 0.

Mask.to_surface(surface, setsurface=surf, unsetsurface=None)
 -> draws on surface, using values from setsurface where the mask is set to 1.

Mask.to_surface(surface, setsurface=surf, unsetcolor=(255,255,255,127))
 -> draws on surface, using values from setsurface where the mask is set to 1.
    and unsetcolor where the mask is set to 0.

Mask.to_surface(surface=None, setsurface=surf)
Mask.to_surface(setsurface=surf)
->creates a new surface because surface=None, based on the setsurface settings.

Mask.to_surface(offset=(10, 10)) -> Surface
```

This would allow drawing on an existing surface, as well as making it easy to create a new surface based on the mask.

`to_surface` is used here because it is the opposite of from_surface.

**Related Docs:** https://www.pygame.org/docs/ref/mask.html# pygame.mask.Mask.to_surface<hr>

#  Comments

# # #  **[illume](https://github.com/illume)* commented at 2019-05-26 11:26:20*

Another approach to think about is to make a Mask be an acceptable Surface. But this is a __LOT__ more work. The idea being that Surface.blit(mask, ...) and pygame.draw.line(mask, ...) would somehow work. Mask could have to have setcolor, and unsetcolor attributes perhaps (so when they are blit onto a surface we know what to draw).
<hr>

# # #  **[DeluxeYang](https://github.com/DeluxeYang)* commented at 2019-05-27 06:44:09*

this would be really helpful
<hr>

# # #  **[charlesej](https://github.com/charlesej)* commented at 2019-05-27 13:20:05*

[...]

PR #2670, opened 2024-01-15
Added `area` kwarg to `mask.to_surface`

Matiiss:
Close https://github.com/pygame-community/pygame-ce/issues/661
(I don't think any of that renaming mentioned there is going to happen, so, adding the area kwarg is about all there's left to it)


MyreMylar:
Should the size of the area rectangle also be used to determine the size of the final surface created here - where no surface is provided?

It feels a bit odd to effectively ask for a chunk of a mask to be turned into a surface and then get more than that chunk back.

MyreMylar (approved):
Going to approve this, despite my slight quibbles about functionality as I think it will work either way. How it currently works is more similar to the `.blit()` version of the area parameter -but it does feel more logical to produce a cropped mask surface if we aren't explicitly supplying a surface.

Matiiss:
> Going to approve this, despite my slight quibbles about functionality as I think it will work either way. How it currently works is more similar to the `.blit()` version of the area parameter -but it does feel more logical to produce a cropped mask surface if we aren't explicitly supplying a surface.

I did use `.blit()`'s implementation as a reference, it made sense for this to be consistent with that at the time.

zoldalma999 (commented):
Sorry for the late response, but I agree with Myre here. It would be better to use the rect's size for creating the surface. Otherwise looks good to me!

Matiiss:
Some TODOs for myself:
- [ ] fix the docs (area rect doesn't affect the destination)
- [ ] add tests for a non-solid (un)setsurface to ensure that they are offset properly as well since their (0, 0) should align with the (0, 0) of the original mask

Other than that, please do check my C code, I went through tons of iterations on this, so something might have got lost or I might have missed something.

Matiiss:
> Some TODOs for myself:
> 
>     * [x]  fix the docs (area rect doesn't affect the destination)
> 
>     * [x]  add tests for a non-solid (un)setsurface to ensure that they are offset properly as well since their (0, 0) should align with the (0, 0) of the original mask
> 
> 
> Other than that, please do check my C code, I went through tons of iterations on this, so something might have got lost or I might have missed something.

Alright, those two items are done (sort of), so yeah, reviews are welcome. Do let me know if I should add some more tests for the case I was describing, where say I mix in some other mask patterns instead of just filled/unfilled masks and maybe where I test using both setsurface and unsetsurface.

src_c/mask.c
```diff
@@ -2158,31 +2163,93 @@ mask_to_surface(PyObject *self, PyObject *args, PyObject *kwargs)
...
+        if (area_rect->x < 0) {
+            // x_dest -= area_rect->x;
+            area_rect->w += area_rect->x;
+            area_rect->x = 0;
+        }
+        if (area_rect->y < 0) {
+            // y_dest -= area_rect->y;
+            area_rect->h += area_rect->y;
+            area_rect->y = 0;
+        }
```
ankith26, line 2234: I don't get the logic here. This is not really clipping as the docs claim, this does a shift of the rect to fit the bounds

ankith26: nvm I was being dumb

ankith26:
regarding expanding the tests, I think the tests are adequeate rn but if you wanna expand it in a future PR feel free

PR #2673, opened 2024-01-16
Try new SDL lib prereleases

Starbuck5:


Starbuck5:
Note to self (or anyone else):

ankith26:
The full releases of both these modules are here.

SDL_ttf update is already done in https://github.com/pygame-community/pygame-ce/pull/2688
SDL_mixer update is happening in https://github.com/pygame-community/pygame-ce/pull/2703 with related updates

PR #2676, opened 2024-01-17
Serving documentation with a simple server.

gresm:
As discussed in [here](https://discord.com/channels/772505616680878080/772940667231928360/1197129403386314782) running ``python -m pygame.docs`` in some circumstances may fail - when a default browser is a non-classic snap package and ``pygame-ce`` library is installed in a ``.hidden`` folder (dotted), as those folders snaps [can't access](https://bugs.launchpad.net/snapd/+bug/1979060). This pull request solves this issue by introducing another method of opening documentation - running a simple local server for this purpose (with ``python -m pygame.docs.serve``) while not changing the original behavior.

### DIFF
--- a/docs/serve.py
+++ b/docs/serve.py
@@ -1,0 +1,42 @@
+# python -m pygame.docs.serve
+
+import sys
+from http.server import ThreadingHTTPServer, SimpleHTTPRequestHandler
+
+from pygame.docs import PKG_DIR, has_local_docs
+
+
+class DocsHandler(SimpleHTTPRequestHandler):
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs, directory=str(PKG_DIR))
+
+
+def serve(address: str, port: int):
+    with ThreadingHTTPServer((address, port), DocsHandler) as httpd:
+        try:
+            httpd.serve_forever()
+        except KeyboardInterrupt:
+            print("\nKeyboard interrupt received, exiting.")
+            sys.exit(0)
+
+
+if __name__ == "__main__":
+    if not has_local_docs():
+        print("ERROR: no local documentation found, cannot serve anything, exiting...")
+        sys.exit(1)
+    print("WARNING: this is not for production use!")
+    import argparse
+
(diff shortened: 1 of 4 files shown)


gresm:
Wasn't really sure whether I should and stubs, as I was changing the part of code without it, but I guess that it isn't that bad idea.

ankith26 (commented):
I'm now thinking we could just make it so that `py -m pygame.docs` creates a local server and opens it in the browser by default. This change shouldn't be too disruptive right?

gresm:
It shouldn't, but I didn't want to change the default behavior.

ankith26:
I'm fine with changing behaviour for this specific case, but let's wait for what others would say and then accordingly move this PR forward

Starbuck5:
What about changing the default behavior on Linux only?

gresm:
@Starbuck5 It probably will be confusing for some people, but I can re-implement this in the way that one can use ``python -m pygame.docs.serve`` to explicitly run server, ``python -m pygame.docs.static`` to explicitly use default behavior and ``python -m pygame.docs`` to let the program decide itself.

docs/serve.py
```diff
@@ -0,0 +1,68 @@
...
+        default=8000,
+        type=int,
+        nargs="?",
+        help="specify alternate port (default: 8000)",
+    )
+    parser.add_argument(
+        "--open",
+        "-o",
+        action="store_true",
+        default=False,
```
ankith26, line 52: Maybe the default here should be `True`? For consistency sake with `pygame.docs.static` which launches the browser by default?

ankith26 (approved):
Left a review, but apart from that this PR LGTM.

Thanks for the PR :partying_face:

docs/serve.py
gresm: Sure.

gresm:
Can someone rerun the test?

oddbookworm:
> Can someone rerun the test?

Running now

MyreMylar (approved):
LGTM 👍

PR #2680, opened 2024-01-19
[ DRAFT ] Optimized alpha_sse2_argb_no_surf_alpha

itzpr3d4t0r:
This WIP PR optimizes one of our sse2 alpha blitters to use a better 4-1 strategy instead of the current 2 if even or 1 at a time otherwise.

The implementation is almost ready but i can't quite figure out how to finish it so i call @Starbuck5 and @MyreMylar for help on this one. 
What i get is that The RGB and alpha calculations taken separately are correct but i still lack the section where i select the src/dest based of src_alpha or dst_alpha. So the result is that when the dst alpha is 0 or the src alpha 255 the result is wrong, otherwise it looks to be ok.

here is a small graph I made to showcase the current improvement ( just to grasp the order of magnitude):

image

Made with the following program:
```Py
import pygame
import json
from timeit import repeat
from datetime import datetime

from pygame import Surface

pygame.init()

MAX_SIZE = 1000
REP = 300
NUM = 1

screen = pygame.display.set_mode((MAX_SIZE, MAX_SIZE))

c1 = [32, 200, 44, 123]
c2 = [1, 12, 159, 32]

s1 = Surface((1, 1), depth=32).convert_alpha()
s2 = Surface((1, 1), depth=32).convert_alpha()

tests = []

for old in [""]:
    stmt = f"s2.blit(s1, (0, 0))"
    tests.append(("sse2_argb_no_surf_alpha NEW" + old, stmt))


def run_tests(tests: list):
    # print hours and minutes and seconds of start time
    start_time = datetime.now()
    print(f"Started at {start_time.strftime('%H:%M:%S')}")
    _m = 3

    for filename, stmt in tests:
        data = {"title": filename, "data": []}
        test_start_time = datetime.now()
        print(f"=====| {filename} |=====")

        print("<", end="")
            amt = _m * size // 100
            print("\r[" + "*" * amt + " " * (_m * 10 - amt) + "]", end="")

            s1 = Surface((size, size), depth=32).convert_alpha()
            s2 = Surface((size, size), depth=32).convert_alpha()
            s1.fill(c1)
            s2.fill(c2)

            g = globals() | locals()
            times = repeat(stmt, number=NUM, repeat=REP, globals=g)
            data["data"].append(times)

        test_end_time = datetime.now()
        print(f"\nFinished {filename} in {test_end_time - test_start_time}")
        print("====================================\n")

        with open(f"{filename}.json", "w") as f:
            json.dump(data, f)

    # print hours and minutes and seconds of end time and delta time
    end_time = datetime.now()
    print(f"Finished at {end_time.strftime('%H:%M:%S')}")
    print(f"Total time it took: {end_time - start_time}")


run_tests(tests)
pygame.quit()
```


MyreMylar:
Sounds like you are hitting the SDL1 wonky formula issue. 

Unfortunately, at some point in SDL1 blitter development somebody noticed one of the classic composition issues with the straight alpha blending formula and rather than shifting whole sale to pre-multiplied alpha (possibly not widely understood at the time), they instead bodged in a fix so that when destination alpha is equal to 0 you copy the source pixel over the destination rather than using the straight alpha blend formula. This hides one of the most commonly noticed problems with the 'straight alpha blending' formula.

That choice, and continuing to support that choice, means a branching formula, or faked branching, of some sort is needed (i.e. you can't do exactly the same thing for every pixel). All of which is bad news for SIMD conversion attempts as we have to handle per pixel information across multiple pixels which means doing a bunch of extra work that is only needed for some fraction of potential pixels.

To fix it in your changed formula you'll need to create a register of which pixels have a destination alpha of 0 with something like `_mm_cmpeq_epi8` and then use that to replace whatever the calculated pixel is with the source pixel. Looking at the old code it looks like doing this was an annoying process the first time around - I suspect you could use a dynamic shuffle to zero out the RGB if we allowed SSE3, but at SSE2 level it is more annoying. Perhaps someone has made an SSE2 version of `_mm_shuffle_epi8` or something similar.

itzpr3d4t0r:
An update: main issues are now fixed but function runtime has increased by about 16%. Now the matter is solving an inexplicable error in the circleci arm runs since those tests seem to fail without reason.
image

PR #2688, opened 2024-01-22
Upgrade `SDL_ttf `to `2.22.0`, add test for a bugfix

Matiiss:


test/font_test.py
```diff
@@ -318,6 +318,23 @@ def test_render_multiline(self):
...
+            return
+
+        f = pygame_font.Font(None, 20)
+        one_newline = f.render("\n", True, "black", "white")
+        two_newlines = f.render("\n\n", True, "black", "white")
+        three_newlines = f.render("\n\n\n", True, "black", "white")
+        self.assertTrue(
+            one_newline.get_height()
+            < two_newlines.get_height()
+            < three_newlines.get_height()
```
ankith26, line 335: could we improve this test, maybe test that the height of `two_newlines` is roughly 2x the height of `one_newline` and stuff?

Matiiss: Alright, I added two more test cases that check exactly that, but really the main thing this test is for is to make sure `font.render` doesn't raise an exception when supplied two or more newlines.

ankith26 (approved):

Starbuck5 (approved):
Thanks!

PR #2684, opened 2024-01-22
SDL 2.29.2 pre-release

oddbookworm:

PR #2690, opened 2024-01-22
SDL 2.29.3 pre-release

Matiiss:
Moved https://github.com/pygame-community/pygame-ce/pull/2684 to an upstream branch


oddbookworm:
Note that there is a small fix in `surface.c` that fixes a couple broken behaviors from <https://github.com/libsdl-org/SDL/commit/04e98d2236c02cbb10d0b88e1782b4dd4bf9d44a>

Starbuck5:
> Note that there is a small fix in surface.c that fixes a couple broken behaviors from https://github.com/libsdl-org/SDL/commit/04e98d2236c02cbb10d0b88e1782b4dd4bf9d44a

Is this a broken behavior we should complain or is this just a change in behavior? Seems like the latter.

oddbookworm:
> Is this a broken behavior we should complain or is this just a change in behavior? Seems like the latter.

Yeah, it seems like it was an intentional change to emit an SDL error message when something went wrong

PR #2692, opened 2024-01-23
Alternate take on `Renderer.to_surface` memory leak fix

ankith26:
PR that takes an idea first implemented in #2665 and tries to make it work


ankith26:
hrmmm, okay this does not work. Let's go with the first PR instead

Issue #2645, opened 2024-01-02
pygame.Color.from_normalized()


JiffyRob:
**Description**
`pygame.Color` objects can convert to and from RGB, HSLA, HSVA, CMY, and I1I2I3 colorspaces.  They can also convert to an OpenGL style 0-1 RGB space by using the normalize() method, but there is no API to convert back.  Granted it's pretty trivial to multiply and round it yourself, but it's an odd gap in the API that should probably be fixed.

Use case: I'm loading obj and mtl files and they store their colors in the 0-1 color space.  I need the 0-255 space for drawing and rendering purposes.

ankith26:
Agreed, for consistency sake this could be a nice little addition.

Adding a "good first issue" for people looking into contributing to C code, this should not be too complex (and you can look around the existing method implementations in `src_c/color.c` to get a better idea)

XFajk:
> **Description** `pygame.Color` objects can convert to and from RGB, HSLA, HSVA, CMY, and I1I2I3 colorspaces. They can also convert to an OpenGL style 0-1 RGB space by using the normalize() method, but there is no API to convert back. Granted it's pretty trivial to multiply and round it yourself, but it's an odd gap in the API that should probably be fixed.
> 
> Use case: I'm loading obj and mtl files and they store their colors in the 0-1 color space. I need the 0-255 space for drawing and rendering purposes.

so I started looking into this and I have a few questions 

I started with the function that unnormalizes the colors because it seems the easiest but I am not sure how I should implement it since the `normalize()` function returns a tuple so should I just create a method that would create a `pygame.Color` object from that kind of tuple or should it just create a new tuple that has the normal `RGBA` values range from  0 to 255 values what would be more fitting 

and should I even ask you I am kinda new to this contributing stuff when I say new I say this is my first issue so should I be asking the issues creator how am I suppose to implement the feature or should I just implement it and then ask if can be like this if they say no then I remake it

Matiiss:
How about we add a property called `normalized` (internally it might as well call `normalize`, but we also need a setter for it then) and then that easily integrates with the existing C API to add a `from_normalized` classmethod.
This would make the Python API more consistent in that `normalized` is basically a color space that you can get, set and create from.

PR #2693, opened 2024-01-23
added from_normalized() and get and set normalized methods for pygame.Color 

XFajk:
this should add the feature requested in #2645 
I added the `from_normalized()` method to `pygame.Color` also I added the getter and setter methods for the `normalized` parameter so you can change it


ankith26 (dismissed):
For backward compatibility reasons, we cannot remove the `normalize()` method. However, we can definitely add a new `from_normalized()` class method, and a new `normalized` property

XFajk:
Added it back

src_c/color.c
```diff
@@ -1477,6 +1489,89 @@ _color_set_cmy(pgColorObject *color, PyObject *value, void *closure)
...
+static int 
+_color_set_normalized(pgColorObject *color, PyObject *value, void *closure) {
+
+    PyObject *item;
+    double frgba[4] = {0.0, 0.0, 0.0, 1.0};
+
+    DEL_ATTR_NOT_SUPPORTED_CHECK("normalized", value);
+
+    if (!PySequence_Check(value) || PySequence_Size(value) < 3) {
+        PyErr_SetString(PyExc_ValueError, "invalid CMY value");
```
Matiiss, line 1514: The error message here says "CMY", should be "normalized"

src_c/color.c
```diff
@@ -706,7 +718,7 @@ _color_from_space(char *space, PyObject *args)
     return (PyObject *)color;
 }
 
-/**
+/** deprecated
```
Matiiss, line 721: I don't think this comment is necessary and the method is not really deprecated yet

Matiiss (changes requested):
Thanks a lot for the PR!

The implementation looks great!

A couple of things that need to be added:
- docs (see the comment about the color_doc.h file)
- tests, need to add those for the new property and classmethods, test file is `test/color_test.py`, just follow the existing tests for these properties and methods
- type stubs, need to add those in `buildconfig/stubs/pygame/color.pyi`, again, just see how it is already done and follow that example

src_c/doc/color_doc.h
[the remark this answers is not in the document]
XFajk: should I add the .. versionadded:: x.y.z to the docs and if yes what version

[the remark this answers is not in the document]
Matiiss: Yep, 2.5.0

docs/reST/ref/color.rst
```diff
@@ -191,6 +191,21 @@
...
 
       .. ## Color.i1i2i3 ##
+
+   .. attribute:: normalized
+
+      | :sl:`Gets or sets the normalized representation of the Color.`
+      | :sg:`normalized -> tuple`
+      
+      The ``Normalized``` representation of the Color. The ``Normalized``` components of are
+      in the ranges specified ``r`` = [0, 1], ``g`` = [0, 1], ``b`` = [0, 1]
```
Matiiss, line 201: > The ``Normalized``` components of are

Something is missing in that sentence there

src_c/color.c
```diff
@@ -1477,6 +1489,89 @@ _color_set_cmy(pgColorObject *color, PyObject *value, void *closure)
...
+    }
+    
+    /*
+    * the -1 is there so the rounding is more simular to int rounding in python
+    * than real world roundig
+    */
+    color->data[0] = (Uint8)round(frgba[0]*255.0)-1;
+    color->data[1] = (Uint8)round(frgba[1]*255.0)-1;
+    color->data[2] = (Uint8)round(frgba[2]*255.0)-1;
+    color->data[3] = (Uint8)round(frgba[3]*255.0)-1;
```
Matiiss, line 1569: The CI is failing (https://github.com/pygame-community/pygame-ce/actions/runs/7647698984/job/20841798096?pr=2693#step:3:2200) because your test is failing and your test is failing mostly because you are subtracting 1 from the rounded value, I'd suggest removing that `-1` subtraction here because if given 1 the result should obviously be 255 not 254
And also because with 0 it will be 0 but then you subtract 1 and it goes up to 255 which also doesn't really make sense.

test/color_test.py
```diff
@@ -844,6 +844,25 @@ def test_from_i1i2i3(self):
 
         with self.assertWarns(DeprecationWarning):
             self.assertEqual(expected_i1i2i3, pygame.Color.from_i1i2i3((0, 0, 0, 0)))
+            
+    def test_from_normalized(self):
+        normal = pygame.Color.from_normalized(1, 1, 1, 1)
+        normal_tuple = pygame.Color.from_normalized((1, 1, 1, 1))
+
+        expected_normal = (255, 255, 255)
```
Matiiss, line 852: This should be
```
expected_normal = (255, 255, 255, 255)
```

Matiiss:
Also make sure you run `python setup.py format`, otherwise the linter is failing as well.

XFajk:
ok I have changed the docs a bit though I am not sure if this is how you wanteded it. So check it out again and tell me if it can be that way. Hopefully, I copied/made all the tests for the normalized representation if I have missed any pls tell me I am not sure I double checked but I may have missed one or two

XFajk:
> LGTM 👍 I assume those other random files were formatted by `python setup.py format`, since the changes are tiny, I assume it's fine to leave them in.

yeah exactly

MyreMylar (approved):
LGTM 👍

PR #2698, opened 2024-01-29
added requirements-dev.txt

oddbookworm:

PR #2699, opened 2024-01-29
Requirements dev file

oddbookworm:
Continuation of #2698, using a branch on this repo instead of my own


requirements-dev.txt
```diff
@@ -0,0 +1,7 @@
+black<=23.7.0
+clang-format<=17.0.6
+numpy<=1.26.3
+opencv-python<=4.8.1.78
```
ankith26, line 4: numpy and opencv aren't really dev dependencies, they're more like optional runtime dependencies

ankith26 (commented):
I'm not sure if CI should be using this requirements dev file? It negatively affects build times, and in some builds this has an extreme effect.

oddbookworm:
> I'm not sure if CI should be using this requirements dev file? It negatively affects build times, and in some builds this has an extreme effect.

I think that's part of the point of having one. If we pin the dep versions, then when something changes that would affect us, the CI would use the new version and would break. Unless we manually pin the versions in the CI too, that wouldn't really fix anything.

requirements-dev.txt
Starbuck5: It also looks like numpy and opencv failed builds on pypy are the cause of most? all? of the current CI failures.

What I'm thinking right now is that opencv should be dropped from here and numpy should be hard pinned to the last version that ships pypy3.8 wheels (1.24.4)

We test in CI with numpy, we don't test with opencv.

oddbookworm: Can't pin to 1.24.4 numpy because it didn't ship with 3.12 wheels

MyreMylar: Probably if we are going to use numpy as a dependency we should synch to their 3.5 years/42 month community python version deprecation schedule as well or we will keep having these out of synch issues. Their schedule is defined in this NEP:

https://numpy.org/neps/nep-0029-deprecation_policy.html

3.8 was deprecated on April 14th 2023 and 3.9 will be deprecated this April.

Starbuck5: I'd rather keep a bit more extended support, as we've been doing.

I guess we should not pin numpy in the requirements file.

Also, numpy 2 is coming out if y'all haven't heard.

narilee2006:
Why not requirements.txt?

Starbuck5:
Requirements-dev is a common way to express development dependencies, I think. These aren't dependencies for pygame-ce's runtime, they are dependencies for building/testing pygame-ce.

Starbuck5:
Does this need to be updated for/because of https://github.com/pygame-community/pygame-ce/pull/2744 ?

oddbookworm:
Yup, about to push a commit updating

MyreMylar (approved):
LGTM 👍

oddbookworm:
I'll hold off on this until #2824 is resolved, needs updates from meson build too

requirements-dev.txt
```diff
@@ -0,0 +1,5 @@
+ruff<=0.4.4
+clang-format<=18.1.1
```
ankith26, line 2: In my understanding, these won't be dependencies, as pre-commit automatically takes care of it behind the scenes.

requirements-dev.txt
```diff
@@ -0,0 +1,5 @@
...
+numpy
+pylint<=3.1.0
+Sphinx<=7.2.6
```
ankith26, line 5: Also, if #2853 is merged this line won't be needed, as the Sphinx dependency will be tracked in `pyproject.toml` along with other build-time dependencies.

requirements-dev.txt
ankith26: What would be nice is to pass the equivalent of `--prefer-binary` specifically on the numpy dependency (and also maybe even opencv if we are adding that). Should fix our issues ig, idk

requirements-dev.txt
```diff
@@ -0,0 +1,3 @@
+numpy
+pylint<=3.1.0
+meson
```
ankith26, line 3: meson is already covered due to specifying `meson-python` in `[build-system] requires` of `pyproject.toml`, so like the rest it need not be re-specified here.

However, you could specify `mypy` here instead, it's used by the stubcheck program we have

.github/workflows/build-windows.yml
```diff
@@ -1,4 +1,3 @@
-name: Windows
```
ankith26, line 1: why was this line removed

oddbookworm: Oops, must've been a wonky merge conflict fix

ankith26:
This PR has been superseded by `dev.py` and therefore I am closing this. But still, thanks for working on it as the work done here still inspired some design and behaviour of the current `dev.py`.

PR #2704, opened 2024-01-30
Fix Window parameter name allow_highdpi -> allow_high_dpi

szabolcsdombi:
Please see:
https://github.com/pygame-community/pygame-ce/blob/ec5e6f80ebe22102f2c6b0fa2ee7875e6c82fea6/src_c/window.c#L857


Starbuck5 (approved):
I checked some old sources, and this actually did used to be `allow_highdpi` when Window was implemented in Cython.

That being said, I'm fine to move forward with is as it is currently named. Thanks for the PR.

ankith26 (approved):

PR #2707, opened 2024-02-04
transform.scale: softstretch SDL2/SDL3 compat

Starbuck5:
In SDL3, SDL_SoftStretch takes a ScaleMode parameter. For this case, it should be "nearest neighbor" because this is our basic scaling function.

With this change, the transform module compiles with SDL3.

### DIFF
--- a/src_c/_pygame.h
+++ b/src_c/_pygame.h
@@ -74,6 +74,9 @@
 
 #define PG_SurfaceHasRLE SDL_SurfaceHasRLE
 
+#define PG_SoftStretchNearest(src, srcrect, dst, dstrect) \
+    SDL_SoftStretch(src, srcrect, dst, dstrect, SDL_SCALEMODE_NEAREST)
+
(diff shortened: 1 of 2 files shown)


MyreMylar (approved):
Looks good 👍 

I wonder if at any point we have compared the results, visually & performance wise between a linear filtered SoftStretch and `pygame.smoothscale()`. Probably we have.

Starbuck5:
I didn't recall ever testing out SDL2's SDL_SoftStretchLinear, so I gave it a quick test to respond to your comment.

I did a small scale test, just looked at one image and did a 1.5x upscale. Visually it looked similar.

Performance-wise, SDL_SoftStretchLinear took 1.28 - 1.31 seconds to do 1000 scales. Our smoothscale does 1000 scales in 0.83 - 0.86 seconds using SSE2.

oddbookworm (approved):
LGTM!

PR #2708, opened 2024-02-04
Add missing special_flags links

itzpr3d4t0r:
Since the addition of a dedicated `special_flags` docs page (https://pyga.me/docs/ref/special_flags_list.html) Surface.blit was the only method that actually contained a link to the page. This PR adds the missing links to blits/fblits/fill.


MyreMylar (approved):
LGTM 👍

Starbuck5 (commented):
Great idea for a PR!

Some of these should be moved onto new lines, we've conventionally kept the line length in the rst files down where possible.

PR #2709, opened 2024-02-04
Add missing geometry versionadded tags

itzpr3d4t0r:
The geometry module was still missing versionadded tags so I added them.


MyreMylar (approved):
LGTM 👍

ankith26 (approved):
I was thinking a single version added at the module level will do, but this works too, thanks for the PR :tada:

PR #2712, opened 2024-02-06
SDL2/SDL3 mutex support

Starbuck5:
This gets the `time` module fully compiling, helps the `event` module as well (but `event` needs other patches).

Note from SDL3's readme-migration on mutexes:
> SDL_LockMutex and SDL_UnlockMutex now return void; if the mutex is valid (including being a NULL pointer, which returns immediately), these functions never fail. If the mutex is invalid or the caller does something illegal, like unlock another thread's mutex, this is considered undefined behavior.


oddbookworm (approved):
LGTM!

ankith26 (approved):
This PR has conflicts that need resolving, but apart that this PR LGTM! Thanks :balloon:

PR #2715, opened 2024-02-11
Adapt test runner count tweak to be 3.12.1 only

MyreMylar:
So when 3.12.1 released we had to add a tweak to our test running machinery because CPython stopped counting skipped tests as 'run'. Apparently this change has now been reverted for 3.12.2 (and 3.13) so we are back to how it was before.

This was causing our CircleCI runs to fail because that CI was grabbing the latest release of 3.12.2. I swear if they change this back again...

### DIFF
--- a/test/test_utils/test_runner.py
+++ b/test/test_utils/test_runner.py
@@ -285,11 +285,11 @@ def run_test(
         print(output.read())
         output.seek(0)
 
-    # change in way skipped tested counted from Python 3.12.1 onwards
+    # change in way skipped tested counted in Python 3.12.1 only (so far)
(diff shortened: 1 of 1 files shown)


MyreMylar:
This is the revert commit:

https://github.com/python/cpython/pull/114994

With the relevant changes being in this chunk here:

https://github.com/python/cpython/blob/e1552fd19de17e7a6daa3c2a6d1ca207bb8eaf8e/Lib/unittest/case.py#L609-L620

Moving:

```
result.startTest(self)
```

...before or after the skip test detecting block there that returns if a test is skipped.

Starbuck5 (approved):
Thanks!

PR #2716, opened 2024-02-12
Add constructing Rects/Frects with just size

MyreMylar:
To enable these shenanigans:

```python
from pygame import Rect

size = (100, 40)
rect = Rect(size).move_to(center=(60, 30))
print(rect)

>> Rect(10, 10, 100, 40)
```


MyreMylar:
Well this doesn't work.

PR #2731, opened 2024-02-26
Reorganize geometry internals

itzpr3d4t0r:
This is a preparation PR to enable us to have functions like Rect/Frect.collidecircle() and in general interoperability with shapes. Specifically it fixes a circular import we had going on.

What this PR really does at a high level is moving some functions that need to get used by rects/circles/lines/polygons into a separate header file so that it can be included everywhere without circular imports.

Deleted:
- collisions .h/.c

Added:
- geometry_common.c/.h (containing common collision functions and PyObject->Shape conversion functions.)

Simplified:
- Some unnecessary imports


src_c/geometry_utils.c
```diff
@@ -0,0 +1,207 @@
+#include "geometry_utils.h"
+
+int
+_pg_circle_set_radius(PyObject *value, pgCircleBase *circle)
+{
+    double radius = 0;
+    if (!pg_DoubleFromObj(value, &radius) || radius <= 0) {
```
oddbookworm, line 7: nitpick, but I think using 0.0 instead of 0 here because `radius` is a double is better practice

src_c/geometry_utils.c
```diff
@@ -0,0 +1,207 @@
...
+                !_pg_circle_set_radius(f_arr[1], out)) {
+                return 0;
+            }
+            return 1;
+        }
+        else {
+            /* Sequences of size other than 3 or 1 are not supported
+            (don't wanna support infinite sequence nesting anymore)*/
+            return 0;
+        }
```
oddbookworm, line 55: The comment in the `else` block is a bit misleading here. Size of 2 is also supported.

oddbookworm, line 55: Would a `switch` statement work here? Would that be worth implementing?

src_c/geometry_utils.c
```diff
@@ -0,0 +1,207 @@
...
+                return 0;
+            }
+            Py_DECREF(tmp);
+            return 1;
+        }
+        else {
+            /* Sequences of size other than 3 or 1 are not supported
+            (don't wanna support infinite sequence nesting anymore)*/
+            return 0;
+        }
```
oddbookworm, line 116: same comments as above about maybe using switch, and making the comment more accurate

src_c/geometry_utils.c
```diff
@@ -0,0 +1,207 @@
...
+    }
+
+    /* Path for objects that have a circle attribute */
+    PyObject *circleattr;
+    if (!(circleattr = PyObject_GetAttrString(obj, "circle"))) {
+        PyErr_Clear();
```
oddbookworm, line 122: Is this necessary?

src_c/geometry_utils.c
```diff
@@ -0,0 +1,207 @@
...
+        return 1;
+    }
+    else if (nargs == 3) {
+        if (!pg_DoubleFromObj(args[0], &out->x) ||
+            !pg_DoubleFromObj(args[1], &out->y) ||
+            !_pg_circle_set_radius(args[2], out)) {
+            return 0;
+        }
+        return 1;
+    }
```
oddbookworm, line 169: switch?

src_c/geometry_utils.c
```diff
@@ -0,0 +1,207 @@
...
+}
+
+/* === Collision Functions === */
+
+inline int
+pgCollision_CirclePoint(pgCircleBase *circle, double Cx, double Cy)
+{
+    double dx = circle->x - Cx;
+    double dy = circle->y - Cy;
+    return dx * dx + dy * dy <= circle->r * circle->r;
```
oddbookworm, line 180: My monkey brain is screaming at me here that we should verify that `circle` is not NULL before trying to use it here

src_c/geometry_utils.c
```diff
@@ -0,0 +1,207 @@
...
+
+inline int
+pgCollision_CircleCircle(pgCircleBase *A, pgCircleBase *B)
+{
+    double dx, dy;
+    double sum_radi;
+
+    dx = A->x - B->x;
+    dy = A->y - B->y;
+    sum_radi = A->r + B->r;
```
oddbookworm, line 191: nitpick: should be `sum_radii`

src_c/geometry_utils.c
```diff
@@ -0,0 +1,207 @@
...
+
+inline int
+pgCollision_RectCircle(double rx, double ry, double rw, double rh,
+                       pgCircleBase *circle)
+{
+    const double cx = circle->x, cy = circle->y;
+    const double r_bottom = ry + rh, r_right = rx + rw;
+
+    const double test_x = (cx < rx) ? rx : ((cx > r_right) ? r_right : cx);
+    const double test_y = (cy < ry) ? ry : ((cy > r_bottom) ? r_bottom : cy);
```
oddbookworm, line 204: more monkey brain screaming that we should verify that `circle` is not NULL

oddbookworm (approved):
I know this is just a bunch of moving stuff around, but I found some nitpicky things to comment on anyway lol. IMO none of these are a pull blocker though, so I'm approving

src_c/geometry_utils.c
itzpr3d4t0r: It's there to comply with the same code in place for Rect.

src_c/geometry_utils.c
itzpr3d4t0r: Removed the comments, they were actually outdated and wrong since we accept sequence nesting.

itzpr3d4t0r: Wouldn't be faster but would clean the code a bit, thanks.

src_c/geometry_utils.c
itzpr3d4t0r: I don't think a switch here would make code cleaner, would just add an indentation level to a lot of code per case. (Btw i changed to a switch in the fast sequence case).

src_c/geometry_utils.c
itzpr3d4t0r: Yep.

src_c/geometry_utils.c
itzpr3d4t0r: These functions are meant to be used at the lowest level possible, so they assume the inputs are valid. This can't be cahnged since the checks you want here are already in place in each collision function. If a shape can't be converted to an SLD_Rect/pg_CircleBase etc. we have already errored out a lot before this function gets called.

src_c/geometry_utils.c
itzpr3d4t0r: sure.

src_c/geometry_utils.c
itzpr3d4t0r: Same as my last comment

ankith26 (approved):
LGTM, thanks for the PR :balloon:

PR #2733, opened 2024-03-02
Add pipewire and libsamplerate+fftw3

ankith26:
Follow up PR for #1997 but for audio stuff. (that PR should be merged first)

So, pipewire will most likely be the default audiodriver in SDL3, so it makes sense to add it early for testing purposes. Again, this does not affect any default behaviour in SDL2.

I'm also taking the opportunity to add libsamplerate support, this library is useful for improving sound quality on linux


Starbuck5:
Just checked, libsamplerate is not in SDL3 FYI. (I know you didn't say it was, just informing you).

I checked the size of the wheels between this and the wheels produced for #1997 (a recent baseline, manylinux x86_64), and found the wheels are about ~16 kb bigger, which is suspiciously little for the addition of two new libraries.

ankith26:
> I checked the size of the wheels between this and the wheels produced for https://github.com/pygame-community/pygame-ce/pull/1997 (a recent baseline, manylinux x86_64), and found the wheels are about ~16 kb bigger, which is suspiciously little for the addition of two new libraries.

This is because only libsamplerate is being bundled, and it is a tiny library. pipewire is getting the dynamic load treatment, so apart from driver code on the SDL side, nothing else is being added

Starbuck5 (approved):

MyreMylar (approved):
LGTM 👍

PR #2738, opened 2024-03-04
Update mixer docs

lispspb:
I think mixer module docs need some clarification.
`A value from 0 to the value of get_num_channels()` is for [0-num_channels), while `the value argument is between 0.0 and 1.0`, `volume in the range of 0.0 to 1.0 (inclusive)` and `a value from 0.0 to 1.0 representing the volume` mean [0.0-1.0].
Should it be "from x to y (inclusive/exclusive)" everywhere?

### DIFF
--- a/docs/reST/ref/mixer.rst
+++ b/docs/reST/ref/mixer.rst
@@ -472,7 +472,7 @@ The following file formats are supported
    | :sg:`Channel(id) -> Channel`
 
    Return a Channel object for one of the current channels. The id must be a
-   value from 0 to the value of ``pygame.mixer.get_num_channels()`` (exclusive).
+   value from 0 up to, but not including, ``pygame.mixer.get_num_channels()``.
 
    The Channel object can be used to get fine control over the playback of
    Sounds. A channel can only playback a single Sound at time. Using channels
--- a/docs/reST/ref/mixer.rst
+++ b/docs/reST/ref/mixer.rst
@@ -431,7 +431,7 @@ The following file formats are supported
       | :sl:`get the playback volume`
       | :sg:`get_volume() -> value`
 
-      Return a value from 0.0 to 1.0 representing the volume for this Sound.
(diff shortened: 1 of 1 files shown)


docs/reST/ref/mixer.rst
```diff
@@ -472,7 +472,7 @@ The following file formats are supported
    | :sg:`Channel(id) -> Channel`
 
    Return a Channel object for one of the current channels. The id must be a
-   value from 0 to the value of ``pygame.mixer.get_num_channels()``.
+   value from 0 to the value of ``pygame.mixer.get_num_channels()`` (exclusive).
```
ankith26, line 475:
```suggestion
   value from 0 upto but not including ``pygame.mixer.get_num_channels()``.
```
The word "exclusive" isn't very clear, and may mislead people to think that both ends aren't included

ankith26 (commented):
Thanks for the PR!

I have left a minor review suggestion at one place, the rest of the changes LGTM

docs/reST/ref/mixer.rst
Starbuck5: Good idea.

But I'd say it as "up to" rather than "upto"

Starbuck5:
I disagree that the "0.0 -1.0" range needs clarification for volume.

docs/reST/ref/mixer.rst
MyreMylar, line 475:
```suggestion
   value from 0 up to, but not including, ``pygame.mixer.get_num_channels()``.
```

ankith26 (approved):
LGTM. Sure, it makes the wording slightly more verbose, but it reduces the room for any confusion I guess. Therefore I think this PR is a merge-able state now.

Thanks for contributing to pygame-ce! :tada:

MyreMylar (approved):
OK 👍

PR #2741, opened 2024-03-07
Add `mixer.get_driver`

ankith26:
While working on #2733 I realised that we don't actually have the `mixer` analogue to `display.get_driver()`. This PR adds `mixer.get_driver` to find out the current audio driver being used.


Starbuck5 (approved):
Seems legit, looks good

- API isn't changing in SDL3 ✔️ (well, the SDL_AUDIODRIVER env is moving to SDL_AUDIO_DRIVER, but we can put in a note for that)
- API doesn't already exist in _sdl2.audio ✔️
- Test passes locally ✔️
- Exhaustive list of drivers checks out against a skim of source code ✔️

Should get `SDL_GetAudioDriver` and `SDL_GetNumAudioDrivers` be ported too, into a method to get a list of available drivers? That way a user setting SDL_AUDIODRIVER could see what the options are.

Starbuck5:
> Should get SDL_GetAudioDriver and SDL_GetNumAudioDrivers be ported too, into a method to get a list of available drivers? That way a user setting SDL_AUDIODRIVER could see what the options are.

It seems like this would help test #2733

MyreMylar (approved):
LGTM 👍 

Tested with:

```
import pygame

pygame.init()
print(pygame.mixer.get_driver())
```
```
>> wasapi
```

ankith26:
> Should get `SDL_GetAudioDriver` and `SDL_GetNumAudioDrivers` be ported too, into a method to get a list of available drivers? That way a user setting `SDL_AUDIODRIVER` could see what the options are.

Sounds like a good idea. We could also give display the same treatment. Maybe we could export this as a list, instead of having to export two separate functions.

Starbuck5:
That's what I had in mind.

Issue #2740, opened 2024-03-06
Segfault when setting a pixel with PixelArray


bilhox:
Original report : https://github.com/pygame/pygame/issues/4163#issue-2152065057

Environment:

Operating system: Linux / Windows (Endeavour)
Python version : 3.11.7
SDL version: 2.28.0+
PyGame version: 2.4.1
Current behavior:

Segfault when changing a pixel in a PixelArray with the syntax ``pixel_array[y][x]``. 
Returns an error ``ValueError: sequence size missmatch`` with the syntax ``pixel_array[x, y]``.

Expected behavior:

Target pixel is set to the requested color in the PixelArray, as per documentation.

Test code

```py
import pygame
test = pygame.Surface([800,800])
testBuf = pygame.PixelArray(test)
testBuf[400][400] = [255,255,0]
```
Stack trace/error output/other error logs
```
$ python3.11
Python 3.11.7 (main, Jan 29 2024, 16:03:57) [GCC 13.2.1 20230801] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import pygame
pygame-ce 2.4.0 (SDL 2.28.5, Python 3.12.0)
>>> test = pygame.Surface([800,800])
>>> testBuf = pygame.PixelArray(test)
>>> testBuf[400][400] = [255,255,0]
Segmentation fault (core dumped)
```

ankith26:
Thanks for the issue, linked PR fixes the segfault.

As for your usecase, replacing `[255, 255, 0]` with `(255, 255, 0)` makes pixelarray interpret the object as a color and behave as you expect

PR #2742, opened 2024-03-07
PixelArray raise error assigning sequence to pixel

ankith26:
fixes https://github.com/pygame-community/pygame-ce/issues/2740

- In the pixelarray module, any sequence (except tuple and `pygame.Color`) isn't considered a "color", but rather a pixel sequence
- The segfault happens because the code expects a pixelarray upon subscripting for assignment, but receives an "integer color"

I am open to suggestions on how to improve the error message, can't think of something descriptive enough ATM.

### DIFF
--- a/src_c/pixelarray.c
+++ b/src_c/pixelarray.c
@@ -1318,6 +1318,12 @@ _pxarray_ass_item(pgPixelArrayObject *array, Py_ssize_t index, PyObject *value)
             if (!tmparray) {
                 return -1;
             }
+            if (!pgPixelArrayObject_Check(tmparray)) {
+                PyErr_SetString(
+                    PyExc_ValueError,
+                    "cannot assign a pixel sequence to a single pixel");
+                return -1;
+            }
             retval =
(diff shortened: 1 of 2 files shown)


oddbookworm (changes requested):
I take it the segfault is reproducible? If so, I'd like to see a regression test

ankith26:
Have added the regression test, thanks for pointing it out

oddbookworm (approved):
LGTM!

narilee2006:
Why not convert the `PyList` to a `PyTuple`?
```c
static int
_pxarray_ass_item(pgPixelArrayObject *array, Py_ssize_t index, PyObject *value)
{
    // Shortened code...

    if (PyList_Check(value))
        value = PyList_AsTuple(value);
```

ankith26:
That is a behaviour change that deviates from the specification of the pixelarray module. According to the docs:
> For assignment, a tuple can only be a color. Any other sequence type is a sequence of colors.

MyreMylar (approved):
LGTM 👍

PR #2743, opened 2024-03-07
Undeprecate draw aaline blend

### LINKED ISSUE #2739: Restore pygame.draw.aaline ` blend` argument
I got a report from Irv Kalb on discord that the removal of the `blend` argument in aaline in https://github.com/pygame-community/pygame-ce/pull/2506 broke some of their code. 

I'm following up with them on how they didn't see (or did they just ignore?) the years of deprecation warnings.

Regardless, they suggested rather than removing the parameter entirely (resulting in crashes), it could just be ignored. I think this is a good idea. It could be ignored and raise a deprecation warning complaining about its use. If we want to remove it entirely, why not do that in a potential future pygame-ce 3?

ALSO, Irv pointed that the draw example at the bottom of the draw docs page still uses the argument. So that needs to be fixed.

oddbookworm:
fixes #2739 and properly deprecates the blend argument so that if it's used, a warning will always be emitted instead of just false

### DIFF
--- a/src_c/draw.c
+++ b/src_c/draw.c
@@ -105,7 +105,7 @@ aaline(PyObject *self, PyObject *arg, PyObject *kwargs)
     PyObject *colorobj, *start, *end;
     SDL_Surface *surf = NULL;
     float startx, starty, endx, endy;
-    PyObject *blend;
+    PyObject *blend = NULL;
     int drawn_area[4] = {INT_MAX, INT_MAX, INT_MIN,
                          INT_MIN}; /* Used to store bounding box values */
     Uint32 color;
--- a/src_c/draw.c
+++ b/src_c/draw.c
@@ -105,16 +105,27 @@ aaline(PyObject *self, PyObject *arg, PyObject *kwargs)
     PyObject *colorobj, *start, *end;
     SDL_Surface *surf = NULL;
     float startx, starty, endx, endy;
+    PyObject *blend;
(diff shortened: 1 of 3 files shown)


ankith26 (commented):
Perhaps the formatting changes that are unrelated to this PR should be a separate PR? It's probably the new version of clang-format messing things up. It has been causing my PRs to fail too

oddbookworm:
> Perhaps the formatting changes that are unrelated to this PR should be a separate PR? It's probably the new version of clang-format messing things up. It has been causing my PRs to fail too

I opened #2744 to apply the formatting changes. I can rebase this pull when that merges

Starbuck5:
I think it’s fine to leave the feature removed (commits don’t need to be reverted). It’s just that the parameter should be un-removed and should always raise a depreciation warning.

ankith26 (approved):

src_c/draw.c
```diff
@@ -105,18 +105,29 @@ aaline(PyObject *self, PyObject *arg, PyObject *kwargs)
     PyObject *colorobj, *start, *end;
     SDL_Surface *surf = NULL;
     float startx, starty, endx, endy;
+    PyObject *blend;
```
Starbuck5, line 108: Shouldn't this start as null?

oddbookworm: oops

Starbuck5 (approved):
Alrighty, thanks! Code looks good, haven't tested locally.

PR #2744, opened 2024-03-09
Formatting updates using newest version of clang-format

oddbookworm:


src_c/freetype/ft_wrap.c
```diff
@@ -38,8 +38,7 @@ ft_wrap_quit(pgFontObject *);
  *
  *********************************************************/
 void
-_PGFT_SetError(FreeTypeInstance *ft, const char *error_msg, FT_Error error_id)
-{
+_PGFT_SetError(FreeTypeInstance *ft, const char *error_msg, FT_Error error_id){
```
ankith26, line 41: The changes in this file are weird... but we will just go with whatever clang-format is doing here

ankith26 (approved):
LGTM, thanks for the PR

ankith26:
There seems to be a new mypy fail too, as it catches a real issue. Maybe that can be handled in this PR as well, just gotta add something like
```py
def colorspace(
    surf: Surface, color: Literal["YUV", "HSV"], dest_surf: Surface = ..., /
) -> Surface: ...
```
to `camera.pyi`

oddbookworm:
> There seems to be a new mypy fail too, as it catches a real issue. Maybe that can be handled in this PR as well, just gotta add something like
> 
> ```python
> def colorspace(
>     surf: Surface, color: Literal["YUV", "HSV"], dest_surf: Surface = ..., /
> ) -> Surface: ...
> ```
> 
> to `camera.pyi`

Updated. I also updated the doc signature to match

novialriptide (approved):

PR #2748, opened 2024-03-11
wasm: unify pygame-web/pyodide/static Freetype init

pmp-p:
both pygbag and pyodide benefit from that change see
https://github.com/pyodide/pyodide/pull/4602#pullrequestreview-1928005211

part of the change already was in anyway https://github.com/pygame-community/pygame-ce/blob/be55232410cb33391be23d0a279ec2632d75820e/src_c/font.c#L71 and the other part was in a patch applied by pygbag because the draft https://github.com/pygame-community/pygame-ce/pull/1967 was incomplete.

### DIFF
--- a/src_c/static.c
+++ b/src_c/static.c
@@ -269,6 +269,12 @@ static struct PyModuleDef mod_pygame_static = {PyModuleDef_HEAD_INIT,
 PyMODINIT_FUNC
 PyInit_pygame_static()
 {
+    // cannot fail here, and font_initialized is already set to 1 in font.c .
+    TTF_Init();
+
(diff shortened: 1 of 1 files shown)


oddbookworm:
@pmp-p Your CI failure is due to `TTF_Init` not being available. https://github.com/pygame-community/pygame-ce/actions/runs/8237122349/job/22525122282?pr=2748#step:7:231

ankith26 (approved):
Not a WASM expert, but this change makes sense, so I'm approving. Thanks :tada:

Starbuck5 (approved):

PR #2762, opened 2024-03-21
Remove pygame.threads

Starbuck5:
The source file doesn't seem to have had a non-linting related update in the past 7 years, it's not documented, and a scour of Google for mentions saw no one using it. I think this is safe to remove.

Also, unlike when this module was started, the standard library now has a way to do a threaded map, in the form of `ThreadPoolExecutor`.


test/test_utils/run_tests.py
```diff
@@ -250,41 +250,33 @@ def sub_test(module):
                 ),
             )
```
Starbuck5, line 252: These run_tests changes only impact the "use subprocess" mode, which we don't use. And regardless, the use subprocess mode doesn't work. I have a patch locally that gets it working but I am not seeing much improvement in runtime by multithreading this.

ankith26 (changes requested):
I think we could deprecate it but not remove it just yet.
On a quick search I found atleast one github repo that uses it: https://github.com/pmtkachev/TMNT-Fight-NES/blob/c297f8aa9d802dc3fdbfc4b1b8726ccc95eeff97/src/game_func.py#L2

Starbuck5:
> I think we could deprecate it but not remove it just yet.
On a quick search I found atleast one github repo that uses it: https://github.com/pmtkachev/TMNT-Fight-NES/blob/c297f8aa9d802dc3fdbfc4b1b8726ccc95eeff97/src/game_func.py#L2

But they're not using the functionality! pygame.threads.Thread is a weird import of threading.Thread. They must've written thread and clicked "fix imports" and picked something from a list. No human would've written that unaided.

This is an undocumented module that hasn't made any progress towards being a real thing since pygame 1.8.1, and has never been documented anywhere. The functionality is available in the standard library now.

I don't see an undocumented experimental module that no one is actually using as being covered by any backwards compatibility concerns.

oddbookworm:
> But they're not using the functionality! pygame.threads.Thread is a weird import of threading.Thread. They must've written thread and clicked "fix imports" and picked something from a list. No human would've written that unaided.

Side note:
I created a [pull request](https://github.com/pmtkachev/TMNT-Fight-NES/pull/2) there to get rid of the weird alias and just use the real thing lol

oddbookworm (approved):
LGTM!

test/image_test.py
```diff
@@ -1356,12 +1357,11 @@ def test_save_extended(self):
...
 
     def threads_load(self, images):
-        import pygame.threads
-
-        for i in range(10):
-            surfs = pygame.threads.tmap(pygame.image.load, images)
-            for s in surfs:
-                self.assertIsInstance(s, pygame.Surface)
+        for _ in range(10):
+            with ThreadPoolExecutor(max_workers=20) as executor:
```
ankith26, line 1361: shouldn't the loop be inside the context block?

The current code is still going to sequentially load stuff because a new executor instance is created for each loop?

ankith26 (changes requested):

test/image_test.py
Starbuck5: This does what the previous code does. It loads all images in parallel, 10 times.

ankith26 (approved):
Okay sure, let's remove this. Thanks for the PR :tada:

PR #2763, opened 2024-03-21
Fix ftfont_test being run directly

Starbuck5:
ftfont_test uses font_test to run, but it was trying to import pygame.freetype for a version check without an explicit import (which pygame.freetype needs). Removed the dependency on pygame.freetype for the ftfont path.

I ran into this while I was trying to get the old multithreaded test runner working, which runs each test in individual processes. This failed under those conditions. Normally something else imports pygame.freetype first so it's in working order by the time it gets here.

### DIFF
--- a/test/font_test.py
+++ b/test/font_test.py
@@ -876,7 +876,7 @@ class FontTypeTest(unittest.TestCase):
                 ("set_underline", (True,)),
                 ("size", ("any Text",)),
             ]
-            skip_methods = set(("get_strikethrough", "set_strikethrough"))
+            skip_methods = {"get_strikethrough", "set_strikethrough"}
         else:
             methods = [
                 ("get_height", tuple()),
@@ -972,7 +972,7 @@ class FontTypeTest(unittest.TestCase):
                 ("italic", True),
                 ("resolution", None),
             ]
-            skip_properties = set(("strikethrough",))
+            skip_properties = {"strikethrough"}
         else:
             properties = [
                 ("name", None),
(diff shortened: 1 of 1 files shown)


oddbookworm (approved):
I don't see any issues, so LGTM

test/font_test.py
```diff
@@ -876,7 +876,7 @@ def test_font_method_should_raise_exception_after_quit(self):
                 ("set_underline", (True,)),
                 ("size", ("any Text",)),
             ]
-            version = pygame.freetype.get_version()
+            skip_methods = set(("get_strikethrough", "set_strikethrough"))
```
ankith26, line 879:
```suggestion
            skip_methods = {"get_strikethrough", "set_strikethrough"}
```
nitpick: can use set syntax here

test/font_test.py
```diff
@@ -979,7 +972,7 @@ def test_font_property_should_raise_exception_after_quit(self):
                 ("italic", True),
                 ("resolution", None),
             ]
-            version = pygame.freetype.get_version()
+            skip_properties = set(("strikethrough",))
```
ankith26, line 975:
```suggestion
            skip_properties = {"strikethrough"}
```
nitpick: set syntax

ankith26 (approved):
Left a couple of nitpicks, but apart from those this PR LGTM! :cake:

test/font_test.py
Starbuck5: Wow, I had forgotten set literals were a thing.

MunaAlaneme:
Issue BFDI

PR #2764, opened 2024-03-22
WIP: add GNU target arch attributes for AVX2 functions

robertpfeiffer:
Please ignore, testing how MSVC handles this in the CI


robertpfeiffer:
If you want, I can guard these behind ifdefs so they only show up on GCC and clang, or I can just close this PR experiment.

ankith26:
Could you explain the benefit of having this PR conditional on clang/gcc being used?

PR #2765, opened 2024-03-23
Fix `pygame.display.get_caption()` Docs in display.rst

kritserv:
Hi, this is a small grammar fix. I believe this PR could be labeled with `docs`.
using `of` instead of `for`

**from**

Returns the title and icontitle **for** the display window. In pygame 2.x these will always be the same value.

**to**

Returns the title and icontitle **of** the display window. In pygame 2.x these will always be the same value.


Starbuck5 (approved):
Sure. I agree the sentence sounds better with this change.

PR #2770, opened 2024-03-25
setup.py: Properly add _sdl2_data_files to data_files

cclauss:
Currently, this code is _ALWAYS_ adding an empty list instead of the data files that were found by `glob.glob()`.

It adds the empty list before filling the list!


ankith26 (approved):
The current code isn't broken, because of mutability of lists (the same list is being filled, so finally it isn't actually empty)

But thanks for the PR, the code is now less-weird and simpler. :tada:

cclauss:
TIL:
```python
>>> big = []
>>> little = []
>>> big.append(("little", little))
>>> big
[('little', [])]
>>> for c in "abc":
...     little.append(c)
...
>>> little
['a', 'b', 'c']
>>> big
[('little', ['a', 'b', 'c'])]
```

oddbookworm (approved):
Now that we support only python 3.8+, I'd love to see more use of the walrus operator when it makes sense to do so

PR #2774, opened 2024-03-26
Update freetype download link because current one is broken (fixes failing circleci builds)

### LINKED ISSUE #2773: build-freetype: http://download.savannah.gnu.org is busy or down
The `linux-arm-wheels` process on [CircleCI](https://app.circleci.com/pipelines/github/pygame-community/pygame-ce) is consistently failing because http://download.savannah.gnu.org is unreachable.

https://freetype.org/download.html lists two mirrors for downloading freetype.  Would it be possible to modify `freetype_build/build-freetype.sh` to alternate between these mirrors to improve build reliability?

ankith26:
fixes #2773

### DIFF
--- a/buildconfig/manylinux-build/docker_base/freetype/build-freetype.sh
+++ b/buildconfig/manylinux-build/docker_base/freetype/build-freetype.sh
@@ -7,7 +7,7 @@ FREETYPE="freetype-2.13.2"
 HARFBUZZ_VER=8.2.1
 HARFBUZZ_NAME="harfbuzz-$HARFBUZZ_VER"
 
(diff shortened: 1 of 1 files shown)


oddbookworm:
I wonder if it would be feasible to set it up to try multiple mirrors in the case that the first one we try fails

cclauss (approved):

oddbookworm (approved):
LGTM! I still think it would be a good idea to maybe try multiple mirrors if the default one fails, but not a blocker for this pull IMO

Starbuck5 (approved):

Starbuck5:
> LGTM! I still think it would be a good idea to maybe try multiple mirrors if the default one fails, but not a blocker for this pull IMO

This link is mirrored, try visiting it. I went 3 times and it sent me three different places.

PR #2775, opened 2024-03-26
build-macos.yml: Test arm64 deps on Apple Silicon

cclauss:
macos-14 runs on an M1 Mac. Anything less runs on Intel.

* https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners/about-github-hosted-runners#standard-github-hosted-runners-for-public-repositories

The M1 should run faster too...

### DIFF
--- a/.github/workflows/build-macos.yml
+++ b/.github/workflows/build-macos.yml
@@ -1,6 +1,6 @@
 name: macOS
 
-# Run CI only when a release is created, on changes to main branch, or any PR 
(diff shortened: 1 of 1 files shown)


ankith26 (commented):
Duplicate of https://github.com/pygame-community/pygame-ce/pull/2494

But thanks for contributing, if you wish you could open a PR to that branch if you feel you can improve it further

PR #2778, opened 2024-03-28
Remove apt upgrade in CI

ankith26:
I can't seem to remember why the update/upgrade was added in CI in the first place. Anyways, I'm now trying to remove it because every once in a while this step likes to break, and I don't think there's a benefit of having this extra step


Starbuck5:
Maybe keep the update but remove the upgrade? https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners/customizing-github-hosted-runners#installing-software-on-ubuntu-runners

ankith26:
In my latest commit upgrade has been removed but update has been kept as it was

Starbuck5 (approved):

PR #2779, opened 2024-03-28
Add armv6 and armv7 to multiarch CI and fix build issues

ankith26:
This PR mainly does ~~3~~ 2 things
- ~~Fixes arm32 specific warnings across the codebase~~ Has been moved out to be done in a later PR
- Fixes compilation on arm32 (both armv6 and armv7) under the meson build system. As a result of the meson PR, the NEON optimizations are enabled by default on armv7 now (in the legacy buildconfig it is controlled via an opt in flag). Now that it is much easier to pass per file flags, the flags needed to build with `sse2neon.h` are only passed to the files that need it. Therefore, I made it so that only the files that require the `sse2neon.h` header include it. If any file that shouldn't be including it tries to include it, the compilation fails.
- And now coming to the good stuff, we finally have testing on CI to verify that we support both armv6 and armv7 based raspberrypi-like systems! The first level of testing tests that a wheel can be built and unit test can be run on armv6 and armv7. The second level of testing tests that an armv7 wheel can simply be renamed to lie that it's an armv6 wheel, and still pass tests on armv6 hardware. This is important because piwheels does exactly this to package `pygame-ce` wheels over at https://www.piwheels.org/project/pygame-ce/

### DIFF
--- a/.github/workflows/build-debian-multiarch.yml
+++ b/.github/workflows/build-debian-multiarch.yml
@@ -35,26 +35,45 @@ concurrency:
   group: ${{ github.workflow }}-${{ github.ref }}-debian-multiarch
   cancel-in-progress: true
 
+# this command is called in two places, so save it in an env first
+env:
+  INSTALL_CMD: |
+    apt-get update --fix-missing
+    apt-get upgrade -y
+    apt-get install build-essential meson -y
+    apt-get install libsdl2-dev libsdl2-image-dev libsdl2-mixer-dev libsdl2-ttf-dev -y
+    apt-get install libfreetype6-dev libportmidi-dev fontconfig -y
+    apt-get install python3-dev python3-pip python3-wheel python3-sphinx -y
+    pip3 install meson-python --break-system-packages
+
 jobs:
   build-multiarch:
-    name: Debian (Bullseye - 11) [${{ matrix.arch }}]
+    name: Debian (Bookworm - 12) [${{ matrix.arch }}]
     runs-on: ubuntu-22.04
 
     strategy:
       fail-fast: false  # if a particular matrix build fails, don't skip the rest
       matrix:
         # maybe more things could be added in here in the future (if needed)
-        arch: [s390x, ppc64le]
+        include:
+          - { arch: s390x, base_image: '' }
+          - { arch: ppc64le, base_image: '' }
+          - { arch: armv6, base_image: '' }
+          # a custom base_image is specified in the armv7 case. This is done because
+          # the armv6 image is just raspbian in disguise. And the wheel built on armv7
+          # is going to be tested on armv6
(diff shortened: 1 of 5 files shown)


.github/workflows/build-debian-multiarch.yml
```diff
@@ -77,7 +77,7 @@ jobs:
         install: |
           apt-get update --fix-missing
           apt-get upgrade -y
-          apt-get install libsdl2-dev libsdl2-image-dev libsdl2-mixer-dev libsdl2-ttf-dev libfreetype6-dev libportmidi-dev libjpeg-dev fontconfig -y
+          apt-get install build-essential libsdl2-dev libsdl2-image-dev libsdl2-mixer-dev libsdl2-ttf-dev libfreetype6-dev libportmidi-dev fontconfig -y
           apt-get install python3-setuptools python3-dev python3-pip python3-wheel python3-sphinx -y
```
cclauss, line 81: A suggestion for readability...
```suggestion
          apt-get install build-essential libsdl2-dev libsdl2-image-dev libsdl2-mixer-dev \
                          libsdl2-ttf-dev libfreetype6-dev libportmidi-dev fontconfig python3-dev \
                         python3-pip python3-setuptools python3-sphinx python3-wheel -y
```

ankith26: Good suggestion, but I vaguely remember trying something like this in the past and it not working for mysterious reasons

ankith26:
This PR is now done from my side. I have edited the opening comment in this PR to describe the PR in more detail

cclauss (approved):

Starbuck5 (approved):
CI stuff is cool but a little over my head.

Meson change makes sense.

Code changes make sense.

After this PR does it mean people on Pi will get NEON acceleration by default?

ankith26:
> After this PR does it mean people on Pi will get NEON acceleration by default?

Basically whichever pi has NEON support in hardware will have NEON acceleration by default (so like all pi except pi 1 and older pi zero). And whichever pi doesn't, runtime detection that we have ensures that the NEON instructions won't be invoked (which will crash with illegal instruction error). It was a long standing speculation that this wouldn't work, but now this PR has tests that pass and therefore show that this approach works.

Starbuck5:
> Basically whichever pi has NEON support in hardware will have NEON acceleration by default (so like all pi except pi 1 and older pi zero). And whichever pi doesn't, runtime detection that we have ensures that the NEON instructions won't be invoked (which will crash with illegal instruction error). It was a long standing speculation that this wouldn't work, but now this PR has tests that pass and therefore show that this approach works.

Very nice!

This does sound like something it would be to good to verify with a tester after our next dev release though, just to be sure.

MyreMylar (approved):
OK, LGTM 👍

PR #2785, opened 2024-04-01
Add example for retro scaling

snowfruit:
Add file "retro_scaling.py" that shows how to render to a surface and scale it to achieve a retro look.

### DIFF
--- a/examples/retro_scaling.py
+++ b/examples/retro_scaling.py
@@ -52,11 +52,11 @@ def main():
         # Scale the canvas-surface.
         canvas = pygame.transform.scale_by(canvas, scale)
 
-        # Calculate the center the screen-surface.
+        # Calculate the center for the screen-surface.
         screen_center_x = math.floor(screen.get_width() / 2)
         screen_center_y = math.floor(screen.get_height() / 2)
 
-        # Calculate the center the canvas-surface.
+        # Calculate the center for the canvas-surface.
         canvas_center_x = math.floor(canvas.get_width() / 2)
         canvas_center_y = math.floor(canvas.get_height() / 2)
 
--- a/examples/retro_scaling.py
+++ b/examples/retro_scaling.py
@@ -1,0 +1,78 @@
+
+import math
+import pygame
+
+
+def new_canvas():
+    # Return a new blank canvas.
+    return pygame.Surface((320, 240))
+
+
+def main():
+    pygame.init()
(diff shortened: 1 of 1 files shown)


ankith26 (changes requested):
Hello, thanks for contributing! :partying_face: 

We use `black` for keeping our code formatted, please run `python3 setup.py format` and commit the resulting changes.

robertpfeiffer:
Hmm. As is, this duplicates existing functionality, but it could be useful as a "Rosetta Stone" if we add another example based on Window and Texture, to show how one would to this with SDL2 machinery.

snowfruit:
> Hello, thanks for contributing! 🥳
> 
> We use `black` for keeping our code formatted, please run `python3 setup.py format` and commit the resulting changes.

I thought I did. I will try again.

snowfruit:
> Hmm. As is, this duplicates existing functionality, but it could be useful as a "Rosetta Stone" if we add another example based on Window and Texture, to show how one would to this with SDL2 machinery.

Are you referring to pygame.SCALED?

snowfruit:
> Yeah, if you're up for it. As is, there are many use cases deliberately not covered by SCALED, and rather than having people poke around in the innards of the display module like @Matiiss likes to do (which could break at any time with the next release), it would make more sense to implement SCALED on top of pygame.Window. Maybe somebody wants to have the background a repeating pattern instead of black, or a nine-patch pattern, or just a different colour. Maybe somebody wants to use SCALED but really, really wants to control the size of the window. Something like that is probably within your abilities (you already did it once) and could be really useful.

I have no idea how that would be done. I don't have the skills for something like that. But my code here https://github.com/snowfruit/pygame-addons/tree/main/pygame_screen might spark some ideas for a person who knows how.

snowfruit:
> Hello, thanks for contributing! 🥳
> 
> We use `black` for keeping our code formatted, please run `python3 setup.py format` and commit the resulting changes.

I did the `python3 setup.py format` on the file. I used the web-interface and hope the file is in the right format now. I think the next step is to press the "Re-request review"-button so that's what I will do.

ankith26 (approved):

I think we could potentially expand this in the future to maybe become more interactive, or integrate it into a bigger example. For now I don't mind merging this change in its current state.

oddbookworm (approved):
LGTM

PR #2786, opened 2024-04-01
Optimized Rect multi-collision methods

itzpr3d4t0r:
This PR swaps a common rect-rect collision function with a simplified version that precalculates some conditions outside the core loop. A side-effect of this PR is also making O(N) cases O(1) where the base rect has 0 width or height.

While this isn't the last step of optimizations for these methods it's still a start and it's a fairly simple change. Future changes will involve using fast sequence looping.

This change affects the following functions:
- `collidelist`  (40% faster)
- `collidelistall` (20-50% faster)
- `collideobjects` ( 6-10% faster)
- `collideobjectsall` (6-8% faster)
- `collidedict` (17% faster)
- `collidedictall` (5-15% faster)

A graph containing all functions both with and without the change:
image

### DIFF
--- a/src_c/rect_impl.h
+++ b/src_c/rect_impl.h
@@ -1292,7 +1292,7 @@ RectExport_colliderect(RectObject *self, PyObject *const *args,
     return PyBool_FromLong(_pg_do_rects_intersect(&self->r, argrect));
 }
 
-#ifndef OPTIMIZED_COLLIDERECT(r)
+#ifndef OPTIMIZED_COLLIDERECT
 #define OPTIMIZED_COLLIDERECT(r)                                 \
     (a < MAX(r->x, r->x + r->w) && b < MAX(r->y, r->y + r->h) && \
      c > MIN(r->x, r->x + r->w) && d > MIN(r->y, r->y + r->h))
--- a/src_c/rect_impl.h
+++ b/src_c/rect_impl.h
@@ -1293,9 +1293,14 @@ RectExport_colliderect(RectObject *self, PyObject *const *args,
 }
 
 #ifndef OPTIMIZED_COLLIDERECT
-#define OPTIMIZED_COLLIDERECT(r)                                      \
-    (left < MAX(r->x, r->x + r->w) && top < MAX(r->y, r->y + r->h) && \
-     right > MIN(r->x, r->x + r->w) && bottom > MIN(r->y, r->y + r->h))
+/* This macro is used to optimize the colliderect function. Makes use of
+ * precalculated values to avoid unnecessary calculations. It also checks
+ * whether the other rect has 0 width or height, in which case we don't
+ * collide. */
+#define OPTIMIZED_COLLIDERECT(r)                                       \
+    (r->w && r->h && left < MAX(r->x, r->x + r->w) &&                  \
+     top < MAX(r->y, r->y + r->h) && right > MIN(r->x, r->x + r->w) && \
+     bottom > MIN(r->y, r->y + r->h))
 #endif
 
 static PyObject *
@@ -1331,11 +1336,6 @@ RectExport_collidelist(RectObject *self, PyObject *arg)
                     "Argument must be a sequence of rectstyle objects.");
             }
 
-            /* If 0 width or height, skip, preserves previous behaviour with
-             * _pg_do_rects_intersect */
-            if (argrect->w == 0 || argrect->h == 0)
-                continue;
-
             if (OPTIMIZED_COLLIDERECT(argrect)) {
                 return PyLong_FromLong(loop);
             }
@@ -1355,11 +1355,6 @@ RectExport_collidelist(RectObject *self, PyObject *arg)
             }
 
             Py_DECREF(obj);
-
-            /* If 0 width or height, skip, preserves previous behaviour with
-             * _pg_do_rects_intersect */
-            if (argrect->w == 0 || argrect->h == 0)
-                continue;
 
             if (OPTIMIZED_COLLIDERECT(argrect)) {
                 return PyLong_FromLong(loop);
@@ -1409,11 +1404,6 @@ RectExport_collidelistall(RectObject *self, PyObject *arg)
                     "Argument must be a sequence of rectstyle objects.");
             }
 
-            /* If 0 width or height, skip, preserves previous behaviour with
-             * _pg_do_rects_intersect */
-            if (!argrect->w || !argrect->h)
-                continue;
-
             if (OPTIMIZED_COLLIDERECT(argrect)) {
                 PyObject *num = PyLong_FromLong(loop);
                 if (!num) {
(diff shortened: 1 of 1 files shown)


src_c/rect_impl.h
```diff
@@ -1292,17 +1292,32 @@ RectExport_colliderect(RectObject *self, PyObject *const *args,
...
+
     if (!PySequence_Check(arg)) {
         return RAISE(PyExc_TypeError,
                      "Argument must be a sequence of rectstyle objects.");
     }
 
+    const PrimitiveType a = MIN(srect->x, srect->x + srect->w);
+    const PrimitiveType b = MIN(srect->y, srect->y + srect->h);
+    const PrimitiveType c = MAX(srect->x, srect->x + srect->w);
+    const PrimitiveType d = MAX(srect->y, srect->y + srect->h);
```
Starbuck5, line 1319: Can these have more descriptive names?

src_c/rect_impl.h
```diff
@@ -1292,17 +1292,32 @@ RectExport_colliderect(RectObject *self, PyObject *const *args,
     return PyBool_FromLong(_pg_do_rects_intersect(&self->r, argrect));
 }
 
+#ifndef OPTIMIZED_COLLIDERECT
+#define OPTIMIZED_COLLIDERECT(r)                                 \
```
Starbuck5, line 1296: Should the a/b/c/d variables be passed into this macro like r (rect)?

src_c/rect_impl.h
```diff
@@ -1400,7 +1437,11 @@ RectExport_collidelistall(RectObject *self, PyObject *arg)
 
             Py_DECREF(obj);
 
-            if (_pg_do_rects_intersect(srect, argrect)) {
+            if (argrect->w == 0 || argrect->h == 0) {
```
Starbuck5, line 1440: I think the early exiting makes sense for the one rect before the loop, but I'm not sure it is a good idea to add more branching into the hot part of the loop here and in the other functions.

src_c/rect_impl.h
```diff
@@ -1358,6 +1382,15 @@ RectExport_collidelistall(RectObject *self, PyObject *arg)
         return NULL;
     }
 
+    if (srect->w == 0 || srect->h == 0) {
+        return ret;
+    }
```
Starbuck5, line 1387: In all these early exit blocks can you leave a comment about why it is early exiting?

Starbuck5:
This is a clever idea for an optimization, nice!

Idea: would it compile better if the argrect 0 length check was embedded into the main boolean logic for whether a rect collided or not? This would be the same logic but it might be easier for the compiler to optimize?

src_c/rect_impl.h
itzpr3d4t0r: I would prefer to leave it a little cleaner and use only `r` as an argument for clarity and let the other necessary variables be inferred.

src_c/rect_impl.h
itzpr3d4t0r: Sure

src_c/rect_impl.h
itzpr3d4t0r: Yep that's the thing.

src_c/rect_impl.h
Starbuck5: I disagree that it's cleaner, but it doesn't matter much to me.

Starbuck5:
@itzpr3d4t0r what do you think about putting the argrect 0 check into the macro? It would reduce a lot of code repetition in this PR.

This is my only reservation about approving this PR, that it could be improved in this way. Otherwise it looks good.

itzpr3d4t0r:
> Idea: would it compile better if the argrect 0 length check was embedded into the main boolean logic for whether a rect collided or not? This would be the same logic but it might be easier for the compiler to optimize?

Thanks for the suggestion, I've tried this but I didn't see any speedup, infact it may have been a ltl slower even.

> what do you think about putting the argrect 0 check into the macro? It would reduce a lot of code repetition in this PR.

Do you have any idea on how? Because the macro rn is basically an inline bool expression that needs to be fed to an if statement, and each if statement has a different body depending on the function so I find it hard to imagine a way to implement what you're asking, only way would be to add another macro just for that but it would only save 2 lines of code.

Starbuck5:
Those two ideas you quoted were the same idea, but framed differently.

```c
#define OPTIMIZED_COLLIDERECT(r)                      \
    (r->w && r->h && left < MAX(r->x, r->x + r->w) && \
     top < MAX(r->y, r->y + r->h) &&                  \
     right > MIN(r->x, r->x + r->w) && \ bottom > MIN(r->y, r->y + r->h))
```
Right?

It would also need a comment explaining why in there. But that seems better to me than having a required bit of code outside the colliderect macro that's necessary for the macro to give correct results.

Starbuck5 (approved):
Looks good to me!

Nice clean speedup.

src_c/rect_impl.h
```diff
@@ -1292,17 +1292,39 @@ RectExport_colliderect(RectObject *self, PyObject *const *args,
...
+
     if (!PySequence_Check(arg)) {
         return RAISE(PyExc_TypeError,
                      "Argument must be a sequence of rectstyle objects.");
     }
 
+    const PrimitiveType left = MIN(srect->x, srect->x + srect->w);
+    const PrimitiveType top = MIN(srect->y, srect->y + srect->h);
+    const PrimitiveType right = MAX(srect->x, srect->x + srect->w);
+    const PrimitiveType bottom = MAX(srect->y, srect->y + srect->h);
```
ankith26, line 1326: This block could be a macro too IMO. something like `OPTIMIZED_COLLIDERECT_SETUP` so that it is clear that it must be called before doing `OPTIMIZED_COLLIDERECT` in code

src_c/rect_impl.h
```diff
@@ -1322,7 +1345,7 @@ RectExport_collidelist(RectObject *self, PyObject *arg)
      * PySequence_GetItem() function to get the items. */
     else {
         for (loop = 0; loop < PySequence_Length(arg); loop++) {
-            PyObject *obj = PySequence_GetItem(arg, loop);
+            PyObject *obj = PySequence_ITEM(arg, loop);
```
ankith26, line 1348: This has the potential to change behaviour as it is now skipping checking of sequence objects.
The thing is `PySequence_Length` probably checks for sequence anyways, so it should probably be first saved in a variable, checked for error and then used in the loop condition, and this will be good to go.

Starbuck5: There's a `PySequence_Check` on line 1318 above.

Although now that I'm looking at this I see the comment on 1345 is now out of date, since it's not using PySequence_GetItem anymore.

src_c/rect_impl.h
```diff
@@ -1322,7 +1345,7 @@ RectExport_collidelist(RectObject *self, PyObject *arg)
      * PySequence_GetItem() function to get the items. */
```
MyreMylar, line 1345:
```suggestion
     * PySequence_ITEM() function to get the items. */
```

MyreMylar (approved):
LGTM 👍

PR #2788, opened 2024-04-02
Remove all .editorconfig duplicates, re-add in project root with updates

ankith26:
~~I don't see a reason for having this file, now that we have `python3 setup.py format`. So I propose a removal in this PR.~~

Even if we do decide to retain this file, we could move it to project root, instead of duplicating it in 4 separate subtrees.

UPDATE: I have updated the file and moved it to project root


robertpfeiffer:
Way back when, somebody insisted on not adding any more dotfiles to the project root. Please have mercy with emacs users.

Starbuck5 (approved):
Thanks.

oddbookworm (approved):
LGTM

PR #2793, opened 2024-04-05
Simplified SIMD fill algorithms

itzpr3d4t0r:
This PR condenses our SIMD fill algorithms a bit by removing the use of two color registers (that were the same register effectively).

I measured no significant difference in performance.


MyreMylar (approved):
OK, LGTM. Seems like a small change to no harm. 👍

Starbuck5 (approved):
Alright, this looks good to me.

PR #2794, opened 2024-04-06
Remove docs non-public module stubs

ankith26:
I am making this PR as a result of the discussion from https://github.com/pygame-community/pygame-ce/pull/2557

@Starbuck5 pointed out that non-public API does not need type-stubs, which is a fair argument. So I'm making this PR to remove these stubs.

The current buildconfig ignores these stubs anyways, so this change can definitely not impact any end user.


MyreMylar (approved):
LGTM 👍  get that file size down!

MyreMylar:
looks like you need to remove from the stubcheck too though!

Starbuck5 (approved):

PR #2795, opened 2024-04-09
black format setup.py

cclauss:
https://github.com/psf/black/blob/main/README.md


ankith26 (commented):
The reason this file (and other stuff in buildconfig) were left unformatted is that we are moving away from it as you are probably aware of.

We don't expect to be making any non-trivial changes to these files anymore, so I see no point in formatting. That being said, I don't strongly oppose this PR.

cclauss:
Makes sense.  Closing.

PR #2796, opened 2024-04-09
Remove lint logic from setup.py

cclauss:
Followup to
* #2772 

Pylint is too slow to run in pre-commit but let's keep it running in GitHub Actions.
This could be sped up radically with `ruff check --select=PL` but that could be a future enhancement.

### DIFF
--- a/.github/workflows/format-lint.yml
+++ b/.github/workflows/format-lint.yml
@@ -2,7 +2,7 @@ name: python3 setup.py lint
 
 # Run lint CI on changes to main branch, or any PR to main. Do not run CI on
 # any other branch.
-# run only if there are changes on files that are linted (C, python and rst files)
+# Run only if there are changes on files that are linted (C, Python and rst files)
 on:
   push:
     branches: main
@@ -43,10 +43,9 @@ jobs:
       - uses: actions/checkout@v4.1.2
 
       - name: Install deps
-        run: python3 -m pip install pylint black clang-format sphinx"<7.2.0"
+        run: python3 -m pip install pylint sphinx"<7.2.0"
 
-      - name: Check code Formatting and Linting
-        run: python3 setup.py lint
+      - run: pylint src_py docs
 
       - name: Check docs changes are checked in
         run: |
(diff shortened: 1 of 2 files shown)


setup.py
```diff
@@ -273,7 +273,7 @@ def consume_arg(name):
     if cython_only:
         sys.exit(0)
 
-no_compilation = any(x in ['lint', 'format', 'docs'] for x in sys.argv)
+no_compilation = any(x in ('docs') for x in sys.argv)
```
ankith26, line 276:
```suggestion
no_compilation = 'docs' in sys.argv
```
simplified the expression

.github/workflows/format-lint.yml
```diff
@@ -43,10 +43,9 @@ jobs:
       - uses: actions/checkout@v4.1.2
 
       - name: Install deps
-        run: python3 -m pip install pylint black clang-format sphinx"<7.2.0"
+        run: python3 -m pip install pylint sphinx"<7.2.0"
 
-      - name: Check code Formatting and Linting
```
ankith26, line 48: This line doesn't need to removed, right?

Maybe just needs updating to say "Check code linting"

cclauss: This is personal taste but when the command is super-simple like `pylint src_py docs`, I prefer to see that instead of some imprecise name like `Check code linting` but let's make the proposed change.

oddbookworm:
> I'd like to give this PR a small wait, just so that the other devs can get onboard the pre-commit strategy and not have any issues with it.

FWIW I'm on board with pre-commit

oddbookworm (approved):
I'm on board with this

ankith26 (approved):
I will go ahead and approve this, but I'd like someone else to merge this just so that this change has more visibility.

Thanks for the PR :100:

MyreMylar (approved):
LGTM 👍 

I'm on board with this pre-commit strategy as well. I mean in principle I'm on board with it, when it inevitably catches me out later when I forget about it I reserve the right to complain petulantly.

Issue #2463, opened 2023-09-21
Add line thickness in gfxdraw


mzivic7:
**Description**

Add line thickness value in gfxdraw.aacircle and other gfxdraw shapes, just like in pygame.draw.

For now i am using this code for aacircle:

```
def draw_circle(surface, color, pos, radius, thickness):
	if thickness != 1:
		for num in range(1, thickness):
			gfxdraw.aacircle(surface, int(pos[0]), int(pos[1]), int(radius)+num, color) 
```
		
But it is not perfect: there can be seen background pixels between two lines, because gfxdraw accepts only int.

novialriptide:
Please correct me if I'm wrong, but I think we have stopped supporting gfxdraw.

I recommend using `pygame.draw.line()` instead.

https://pyga.me/docs/ref/draw.html#pygame.draw.line

ankith26:
The thing is we don't actively maintain `gfxdraw`, but because many people use it and it has some stuff that `pygame.draw` doesn't, `gfxdraw` isn't being removed anytime soon. The long term plan is to add all of `gfxdraw`s features to draw and then we could deprecate/remove gfxdraw

igordsm:
@ankith26   Would adding a thickness parameter to aacircle and change its implementation to use pygame.draw instead be a good solution for this? Perhaps adding a deprecation message would also be a good ideia.

oddbookworm:
I'm adding some tasks to the original message to close this issue when complete

Issue #2450, opened 2023-09-12
gfxdraw.aacircle weird lines when large radius


mzivic7:
**Environment:**

- **Operating system:** Linux(Arch), Windows
- **Python version:** 3.11.5
- **SDL version:** 2.26.5
- **PyGame version:** 2.3.2

**Current behavior:**

gfxdraw.aacircle draws weird lines when radius is larger than about 1000.

**Expected behavior:**

To draw nice and round anti-aliased circle.

**Screenshots:**

![Screenshot from 2023-09-12 13-38-24](https://github.com/pygame-community/pygame-ce/assets/92328096/7167ce73-96ff-4649-b38f-8dd0600c13ba)

**Test code:**

```
import pygame
from pygame import gfxdraw
pygame.init()
screen = pygame.display.set_mode((1000, 800))
run = True
while run:
    for e in pygame.event.get():
        if e.type == pygame.QUIT: run = False
    screen.fill((0, 0, 0))
    gfxdraw.aacircle(screen, -500, 1000, 1200, (255,255,255))   # radius = 1200
    pygame.display.flip()
pygame.quit()
```

Temmie3754:
Able to replicate this behaviour

oddbookworm:
As mentioned in #2463 `gfxdraw` is supposed to be deprecated eventually. With the merge of #2800 looking like it'll come soon, I think we can then consider this as resolved as it's going to get. I'll add a checkbox to the original message so that this can be closed when gfxdraw is formally deprecated

PR #2800, opened 2024-04-14
draw.aacircle()

mzivic7:
Pygame-ce has only `gfxdraw.aacircle` function for drawing antialiased circles, but it has several problems:
- It's arguments are different from draw.circle
- It does not accept float as input
- It can't draw circle quadrants
- There is no line thickness option
   Related issue: #2463
- Draws incorrect lines when radius is larger than about 1000
   Related issue: #2450

and it is intended to be deprecated/removed in future.

All these problems are fixed in `draw.aacircle`, that is implemented in this commit.

Changes:
- Added `draw.aacircle` function that has same inputs and outputs as `draw.circle`
- Added algorithms for drawing thin and thick antialiased circles, adapted from:
https://cgg.mff.cuni.cz/~pepca/ref/WU.pdf
- Added to documentation
- Added to code_examples
- Added to tests


oddbookworm:
Looks like you need to run `python setup.py format`

buildconfig/stubs/pygame/draw.pyi
```diff
@@ -31,6 +31,17 @@ def circle(
...
+    surface: Surface,
+    color: ColorValue,
+    center: Coordinate,
+    radius: float,
+    width: int = 0,
+    draw_top_right: bool = False,
+    draw_top_left: bool = False,
+    draw_bottom_left: bool = False,
+    draw_bottom_right: bool = False,
+) -> Rect: ...
```
oddbookworm, line 44: I think it would actually be better to have two signatures here. One without the bools, and one with

src_c/draw.c
```diff
@@ -735,6 +743,123 @@ circle(PyObject *self, PyObject *args, PyObject *kwargs)
...
+static PyObject *
+aacircle(PyObject *self, PyObject *args, PyObject *kwargs)
+{
+    pgSurfaceObject *surfobj;
+    PyObject *colorobj;
+    SDL_Surface *surf = NULL;
+    Uint32 color;
+    SDL_Rect cliprect;
+    PyObject *posobj, *radiusobj;
+    int posx, posy, radius;
```
oddbookworm, line 755: Here you list radius as an int, but in the stub you put float

src_c/draw.c
```diff
@@ -2345,6 +2470,120 @@ draw_circle_filled(SDL_Surface *surf, int x0, int y0, int radius, Uint32 color,
...
 }
 
+static void
+eight_pixels(SDL_Surface *surf, int x0, int y0, Uint32 color, int x, int y, 
+             double opacity, int top_right, int top_left, int bottom_left, 
+             int bottom_right, int *drawn_area) {
+    opacity = opacity / 255;
+    Uint32 pixel_color;
+    if (top_right == 1) {
+        pixel_color = get_antialiased_color(surf, x0 + x, y0 - y, color, opacity);
```
oddbookworm, line 2480: CI is complaining about an implicit conversion from double to float here, either use a float for opacity or explicitly cast. I would be careful casting from double to float though because of the chance that the original might be outside the range of the destination type without validation

src_c/draw.c
```diff
@@ -2345,6 +2470,120 @@ draw_circle_filled(SDL_Surface *surf, int x0, int y0, int radius, Uint32 color,
     }
 }
 
+static void
+eight_pixels(SDL_Surface *surf, int x0, int y0, Uint32 color, int x, int y,
```
oddbookworm, line 2474: What does `eight_pixels` mean? I recommend making the name more clear

src_c/draw.c
```diff
@@ -2345,6 +2470,120 @@ draw_circle_filled(SDL_Surface *surf, int x0, int y0, int radius, Uint32 color,
...
+                double opacity = 255 * (ceil(height) - height);
+                if (opacity < prev_opacity) {
+                    y--;
+                }
+                prev_opacity = opacity;
+                eight_pixels(surf, x0, y0, color, x, y, 255,
+                             top_right, top_left, bottom_left, bottom_right, drawn_area);
+                eight_pixels(surf, x0, y0, color, x, y - 1, opacity,
+                             top_right, top_left, bottom_left, bottom_right, drawn_area);
+                x++;
```
oddbookworm, line 2530: Some people might complain about post-increment/decrement being used instead of pre-increment/decrement due to the post version making a copy yadda yadda yadda. I don't personally think it matters because of modern compiler optimization, but feel free to performance test it lol

test/draw_test.py
```diff
@@ -5739,6 +5740,765 @@ class DrawCircleTest(DrawCircleMixin, DrawTestCase):
     """
 
 
+### AACircle Testing ############################################################
```
oddbookworm, line 5743: nitpick:
all the method names and the docstrings used `circle` and not `aacircle`

oddbookworm (commented):
Just a few comments, looking good so far!

src_c/draw.c
mzivic7: but draw.circle does exactly same thing, this is because algorithms must use integer values for position and radius, but function still accepts floats

mzivic7: `pg_IntFromObj` [here](https://github.com/pygame-community/pygame-ce/pull/2800/files/fb9ea0e7cb43297701a2fd4cb129f938cb9836eb#diff-d40ef6601363eeb9acdc38f0d143fc5a577b6c98f702eb13f68116af0b1ad7d6R783) converts python float object to integer and  stores it in radius variable.

buildconfig/stubs/pygame/draw.pyi
```diff
@@ -31,6 +31,24 @@ def circle(
...
     draw_bottom_right: bool = False,
 ) -> Rect: ...
+def aacircle(
+    surface: Surface,
+    color: ColorValue,
+    center: Coordinate,
+    radius: float,
+    width: int = 0,
+) -> Rect: ...
+def aacircle(
```
oddbookworm, line 41: I suspect that your stub check is failing due to the lack of `@overload` from `typing`

[docs](https://docs.python.org/3/library/typing.html#overload)

src_c/draw.c
```diff
@@ -2514,48 +2528,56 @@ draw_circle_xaolinwu(SDL_Surface *surf, int x0, int y0, int radius,
...
+                    surf, x0, y0, color, x - 1, y - 1, opacity, top_right,
+                    top_left, bottom_left, bottom_right, drawn_area);
             }
-        } else if (radius_ == radius) {
+        }
+        else if (radius_ == radius) {
             while (x < y) {
-                double height = sqrt(pow(radius_, 2) - pow(x, 2));
-                double opacity = 255 * (ceil(height) - height);
+                x++;
```
oddbookworm, line 2551: you changed where this is from my last review, is there a reason for that?

src_c/SDL_gfx/SDL_gfxPrimitives.c
```diff
@@ -4226,8 +4226,7 @@ lrint(double flt)
 #include <armintr.h>
 #pragma warning(push)
 #pragma warning(disable : 4716)
-__declspec(naked) long int
-lrint(double flt)
+__declspec(naked) long int lrint(double flt)
```
oddbookworm, line 4229: what version of `clang-format` do you have installed in your environment? There are a bunch of files here that you didn't modify, but got formatted. I want to make sure that we're using the correct version

oddbookworm (changes requested):

src_c/SDL_gfx/SDL_gfxPrimitives.c
mzivic7: clang-format version 17.0.6
clang is up-to-date with Arch repos

src_c/draw.c
mzivic7: I changed from post-increment to pre-increment, as you said [here](https://github.com/pygame-community/pygame-ce/pull/2800#discussion_r1564392476)

oddbookworm: Ah, I see the misunderstanding. By pre-increment, I meant `++x` instead of `x++` (post-increment)

src_c/SDL_gfx/SDL_gfxPrimitives.c
mzivic7: black was missing

mzivic7: I installed clang-format 18.1.3 with pipenv, its all good now

buildconfig/stubs/pygame/draw.pyi
Starbuck5: Why? circle above just has one signature.

src_c/draw.c
```diff
@@ -735,6 +743,127 @@ circle(PyObject *self, PyObject *args, PyObject *kwargs)
...
+    if (!pg_TwoIntsFromObj(posobj, &posx, &posy)) {
+        PyErr_SetString(PyExc_TypeError,
+                        "center argument must be a pair of numbers");
+        return 0;
+    }
+
+    if (!pg_IntFromObj(radiusobj, &radius)) {
+        PyErr_SetString(PyExc_TypeError, "radius argument must be a number");
+        return 0;
+    }
```
Starbuck5, line 786: Check out the `RAISE` macro we use for a lot of these"set error and return" cases.

src_c/draw.c
```diff
@@ -2590,20 +2588,20 @@ draw_circle_xaolinwu_thin(SDL_Surface *surf, int x0, int y0, int radius,
...
         }
         prev_opacity = opacity;
-        draw_eight_symetric_pixels(surf, x0, y0, color, x, y, 255 - opacity,
-                                   top_right, top_left, bottom_left,
-                                   bottom_right, drawn_area);
-        draw_eight_symetric_pixels(surf, x0, y0, color, x, y - 1, opacity,
-                                   top_right, top_left, bottom_left,
-                                   bottom_right, drawn_area);
+        draw_eight_symetric_pixels(surf, x0, y0, color, x, y,
+                                   255 - (float)opacity, top_right, top_left,
```
oddbookworm, line 2600:
```suggestion
                                   255.0f - (float)opacity, top_right, top_left,
```

oddbookworm: The reason I suggest that is to be more explicit that we're subtracting floats specifically

src_c/draw.c
```diff
@@ -2590,20 +2588,20 @@ draw_circle_xaolinwu_thin(SDL_Surface *surf, int x0, int y0, int radius,
 {
     int x = 0;
     int y = radius;
-    float prev_opacity = 0;
+    double prev_opacity = 0;
```
oddbookworm, line 2591:
```suggestion
    double prev_opacity = 0.0;
```

oddbookworm (changes requested):

src_c/draw.c
mzivic7: should I do the same for multiplications? 
I thought that in C basic math operations, when given an operand of type float and int, the int is converted to float first.

src_c/draw.c
mzivic7: should I do the same for multiplications?
I thought that in C basic math operations, when given an operand of type float and int, the int is converted to float first.

oddbookworm: I don't like implicit conversions in C. I prefer to be explicit whenever possible, especially when I'm mixing floating point and integral types

src_c/draw.c
```diff
@@ -2345,6 +2472,141 @@ draw_circle_filled(SDL_Surface *surf, int x0, int y0, int radius, Uint32 color,
...
+draw_circle_xaolinwu_thin(SDL_Surface *surf, int x0, int y0, int radius,
+                          Uint32 color, int top_right, int top_left,
+                          int bottom_left, int bottom_right, int *drawn_area)
+{
+    int x = 0;
+    int y = radius;
+    double prev_opacity = 0.0;
+    while (x < y) {
+        double height = sqrt(pow(radius, 2) - pow(x, 2));
+        double opacity = 255.0f * (ceil(height) - height);
```
oddbookworm, line 2595: nitpick: `255.0f` is a 32-bit float, not a 64-bit double. C would interpret `255.0` as a double

src_c/draw.c
```diff
@@ -2345,6 +2472,141 @@ draw_circle_filled(SDL_Surface *surf, int x0, int y0, int radius, Uint32 color,
...
+/* Xaolin Wu Circle Algorithm
+ * adapted from: https://cgg.mff.cuni.cz/~pepca/ref/WU.pdf
+ * with additional line width parameter and quadrants option
+ */
+static void
+draw_circle_xaolinwu(SDL_Surface *surf, int x0, int y0, int radius,
+                     int thickness, Uint32 color, int top_right, int top_left,
+                     int bottom_left, int bottom_right, int *drawn_area)
+{
+    for (int radius_ = radius - thickness; radius_ <= radius; radius_++) {
```
oddbookworm, line 2526: I think maybe a better name for `radius_` would be preferred because it's too close to `radius`

oddbookworm (approved):
I got a couple small comments, but this looks good!

MyreMylar (approved):
This looks really good to me! 👍 

Will enable us to shim the old gfxdraw version completely at some point and has features on top of that that make it consistent with draw.circle.

No notes! 🥇

PR #2801, opened 2024-04-14
Unify freetype init codepaths

Starbuck5:
The freetype.Font init code has a windows codepath and an "everything else" codepath. I think that the windows codepath would work everywhere, so let's try using it everywhere.

Why am I mucking around with this? Because SDL3 changes rwops stuff and freetype looks like it might be the most complex use of rwops we have. If we can make it simpler that will help the transition. This is similar to #2717 in that regard.

### DIFF
--- a/src_c/_freetype.c
+++ b/src_c/_freetype.c
@@ -748,84 +748,6 @@ _ftfont_init(pgFontObject *self, PyObject *args, PyObject *kwds)
         }
     }
 
-#if !defined(WIN32)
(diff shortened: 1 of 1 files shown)


ankith26 (approved):
Always a little scary to remove old and less-understood code, but I don't see a reason the windows path should break in the non-windows.

Thanks for the PR, this is a nice cleanup! :tada:

oddbookworm (approved):
Like Ankith, I don't see any reason this shouldn't work.

PR #2802, opened 2024-04-14
(Freetype) use calloc where possible

Starbuck5:
I was poking around freetype and saw some malloc+memset pairs, which could instead be expressed as a calloc.

This is slightly simpler code to do, and might be faster. *Not that perf really matters for these calls, they are infrequent.


ankith26 (approved):

MyreMylar (approved):
LGTM 👍

PR #2803, opened 2024-04-14
Fix sdist to include meson stuff and minor updates

ankith26:
This PR ensures three files are bundled in our sdist wheels: `pyproject.toml`, `meson.build` and `meson_options.txt`.

With this change, anyone attempting an sdist install will be making use of the new buildconfig by default (though the old one can still be used with some flags or workarounds)

I also added excludes of files unneeded in sdist builds: pyc files, mypy cache files and mac ci files.


MyreMylar (approved):
LGTM 👍

MyreMylar:
hmm though it looks like this is failing to build on Ubuntu 22.04.... probably in the meson build somewhere but we don't seem to get any logs for that just the ending.

ankith26:
Fixed the issue by adding a pip upgrade

Starbuck5 (approved):

PR #2804, opened 2024-04-14
Use macros for (Bytes|Bits)PerPixel compat in SDL3

Starbuck5:
This doesn't fix every read/write of these struct members, but it gets 99% of them. And this is hard to test (SDL3wise) without everything else compiling on SDL3, which it doesn't without a lot more patches.

For SDL2, this doesn't change how it compiles at all.


ankith26 (approved):
LGTM, thankies :+1:

MyreMylar (approved):
LGTM 👍

Issue #2799, opened 2024-04-12
pygame.mouse.set_pos() is not working under wayland


mzivic7:
**Environment:**
- **Operating system**: Arch Linux x86_64 (wayland)
- **Python version**: 3.11.8
- **SDL version**: SDL 2.28.5
- **pygame-ce version**: 2.4.1 (through pipenv)
- **Also tried**: pygame-ce 2.5.0.dev1 (SDL 2.30.2, Python 3.11.8) (through pipenv)

**Issue:**

pygame.mouse.set_pos() is not working under wayland. Yes, wayland does not allow applications to set cursor position, but application ran through xwayland can (as stated [here](https://github.com/py-sdl/py-sdl2/blob/master/sdl2/ext/mouse.py#L179)).

As a proof there is blender, it also uses system SDL2 (2.30.2), [here](<https://github.com/blender/blender/blob/main/extern/sdlew/include/SDL2/SDL_mouse.h#L40>) is code for setting mouse position in blender source.

And [here](https://github.com/pygame-community/pygame-ce/blob/main/src_c/mouse.c#L61) is pygame-ce (SDL 2.28.5) code for same thing. 
When i run pygame it says that it "seems to be running through X11 on top of wayland", which I assume is xwayland.

Looking at xwayland source, it indeed supports cursor warping, [here](https://gitlab.freedesktop.org/xorg/xserver/-/blob/xwayland-21.1/hw/xwayland/xwayland-cursor.c#L410).

To check if this is pipenv problem, I installed python-pygame 2.5.2 (that uses system SDL 2.30.2) package through pacman, because python is externally managed on Arch and pygame-ce is not available on pacman (even on AUR). 
pygame.mouse.set_pos() does not work here too, so pipenv is not causing problems.

Same issue is happening under weston running with `--xwayland` flag.

I built pygame-ce from source (with missing porttime if that matters).
Now its `pygame-ce 2.5.0.dev1 (SDL 2.30.2, Python 3.11.8)` still in pipenv.
No more reports about running through X11, but pygame.mouse.set_pos() is not working here too.

`wayland-debug -r pipenv run python mouse_test.py` shows no output (mouse_test.py is code that has pygame.mouse.set_pos() in it).
Application ran through wayland would show huge amount of wayland events. 
This is suggesting that pygame is still running through xwayland.

Running same code under just `weston` opens new window outside simulated area, but `weston --xwayland` opens window inside that area, which confirms that pygame is running through xwayland.

[...]

PR #2807, opened 2024-04-15
Fixed mouse.set_pos under wayland

mzivic7:
I found the reason why mouse.set_pos() is not working under wayland (xwyland), in one of blender [bug reports](https://projects.blender.org/blender/blender/issues/53004#issuecomment-551561):
> The issue is that the cursor cannot be warped as long as it is visible. The way things work currently is, if a cursor is hidden and it is warped, we lie to X11 clients and say it is actually warped. While doing so, we lock the pointer via the available Wayland protocols. When the cursor is made visible or that for some other reason the lock is broken (for example the user Alt-tab:ed away), we stop lying and where the visible cursor is displayed is the position we tell the clients about. This works for the majority of use cases we have run into (including Blender uses, excluding the one discussed in this report).
We handle confinement in a more direct way.
In other words, if the patch implements a behavior where the pointer cursor is warped while visible, it wont work. If the pointer cursor is confined using X11 protocol, or hidden then warped, and it doesn't work, it could very much be an Xwayland issue.

I tried this:
```py
pygame.mouse.set_visible(False)
pygame.mouse.set_pos(300, 300)
pygame.mouse.set_visible(True)
```
and it worked!

So, I just added `PG_HideCursor();` before and `PG_ShowCursor();` after [this](https://github.com/pygame-community/pygame-ce/blob/main/src_c/mouse.c#L61) line.

### DIFF
--- a/src_c/SDL_gfx/SDL_gfxPrimitives.c
+++ b/src_c/SDL_gfx/SDL_gfxPrimitives.c
@@ -4226,7 +4226,8 @@ lrint(double flt)
 #include <armintr.h>
 #pragma warning(push)
 #pragma warning(disable : 4716)
-__declspec(naked) long int lrint(double flt)
+__declspec(naked) long int
+lrint(double flt)
 {
     __emit(0xEC410B10);  // fmdrr  d0, r0, r1
     __emit(0xEEBD0B40);  // ftosid s0, d0
--- a/src_c/SDL_gfx/SDL_gfxPrimitives.c
+++ b/src_c/SDL_gfx/SDL_gfxPrimitives.c
@@ -4226,8 +4226,7 @@ lrint(double flt)
 #include <armintr.h>
 #pragma warning(push)
 #pragma warning(disable : 4716)
-__declspec(naked) long int
-lrint(double flt)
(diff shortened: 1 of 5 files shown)


mzivic7:
Closes #2799

mzivic7:
I  reported issue upstream to SDL: [here](https://github.com/libsdl-org/SDL/issues/9539)

Starbuck5:
So SDL has fixed it, but only for SDL3. Great research and effort @mzivic7!

I just asked on your issue there whether they think it's worth back-porting to SDL2. If they do back-port, I say we let them handle this hack instead of putting it into our own code.

If they don't want to back-port, we'll have to think further about how much we care about this. One thing for sure though, if we put this into our code I want comments in the codebase explaining exactly why.

ankith26:
Update: the SDL devs have backported the fix to SDL2

Thanks @mzivic7 for reporting the issue upstream and suggesting the fix. Now to get this fix on our side we just gotta bump SDL when they release the next release.

mzivic7:
Great, I will close this pr and issue, since this is no longer related to pygame-ce code.

PR #2813, opened 2024-04-20
Surface properties

### LINKED ISSUE #2758: Idea: start adding a small number of properties to Surface
The `Surface` class has many getters and setters. Some of them are very commonly used and I feel that their verbosity can be cumbersome.

I'm thinking of
```py
.get_size()
.get_width()
.get_height()
```

I think it would quite elegant to be able to do
```py
.size
.width
.height
```

These would be read-only properties, I don't think it makes sense to try implement scaling when these are changed.

Why am I thinking of these and not other getters/setters? These are the good starting ones in my opinion, because they're common and seem pretty unambiguous.

What do other people think?

ScriptLineStudios:
Closes #2758


test/surface_test.py
```diff
@@ -674,6 +674,22 @@ def test_get_width__size_and_height(self):
...
+    def test_attributes(self):
+        """Test width, height, and size attributes of surface"""
+        s = pygame.Surface((100, 50))
+        self.assertEqual(s.w, 100)
+        self.assertEqual(s.width, 100)
+        self.assertEqual(s.h, 50)
+        self.assertEqual(s.height, 50)
+        self.assertEqual(s.size, (100, 50))
+
+        with self.assertRaises(AttributeError):
```
Starbuck5, line 686: I think this will test whether at least one of them raises an AttributeError. It would be better to test if all of them raise an attribute error.

docs/reST/ref/surface.rst
```diff
@@ -87,6 +87,14 @@
    entire Surface. If it is changed, all drawing operations will only effect
    the smaller area.
 
+   Surfaces contain the following read-only attributes:
```
Starbuck5, line 90: This should specify a version added somehow.

src_c/surface.c
Starbuck5, line 1: What a simple implementation! Nice!

ankith26 (commented):
Interesting idea!

This PR is adding both `.w` and `.width` and `.h` and `.height`. Is this actually required? Is it being done for consistency with rect? If we add this we will end up having like atleast 3 ways of doing the exact same thing, which isn't very ideal IMHO

ScriptLineStudios:
> Interesting idea!
> 
> This PR is adding both `.w` and `.width` and `.h` and `.height`. Is this actually required? Is it being done for consistency with rect? If we add this we will end up having like atleast 3 ways of doing the exact same thing, which isn't very ideal IMHO

Yeah I also felt it was a little redundant, but yes did it for consistency with Rect.

test/surface_test.py
ScriptLineStudios: Done!

docs/reST/ref/surface.rst
ScriptLineStudios: Done!

Starbuck5:
Let's just do `width`, `height`, and `size` for now. (Removing `w` and `h`).

I think w and h would be fine to have, but in the interest of making this PR easy to merge let's remove them for now.

They can always be added in the future, but once they're added it's quite hard to remove.

itzpr3d4t0r (approved):
Agree with previous suggestions, that aside LGTM!

ScriptLineStudios:
> Let's just do `width`, `height`, and `size` for now. (Removing `w` and `h`).
> 
> I think w and h would be fine to have, but in the interest of making this PR easy to merge let's remove them for now.
> 
> They can always be added in the future, but once they're added it's quite hard to remove.

Sounds good!

buildconfig/stubs/pygame/surface.pyi
```diff
@@ -42,6 +42,16 @@ _ViewKind = Literal[
...
+    @property
+    def w(self) -> int: ...
+    @property
+    def width(self) -> int: ...
+    @property
+    def h(self) -> int: ...
+    @property
+    def height(self) -> int: ...
+    @property
+    def size(self) -> Tuple[int, int]: ...
```
itzpr3d4t0r, line 54: Still missing removal of "w" and "h" here

docs/reST/ref/surface.rst
```diff
@@ -87,6 +87,15 @@
...
+
+   ::
+      
+      width, height
+      size
+
+   .. versionadded:: 2.5.0
```
ankith26, line 97: I would suggest putting these under their own `.. attribute` and versionadded under that.

Starbuck5: I think it's fine to keep them low key like this, if it's a problem in the future they can get added properly.

This is consistent with the Rect docs.

Starbuck5: @ScriptLineStudios ?

ankith26: The problem with this though, is that it looks as though the `versionadded` applies to the whole class, and not the attributes, and we don't want this confusion

novialriptide (approved):
I think everything here is good.

Starbuck5 (approved):
Andrew and I added some commits to get this fully up and good w/ Ankith's feedback, I'm still happy with it.

docs/reST/ref/surface.rst
ankith26: The new additions LGTM, but this block is still here and can be removed now

ankith26 (approved):
LGTM. This is a nice simple change that should make things more convenient for users, thanks! :cake:

PR #2816, opened 2024-04-22
Add pygame.display.get/set_window_position()

Damus666:
I implemented what is said in the title of the pull request plus tests, stubs and documentation, I tested locally the functions.

Why?
- The window API should not have this feature restricted to it, without it seems incomplete
- The SDL code to achieve this is minimal so I see no reason to exclude this functionality
- The environment variable won't allow position changes after initialization and is nor pythonic nor clear
- The window API currently doesn't support OpenGL among the many features, so adding this feature to the display API is important
- Really, why not? It's not bloating nor complicating rather implementing existing and important functions provided by SDL

I'm very open to feedback and critics, I got inspired by the rest of `display.c` to write the implementation :)
Sorry for the double pull request, I messed up the previous branch, the current was forked directly from pygame-community/pygame-ce:main


src_c/display.c
```diff
@@ -1447,6 +1447,32 @@ pg_window_size(PyObject *self, PyObject *_null)
...
+    return Py_BuildValue("(ii)", x, y);
+}
+
+static PyObject *
+pg_set_window_pos(PyObject *self, PyObject *arg)
+{
+    SDL_Window *win = pg_GetDefaultWindow();
+    int x, y = 0;
+
+    if (!PyArg_ParseTuple(arg, "i|i", &x, &y))
```
oddbookworm, line 1467: Why is the y coordinate optional?

buildconfig/stubs/pygame/display.pyi
```diff
@@ -82,6 +82,8 @@ def get_caption() -> Tuple[str, str]: ...
 def set_palette(palette: Sequence[ColorValue], /) -> None: ...
 def get_num_displays() -> int: ...
 def get_window_size() -> Tuple[int, int]: ...
+def get_window_pos() -> Tuple[int, int]:...
+def set_window_pos(x: int, y: int) -> None:...
```
oddbookworm, line 86: I think this should be generalized to also accept any sequence of two integers as a single argument

src_c/display.c
```diff
@@ -1447,6 +1447,32 @@ pg_window_size(PyObject *self, PyObject *_null)
...
+        return NULL;
+
+    if (win)
+        SDL_SetWindowPosition(win, x, y);
```
oddbookworm, line 1471: In SDL3, this will return something based on whether it was successful or not ([SDL3 docs](https://wiki.libsdl.org/SDL3/SDL_SetWindowPosition)). @Starbuck5 @ankith26 @MyreMylar what do y'all think about this? Should there be some version checks built into new code that can support SDL3 more directly in the future, or should that be added in with other SDL3-compat stuff?

The same also applies to `SDL_GetWindowPosition`

docs/reST/ref/display.rst
```diff
@@ -715,6 +715,28 @@ required).
...
+
+   | :sl:`Set the current window position`
+   | :sg:`set_window_pos(int, int) -> None`
+
+   Sets the position of the window initialized with :func:`pygame.display.set_mode()`.
+   This differs from updating environment variables as this function can be called after the display has been initialised.
+   The user will still be able to move the window after this call. See also :func:`pygame.display.get_window_pos()`.
+
+   .. ## pygame.display.set_window_pos ##
+
```
oddbookworm, line 739: What about multi-monitor setups? Is that relative to the primary monitor? Relative to the top left corner of the leftmost monitor?

oddbookworm (changes requested):

src_c/display.c
Damus666: I copied the parse tuple from set_caption, without realizing in there the second argument was optional!

docs/reST/ref/display.rst
Damus666: I already noticed a way to improve the docs, and the position for me is relative to my only monitor, how can I test it with multiple monitors if I don't have them?

Damus666:
Would re-running checks magically fix the Pypy exception (which was not raised by code edited by me)? I remember that happening with another pull request of mine. (how would I do that?)

oddbookworm:
> Would re-running checks magically fix the Pypy exception (which was not raised by code edited by me)? I remember that happening with another pull request of mine. (how would I do that?)

Potentially, a rerun would have worked. Sometimes the fails are intermittent and unrelated to the changes. A member can rerun (most) failed checks (I haven't figured out how to rerun circleci yet). You can force rerun all CI by closing and reopening the PR, but I don't like doing that if I don't have to

docs/reST/ref/display.rst
oddbookworm: I'll test it for you

Damus666: Thank you very much :)

oddbookworm: On windows 11, looks like it's relative to the primary monitor's top-left corner. On my setup, that's my middle monitor, and moving the window to my left monitor gives me `(-1512, 397)`. There's some weirdness when the window crosses monitor boundaries (at least for me, where I have mismatched resolutions and framerates between monitors), but it seems to be working fine otherwise

Damus666: Okay, thank you!

docs/reST/ref/display.rst
```diff
@@ -715,6 +715,31 @@ required).
...
 
+.. function:: get_window_pos
+
+   | :sl:`Return the position of the window or screen`
+   | :sg:`get_window_pos() -> tuple`
+
+   Returns the position of the window initialized with :func:`pygame.display.set_mode()`.
+   The position will change when the user moves the window or when the position is set manually with :func:`pygame.display.set_window_pos()`.
+   Coordinates could be negative or outside of the desktop size bounds.
+   The position is relative to the topleft of the monitor the window is contained in and the y coordinate ignores the window frame.
```
oddbookworm, line 726: Primary monitor*

oddbookworm (changes requested):

docs/reST/ref/display.rst
Damus666: I just pushed a commit about that, those docs were created before your tests :)

oddbookworm (approved):
LGTM! Tested locally on windows 11, and it works.

buildconfig/stubs/pygame/display.pyi
```diff
@@ -82,6 +82,8 @@ def get_caption() -> Tuple[str, str]: ...
 def set_palette(palette: Sequence[ColorValue], /) -> None: ...
 def get_num_displays() -> int: ...
 def get_window_size() -> Tuple[int, int]: ...
+def get_window_pos() -> Tuple[int, int]:...
+def set_window_pos(pos: Sequence[int]) -> None:...
```
ankith26, line 86:
```suggestion
def set_window_pos(pos: Coordinate) -> None:...
```
typing it as a `Coordinate` is more generic and accurate

docs/reST/ref/display.rst
```diff
@@ -715,6 +715,31 @@ required).
...
+.. function:: set_window_pos
+
+   | :sl:`Set the current window position`
+   | :sg:`set_window_pos((x, y)) -> None`
+
+   Sets the position of the window initialized with :func:`pygame.display.set_mode()`.
+   This differs from updating environment variables as this function can be called after the display has been initialised.
+   The position is expected to be relative to the topleft of the primary monitor.
+   The y coordinate will ignore the window frame (y = 0 means the frame is hidden).
+   The user will still be able to move the window after this call. See also :func:`pygame.display.get_window_pos()`.
```
ankith26, line 739: I'd like to have it noted here that there is a chance that this function can just fail on some platforms/configurations (example: direct wayland on linux)
So we shouldn't encourage users to rely on this function.

src_c/display.c
ankith26: I think we should just put a comment here for now saying that this function will start raising errors, and deal with it during the porting

ankith26 (commented):
I'm not strongly in favour or strongly against this PR, but I have left a few reviews for your consideration.

Damus666:
@ankith26 I know it makes no sense, but I could be against my pull request. I made it because I knew the display API was gonna stay and I didn't understand why it was missing those functions. But then we started talking quite a lot about the window module.

From what I understood, the window API will be THE way of making the window. The display module will *only* contain monitor/screen related functions, meaning that `set_mode`, `get_window_size`, `get_surface` (I might be missing something) will be deprecated and only available with the Window object.
Feel free to correct me if what I understood is not planned, but *if* it was, having `get/set_window_pos` added and then *immediately* deprecated doesn't feel like a good move.

I use this comment as an occasion to ask you what's the future of the display/window API. If display will always remain as an option to create a singleton window, then I'm strongly in favor of my pull request (who would have guessed) but otherwise, then we should finish the window API and deprecate display functions related to the window.

NOTE: I'm still gonna modify it like you suggested as it makes a lot of sense, but I think this are important considerations.

ankith26:
> I use this comment as an occasion to ask you what's the future of the display/window API. If display will always remain as an option to create a singleton window, then I'm strongly in favor of my pull request (who would have guessed) but otherwise, then we should finish the window API and deprecate display functions related to the window.

The display module will probably never be removed. Almost all pygame code written relies on it. What we will most likely do once the window API is finalized and published, is recommend users to write any new code in it. The old code that uses display will continue to work.

Damus666:
> The display module will probably never be removed. Almost all pygame code written relies on it. What we will most likely do once the window API is finalized and published, is recommend users to write any new code in it. The old code that uses display will continue to work.

Thanks for the information.

docs/reST/ref/display.rst
Damus666: I saw that the similar functions in window.c call the same SDL functions, doesn't that mean that they could also fail on those platforms/configurations?

buildconfig/stubs/pygame/display.pyi
```diff
@@ -82,6 +82,8 @@ def get_caption() -> Tuple[str, str]: ...
 def set_palette(palette: Sequence[ColorValue], /) -> None: ...
 def get_num_displays() -> int: ...
 def get_window_size() -> Tuple[int, int]: ...
+def get_window_pos() -> Tuple[int, int]:...
```
novialriptide, line 85: Why `window_pos` instead `window_position`?

Damus666: When I wrote it I gave it that name because it was shorter but Window has the position attribute and not pos so it makes sense to have position. I'll edit this when I'm back in Italy in 14 hours

src_c/display.c
```diff
@@ -1447,6 +1447,39 @@ pg_window_size(PyObject *self, PyObject *_null)
...
+    SDL_Window *win = pg_GetDefaultWindow();
+    PyObject *pos = NULL;
+    int x, y = 0;
+
+    if (!PyArg_ParseTuple(arg, "O", &pos))
+        return NULL;
+
+    if (pos != NULL) {
+        if (!pg_TwoIntsFromObj(pos, &x, &y))
+            return RAISE(PyExc_TypeError, "pos must be two numbers");
```
novialriptide, line 1473: Please change this as well.

Damus666: My bad I didn't see it thanks

oddbookworm (approved):
Still looks good

MyreMylar (approved):
OK, I don't mind this going in. 

It doesn't really create more work for the eventual plan of  Window supplanting display because Window already supports this functionality. I guess it slightly takes a little shine of Window for having features that display doesn't. But eh, that's not a great reason not to do it at the minute.

If display eventually becomes a python shim around the new Window module then this will only add a tiny bit more code to that shim.

PR #2819, opened 2024-04-24
Remove use of Py_BuildValue("(ii)")

itzpr3d4t0r:
This PR removes all occurrences of `Py_BuildValue("(ii)")` and swaps them with the faster pg_tuple_couple_from_values_int.


Damus666:
I don't know if my recent PR will be approved but I'm also gonna replace that function, thanks for the "idea"

zoldalma999 (approved):
LGTM!

PR #2820, opened 2024-04-25
Track version only in pyproject.toml

ankith26:
Currently, every time we bump the version, we have to update the version in like 4 files, and in some cases multiple lines within these files.

We already have a mechanism to read the version info defined in `pyproject.toml` in python. This mechanism was recently implemented to be used in the meson buildconfig so that we don't have to re-define the version there.

I'm taking that a step further in this PR, and now this file can be used in all places that needs a version. After this PR, there will be only one line in the codebase that defines the version (in the `pyproject.toml` file)

### DIFF
--- a/buildconfig/get_version.py
+++ b/buildconfig/get_version.py
@@ -29,4 +29,29 @@ else:
 
     version = ast.literal_eval(finds[0].strip())
 
-print(version)
+
+_splits = version.split(".")
+
+# handle optional dev tag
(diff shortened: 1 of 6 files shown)


oddbookworm (approved):
I don't see a reason not to, let's reduce the chance of something getting missed in future version changes lol

MyreMylar (approved):
LGTM 👍 

Nice change to speed up release fiddling.

PR #2824, opened 2024-04-26
Format Python code with ruff format instead of black

cclauss:
Failing tests are fixed in:
* #2796 

We want all tools run in `pre-commit` to be as fast as possible to enhance contributor productivity.

Let's consider formatting Python code with a highly compatible yet much faster formatter than Black.

https://docs.astral.sh/ruff/formatter/#philosophy
> The formatter is designed as a drop-in replacement for [Black](https://github.com/psf/black), but with an excessive focus on performance and direct integration with Ruff.  Given Black's popularity within the Python ecosystem, targeting Black compatibility ensures that formatter adoption is minimally disruptive for the vast majority of projects.

From: Known Deviations from Black
* https://docs.astral.sh/ruff/formatter/black/#hex-codes-and-unicode-sequences
* https://docs.astral.sh/ruff/formatter/black/#module-docstrings


.pre-commit-config.yaml
```diff
@@ -5,20 +5,17 @@
...
-                | ^docs/reST/.*$
-                | setup.py
-            )$
-
+  - repo: https://github.com/astral-sh/ruff-pre-commit
+    rev: v0.4.2
+    hooks:  # See pyproject.toml for configuration options.
+      #- id: ruff  # Run the linter
+      #  types_or: [ python, pyi, jupyter ]
+      #  args: [ --fix ]
```
oddbookworm, line 13: Are these lines here for any particular reason if they're commented out?

oddbookworm (commented):
Personally, I've been using ruff more lately, so I'm not against switching to it over black. But, I consider #2796 a blocker for this pull because of the failing lint check. I don't want every commit/pull to have a failing lint check while we wait on a fix pull to be merged.

I also would like to see opinions from the pygame-ce steering council.

.pre-commit-config.yaml
cclauss: We will enable them in future PRs.  Better to have the appropriate syntax at the ready.

oddbookworm:
@cclauss I haven't seen you on the pygame community discord server, where a lot of contribution discussions happen. If you'd like to join in those conversations, you can join the server at https://discord.gg/pygame

.pre-commit-config.yaml
oddbookworm: I'm not going to require this, but one small suggestion would be to add a comment that it'll be enabled in the future

cclauss, line 13:
```suggestion
```

MyreMylar (approved):
OK, this looks good to me. 

AFAICT, for our purposes ruff is just black (with a couple of super minor formatting differences) that will run faster because the formatting engine is re-written in a native code language (Rust in this case) rather than pure python.

No good reason not to have our formatter run faster (other than my general grumpy dislike of change.)

cclauss:
> a couple of super minor formatting differences

https://docs.astral.sh/ruff/formatter/#black-compatibility

cclauss:
Please re-run the failing test.

oddbookworm (approved):
LGTM!

cclauss:
Do we need more approvals?

Damus666:
@cclauss  Don't think so, I think someone just need to press some magical merge button

test/freetype_test.py
```diff
@@ -921,15 +921,17 @@ def test_freetype_Font_render_to_mono(self):
...
+                        bottomleft, surf.get_bitsize(), fg_color
+                    ),
                 )
                 bottomright = rrect.width - 1, rrect.height - 1
                 self.assertEqual(
                     surf.get_at(bottomright),
                     r_fg_color,
-                    "Position: {}. Depth: {}."
-                    " fg_color: {}.".format(bottomright, surf.get_bitsize(), fg_color),
+                    "Position: {}. Depth: {}." " fg_color: {}.".format(
```
ankith26, line 932: I know you are just applying an auto-formatted patch, but now that this line is being modified might as well concatenate the strings in this file

cclauss: There are https://docs.astral.sh/ruff/rules/#flake8-implicit-str-concat-isc and https://docs.astral.sh/ruff/rules/#flynt-fly and https://docs.astral.sh/ruff/rules/f-string rules that can all be applied to lines like this.  If we apply too many rules at once then pull requests never get reviewed.

ankith26 (approved):

PR #2825, opened 2024-04-26
`Surface.fblits()` image caching for performace

itzpr3d4t0r:
**Problem: Optimizing Image Blitting**

While experimenting with SIMD (Single Instruction, Multiple Data) operations, I realized that repeatedly reloading the same image from memory for each position we want to draw is inefficient. The current process involves:

1. Loading 8 pixels from memory (both top and bottom images).
2. Manipulating the pixels.
3. Storing the result back into memory.

This process is repeated for every position we want to draw to, resulting in a total of 2N operations (N surfaces, N loads + N stores).

**Solution: Register Cache**

To improve performance, I propose the following approach:

1. Create a register cache (of __m256i) and load the Surface’s pixels once.
2. Manipulate and store the results at each position in the draw sequence.

Resulting in 1+ N operations (1 load, N stores), significantly speeding up the blitting process.

By avoiding repeated memory reads, we significantly reduce blitting time. To implement this, pass a tuple in the format `(Surface_To_Draw, List_Of_Positions_To_Draw)`.

This PR is a draft, and I’m actively working on further improvements and additions, like for example support for blend flags and colorkey / alpha. The results so far are very promising (blitcopy) :

- `fblits()` with caching: 58-65 fps
- `fblits()` without caching: 12 fps
- `blits()`: 11-12 fps

Here is a little program I've been experimenting with to get these numbers:
```Py
from random import randint, random

import pygame

pygame.init()

screen_size = 1000

screen = pygame.display.set_mode((screen_size, screen_size))

size = 50
flag = 0

s = pygame.Surface((size, size))
pygame.draw.circle(s, (100, 0, 255), (size // 2, size // 2), size // 2)


class Particle:
    def __init__(self, x, y, vx, vy):
        self.x = x
        self.y = y
        self.vx = vx
        self.vy = vy

    def update(self, dt):
        self.x += self.vx * dt
        self.y += self.vy * dt

        if self.x < 0:
            self.x = 0
            self.vx *= -1
        elif self.x > screen_size - size - 1:
            self.x = screen_size - size - 1
            self.vx *= -1

        if self.y < 0:
            self.y = 0
            self.vy *= -1
        elif self.y > screen_size - size - 1:
            self.y = screen_size - size - 1
            self.vy *= -1

        return self.x, self.y


particles = [
    Particle(randint(0, screen_size - size), randint(0, screen_size - size),
             random() * 2 - 1, random() * 2 - 1)
    for _ in range(10000)
]

clock = pygame.time.Clock()

font = pygame.font.SysFont("Arial", 25, True)

modes = ["fblits", "blits", "fblits cached"]
mode_index = 0
mode = modes[mode_index]

while True:
    dt = clock.tick_busy_loop(1000) * 60 / 1000
        if event.type == pygame.QUIT:
            pygame.quit()
            exit()
        elif event.type == pygame.MOUSEWHEEL:
            mode_index = (mode_index + event.y) % len(modes)
            mode = modes[mode_index]
            if event.key == pygame.K_0:
                flag = 0

    screen.fill(0)

    if mode == "fblits":
        screen.fblits([(s, p.update(dt)) for p in particles], flag)
    elif mode == "blits":
        screen.blits([(s, p.update(dt), None, flag) for p in particles])
    elif mode == "fblits cached":
        screen.fblits(
            [
                (s, [p.update(dt) for p in particles])
            ],
            flag,
            True
        )

    fps_text = font.render(
        "FPS: " + f"{int(clock.get_fps())}" + f"| N: {len(particles)}",
        True, (255, 255, 255))
    pygame.draw.rect(screen, (0, 0, 0),
                     fps_text.get_rect(center=(screen_size / 2, screen_size / 2)))
    screen.blit(fps_text, fps_text.get_rect(center=(screen_size / 2, screen_size / 2)))

    mode_text = font.render(f"{mode}", True, (255, 255, 255))
    pygame.draw.rect(screen, (0, 0, 0),
                     mode_text.get_rect(center=(screen_size / 2, screen_size / 2 + 25)))
    screen.blit(mode_text,
                mode_text.get_rect(center=(screen_size / 2, screen_size / 2 + 25)))

    pygame.display.flip()
```


Starbuck5:
I don't really understand the proposal, but piping in on this:

> Create a register cache (of __m256i) and load the Surface’s pixels once.

It sounds like you want to store an arbitrary amount of data in SIMD registers-- that's not possible. There are a set amount of these registers in each processor core, like 16. http://www.infophysics.net/amd64regs.png

XMM are the SSE registers, YMM are the AVX registers, ZMM (not shown in that diagram) are the AVX512 registers.

MyreMylar:
Had a little peek at this. 

Are you intending to add alpha blending support, or special blend mode support, in this PR or in a future one?

PR #2827, opened 2024-04-26
Cleanup codebase, docs and examples with American spelling

Damus666:
I used Visual Studio Code search tool to find all the occasions for non American spelling, looked quickly at each one and chose to replace them or not.
I ignored file names, files, links and the C/PY color dict files (where both spellings are intended)
The reason is:
- the codebase should use one convention and stick to it
- the American spelling is preferred to match the actual `pygame` classes and methods such as `Color` and `grayscale()`

For now I considered the following non American spelling (you can request more):
- grey -> gray
- colour -> color

If all checks pass it means I didn't break any variable name.

### DIFF
--- a/docs/reST/ref/color.rst
+++ b/docs/reST/ref/color.rst
@@ -348,7 +348,7 @@
       | :sl:`returns a Color where the r,g,b components have been multiplied by the alpha.`
       | :sg:`premul_alpha() -> Color`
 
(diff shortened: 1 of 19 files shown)


ankith26 (approved):
LGTM, thanks for the PR :+1:

oddbookworm (approved):
LGTM!

Issue #2452, opened 2023-09-16
Port TTF_SetFontOutline


Starbuck5:
**Description**

SDL_TTF has outlining functionality! https://wiki.libsdl.org/SDL2_ttf/TTF_SetFontOutline.

We should port this functionality such that people can effectively render outlines on their text.

To do this, we could directly port the attribute and get/set it on the font, but then people will need two render calls, one for the outline and one for the text. For example:
```py
font.outline = 2
text = font.render("Hello world", True, "orange")
win.blit(text, (10, 10))

font.outline = 0
text = font.render("Hello world", True, "black")
win.blit(text, (10 + width, 10 + width))
```

We could also have something higher level, where it auto renders the outline in font.render. Random notes of that:
```
font.outline_width (attribute, get/set with TTF_ outline)
font.outline_color (attribute, defaults to black)

Render with an outline:
font.render

Render only the outline:
font.render(outline_only kwarg) | font.render_outline_only() | font.outline_only
```

Starbuck5:
I have the most basic bits of the implementation in https://github.com/Starbuck5/pygame-ce/tree/font-outline-basic. No types, no tests, no docs, just the c code.

PR #2828, opened 2024-04-26
Add outline option for rendering fonts

narilee2006:
Hello! I tried doing https://github.com/pygame-community/pygame-ce/issues/2452. Let me know if there are any issues.


src_c/font.c
```diff
@@ -829,6 +829,34 @@ font_getter_style_name(PyObject *self, void *closure)
...
+    if (!PgFont_GenerationCheck(self)) {
+        RAISE_FONT_QUIT_ERROR_RETURN(-1);
+    }
+
+    TTF_Font *font = PyFont_AsFont(self);
+    int val;
+
+    DEL_ATTR_NOT_SUPPORTED_CHECK("outline", value);
+
+    val = PyLong_AsLong(value);
```
oddbookworm, line 844: One note is that [`PyLong_AsLong`](https://docs.python.org/3/c-api/long.html#c.PyLong_AsLong) returns `-1` on failure, so you should probably do what those docs say and check if an error occurred, and return -1 if one did. Something like this should work. Note that no additional exception needs to be set because it's already set by PyLong_AsLong
```c
if ((val == -1) && (PyErr_Occurred() != NULL))
{
    return -1;
}
```

buildconfig/stubs/pygame/font.pyi
```diff
@@ -56,6 +56,8 @@ class Font:
     def point_size(self) -> int: ...
     @point_size.setter
     def point_size(self, value: int) -> None: ...
+    @property
+    def outline(self) -> int: ...
```
oddbookworm, line 60: You also need to stub the setter
```python3
@outline.setter
def outline(self, value: int) -> None: ...
````

File: src_c/font.c [hunk 3]
```diff
@@ -829,6 +829,34 @@ font_getter_style_name(PyObject *self, void *closure)
...
+}
+
+static PyObject *
+font_getter_outline(PyObject *self, void *closure)
+{
+    if (!PgFont_GenerationCheck(self)) {
+        return RAISE_FONT_QUIT_ERROR();
+    }
+
+    return PyLong_FromLong(TTF_GetFontOutline(PyFont_AsFont(self)));
```
oddbookworm, line 857: I don't really like this chain like this, but I don't think it's technically wrong. Neither inner function call returns something specific on failure to check, so I guess it's fine

test/font_test.py
```diff
@@ -992,6 +992,10 @@ def test_font_property_should_raise_exception_after_quit(self):
                 properties.append(("point_size", 1))
             else:
                 skip_properties.add("point_size")
+            if version >= (2, 0, 12):
+                properties.append(("outline", 2006))
+            else:
+                skip_properties.add("outline")
```
oddbookworm, line 998: Don't do it like this lol. Add it back as a property in the list, but only for the shorter list. It won't work for ftfont because ftfont inherits from _freetype.Font, not font.Font

oddbookworm (changes requested):

test/font_test.py
oddbookworm: The reason you don't need to do it like this IMO is that we support TTF 2.20.0 or something like that, well past 2.0.12

test/font_test.py
```diff
@@ -1057,6 +1058,7 @@ def query(
         underline=False,
         strikethrough=False,
         antialiase=False,
+        outline=False,
```
oddbookworm, line 1061: Isn't outline an int, not a bool?

test/font_test.py
```diff
@@ -981,6 +981,7 @@ def test_font_property_should_raise_exception_after_quit(self):
                 ("italic", True),
                 ("underline", True),
                 ("strikethrough", True),
+                ("outline", 2006),
```
oddbookworm, line 984: 2006? That's both oddly specific and very high. Why not something reasonable, like 1 or 2?

oddbookworm (commented):
I have just a couple comments

test/font_test.py
narilee2006: I tried sneaking in an easteregg of myself. It's the year I am born.

oddbookworm (approved):
LGTM!

Starbuck5 (changes requested):
The challenge of #2452 is to design a good API for this, and this PR doesn’t accomplish that in my opinion.

- Outline color isn’t supported
- Looking at my notes from before (haven’t run this yet), doesn’t this PR mean that people have to call font.render twice? Once for the text, then once for the outline. That seems inconvenient for users.

narilee2006:
> Outline color isn’t supported

It actually is, you need to call it in the font.render() method.

> Looking at my notes from before (haven’t run this yet), doesn’t this PR mean that people have to call font.render twice? Once for the text, then once for the outline. That seems inconvenient for users.

Yep, that's how SDL2 implemented it.

narilee2006:
Perhaps `outline` as a kwarg so the developer doesn't need to use two lines to create the font?

oddbookworm:
Maybe perhaps
```
pygame.font.Font(filename=None, size=20, outline=0)
```
for the font constructor and
```
pygame.font.Font.render(text, antialias, color, bgcolor=None, wraplength=0, outline_color=None)
```
for the render call?

But I'm not sure what to do if `outline_color` defaults when `outline` was set to something nonzero

Starbuck5:
> Perhaps outline as a kwarg so the developer doesn't need to use two lines to create the font?

Well then how do you do the color? Another kwarg? Kwargs are expensive to parse in functions that get called all the time, too.

I think my second proposal in the issue would work well (minus the render outline only thing, I think that’s unnecessary now)

‘’’
font.outline_width (attribute, get/set with TTF_ outline)
font.outline_color (attribute, defaults to black)

Render with an outline:
font.render
‘’’

narilee2006:
I'm personally against the outline_color parameter as existing draw methods have the `width` parameter to determine their border size; if the `width` parameter is greater than 0 it won't fill the shape and only draw the borders.

https://pyga.me/docs/ref/draw.html

narilee2006:
> Well then how do you do the color? Another kwarg? Kwargs are expensive to parse in functions that get called all the time, too.

@Starbuck5, sorry I meant the `outline` kwarg in the constructor of the Font object.

MyreMylar:
I agree with @Starbuck that this should be implemented as a styling attribute to Font - similar to the current styling attributes bold, italic & underline. The current `.render()` function is already overlong with parameters like the 'antialias' parameter that definitely should have been an attribute defaulting to the most common usage (true). 

Proposed test program:

```python
import pygame
from pygame import Color, Font, QUIT

pygame.init()
window_surf = pygame.display.set_mode((400, 200))

impact_font = Font(filename="fonts/impact.ttf", size=64)

impact_font.outline_width = 2  # defaults to 0
impact_font.outline_color = Color("red")  # defaults to black

text_with_outline = impact_font.render(
    text="Hello World", antialias=True, color=Color("white")
)

running = True
while running:
        if event.type == QUIT:
            running = False

    window_surf.fill(Color("black"))

    window_surf.blit(text_with_outline, (50, 50))

    pygame.display.update()


```

font.render can just check if outline_width is greater than 0 or not - and if so draw the outline before or after the main font rendering code - whichever looks better.

As an aside maybe we could make color and antialias into arguments with default values. We almost certainly would have already converted 'antialias' if it was on the other side of 'color' in the argument list. At least that way you could do:

```
font.render("Hello World", color=Color("white"))
```

Which isn't a typing saving over:

```
font.render("Hello World", True, Color("white"))
```

But probably clearer to read.

Might also open up:

```
font.default_color = Color("red")

hw_text_surf = font.render("Hello World in red")
other_text_surf = font.render("Other text, also in red")
```

MyreMylar (changes requested):
I think 80% of the time when people use this they are going to want to draw an outline around actual visible text. 

A possible upgrade to `font.render()` that might be nice in concert with this is to allow it to set alpha pixels on the produced surface. I believe at the minute text surfaces come out  of render in a slightly odd format that looks like normal 32 bit 'ARGB' - but is subtly not that. You can see this by doing:
```
text_surf  = font.render("Hello World", True, Color("white"))
# text_surf = text_surf.convert_alpha()
text_surf = text_surf.premul_alpha() 

running = True
while running:
        if event.type == QUIT:
            running = False

    window_surf.fill(Color("grey"))

    window_surf.blit(text_surf, (50, 50), special_flags=BLEND_PREMULTIPLIED)

    pygame.display.update()
```
and seeing the difference after uncommenting the `.convert_alpha()` line. I expect at some point it was marginally cheaper to blit whatever, not-quite-full-transparency format comes out of font.render() but I doubt it still is. Anyway, if we could render semi- and full transparency text then you would get the benefits of outline only text without having to change the API.

narilee2006:
Okay... I'll get to work on this.

docs/reST/ref/font.rst
```diff
@@ -306,6 +306,28 @@ solves no longer exists, it will likely be removed in the future.
 
       .. ## Font.point_size ##
 
+   .. attribute:: outline_size
```
Starbuck5, line 309: I'd rather do outline_width

docs/reST/ref/font.rst
```diff
@@ -306,6 +306,28 @@ solves no longer exists, it will likely be removed in the future.
...
+
+      | :sl:`Gets or sets the font's outline size`
+      | :sg:`outline_size -> int`
+
+      Returns the size of the outline.
```
Starbuck5, line 314: But it doesn't just return it, it can be set here too. What units is the outline size/width in? What does setting an outline size/width do?

docs/reST/ref/font.rst
```diff
@@ -306,6 +306,28 @@ solves no longer exists, it will likely be removed in the future.
...
+      .. versionadded:: 2.5.0
+
+      .. ## Font.outline_size ##
+
+   .. attribute:: outline_color
+
+      | :sl:`Gets or sets the font's outline color`
+      | :sg:`outline_size -> RGB`
+
+      Returns the color of the outline.
```
Starbuck5, line 325: Ditto this doesn't just return it, it can be set here too. What does this default to?

Plus one of these attributes should have a small example of how to use this, with an image of what the output looks like.

src_c/font.c
```diff
@@ -38,11 +38,11 @@
 #include "structmember.h"
 
 #define RAISE_TEXT_TYPE_ERROR() \
-    RAISE(PyExc_TypeError, "text must be a unicode or bytes");
+    RAISE(PyExc_TypeError, "text must be a unicode or bytes")
 
 #define RAISE_FONT_QUIT_ERROR_RETURN(r) \
     RAISERETURN(pgExc_SDLError,         \
-                "Invalid font (font module quit since font created)", r)
+                "Invalid font (font module quit since font created)", r);
```
Starbuck5, line 45: Why change this?

src_c/font.c
```diff
@@ -618,50 +582,150 @@ font_render(PyObject *self, PyObject *args, PyObject *kwds)
...
+    static char *kwlist[] = {"text",    "antialias",  "color",
+                             "bgcolor", "wraplength", NULL};
+
+    if (!PyArg_ParseTupleAndKeywords(args, kwds, "OpO|Oi", kwlist, &text,
+                                     &antialias, &fg_rgba_obj, &bg_rgba_obj,
+                                     &wraplength)) {
+        return NULL;
+    }
+
+    // 글꼴 생성
```
Starbuck5, line 661: Comments should be in English

src_c/font.c
```diff
@@ -538,78 +538,42 @@ font_set_strikethrough(PyObject *self, PyObject *arg)
...
-            return NULL;
+        if (astring == NULL) {
+            return 0;  // exception already set.
         }
         if (strlen(astring) != (size_t)_size) {
-            return RAISE(PyExc_ValueError,
-                         "A null character was found in the text");
+            PyErr_Format(PyExc_ValueError,
+                         "A null character was found in the text.");
+            return 0;
```
Starbuck5, line 568: Why do these changes exist?

src_c/include/pygame_font.h
```diff
@@ -29,6 +29,10 @@ typedef struct {
     PyObject_HEAD TTF_Font *font;
     PyObject *weakreflist;
     int ptsize;
+
+    SDL_Color outline_color;
+    int outline_size;
+
```
Starbuck5, line 35: The spaces around these two are jarring with the lack of space between the other ones.

test/font_test.py
```diff
@@ -1043,7 +1052,7 @@ def setUp(self):
             self.screen = pygame.display.set_mode((600, 200))
             self.screen.fill((255, 255, 255))
             pygame.display.flip()
-            self.f = pygame_font.Font(None, 32)
+            self.f = pygame_font.Font(None, 96)
```
Starbuck5, line 1055: Why?

src_c/font.c
```diff
@@ -538,78 +538,42 @@ font_set_strikethrough(PyObject *self, PyObject *arg)
     Py_RETURN_NONE;
 }
 
-static PyObject *
-font_render(PyObject *self, PyObject *args, PyObject *kwds)
+static int
+_create_font_surface(TTF_Font *font, PyObject *text, int antialias,
+                     SDL_Color foreg, SDL_Color backg, int draw_backg,
+                     int wraplength, SDL_Surface **dst_surf)
```
Starbuck5, line 544: I'm skeptical of the division of effort between font_render and this function.

I haven't dived much into it yet, but it seems like it would be simpler to have more of the argument parsing and error checking in font_render. Like why does this function need to parse the C string out of the Pyobject, or check the wraplength.

In fact, if the text has an outline, this code has to parse the C string out of the PyObject twice because it calls this function twice, rather than extracting it in font_render and then passing it in.

Also, it would be nice from an organizational perspective to have the "Python API level function" that deals with Python, and then have this function be C-only. It's something I want to do more in our code.

src_c/font.c
robertpfeiffer: Is this a clang-format version mismatch?

src_c/font.c
```diff
@@ -618,50 +582,150 @@ font_render(PyObject *self, PyObject *args, PyObject *kwds)
...
+        return NULL;
+    }
+
+    // 글꼴 생성
+    if (!pg_RGBAFromObjEx(fg_rgba_obj, rgba, PG_COLOR_HANDLE_ALL)) {
+        return 0;  // exception already set
+    }
+
+    SDL_Color foreg = {rgba[0], rgba[1], rgba[2], SDL_ALPHA_OPAQUE};
+    SDL_Color backg = {0, 0, 0, SDL_ALPHA_OPAQUE};
```
robertpfeiffer, line 667: should we warn here when outline colour is the same as background or foreground?

src_c/font.c
```diff
@@ -618,50 +582,150 @@ font_render(PyObject *self, PyObject *args, PyObject *kwds)
...
+    Uint8 rgba[] = {0, 0, 0, 0};
+
+    if (!PgFont_GenerationCheck(self)) {
+        return RAISE_FONT_QUIT_ERROR()
+    }
+
+    static char *kwlist[] = {"text",    "antialias",  "color",
+                             "bgcolor", "wraplength", NULL};
+
+    if (!PyArg_ParseTupleAndKeywords(args, kwds, "OpO|Oi", kwlist, &text,
```
robertpfeiffer, line 655: having foreground colour and background colour as params, but outline colour as a property feels inconsistent. I would like to request some feedback from the maintainers on this. Maybe we should fix this now, maybe we need to make an issue for this after merge.

src_c/font.c
```diff
@@ -618,50 +582,150 @@ font_render(PyObject *self, PyObject *args, PyObject *kwds)
...
+                                     &wraplength)) {
+        return NULL;
+    }
+
+    // 글꼴 생성
+    if (!pg_RGBAFromObjEx(fg_rgba_obj, rgba, PG_COLOR_HANDLE_ALL)) {
+        return 0;  // exception already set
+    }
+
+    SDL_Color foreg = {rgba[0], rgba[1], rgba[2], SDL_ALPHA_OPAQUE};
```
robertpfeiffer, line 666: I don't understand the API here, shouldn't the background be transparent by default? I know this matches the old behaviour, but this probably warrants a comment.

src_c/font.c
```diff
@@ -618,50 +582,150 @@ font_render(PyObject *self, PyObject *args, PyObject *kwds)
...
+    }
+
+    if (!_create_font_surface(font, text, antialias, foreg, backg, draw_backg,
+                              wraplength, &surf)) {
+        return NULL;
+    }
+
+    SDL_Surface *filled_with_outline_surf = NULL;
+
+    if (outline_size > 0) {
```
robertpfeiffer, line 688: Does the order of operations matter here? Would the drawn pixels look different if you rendered the outline first, and blitted the letters onto that (except surface size)

robertpfeiffer (changes requested):

robertpfeiffer:
Overall, this is promising.

docs/reST/ref/font.rst
narilee2006: Example as in the examples directory? Or an example as an example image in the documentation?

src_c/font.c
narilee2006: During my time learning the code I accidentally deleted those ;s.

src_c/font.c
narilee2006: Return type is no longer a `PyObject *`.

src_c/include/pygame_font.h
narilee2006: Could you explain further? I don't understand.

test/font_test.py
narilee2006: It's hard to see the outline when the font is small.

docs/reST/ref/font.rst
```diff
@@ -306,6 +306,28 @@ solves no longer exists, it will likely be removed in the future.
 
       .. ## Font.point_size ##
 
+   .. attribute:: outline_size
+
+      | :sl:`Gets or sets the font's outline size`
+      | :sg:`outline_size -> int`
```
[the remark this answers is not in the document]
narilee2006: Nope, it defaults to 0. I'll add this to the docs.

src_c/font.c
narilee2006: Yes, the order matters as the size of the outline will increase inward and outward of the text's outlines.

src_c/font.c
narilee2006: I learned from my professor that you cannot assign `NULL` to `typedef struct`s. Correct me if she's wrong.

src_c/font.c
narilee2006: I'll eventually touch on this and refactor the code.

src_c/font.c
narilee2006: I don't think that'll be necessary as printing content to the console is already a controversial topic.

https://github.com/pygame/pygame/issues/1468

src_c/font.c
robertpfeiffer: This should probably be a comment here then.

/* order of operations matters. \*/
/* blitting foreground over outline results in smaller outline than otherwise \*/

something like that

src_c/font.c
robertpfeiffer: No, I meant I don't understand why the alpha isn't 0. It was like that before, but this part of the code is unclear to me.

docs/reST/ref/font.rst
Starbuck5: Example as in a code block in the docs showing how it works, with a picture of the rendered output.

src_c/font.c
Starbuck5: RAISE doesn't return a PyObject, it returns NULL. And NULL is 0, so this is the same code?

src_c/include/pygame_font.h
Starbuck5:
```
PyObject *weakreflist;
int ptsize;
*SPACE*
SDL_Color outline_color;
int outline_size;
*SPACE*
unsigned int ttf_init_generation;
```
There aren't any other spaces in the struct definition.

docs/reST/ref/font.rst
```diff
@@ -306,6 +306,28 @@ solves no longer exists, it will likely be removed in the future.
 
       .. ## Font.point_size ##
 
+   .. attribute:: outline_width
+
+      | :sl:`Gets or sets the font's outline size`
+      | :sg:`outline_width -> int`
+
+      Gets or sets the size of the outline in pixels defaulting to 0.
```
Starbuck5, line 314: I tested this with Arial, and I saw that it actually added 2 pixels of outline with width=1, and 4 pixels of outline with width=2.

`Gets or sets the size of the outline in pixels defaulting to 0.`
Size still?
Grammatically it should be `Gets or sets the size of the outline in pixels, defaulting to 0.`

You could also mention details like it has to be a positive number or zero, zero means no outline.

src_c/font.c
```diff
@@ -829,6 +892,93 @@ font_getter_style_name(PyObject *self, void *closure)
...
+#if SDL_TTF_VERSION_ATLEAST(2, 0, 12)
+    return PyLong_FromLong(self->outline_width);
+#else
+    return RAISE(pgExc_SDLError,
+                 "Incorrect SDL_TTF version (requires 2.0.12)");
+#endif
+}
+
+static int
+font_setter_outline_width(PyFontObject *self, PyObject *value, void *closure)
```
Starbuck5, line 910: When I try to set a float it gives me this:

```
C:\Users\charl\Desktop\testr.py:8: DeprecationWarning: an integer is required (got type float).  Implicit conversion to integers using __int__ is deprecated, and may be removed in a future version of Python.
  arial.outline_width = 1.5
```

src_c/font.c
narilee2006: It was originally `SDL_ALPHA_OPAQUE` before. https://github.com/pygame-community/pygame-ce/pull/2828/files#diff-75538c02a9bfe6511e8b72324ec5736b4d22aae6c5a99c52a83cef0dc70df300L575

src_c/font.c
narilee2006: This does not happen on my machine.
```
ERROR: test_set_outline_width_as_float (__main__.FontTypeTest.test_set_outline_width_as_float)
  File "/home/nari/Documents/pygame-ce/test/font_test.py", line 392, in test_set_outline_width_as_float
    f.outline_width = 20.06
    ^^^^^^^^^^^^^^^
TypeError: 'float' object cannot be interpreted as an integer

Ran 72 tests in 13.958s
```

docs/reST/ref/font.rst
narilee2006: > I tested this with Arial, and I saw that it actually added 2 pixels of outline with width=1, and 4 pixels of outline with width=2.

The unit is actually radius, not pixels.

docs/reST/ref/font.rst
```diff
@@ -306,6 +306,30 @@ solves no longer exists, it will likely be removed in the future.
 
       .. ## Font.point_size ##
 
+   .. attribute:: outline_width
```
novialriptide, line 309: `outline_radius` is a better name if you're already going to mention radius in the docs description.

docs/reST/ref/font.rst
```diff
@@ -306,6 +306,30 @@ solves no longer exists, it will likely be removed in the future.
...
+      If the value is set to 0, it shows no outline.
+
+      .. versionadded:: 2.5.0
+
+      .. ## Font.outline_width ##
+
+   .. attribute:: outline_color
+
+      | :sl:`Gets or sets the font's outline color`
+      | :sg:`outline_width -> RGB`
```
novialriptide, line 325: You say (0, 0, 0, 255), but that's RGBA.

src_c/font.c
novialriptide: Resolving as this code no longer exists.

damusss:
I'm taking the liberty to close this pull request as a different approach has already been merged and published. Still grateful it has been worked on though!

PR #2832, opened 2024-04-28
Try sleeping in window autoresize test

MyreMylar:
An idea to hopefully reduce number of CI failures on this test.

Example failures:
https://github.com/pygame-community/pygame-ce/actions/runs/8573054619/job/23496953618?pr=2793
https://github.com/pygame-community/pygame-ce/actions/runs/8506569444/job/23297052392?pr=2766

I'm assuming that it is failing because the surface event has not yet been processed by the system at the time of checking and thus the callback has yet to be called for our window surface.

Adding a small delay after resizing the window should allow the system a chance to catch up.

If this doesn't work - we could also just try disabling it on Pypy where it seem to happen most often.

### DIFF
--- a/test/window_test.py
+++ b/test/window_test.py
@@ -1,6 +1,7 @@
 import unittest
 import pygame
 import os
+import time
 
 from pygame import Window
 from pygame.version import SDL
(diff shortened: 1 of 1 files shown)


Damus666 (approved):
I really do hope it reduces/fix the fails, thanks

Starbuck5 (approved):
I re-ran the Windows CI to give it an opportunity to fail again, and it still didn't.

So I'm down to give this a try.

itzpr3d4t0r (approved):

Issue #2821, opened 2024-04-26
`pygame.Surface.fblits` slower in 2.5.0.dev2 vs 2.4.1


oddbookworm:
@bigwhoopgames mentioned on discord that he's getting a 25%-33% fps drop when using the 2.5.0.dev2 prerelease vs the 2.4.1 release. I did some testing by giving him special wheels with specific changes to see if that fixed it, but didn't really see much difference. List of things I tried:
* Reverting #2679
* Reverting #2679 and #2667
* Adding `/Ox` to the MSVC build in meson
* Setting `optimization` to `3` in meson

Image provided by Big Whoop showing some comparisons from a performance profile
![image](https://github.com/pygame-community/pygame-ce/assets/49015102/289b9d4d-dbbc-4b7b-9e62-89d5781ccb73)

bigwhoopgames:
Tested with my youtube particles tutorial. Running on 2.4.1 ~20K particles at 60FPS and on 2.5.0 getting about 6800.

https://github.com/bigwhoopgames/youtube-particles/blob/master/particles.py

Line 88 needs changing to work with python 3.12, just wrap values passed to random as ints.

itzpr3d4t0r:
Could it be SDL's fault? We got `SDL 2.28.5` on `2.4.1` and `SDL 2.30.2` on `2.5.0 -dev2`. It's basically the only thing that changed between our releases with fblits.

itzpr3d4t0r:
Confirmed, i tested installing the latest pygame-ce version with SDL 2.28.5 and the fps in my test program are back to pygam.ce 2.4.1 territory. (got 400 fps in 2.4.1 and 260-270 in 2.5.0 dev2 and 350-400 with 2.5.0 with the old SDL version)

oddbookworm:
How hard would making an SDL reproducer be so I can report it upstream to them?

robertpfeiffer:
Do you even need to recompile?
Just compile with SDL SDL 2.28.5 and substitute with 2.30.2 through the dynamic API replacer

SDL3_DYNAMIC_API=/my/actual/libSDL2.so.30.2 python3 benchmark.py

Starbuck5:
Information from discord-

itzpr said
> To be specific I restricted the issue between 2.28.5 and 2.29.3
Couldn't really install 2.29.1 but i guess 2.29.0 introduced the issue

oddbookworm posted a C reproducer attempt
```c
#include <stdio.h>
[...]

PR #2835, opened 2024-05-03
Use HasColorKey instead of GetColorKey in pgSurface_Blit

Starbuck5:
This routine doesn't use the value of GetColorKey, it just wants to know whether the surface has one.

Internally, this was setting a bunch of SDL errors ("surf doesn't have colorkey"-esque), which became much more expensive in SDL 2.29.3 because of a new logging feature.

Therefore lets replace it with HasColorKey (a newer function, the code was originally written before this existed). This fixes the performance regression here and even makes the performance slightly better than it was before, at least in my test scenario.

Credit to @itzpr3d4t0r for narrowing this down to SDL_GetColorKey

Relevant to #2821

### DIFF
--- a/src_c/surface.c
+++ b/src_c/surface.c
@@ -3822,7 +3822,6 @@ pgSurface_Blit(pgSurfaceObject *dstobj, pgSurfaceObject *srcobj,
     int result, suboffsetx = 0, suboffsety = 0;
     SDL_Rect orig_clip, sub_clip;
     Uint8 alpha;
-    Uint32 key;
 
     /* passthrough blits to the real surface */
     if (((pgSurfaceObject *)dstobj)->subsurface) {
@@ -3860,7 +3859,7 @@ pgSurface_Blit(pgSurfaceObject *dstobj, pgSurfaceObject *srcobj,
     pgSurface_Prep(srcobj);
 
     if ((blend_flags != 0 && blend_flags != PYGAME_BLEND_ALPHA_SDL2) ||
-        ((SDL_GetColorKey(src, &key) == 0 || _PgSurface_SrcAlpha(src) == 1) &&
+        ((SDL_HasColorKey(src) || _PgSurface_SrcAlpha(src) == 1) &&
          /* This simplification is possible because a source subsurface
             is converted to its owner with a clip rect and a dst
             subsurface cannot be blitted to its owner because the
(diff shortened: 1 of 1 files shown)


Starbuck5:
Test script:

```py
import random
import time

import pygame

random.seed(36)

width = 600
height = 400

screen = pygame.Surface((width, height))

def make_particle():
    p_surf = pygame.Surface((5,5))
    p_surf.fill((random.randint(0,255), random.randint(0,255), random.randint(0,255)))

    return (p_surf, (random.randint(0, width), random.randint(0, height)))

particles = [make_particle() for _ in range(1000000)]

start = time.time()
screen.fblits(particles)
print(time.time() - start)
```

Results:
- main: 1.13 seconds
- this PR: 0.17 seconds
- 2.4.1 (from pip): 0.27 seconds

Starbuck5:
An example of how the issue ( #2821 ) isn't fully fixed by this PR--

If you take the previous example and swap out the line defining `p_surf` for `p_surf = pygame.Surface((5,5), pygame.SRCALPHA)`, to make it an alpha blit, there is still a performance problem.

On 2.4.1 it's 0.30 seconds, on main it's 1.60 seconds, and on this PR it's 0.63 seconds. So an improvement, but the SDL_GetColorKey in `SoftBlitPyGame` also needs to be re-evaluated

itzpr3d4t0r (approved):
Great! Tested this branch and compared it to 2.4.1's performance with fblits drawing 10 surfaces of increasingly bigger size up to 1000x1000 and got the following results:
Figure_1
A very good improvement for small (up to 200x200) surfaces!

itzpr3d4t0r:
> An example of how the issue ( #2821 ) isn't fully fixed by this PR--
> On 2.4.1 it's 0.30 seconds, on main it's 1.60 seconds, and on this PR it's 0.63 seconds. So an improvement, but the SDL_GetColorKey in `SoftBlitPyGame` also needs to be re-evaluated

About this, if SDL can do something about SDL_GetColorkey's perf we're good, otherwise could we implement our own function for that? I know SDL_Surface.map is private and it wouldn't be the best practice but what woud be the altrnatives?

robertpfeiffer (approved):
LGTM

ankith26 (approved):

PR #2840, opened 2024-05-05
flood fill

robertpfeiffer:
[Flood fill](https://en.wikipedia.org/wiki/Flood_fill) drawing operation, like in GIMP and ASEprite.

This still needs docs and tests, and perhaps some more special-case handling of indexed surfaces. Can take a pattern surface as color argument. If you like this, maybe we can add pattern surfaces to polygons also, but in a pinch, you can draw a polygon and then use flood fill, or draw a polygon on an intermediate surface, use flood fill, and then blit the surface.

One thing this can't do that you can do in GIMP is gradients.

<details>
<summary>Test Program</summary>

```python
import pygame

pygame.init()

window = pygame.display.set_mode((800, 600))

running = True

test_rect = pygame.Rect(50, 50, 50, 50)

pattern = pygame.Surface((2, 2))
pattern.fill((255, 255, 255))
pattern.set_at((0, 0), (255, 0, 0))
pattern.set_at((1, 1), (0, 0, 255))

should_fill_color = False
should_fill_pattern = False
should_fill_background = False

while running:
        if event.type == pygame.QUIT:
            running = False

        if event.type == pygame.KEYDOWN and event.key == pygame.K_SPACE:
            test_rect.width *= 2.0
            test_rect.height *= 2.0

        if event.type == pygame.KEYDOWN and event.key == pygame.K_p:
            should_fill_pattern = True
            should_fill_color = False

        if event.type == pygame.KEYDOWN and event.key == pygame.K_c:
            should_fill_color = True
            should_fill_pattern = False

        if event.type == pygame.KEYDOWN and event.key == pygame.K_b:
            should_fill_background = True

    window.fill((0, 0, 0))

    pygame.draw.rect(window, pygame.Color("white"), test_rect, width=1)
    
    if should_fill_color:
        pygame.draw.flood_fill(window, pygame.Color("blue"), (85, 62))

    if should_fill_pattern:
        pygame.draw.flood_fill(window, pattern, (52, 52))

    if should_fill_background:
        pygame.draw.flood_fill(window, pygame.Color("yellow"), (285, 362))

    pygame.display.update()
```

</details>

## Summary by CodeRabbit

* **New Features**
  * Added flood_fill to the drawing module: fills a connected area from a start point using a solid color or a pattern and returns a bounding Rect; supports a start_pos keyword.

* **Documentation**
  * Added usage and behavior details for flood_fill, including parameter options, pattern support, version info, and return semantics.

* **Tests**
  * Added automated tests for solid-color, pattern-based, and shape flood fills (duplicate test class entry present).


Damus666 (approved):

src_c/draw.c
```diff
@@ -85,6 +85,13 @@ draw_round_rect(SDL_Surface *surf, int x1, int y1, int x2, int y2, int radius,
                 int width, Uint32 color, int top_left, int top_right,
                 int bottom_left, int bottom_right, int *drawn_area);
 
+static int
+flood_fill_inner(SDL_Surface *surf, int x1, int y1, Uint32 new_color,
+                 SDL_Surface *pattern, int *drawn_area);
+
+static void
+unsafe_set_at(SDL_Surface *surf, int x, int y, Uint32 color);
```
oddbookworm, line 93: I can't actually find a definition for this?

oddbookworm (changes requested):

src_c/draw.c
```diff
@@ -988,6 +1082,33 @@ swap(float *a, float *b)
     *b = temp;
 }
 
+#define WORD_BITS (8 * sizeof(unsigned int))
```
oddbookworm, line 1085: Hmmmmm, I'm not convinced that there isn't either a bug or maybe a bad variable name here. I know the vast majority of systems have a word size of 32 or 64 nowadays, but even a byte can technically be something else right? Do we care enough to support those extremely odd specific cases?

src_c/draw.c
```diff
@@ -988,6 +1082,33 @@ swap(float *a, float *b)
...
+
+static inline SDL_bool
+_bitarray_get(unsigned int *bitarray, size_t idx)
+{
+    if (bitarray[idx / WORD_BITS] & (1 << (idx % WORD_BITS)))
+        return SDL_TRUE;
+    else
+        return SDL_FALSE;
+}
+
```
oddbookworm, line 1111: Does endianness matter here? I always get nervous when bitshifting and pixels are involved

src_c/draw.c
```diff
@@ -1648,6 +1769,184 @@ draw_line(SDL_Surface *surf, int x1, int y1, int x2, int y2, Uint32 color,
...
+    Uint8 *pix;
+
+    // Von Neumann neighbourhood
+    int VN_X[] = {0, 0, 1, -1};
+    int VN_Y[] = {1, -1, 0, 0};
+
+    if (!(x1 >= cliprect.x && x1 < (cliprect.x + cliprect.w) &&
+          y1 >= cliprect.y && y1 < (cliprect.y + cliprect.h))) {
+        free(frontier);
+        free(mask);
```
oddbookworm, line 1842: why not `frontier_next`?

oddbookworm (commented):
A couple of questions

src_c/draw.c
```diff
@@ -978,6 +985,93 @@ rect(PyObject *self, PyObject *args, PyObject *kwargs)
...
+    flood_fill_result =
+        flood_fill_inner(surf, startx, starty, color, pattern, drawn_area);
+
+    if (pattern != NULL) {
+        SDL_FreeSurface(pattern);
+    }
+
+    if (did_lock) {
+        if (!pgSurface_Unlock(surfobj)) {
+            return RAISE(PyExc_RuntimeError, "error unlocking surface");
```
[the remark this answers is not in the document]
robertpfeiffer: I'll change that.

src_c/draw.c
robertpfeiffer: I don't think a byte can be something else than 8 bits in C. I don't think we need to worry about machines that predate the IBM system 360

src_c/draw.c
robertpfeiffer: I don't see how endianness can mess this up. We always use the same expression to find the bit for the index, and this bitarray is not saved to disk or sent over the network to machines with different endianness.

src_c/draw.c
robertpfeiffer: That was an oversight. Thanks for catching that.

robertpfeiffer: (It's a holdover from the extremely "clever" ring buffer code I showed you earlier. I could probably still go back to a single FIFO buffer, but I like the explicit steps here. It doesn't matter perf wise.)

src_c/draw.c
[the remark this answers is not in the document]
robertpfeiffer: Wait, no, I won't. This is consistent with the rest of draw.c

buildconfig/stubs/pygame/draw.pyi
```diff
@@ -568,3 +568,28 @@ def aalines(
...
+  def flood_fill(
+    surface: Surface,
+    color: Union[ColorLike, Surface],
+    start_point: Point
+) -> Rect:
+   """Replace the color of a cluster of connected same-color pixels, beginning
+   from the starting point, with a repeating pattern or solid single color
+
+   :param Surface surface: surface to draw on
+   :param color: color to draw with, the alpha value is optional if using a
```
MyreMylar:
```suggestion
   :param color: color, or surface pattern, to draw with. The alpha value is optional if using a
```

src_c/doc/draw_doc.h
```diff
@@ -10,3 +10,4 @@
 #define DOC_DRAW_LINES "lines(surface, color, closed, points, width=1) -> Rect\nDraw multiple contiguous straight line segments."
 #define DOC_DRAW_AALINE "aaline(surface, color, start_pos, end_pos, width=1) -> Rect\nDraw a straight antialiased line."
 #define DOC_DRAW_AALINES "aalines(surface, color, closed, points) -> Rect\nDraw multiple contiguous straight antialiased line segments."
+#define DOC_DRAW_FLOODFILL "flood_fill(surface, color, start_pos) -> Rect\nflood_fill(surface, pattern_surface, start_pos) -> Rect\nFill a connected area of same-color pixels."
```
MyreMylar:
```suggestion
#define DOC_DRAW_FLOODFILL "flood_fill(surface, color, start_point) -> Rect\nflood_fill(surface, pattern_surface, start_point) -> Rect\nFill a connected area of same-color pixels."
```

src_c/draw.c
```diff
@@ -1290,6 +1297,93 @@ rect(PyObject *self, PyObject *args, PyObject *kwargs)
...
+    SDL_Surface *surf = NULL;
+    int startx, starty;
+    Uint32 color;
+    SDL_Surface *pattern = NULL;
+    SDL_bool did_lock = SDL_FALSE;
+    int flood_fill_result;
+
+    int drawn_area[4] = {INT_MAX, INT_MAX, INT_MIN,
+                         INT_MIN}; /* Used to store bounding box values */
+    static char *keywords[] = {"surface", "color", "start_pos", NULL};
```
MyreMylar, line 1317:
```suggestion
    static char *keywords[] = {"surface", "color", "start_point", NULL};
```

src_c/draw.c
```diff
@@ -1290,6 +1297,93 @@ rect(PyObject *self, PyObject *args, PyObject *kwargs)
...
+        SDL_SetSurfaceRLE(pattern, SDL_FALSE);
+
+        color = 0;
+    }
+    else {
+        CHECK_LOAD_COLOR(colorobj);
+    }
+
+    if (!pg_TwoIntsFromObj(start, &startx, &starty)) {
+        return RAISE(PyExc_TypeError, "invalid start_pos argument");
```
MyreMylar, line 1352:
```suggestion
        return RAISE(PyExc_TypeError, "invalid start_point argument");
```

src_c/draw.c
```diff
@@ -1331,19 +1332,10 @@ flood_fill(PyObject *self, PyObject *arg, PyObject *kwargs)
...
-            return RAISE(PyExc_RuntimeError, "error converting pattern surf");
-        }
-
-        SDL_SetSurfaceRLE(pattern, SDL_FALSE);
-
-        color = 0;
-    }
-    else {
+        pat_surfobj = (pgSurfaceObject *)colorobj;
+        pattern = pat_surfobj->surf;  /* No conversion: we will map per-pixel */
```
robertpfeiffer: Why map per pixel? I assumed the pattern is small, and converting is faster. Think of filling an area with a checkerboard pattern, or stripes, if that helps. I don't want to die on this hill, but if you did this for perf reasons, I'll benchmark.

MyreMylar: I switched it back with the PG_ macro.

MyreMylar (approved):
OK, this seems to be compiling nicely for the CI now and I've tested it locally as well with the unit tests and some test code.

There are probably further extensions we could make to this idea (like color difference thresholds when propagating the fill) but I think this is useful - I could have used it when making my 'Pygame Paint' app. We could also tweak the implementation somewhat and probably the type hints, but what we have here seems workable to me.

I will leave it for a little bit longer for anymore feedback, before merging. Just had the idea to consider renaming the 'color' parameter to 'fill' to encompass that it can be a colour or a Surface...

Issue #2842, opened 2024-05-07
New test fail in `set_system_cursor` when integer value passed is out of range


ankith26:
**Environment:**

- **Operating system** (e.g. Windows, Linux(Debian), Linux(Ubuntu), Mac): Ubuntu 24.04
- **Python version** (e.g. 3.11.1, 3.8.5) : 3.12.3
- **SDL version** (e.g. SDL 2.0.12): 2.30.2
- **pygame-ce version** (e.g. 2.4.0.dev4, 2.1.3): 2.5.0.dev2

I recently updated to ubuntu 24.04 and now I'm getting a new test fail for reasons I cannot understand

```
ankith@AnkithLaptop:~$ py -m pygame.tests.mouse_test
pygame-ce 2.5.0.dev2 (SDL 2.30.2, Python 3.12.3)
/home/ankith/base_env/lib/python3.12/site-packages/pygame/tests/mouse_test.py:29: Warning: PyGame seems to be running through X11 on top of wayland, instead of wayland directly
  pygame.display.set_mode((500, 500))
...........F..
======================================================================
FAIL: test_set_system_cursor (__main__.MouseModuleTest.test_set_system_cursor)
Ensures set_system_cursor works correctly.
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/home/ankith/base_env/lib/python3.12/site-packages/pygame/tests/mouse_test.py", line 126, in test_set_system_cursor
    with self.assertRaises(pygame.error):
AssertionError: error not raised

----------------------------------------------------------------------
Ran 14 tests in 0.137s

FAILED (failures=1)
```

The testcase in question that is failing is
```py
# Right type, invalid value
with self.assertRaises(pygame.error):
    pygame.mouse.set_system_cursor(2000)
```

PR #2843, opened 2024-05-07
Add explicit check and error for system cursor

ankith26:
PR to fix https://github.com/pygame-community/pygame-ce/issues/2842

I am not sure how this issue is coming up only now, I went through the SDL code for it and there is no clear error raising path for this case.


Starbuck5 (approved):
Seems like a good change.

oddbookworm (approved):
LGTM! Maybe an issue on the SDL side is also a good idea?

Issue #2516, opened 2023-10-14
Possible benefit in restructuring `pgSurface_Blit` internals


itzpr3d4t0r:
In some of my testing I've observed that the setup and if chain needed to blit 32bit opaque (non alpha) surfaces was causing a 16% performance loss. Basically removing the whole if chain in `pgSurface_Blit` and only leaving a `result = SDL_BlitSurface(src, srcrect, dst, dstrect);` in the last "else" block was yielding a 16% performance benfit when using fblits.

I believe we could add a separate case (before the others) to check the "32bit opaque non alpha" case, since that's what people are most probably using most of the time. Generally speaking, i also believe we could add a separate version of `pgSurface_Blit` for the multi-blit fuctions that precalculate some conditions and lock/unlock the destination only once to save more performance.  I think more investigation is needed to check if this is actually the way to go or there's some other way.

The function as it stands now for reference:
```C
int
pgSurface_Blit(pgSurfaceObject *dstobj, pgSurfaceObject *srcobj,
               SDL_Rect *dstrect, SDL_Rect *srcrect, int blend_flags)
{
    SDL_Surface *src = pgSurface_AsSurface(srcobj);
    SDL_Surface *dst = pgSurface_AsSurface(dstobj);
    SDL_Surface *subsurface = NULL;
    int result, suboffsetx = 0, suboffsety = 0;
    SDL_Rect orig_clip, sub_clip;
    Uint8 alpha;
    Uint32 key;

    /* passthrough blits to the real surface */
    if (((pgSurfaceObject *)dstobj)->subsurface) {
        PyObject *owner;
        struct pgSubSurface_Data *subdata;

        subdata = ((pgSurfaceObject *)dstobj)->subsurface;
        owner = subdata->owner;
        subsurface = pgSurface_AsSurface(owner);
        suboffsetx = subdata->offsetx;
        suboffsety = subdata->offsety;

        while (((pgSurfaceObject *)owner)->subsurface) {
            subdata = ((pgSurfaceObject *)owner)->subsurface;
            owner = subdata->owner;
            subsurface = pgSurface_AsSurface(owner);
            suboffsetx += subdata->offsetx;
            suboffsety += subdata->offsety;
        }

        SDL_GetClipRect(subsurface, &orig_clip);
        SDL_GetClipRect(dst, &sub_clip);
        sub_clip.x += suboffsetx;
        sub_clip.y += suboffsety;
[...]

PR #2844, opened 2024-05-08
Always call HasColorKey before GetColorKey

### LINKED ISSUE #2821: `pygame.Surface.fblits` slower in 2.5.0.dev2 vs 2.4.1

Starbuck5:
SDL sets an error if GetColorKey is called on a surface that doesn't have colorkey, which is much more expensive performance-wise in the newest SDL versions. Therefore, let's call SDL_HasColorKey first to avoid the error setting cost.

Successor to #2835
Closes #2821

### DIFF
--- a/src_c/alphablit.c
+++ b/src_c/alphablit.c
@@ -114,7 +114,9 @@ SoftBlitPyGame(SDL_Surface *src, SDL_Rect *srcrect, SDL_Surface *dst,
         info.src = src->format;
         info.dst = dst->format;
         SDL_GetSurfaceAlphaMod(src, &info.src_blanket_alpha);
-        info.src_has_colorkey = SDL_GetColorKey(src, &info.src_colorkey) == 0;
+        if ((info.src_has_colorkey = SDL_HasColorKey(src))) {
+            SDL_GetColorKey(src, &info.src_colorkey);
+        }
         if (SDL_GetSurfaceBlendMode(src, &info.src_blend) ||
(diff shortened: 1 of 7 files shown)


ankith26 (approved):

itzpr3d4t0r (approved):
LGTM!

itzpr3d4t0r:
I suppose this closes #2516 as well since the performance gain over 2.4.1 matches what I observed there.

oddbookworm:
> I suppose this closes #2516 as well since the performance gain over 2.4.1 matches what I observed there. Apparently this regression allowed us to find a long-standing performance gap.

I would like to keep that issue open until we have a prerelease that has had time to be tested on actual projects

oddbookworm:
Gonna see if @bigwhoopgames can test a dev wheel from this pull and make sure that the performance regression is gone

itzpr3d4t0r:
I made one here: <https://discord.com/channels/772505616680878080/772505616680878083/1237761275795345552>

Issue #2849, opened 2024-05-13
Unhelpful error message in pygame.transform.gaussian_blur() with zero-dimensioned surfaces


jeremycryan:
**Environment:**

- **Operating system** Windows 7
- **Python version** 3.9.2
- **SDL version** 2.28.5
- **pygame-ce version** 2.4.1
- **Relevant hardware** N/A

**Current behavior:**

When a 0-by-0 surface is passed as an argument to `pygame.transform.gaussian_blur()`, it raises the following exception:

`ValueError: Blur routines do not support dest_surfaces that share pixels with the source surface. Likely the surfaces are the same, one of them is a subsurface, or they are sharing the same buffer.`

This is doubly misleading because:
- It doesn't help the user solve the actual problem (a size-zero surface as an argument)
- It misleadingly directs the user to a specific keyword argument that is not part of the issue, which they might not even be using

This also happens for any surface with 0 as only one of its dimensions (10x0, 0x30).

Code snippet to test this behavior:

```
import pygame
pygame.init()
surf = pygame.Surface((0, 0))
surf = pygame.transform.gaussian_blur(surf, 20)
```

**Expected behavior:**

Two potential fixes:

1. Don't raise an exception. If a 0-by-0 surface is passed in, just return a new 0-by-0 surface through a copy or similar.
2.  Raise a ValueError with a different error message that specifically mentions the dimensions of the surface argument.

I'd probably personally prefer #1, since it's less likely to cause errors in code flows that might produce surfaces of variable (and sometimes zero) size, and it's a fairly intuitive way to handle size-zero surfaces. I encountered this after trying to blur the output of `Pygame.font.Font.render()` for an empty string.

oddbookworm:
I didn’t even think it was possible to construct a surface with a zero dimension

jeremycryan:
> I didn’t even think it was possible to construct a surface with a zero dimension

It certainly isn't useful most of the time --- but lots of existing surface operations handle it gracefully and as expected (usually doing nothing).

The big one is probably `pygame.font.Font.render()` which you might want to draw nothing when a string is empty (such as a user not having typed anything in a text box yet).

PR #2852, opened 2024-05-13
Using blur on a surface with either width or height equal to 0 won't raise a ValueError

Damus666:
I tried to fix issue https://github.com/pygame-community/pygame-ce/issues/2849 and I tested locally that surfaces with either width == 0 or height == 0 won't raise a misleading error message. I also moved the comparison between `dst` and `src` sizes as if `src` is (30, 50) and `dst` is (0, 20), without moving the check, the function would have passed returning a (0, 20) sized surface which would probably cause bugs, both `src` and `dst` must have the same zero sizes for it to return gracefully.
Why not an exception? Like the issue creator said, all transform functions handle zero sized surfaces which are even returned some times (`pygame.Font.render`) so this one should be consistent. If you find anything wrong with my changes let me know.
I also modified the docs and added a test.

### DIFF
--- a/docs/reST/ref/transform.rst
+++ b/docs/reST/ref/transform.rst
@@ -245,6 +245,9 @@ Instead, always begin with the original image and scale to the desired size.)
    .. versionchanged:: 2.3.0
       Passing the calling surface as destination surface raises a ``ValueError``
 
+   .. versionchanged:: 2.5.0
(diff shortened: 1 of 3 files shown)


ankith26 (approved):

oddbookworm (approved):
LGTM!

PR #2853, opened 2024-05-13
Deprecate `setup.py docs`, make docs in meson build

ankith26:
This is another small step in our move away from `setup.py`.

`python setup.py docs` internally executes `python buildconfig/makeref.py`, so this PR makes the straightforward replacement.

Also, another change that this PR makes is that it adds an invocation of above into the meson build itself. This makes the build process further streamlined under one command.

Should also help fix the fails in #2850


MyreMylar:
Should we take the opportunity to rename 'makeref.py' if it is going to be user facing? That is kind of a terrible name for a 'build the docs' script file.

ankith26:
How about just renaming it to just `docs.py`? or `make_docs.py`?

MyreMylar (approved):
OK, that works for me 👍

zoldalma999 (approved):
Looks good to me, nice to see progress on moving away from setup.py

PR #2854, opened 2024-05-14
Bundle only the required pygame icon files

### LINKED ISSUE #2780: Exclude mac specific assets from other builds
This should probably happen after https://github.com/pygame-community/pygame-ce/pull/2557, let that get resolved first.

We have mac specific assets:
src_py/pygame_icon_mac.bmp
src_py/pygame_icon.icns

Combined, they are 500kb.

~~If we can remove that from the 40 non-mac builds we do, we can reduce the size of a release by 20mb.~~
Edit: forgot about whl compression. They are 286 kb combined compressed, so the 40 builds would save a total 11.4 mb. That's almost as much as getting rid of a full wheel.

On another note, what are pygame_icon.icns and pygame.ico even doing?

ankith26:
fixes #2780

### DIFF
--- a/src_py/meson.build
+++ b/src_py/meson.build
@@ -29,11 +29,12 @@ endif
 
 data_files = files(
     'freesansbold.ttf',
-    'pygame.ico',
     'pygame_icon.bmp',
-    'pygame_icon.icns',
-    'pygame_icon_mac.bmp',
 )
+if plat == 'mac'
+    data_files += 'pygame_icon_mac.bmp'
+endif
+
 install_data(data_files, install_dir: pg_dir, install_tag: 'pg-tag')
 
 subdir('_sdl2')


src_py/meson.build
```diff
@@ -29,11 +29,12 @@ endif
 
 data_files = files(
     'freesansbold.ttf',
-    'pygame.ico',
     'pygame_icon.bmp',
-    'pygame_icon.icns',
```
oddbookworm, line 34: Are the first and third files just not needed?

ankith26: They aren't needed, atleast to the best of my understanding.

Starbuck5:
Does this also need changes in setup.py?

ankith26:
I could do the corresponding changes in `setup.py`, but my idea is: now that we are moving away from it, we don't need to port any enhancements to it - just porting bugfixes will do

Starbuck5 (approved):
Ok then, this should be nice!

Issue #2159, opened 2023-05-14
New features with ``pygame.Surface.scroll`` method


bilhox:
Hello,

Last year, I experienced OpenGL and repeated textures, and I thought it would be interesting to have a similar feature in pygame. The way we do it right now with current means isn't that good (not optimized). So after discussions, the best would be to add a boolean kwarg to ``pygame.Surface.scroll`` named ``repeat``. When it's set to ``True``, it shifts surface pixels by an offset and all out of range pixels come back in the other side. As you can see below at the left (In a future, it would be interesting to implement the mirror repeated case):

![](https://learnopengl.com/img/getting-started/texture_wrapping.png)

Finally, we shall be able to use ``pygame.Vector2`` or ``tuple`` / ``list`` as arguments, rather than only 2 arguments ``dx`` and ``dy`` for offset.

This is an example summarizing everything I said:
```py
foo = pygame.image.load("foo.png")
offset = pygame.Vector2(10, 20)
foo.scroll(offset, repeat=True) # repeated
foo.scroll(offset) # not repeated
```

PR #2855, opened 2024-05-14
Allow erasing pixels in pygame.Surface.scroll and add repeat functionality

damusss:
The shifting was incorrect, not setting the space created by the shifted pixels to black. Now it does that [EDIT: now it does that with the erase flag only], but most importantly I think I implemented this feature idea: https://github.com/pygame-community/pygame-ce/issues/2159 implementing a way to allow for repeated shifting. The video proof is here: https://discord.com/channels/772505616680878080/772940667231928360/1240055647584911443

I would love feedback about the C code as I did quite a lot of pointer operations.
Why?
- The erasing functionality is probably expected to happen so it's nice to have it
- The repeat functionality gives to this function some reason to exist
- It's cool

Question: how can I add a test for it? Edit: I modified the test, tell me if it's incomplete

You can use the following code with my request to test locally:

```python
import pygame
pygame.init()
win = pygame.Window(size=(700, 700))
win.get_surface()
clock = pygame.Clock()

surf = pygame.image.load("B:/py/contributing/test.png").convert_alpha() # your image
surf = surf.subsurface(surf.get_bounding_rect())

main = pygame.Surface((600, 600))
main.blit(surf, surf.get_rect(center = (main.get_width()/2, main.get_height()/2)))
#main.set_clip(pygame.Rect(50, 50, 400, 400))

FLAG = pygame.SCROLL_REPEAT
SPEED = 4

while True:
    for e in pygame.event.get():
        if e.type == pygame.QUIT:
            pygame.quit()
            exit()
            
    win.get_surface().fill("black")
    screen = win.get_surface()    
    
    dir = [0, 0]
    if keys[pygame.K_LEFT]:
        dir[0] -= SPEED
    if keys[pygame.K_RIGHT]:
        dir[0] += SPEED
    if keys[pygame.K_UP]:
        dir[1] -= SPEED
    if keys[pygame.K_DOWN]:
        dir[1] += SPEED
    main.scroll(dir[0], dir[1], FLAG)
    
    screen.blit(main, main.get_rect(center=(win.size[0]/2, win.size[1]/2)))
    
    win.flip()
    dt = clock.tick(60)/1000
    win.title = str(clock.get_fps())
    
```


Damus666:
Could the 2 faulty tests be rerun?

MyreMylar:
> The shifting was incorrect, not setting the space created by the shifted pixels to black.

Why do you think this is incorrect? The docs currently say:

> Move the image by dx pixels right and dy pixels down. dx and dy may be negative for left and up scrolls respectively. **Areas of the surface that are not overwritten retain their original pixel values.**

Which looks to me like what the current behaviour is.

e.g.

Current main branch scroll a 200pixel wide surface right by 100 pixels:

image


After this PR:

image

I think we should retain the current behaviour as the default, and if we want the original pixels not overwritten by scrolled pixels cleared to a colour (and why specifically black?) that should be an option/parameter. Otherwise we will probably accidentally break somebody's code.

----

As an addendum - do we think this function is likely to be a performance sensitive one like blit, or one used less frequently? I'm wondering whether it should support keyword arguments or not, now it has 3 params (possibly 4 after the above).

MyreMylar (changes requested):
See comment above.

Damus666:
@MyreMylar yeah that was a misunderstanding on my side. I don't think 4 arguments are too much. If they are, what do you suggest? Another function for repeat? only positional?

bilhox:
Looking at how you implemented it, I believe the best option would be to use the same system used for flags on surfaces and window. Like that we can a fair amount of arguments for this function.

Damus666:
> Looking at how you implemented it, I believe the best option would be to use the same system used for flags on surfaces and window. Like that we can a fair amount of arguments for this function.

That would require 2 new constants pygame.SCROLL_REPEAT and pygame.SCROLL_ERASE i believe. the new arguments would go from 2 to 1, i guess

bilhox:
> > Looking at how you implemented it, I believe the best option would be to use the same system used for flags on surfaces and window. Like that we can a fair amount of arguments for this function.
> 
> That would require 2 new constants pygame.SCROLL_REPEAT and pygame.SCROLL_ERASE i believe. the new arguments would go from 2 to 1, i guess

I prefer this approach.

bilhox:
I think there were separate functions for the actual process of scrolling and the function definition, what about doing this to keep some kind of readability ?

src_c/surface.c
```diff
@@ -2338,43 +2336,214 @@ surf_scroll(PyObject *self, PyObject *args, PyObject *keywds)
...
+            Py_RETURN_NONE;
+        }
+    }
+    // Repeated scrolls are periodic so we can delete the exceeding value
+    if (dx >= w || dx <= -w) {
+        dx = dx % w;
+    }
+    if (dy >= h || dy <= -h) {
+        dy = dy % h;
+    }
```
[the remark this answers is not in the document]
damusss: It's not necessary. When dx or dy are >= w or h, and there is no repeat, that function returns early, so that code is never reached.

bilhox:
Code example so you understand how it simplifies the life a lot : https://gist.github.com/bilhox/ce024b994e536338a750a760024b5939

MyreMylar (approved):

docs/reST/ref/surface.rst
```diff
@@ -317,17 +317,27 @@
...
+         * ``pygame.SCROLL_ERASE``: the space created by the shifting pixels
+           is filled with black or transparency.
+
+         * ``pygame.SCROLL_REPEAT``: the pixels that disappear out of the
+           surface or clip bounds are brought back on the opposite side
+           resulting in an infinitely scrolling and repeating surface.
 
       .. versionaddedold:: 1.9
 
+      .. versionchanged:: 2.5.1 Add repeating scroll and allow erasing pixels
```
MyreMylar, line 339:
```suggestion
      .. versionchanged:: 2.5.2 Add repeating scroll and allow erasing pixels
```

MyreMylar:
All looks good to me now, the three modes work as I expect and add some nice functionality to this function which I expect is not often used in it's current form.

Perhaps it will see a resurgence? 👍

src_c/surface.c
```diff
@@ -2273,19 +2270,192 @@ surf_fblits(pgSurfaceObject *self, PyObject *const *args, Py_ssize_t nargs)
     return RAISE(PyExc_TypeError, "Unknown error");
 }
 
+static int
+scroll(SDL_Surface *surf, int dx, int dy, int x, int y, int w, int h,
```
bilhox, line 2274: I had loved if there was two distinct function ``scroll_repeat`` and ``scroll_erase``. I believe it's much more understandable than all in a single function. If you're not convinced, check ``pg_set_mode`` code 💀 .

bilhox (changes requested):
except what I mentionned, nice, it works correctly now !

src_c/surface.c
damusss: UH I'll try

docs/reST/ref/surface.rst
```diff
@@ -317,17 +317,27 @@
...
-      are not overwritten retain their original pixel values. Scrolling is
-      contained by the Surface clip area. It is safe to have dx and dy values
-      that exceed the surface size.
+      Move the Surface by dx pixels right and dy pixels down. dx and dy may be
+      negative for left and up scrolls respectively.
+
+      Scrolling is contained by the Surface clip area. It is safe to have dx
+      and dy values that exceed the surface size.
+
+      The scroll flag can be ``0`` (normal behavior) or:
```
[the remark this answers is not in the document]
damusss: You really don't need to request changes when some requested changes are already in action you know that right 😭 

does this satisfy you?

bilhox (approved):
LGTM !

damusss:
btw I really suggest squash and merge, 32 commits are not worth this change!!

bilhox:
except myremylar, we're only 2 who contributed to this PR, so I'm waiting tomorrow before merging it.

PR #2858, opened 2024-05-16
Fix pylint undefined usage error in sysfont

ankith26:
Fixes the recent pylint fails, by moving around the conditional imports to the functions where they are actually invoked.

### DIFF
--- a/src_py/sysfont.py
+++ b/src_py/sysfont.py
@@ -29,11 +29,6 @@ from os.path import basename, dirname, exists, join, splitext
 from pygame.font import Font
 from pygame import __file__ as pygame_main_file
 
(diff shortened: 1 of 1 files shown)


Damus666 (approved):

novialriptide (approved):
LGTM!

Starbuck5:
This PR makes me grumpy, I’m tired of package updates breaking our CI over and over.

So I’d hope something like this can be used- https://github.com/pygame-community/pygame-ce/pull/2699

Starbuck5 (approved):
I’m on my phone right now so I can’t really verify, but I trust that Ankith made this patch properly.

PR #2860, opened 2024-05-19
Run display autoinit on window API, fixes

### LINKED ISSUE #2605: Maximizing a window invalidates the window surface

ankith26:
fixes #2605

This PR takes the `display.set_mode`-like approach and implicitly calls display init whenever a window is constructed.

### DIFF
--- a/src_c/window.c
+++ b/src_c/window.c
@@ -7,6 +7,8 @@
 #include "doc/sdl2_video_doc.h"
 #include "doc/window_doc.h"
 
+static int is_window_mod_init = 0;
+
 #if !defined(__APPLE__)
 static char *icon_defaultname = "pygame_icon.bmp";
 static int icon_colorkey = 0;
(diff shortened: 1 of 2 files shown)


src_c/window.c
```diff
@@ -7,6 +7,8 @@
 #include "doc/sdl2_video_doc.h"
 #include "doc/window_doc.h"
 
+int is_window_mod_init = 0;
```
oddbookworm, line 10: Should this be static?

ankith26: It doesn't really make much of a difference in a context like this, but sure, I can make the change

MyreMylar (approved):
OK, LGTM 👍

oddbookworm (approved):
LGTM

robertpfeiffer:
is this really how we want to to things? I added this behaviour to pygame.display as a hack to keep backward compat with pygame 1.9, which required a call to set_mode after a VIDEORESIZE event.

ankith26:
I would not have minded the raise warning/error approach, but then I also don't see a problem with the currently implemented approach

PR #2862, opened 2024-05-20
Add a briefcase template to pygame-ce

Starbuck5:
This is similar to the PyInstaller hook we have, but for Briefcase, another distribution solution. (https://beeware.org/project/projects/tools/briefcase/)

The template and strategy are largely copied from the existing entrypoints in the briefcase library.

Worked on this during the Beeware sprint at Pycon US 2024.

I think this supersedes https://github.com/beeware/briefcase/pull/1804

To test, install briefcase with pip, install pygame-ce from this branch, and run `briefcase new`. After going through the briefcase options, it will have a "GUI Framework" option that will now report pygame-ce as an option. This works as a template for pygame-ce in briefcase, but is not necessary for supporting distributing pygame-ce apps in briefcase.

After this PR (in `briefcase new`):
```
-- GUI Framework -------------------------------------------------------------

What GUI toolkit do you want to use for this project?

  1) Toga
  2) PySide6       (does not support iOS/Android deployment)
  3) PursuedPyBear (does not support iOS/Android deployment)
  4) Pygame        (does not support iOS/Android deployment)
  5) pygame_ce     (does not support iOS/Android deployment)
  6) None
```


src_py/__briefcase/pygame_ce.py
```diff
@@ -0,0 +1,154 @@
...
+        screen.fill(BGCOLOR)
+        pygame.display.flip()
+
+    pygame.quit()
+"""
+
+    def pyproject_table_briefcase_app_extra_content(self):
+        return """
+requires = [
+    "pygame-ce",
```
Starbuck5, line 58: The template code here doesn't require any special version of pygame-ce, it's pretty basic code.

src_py/__briefcase/pygame_ce.py
```diff
@@ -0,0 +1,154 @@
...
+]
+test_requires = [
+{% if cookiecutter.test_framework == "pytest" %}
+    "pytest",
+{% endif %}
+]
```
oddbookworm, line 64: Is this required?

oddbookworm (approved):
LGTM! I just have one small thing that might be fine to remove

src_py/__briefcase/pygame_ce.py
Starbuck5: Talked to the Beeware people and they said not really, so I'll remove it.

src_py/__briefcase/pygame_ce.py
```diff
@@ -0,0 +1,149 @@
...
+
+class PygameCEGuiBootstrap(BaseGuiBootstrap):
+    display_name_annotation = "does not support iOS/Android deployment"
+
+    def app_source(self):
+        return """\
+import importlib.metadata
+import os
+import sys
+from pathlib import Path
```
MyreMylar, line 12:
```suggestion
```
looks like this import is unused.

MyreMylar (approved):
OK, LGTM 👍 

Tested it on my machine and got a hello world briefcase test program running.

ankith26 (commented):
I just went through the process of created a `.deb` package, and it went really smooth! I'm really starting to like this tool.

robertpfeiffer:
Does the .deb depend on pygame-ce, or does it bundle its own SDL, pygame, and so on?

ankith26:
It bundles the stuff in the `pygame-ce` manylinux wheel (so in a sense it is a self contained deb-file that doesn't use system dependencies, which is not very deb-file like)

If a `pygame-ce` debian package existed, we could make it depend on that, and that would remove the need for copying/duplication of `pygame-ce`

ankith26 (approved):
This PR has my approval. Even though I left a question, I feel it's better to merge this in for 2.5.0 anyways, as we can polish this up in future releases.

oddbookworm:
> Does the .deb depend on pygame-ce, or does it bundle its own SDL, pygame, and so on?

I think it is completely self-contained, I might do some testing later with a couple of ubuntu VM's that don't already have things installed

src_py/__briefcase/pygame_ce.py
```diff
@@ -0,0 +1,154 @@
...
+    "pytest",
+{% endif %}
+]
+"""
+
+    def pyproject_table_macOS(self):
+        return """\
+universal_build = true
+requires = [
+    "std-nslog~=1.0.0",
```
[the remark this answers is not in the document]
Starbuck5: It's a beeware thing apparently: https://github.com/beeware/std-nslog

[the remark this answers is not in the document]
freakboy3742: FYI: A macOS app doesn't have a stdout/stderr that is visible for diagnostic purposes. `std-nslog` is a package that redirects stdout and stderr to the system log so that any diagnostic information can be captured.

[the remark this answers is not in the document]
Starbuck5: Thanks for the clarification Russell. :)

PR #2863, opened 2024-05-20
Added shim to fix portmidi break

oddbookworm:
Portmidi broke due to a signature change, and it manifested recently with the 2.0.4 release being installed on my sytem. This shim fixes the cython so that we don't get a compiler error when building with newer portmidi

### DIFF
--- a/src_c/cython/pygame/pypm.pyx
+++ b/src_c/cython/pygame/pypm.pyx
@@ -523,7 +523,7 @@ cdef class Output:
 # in commit 64314cc3d1a6fdddfc6ff5408a3f83af685b8cea
 # portmidi changed the signature of Pt_Time from `PMEXPORT PtTimestamp Pt_Time()` to `PMEXPORT PtTimestamp Pt_Time(void)`
 # this change is significant in that no args in a C function declaration is treated differently than void
-cdef PtTimestamp compatShimDueToMidiChange(void* arg) noexcept:
+cdef PtTimestamp pgCompat_Pt_Time(void* arg) noexcept:
     return Pt_Time()
 
 cdef class Input:
@@ -546,7 +546,7 @@ cdef class Input:
         self.debug = 0
 
         err = Pm_OpenInput(&(self.midi), input_device, NULL, buffersize,
-                           &compatShimDueToMidiChange, NULL)
+                           &pgCompat_Pt_Time, NULL)
         if err < 0:
             raise Exception(Pm_GetErrorText(err))
 
--- a/src_c/cython/pygame/pypm.pyx
+++ b/src_c/cython/pygame/pypm.pyx
@@ -106,14 +106,13 @@ cdef extern from "portmidi.h":
                          PmDeviceID inputDevice,
                          void *inputDriverInfo,
                          long bufferSize,
-                         long (*PmPtr) (), # long = PtTimestamp
+                         PmTimeProcPtr time_proc, # long = PtTimestamp
                          void *time_info)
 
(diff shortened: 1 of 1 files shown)


Starbuck5 (approved):

src_c/cython/pygame/pypm.pyx
```diff
@@ -521,6 +520,11 @@ cdef class Output:
         while Pt_Time() == cur_time:
             pass
 
+# in commit 64314cc3d1a6fdddfc6ff5408a3f83af685b8cea
+# portmidi changed the signature of Pt_Time from `PMEXPORT PtTimestamp Pt_Time()` to `PMEXPORT PtTimestamp Pt_Time(void)`
+# this change is significant in that no args in a C function declaration is treated differently than void
+cdef PtTimestamp compatShimDueToMidiChange(void* arg) noexcept:
```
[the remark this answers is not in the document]
oddbookworm: renamed

src_c/cython/pygame/pypm.pyx
```diff
@@ -542,7 +546,7 @@ cdef class Input:
         self.debug = 0
 
         err = Pm_OpenInput(&(self.midi), input_device, NULL, buffersize,
-                           &Pt_Time, NULL)
+                           &pgCompat_Pt_Time, NULL)
```
ankith26, line 549:
```suggestion
                           <PmTimeProcPtr>&Pt_Time, NULL)
```
Untested fix for this problem that doesn't involve adding that compat shim. I was going though the code for other uses of `Pt_Time` and found this cast style already being used

Starbuck5: I had that instinct too, but I couldn't get it to work. Your patch:

```
  Error compiling Cython file:
  ------------------------------------------------------------
  ...
          cdef PmError err
          self.device = input_device
          self.debug = 0

          err = Pm_OpenInput(&(self.midi), input_device, NULL, buffersize,
                             <PmTimeProcPtr>&Pt_Time, NULL)

                             ^

  ------------------------------------------------------------

  C:\Users\charl\Desktop\pygame-ce\src_c\cython\pygame\pypm.pyx:545:27: Cannot assign type 'PmTimeProcPtr' (alias of 'PmTimestamp (*)(void *) noexcept') to 'long (*)(void) noexcept'
```

Plus I actually feel better about a compat shim that will definitely work versus a cast that just hits it until it works.

ankith26: Then how is the exact same cast not raising the same error on this line: https://github.com/pygame-community/pygame-ce/blob/89f98480116bf71b913bd40c4b04a5ef85b0c8a9/src_c/cython/pygame/pypm.pyx#L286

oddbookworm: Seems like it's a bug in cython potentially? I've reported it [here](https://github.com/cython/cython/issues/6212) because casting it in the generated C file before compiling does work

Starbuck5: I like the current strategy, it works, the suggested patches don't. Possibly either of you will figure out a way to get the cast to generate in Cython, but I'd still rather do it in the current fashion.

Since neither of you (or me) have figured out how to do the cast, I'm resolving this thread.

Starbuck5: I'd like to get this merged sooner rather than later so we can compile on these new systems with new portmidi.

MyreMylar (approved):
LGTM 👍

PR #2865, opened 2024-05-21
Force window surface update after set_size, hopefully proper fix for that pesky frequent pypy test fail

ankith26:
Reverts the hacky test fix applied in https://github.com/pygame-community/pygame-ce/pull/2832 that didn't work out

This new approach essentially forces the update via `SDL_GetWindowSurface` to happen immediately, so we are now ensuring that the window surface is updated before we even leave the function call.

### DIFF
--- a/src_c/window.c
+++ b/src_c/window.c
@@ -570,6 +570,11 @@ window_set_size(pgWindowObject *self, PyObject *arg, void *v)
     }
 
     SDL_SetWindowSize(self->_win, w, h);
+    if (self->surf) {
+        /* Ensure that the underlying surf is immediately updated, instead of
+         * relying on the event callback */
+        self->surf->surf = SDL_GetWindowSurface(self->_win);
+    }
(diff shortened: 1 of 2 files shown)


Damus666 (approved):

MyreMylar (approved):
Lets hope this is second time lucky 👍

oddbookworm (approved):
Hopefully this fixes that pesky intermittent fail

robertpfeiffer:
is this really how we want to to things? I added this behaviour to pygame.display as a hack to keep backward compat with pygame 1.9, which required a call to set_mode after a VIDEORESIZE event.

ankith26:
There is absolutely no behaviour change happening here. All it does is add an extra call to `SDL_GetWindowSurface` that runs before the one in the event watch runs (and when that runs again, there is no harm there)

Did you mean to comment this on another issue?

PR #2869, opened 2024-05-23
remove superfluous window creation from font and draw tests

robertpfeiffer:
Let's just let this run through CI as a separate PR instead.

### DIFF
--- a/test/draw_test.py
+++ b/test/draw_test.py
@@ -1070,11 +1070,11 @@ class BaseLineMixin:
         # Create some surfaces with different sizes, depths, and flags.
         surfaces = []
         for size in ((49, 49), (50, 50)):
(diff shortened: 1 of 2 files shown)


robertpfeiffer:
@oddbookworm I have isolated the segfaults into a separate PR. This is clearly an improvement over the previous version. It should not segfault.

Gabryel-lima (commented):
Removed several lines. Sinister

oddbookworm:
> @oddbookworm I have isolated the segfaults into a separate PR. This is clearly an improvement over the previous version. It should not segfault.

:skull: looks like it still is segfaulting on some of the windows CI runs

robertpfeiffer:
I think this is related to 16-bit surfaces.

PR #2871, opened 2024-05-24
Remove duplicate images in tutorials

MyreMylar:
Seems like we had multiple copies of some images in the docs source files.


Gabryel-lima:
> Seems like we had multiple copies of some images in the docs source files.

Is it just a case of transporting the assets dir, isolating it, passing the new readme paths and committing back? How did I see what you suggested? Because your observation was very good, I hadn't even noticed it when I was translating. Can I change? I'm sorry for the amount of questions, I'm just learning. I had never contributed to open source before.

zoldalma999 (approved):
Looks good to me! Saves about 0.9 mb as well (my generated folder went from 22.1 mb on current main to 21.2 mb on this branch) which is nice

Gabryel-lima (approved):
Yes, great! I'm even going to adjust my translation branch. I will even adjust my translation branch, if it is approved here. I strongly believe so

ankith26 (approved):

PR #2872, opened 2024-05-24
Add pygame.transform.bloom function

damusss:
I implemented the bloom algorithm to add it to `pygame.transform`.
I also added documentation and stubs.

I think this is the best implementation. This is the steps, if someone needs them to review:
- Create a bright pass filter surface
- Iterate the source surface and keep only the pixels with the luminance within the threshold
- Run either the box or blur algorithm
- On the lines where the blurred color is applied, also blit it to the original surface with additive blend

Please tell me if there are any C improvements, structure improvements or performance improvements.


src_c/transform.c
```diff
@@ -3382,6 +3382,288 @@ surf_gaussian_blur(PyObject *self, PyObject *args, PyObject *kwargs)
...
+    Uint8 *bpfcurrent = (Uint8 *)bpfsurf->pixels;
+
+    int x, y;
+    float c_mul = 255.0f * intensity;
+    for (y = 0; y < src->h; y++) {
+        for (x = 0; x < src->w; x++) {
+            Uint8 src_r = srccurrent[0], src_g = srccurrent[1],
+                  src_b = srccurrent[2];
+            float r = (float)src_r / 255.0f, g = (float)src_g / 255.0f,
+                  b = (float)src_b / 255.0f;
```
itzpr3d4t0r, line 3580: I don't think this will work with different formats, like RGBA ARGB etc.

src_c/transform.c
```diff
@@ -3382,6 +3382,288 @@ surf_gaussian_blur(PyObject *self, PyObject *args, PyObject *kwargs)
...
+    int smoothed_radius = blur_radius;
+    if (smooth_level > 1) {
+        smoothed_radius =
+            (int)sqrt(smooth_level * (blur_radius * blur_radius));
+    }
+
+    int bpp = src->format->BytesPerPixel;
+    int pitchdiff = src->pitch - src->w * bpp;
+
+    SDL_Surface *bpfsurf = newsurf_fromsurf(src, src->w, src->h);
```
itzpr3d4t0r, line 3568: There's a memory leak since you don't free this surface after use.

src_c/transform.c
Damus666: I usually only use rgba. are those shifts like 0, 1, 2, 3? either way I'll test it and address it

src_c/transform.c
Damus666: Forgot about it, thanks

src_c/transform.c
```diff
@@ -3382,6 +3382,288 @@ surf_gaussian_blur(PyObject *self, PyObject *args, PyObject *kwargs)
...
+
+    if (retsurf->w == 0 || retsurf->h == 0) {
+        return retsurf;
+    }
+
+    int smoothed_radius = blur_radius;
+    if (smooth_level > 1) {
+        smoothed_radius =
+            (int)sqrt(smooth_level * (blur_radius * blur_radius));
+    }
```
itzpr3d4t0r, line 3563: Isn't this valid only with gaussian blur?

Damus666: > Isn't this valid only with gaussian blur?

... maybe... but is it that important... is it there an equivalent for the box blur? Otherwise it could be safe to remove the argument. The user can expand the blur radius for smoothing themselves

src_c/transform.c
```diff
@@ -3382,6 +3382,288 @@ surf_gaussian_blur(PyObject *self, PyObject *args, PyObject *kwargs)
...
+            float luminance = r * 0.299f + g * 0.587f + b * 0.114f;
+
+            if (luminance > threshold && luminance != 0) {
+                float c = ((luminance - threshold) / luminance) * c_mul;
+                Uint8 new_r = (Uint8)(r * c);
+                Uint8 new_g = (Uint8)(g * c);
+                Uint8 new_b = (Uint8)(b * c);
+                bpfcurrent[0] = new_r > 255 ? 255 : new_r;
+                bpfcurrent[1] = new_g > 255 ? 255 : new_g;
+                bpfcurrent[2] = new_b > 255 ? 255 : new_b;
```
itzpr3d4t0r, line 3590: I don't think manually clamping here is necessary, a cast should suffice (from my tests it looks like it's working at least).

src_c/transform.c
```diff
@@ -3382,6 +3396,176 @@ surf_gaussian_blur(PyObject *self, PyObject *args, PyObject *kwargs)
...
+
+    if ((retsurf->w) != (src->w) || (retsurf->h) != (src->h)) {
+        return RAISE(PyExc_ValueError,
+                     "Destination surface not the same size.");
+    }
+
+    if (src->format->BytesPerPixel < 3 || retsurf->format->BytesPerPixel < 3) {
+        return RAISE(PyExc_ValueError,
+                     "Bloom is only allowed for 24 or 32 bit surfaces.");
+    }
```
itzpr3d4t0r, line 3464: If a destination is supplied you should also check that they have the same BytesPerPixel. Same Format is not necessary here as long as this is true.

src_c/transform.c
```diff
@@ -3382,6 +3396,209 @@ surf_gaussian_blur(PyObject *self, PyObject *args, PyObject *kwargs)
...
+
+                *dstp = new_pixel;
+            }
+
+            srcp++;
+            dstp++;
+        }
+        srcp += src_skip;
+        dstp += dst_skip;
+    }
```
itzpr3d4t0r, line 3531: Given that this takes a big chunk of code lets put it in a separate function, somethign like:
```C
void
filter_luminance(SDL_Surface *src, SDL_Surface *dst, float intensity, float threshold)
```

src_c/transform.c
```diff
@@ -3382,6 +3396,209 @@ surf_gaussian_blur(PyObject *self, PyObject *args, PyObject *kwargs)
...
+    if (src->format->BytesPerPixel != retsurf->format->BytesPerPixel) {
+        return RAISE(PyExc_ValueError,
+                     "Destination surface has different depth");
+    }
+
+    if (retsurf->w == 0 || retsurf->h == 0) {
+        return retsurf;
+    }
+
+    SDL_Surface *bpfsurf = newsurf_fromsurf(src, src->w, src->h);
```
itzpr3d4t0r, line 3475: Not checking that bpfsurf is npot NULL here.

src_c/transform.c
```diff
@@ -3382,6 +3396,209 @@ surf_gaussian_blur(PyObject *self, PyObject *args, PyObject *kwargs)
...
+
+    if (src->format->BytesPerPixel < 3 || retsurf->format->BytesPerPixel < 3) {
+        return RAISE(PyExc_ValueError,
+                     "Bloom is only allowed for 24 or 32 bit surfaces.");
+    }
+
+    if (src->format->BytesPerPixel != retsurf->format->BytesPerPixel) {
+        return RAISE(PyExc_ValueError,
+                     "Destination surface has different depth");
+    }
```
itzpr3d4t0r, line 3469: Given the recent changes this actually needs to check that the source and destination formats are the same to ensure the algorithm works. Something along these lines:
image

src_c/transform.c
```diff
@@ -3382,6 +3382,336 @@ surf_gaussian_blur(PyObject *self, PyObject *args, PyObject *kwargs)
...
+surf_bloom(PyObject *self, PyObject *args, PyObject *kwargs)
+{
+    pgSurfaceObject *dst_surf_obj = NULL;
+    pgSurfaceObject *src_surf_obj;
+    SDL_Surface *new_surf = NULL;
+    const char *blur_type_str = "box";
+
+    float intensity = 1, threshold = 0.5;
+    int blur_radius = 5;
+    char blur_type = 'b';
```
itzpr3d4t0r, line 3668: personally, I'd prefer gaussian_blur to be default since it's more precise. I believe we should be as precise as possible by default and if the uses wants less precision he can change it.

src_c/transform.c
```diff
@@ -3382,6 +3382,336 @@ surf_gaussian_blur(PyObject *self, PyObject *args, PyObject *kwargs)
...
+                             "blur_type",
+                             "dest_surface",
+                             0};
+
+    if (!PyArg_ParseTupleAndKeywords(
+            args, kwargs, "O!|ffisO!", kwlist, &pgSurface_Type, &src_surf_obj,
+            &intensity, &threshold, &blur_radius, &blur_type_str,
+            &pgSurface_Type, &dst_surf_obj)) {
+        return NULL;
+    }
```
itzpr3d4t0r, line 3683: The only mandatory argument is the surface to bloom, I think either just blur_radius or/and intensity should be mandatory as well.
And I think the argumets "order" should be changed a bit and ordered by importance, in case one doesn't use kwargs.
I'd do something like this:
`"surface" -> "blur_radius" (mandatory) ->  "intensity" (mandatory) -> "luminance_threshold" (optional) -> "blur_type" (optional) -> "dest_surface" (optional)`.

itzpr3d4t0r: The reordering isn't a strong requirement.

src_c/transform.c
Damus666: > personally, I'd prefer gaussian_blur to be default since it's more precise. I believe we should be as precise as possible by default and if the user wants less precision they can change it.

I set box as default because gassian blur is 5 or 6 times slower, but I guess the default should be the best looking. I think on a real usage, you would use box_blur for realtime effects (on not-too-big surfaces) while gaussian_blur for cached results. the gaussian bloom of realtime big surfaces is out of question, but the same logic applies to gaussian_blur without bloom

src_c/transform.c
Damus666: Yeah I like this 👍 
@itzpr3d4t0r I'll fix this 2 comments once I know if I should keep the buffer or keep the macros (I left a message on the discord server)

src_c/transform.c
itzpr3d4t0r: And on this note you could add a note about all this in the docs.

Damus666: > And on this note you could add a note about all this in the docs.

Wonderful idea

test/transform_test.py
```diff
@@ -1613,6 +1613,28 @@ def test_blur_zero_size_surface(self):
...
+                        for kwargs in [{}, {"dest_surface": dest_surface}]:
+                            pygame.transform.bloom(
+                                surface,
+                                blur_radius,
+                                intensity,
+                                threshold,
+                                blur_type,
+                                **kwargs,
+                            )
+
```
robertpfeiffer, line 1637: I think this should actually test the result of the operation. I suggest a test that checks that some pixels outside of the margin are still black, pixels inside the drawn area are grey, and pixels in the circle are white.

Damus666:
> I see no reason not to merge this, but I think this might be more flexible as three separate operations: Threshold, blur, and blit. At least, I think this could be more useful as a function that computes the bloom overlay with a black or transparent background.

Maybe it would be more flexible but not more performant as the blit is implemented within the blur using only 2 loops. I'd like for @itzpr3d4t0r to have a word about what you mentioned. I can always modify it as requested.

Damus666:
To add to what I just said, what i think would be best is to keep bloom as a function and maybe add the luminance filter as another function that can reuse the same code this pull request is using.

docs/reST/ref/transform.rst
```diff
@@ -279,6 +279,29 @@ Instead, always begin with the original image and scale to the desired size.)
...
+   An optional destination surface can be passed which is faster than creating a new
+   Surface. This destination surface must have the same dimensions (width, height) and
+   depth and format as the source Surface.
+
+   .. note:: A higher blur radius will be smoother but will be slower. The default
+      gaussian blur will be more precise but box blur is many times faster, which
+      is preferred for real-time effects. Using real-time bloom on very big surfaces
+      is generally not advised, caching is preferred.
+
+   .. versionadded:: 2.5.1
```
MyreMylar, line 303:
```suggestion
   .. versionadded:: 2.5.2
```

docs/reST/ref/transform.rst
```diff
@@ -279,6 +279,29 @@ Instead, always begin with the original image and scale to the desired size.)
 
    .. ## pygame.transform.gaussian_blur ##
 
+.. function:: bloom
+
+   | :sl:`apply the bloom effect to a surface`
+   | :sg:`bloom(surface, blur_radius, intensity, luminance_threshold=0.5, blur_type='gaussian', dest_surface=None) -> Surface`
+
+   Returns a surface where the bright pixels are blurred and added to the original
+   surface resulting in a bloom effect.
```
MyreMylar, line 288: After this section I think some details on the expected parameter ranges would be useful, particularly on the valid values for `blur_type`, which I believe are `gaussian` and `box`. Without some guidance it is hard to know if values like 'intensity` should normally be in the range of 0-1000, or 0.0 to1.0.

MyreMylar (commented):
I gave this the full walt test:

image

with this test program:

```python
import pygame


display_surf = pygame.display.set_mode((1280, 600))

walt_2 = pygame.transform.bloom(
    walt, blur_radius=2, intensity=1.5, luminance_threshold=0.6
)
walt_3 = pygame.transform.bloom(
    walt, blur_radius=15, intensity=4, luminance_threshold=0.4, blur_type="box"
)

pygame.init()

clock = pygame.time.Clock()

running = True

while running:
        if event.type == pygame.QUIT:
            running = False

    display_surf.blit(walt, (0, 100))
    display_surf.blit(walt_2, (417, 100))
    display_surf.blit(walt_3, (834, 100))

    pygame.display.flip()

    clock.tick(60)

```

I think it is a useful effect that developers will use. They'll probably use it, and all of the blurs more if we can eventually get them SIMD accelerated in some manner, but I have no problem adding this among them. Especially as it is reusing a lot of the code from the blurs.

My main criticism is that the input parameters are quite loosely defined right now in the documentation with no expected ranges, and not much description of what they effect in the output.

damusss:
> My main criticism is that the input parameters are quite loosely defined right now in the documentation with no expected ranges, and not much description of what they effect in the output.

you have a point. I'll improve the documentation and explain what each argument does

MyreMylar (approved):
Nice docs, LGTM 👍

src_c/transform.c
```diff
@@ -3778,183 +3778,192 @@ surf_average_color(PyObject *self, PyObject *args, PyObject *kwargs)
...
+    int kernel_radius = sigma * 2;                                            \
+    float lut_sum = 0.0;                                                      \
+    float *overall_buf =                                                      \
+        malloc(sizeof(float) * (dst_pitch * 2 + kernel_radius + 1));          \
+    float *buf = overall_buf;                                                 \
+    float *buf2 = overall_buf + dst_pitch;                                    \
+    float *lut = overall_buf + dst_pitch * 2;                                 \
+    if (overall_buf == NULL) {                                                \
+        return -1;                                                            \
+    }                                                                         \
```
oddbookworm, line 3798: I would move this up two lines so that you're not doing operations on a NULL pointer directly. Yeah, it's just arithmetic, but if `overall_buf` is NULL, then you're going to return anyway. Might as well bail immediately

src_c/transform.c
```diff
@@ -3778,183 +3778,192 @@ surf_average_color(PyObject *self, PyObject *args, PyObject *kwargs)
...
+                                                                              \
+    for (i = 0; i <= kernel_radius; i++) { /* init gaussian lut*/             \
+        /* Gaussian function */                                               \
+        lut[i] =                                                              \
+            expf(-powf((float)i, 2.0f) / (2.0f * powf((float)sigma, 2.0f)));  \
+        lut_sum += lut[i] * 2;                                                \
```
oddbookworm, line 3804: This is a nitpick, but I would make the 2 a 2.0f to be consistent with representing literals with the correct data type. Also applies to any other int literals that should actually be floats.

src_c/transform.c
```diff
@@ -3778,183 +3778,192 @@ surf_average_color(PyObject *self, PyObject *args, PyObject *kwargs)
...
 }
 
+#define GAUSSIAN_BLUR(src, dst, sigma, repeat, assign_code)                   \
+    Uint8 *srcpx = (Uint8 *)src->pixels;                                      \
+    Uint8 *dstpx = (Uint8 *)dst->pixels;                                      \
+    Uint8 nb = PG_SURF_BytesPerPixel(src);                                    \
+    int w = dst->w, h = dst->h;                                               \
+    int dst_pitch = dst->pitch;                                               \
+    int src_pitch = src->pitch;                                               \
+    int i, j, x, y, color;                                                    \
```
oddbookworm, line 3788: Is there a reason you're predeclaring all these iteration indices? The only thing I can think of is that you want to avoid creation time later, but these are primitives that need to be created anyway, so I don't see the need to add the extra step of needing to trace through code to find if these are persistent between parts of the code

oddbookworm (changes requested):
I've got a few comments here, but the code works as expected, so they're not that big

PR #2875, opened 2024-05-25
Deprecate `setup.py sdist`, suggest and use replacement pypa/build based command

ankith26:
We can use pypa/build to make an sdist file, now that we have the meson buildconfig.

I added a `.gitattributes` file, to skip a few files and ensure that the generated sdist is identical* to the one generated by `setup.py sdist`

\* There is one point of difference, the `docs/generated` folder is included in the sdist generated by `setup.py sdist`, but not the pypa/build sdist. This is one of the things that will be taken care due to #2853 being merged.

PS: this PR also fixes a little bug in #2853 (so this is a sort of a successor PR to that)

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,12 @@
+# define files/directories that won't be exported in an sdist
+# Keep in sync with MANIFEST.in
+.circleci/ export-ignore
+.github/ export-ignore
+buildconfig/ci/ export-ignore
+buildconfig/macdependencies/ export-ignore
+buildconfig/manylinux-build/ export-ignore
+
+.editorconfig export-ignore
+.gitattributes export-ignore
+.gitignore export-ignore
+.pre-commit-config.yaml export-ignore
--- a/.github/workflows/build-ubuntu-sdist.yml
+++ b/.github/workflows/build-ubuntu-sdist.yml
@@ -60,15 +60,14 @@ jobs:
       # https://github.com/orgs/community/discussions/47863
       run: |
         sudo apt-get update --fix-missing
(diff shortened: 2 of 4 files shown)


MyreMylar (approved):
LGTM 👍

.gitattributes
```diff
@@ -0,0 +1,12 @@
+# define files/directories that won't be exported in an sdist
+# Keep in sync with MANIFEST.in
```
Starbuck5, line 2: So once setup.py sdist is removed MANIFEST.in can be removed?

Can you point me to a piece of documentation that says this is how pypa/build expects to see files to ignore? I can't find any mention of this in my searches.

Does it need another new file in the root?

Starbuck5:
> * There is one point of difference, the docs/generated folder is included in the sdist generated by setup.py sdist, but not the pypa/build sdist. This is one of the things that will be taken care due to https://github.com/pygame-community/pygame-ce/pull/2853 being merged.

If that's merged, how does it need taking care of? I don't understand.

.gitattributes
Starbuck5: Ugh I expected it to be a pypa/build thing, not a meson thing. Okay.

ankith26:
> I ran this locally and saw the generated docs weren't in the sdist.

Yes, that's intended. And the purpose of #2853 is to ensure that the setup docs command runs atleast once as part of the build process itself (and therefore, runs when the sdist is installed on the user side) so in the end, the user still has the generated docs as before.

Starbuck5 (approved):
Ok, checked that out and it works.

Cool!

Issue #2751, opened 2024-03-17
set_source_location fails at runtime on some audio setups


Starbuck5:
This hasn't been reported on this repo yet, but we really need to do something about it, so making this issue.

When calling `set_source_location` on certain mixer setups
```
pygame.error: Unsupported audio channels
```

Here's the SDL_mixer issue for context: https://github.com/libsdl-org/SDL_mixer/issues/573

This has hit both bigwhoop and dafluffypotato when distributing their games to people where mixer defaults to 7.1 surround on their setup.

- Should we document how to force stereo in the docs for this function and tell people to force stereo?
- Can we prevent mixer from opening in 7.1 mode somehow?
- We could make this a silent fail instead of an exception? Maybe not a good idea.

Starbuck5:
Added a warning/reference to the docs to get this as resolved as I think it can be for 2.5.0.

PR #2881, opened 2024-05-28
Add warning about 7.1 stereo on set_source_location

Starbuck5:
For #2751

### DIFF
--- a/docs/reST/ref/mixer.rst
+++ b/docs/reST/ref/mixer.rst
@@ -576,6 +576,23 @@ The following file formats are supported
       `angle`: Angle is in degrees.
       
       `distance`: Range from 0 to 255.
+
+      .. warning:: This function currently fails and raises a
+         :exc:`pygame.error` when using 7.1 surround sound. 
+         By default, the mixer module will use what the hardware is best
+         suited for, so this leads to hardware specific exceptions when using
+         this function.
+
+         One way of avoiding this is only using :func:`set_source_location`
+         with forced stereo. For example:
+
+         ::
+
+            pygame.mixer.pre_init(
+               channels=2,
+               allowedchanges=pygame.AUDIO_ALLOW_FREQUENCY_CHANGE,
+            )
+            pygame.init()
       
       .. versionadded:: 2.3.0
       


MyreMylar:
What about testing the available hardware/audio device channels as a safety wrapper?

e.g.

```
if (pygame.mixer.get_init()[2] != 8):
     channel.set_source_location(90, 255)
```

Or we could handle this ourselves in the code for `chan_set_source_location()` using `Mix_QuerySpec()` and just skipping positional audio with 8 hardware channels until SDL resolves this?

docs/reST/ref/mixer.rst
```diff
@@ -576,6 +576,22 @@ The following file formats are supported
       `angle`: Angle is in degrees.
       
       `distance`: Range from 0 to 255.
+
+      .. warning:: This function currently fails to position audio
```
MyreMylar, line 580:
```suggestion
      .. warning:: This function currently raises an exception
```

just to be clear?

docs/reST/ref/mixer.rst
```diff
@@ -576,6 +576,22 @@ The following file formats are supported
...
+         when using 7.1 surround sound. By default, the mixer module
```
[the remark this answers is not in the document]
Starbuck5: I didn't want to bring up "channel" more than necessary in this description, since this in the docs for a "Channel" object, but 8 channel is completely unrelated to those objects.

Starbuck5:
> What about testing the available hardware/audio device channels as an advised safety wrapper?
> 
> e.g.
> 
> ```
> num_dev_channels = pygame.mixer.get_init()[2]
> if (num_dev_channels  != 8):
>      channel.set_source_location(90, 255)
> ```
> 
> Or we could handle this ourselves in the code for `chan_set_source_location()` using `Mix_QuerySpec()` and just skipping positional audio with 8 hardware channels until SDL resolves this?

That isn't a bad idea. I went with my current recommendation because it will (theoretically) keep the source position working. If someone has some crazy fancy setup that supports 7.1 audio, and SDL emulates stereo on top of it, and the source location is set, they should still be able to hear the sound coming from an angle.

I'm hesitant to ignore an exception (or never even try) based on Mix_QuerySpec, since SDL will eventually fix this and then we'll need to remove that code. Also it gives no feedback to the user on why the function isn't working.

MyreMylar (approved):
OK LGTM 👍 

Hopefully it gets fixed SDL side at some point and we can un-caveat this function.

docs/reST/ref/mixer.rst
```diff
@@ -576,6 +576,23 @@ The following file formats are supported
       `angle`: Angle is in degrees.
       
       `distance`: Range from 0 to 255.
+
+      .. warning:: This function currently fails and raises an
```
Matiiss, line 580: > raises an

probably should be
> raises **a**

Matiiss: and I forgot to select the next line for context, but it's _:exc:`pygame.error`_

Matiiss (approved):
Otherwise LGTM of course 👍

docs/reST/ref/mixer.rst
Starbuck5: Yep you're right. There's a bunch of instances of what I did in the docs already if you want to make an extremely important PR. :D

PR #2882, opened 2024-05-28
Add test & fix for unusual pitch in `surface.premul_alpha()`

### LINKED ISSUE #2750: premul_alpha doesn't respect weird surface pitches
I found this while reviewing the code of #2615

If the surface pitch is not aligned to 4 bytes, it doesn't properly skip through rows.

For example:
```py
import pygame

surf_height = 5


def create_surface_from_byte_width(byte_width):
    byte_data = bytes(byte_width * surf_height)
    surf_width = byte_width // 4

    dest = pygame.image.frombuffer(
        byte_data, (surf_width, surf_height), "RGBA", pitch=byte_width
    )
    dest.fill((120, 50, 70, 200))
    return dest


dest = create_surface_from_byte_width(8)  # 2 pixels wide
print("Surface 1: pitch =", dest.get_pitch())
print("Before premul:", dest.get_at((dest.get_width() - 1, dest.get_height() - 1)))
dest = dest.premul_alpha()
print("After premul:", dest.get_at((dest.get_width() - 1, dest.get_height() - 1)))

dest = create_surface_from_byte_width(10)  # 2.5 pixels wide
print("Surface 2: pitch =", dest.get_pitch())
print("Before premul:", dest.get_at((dest.get_width() - 1, dest.get_height() - 1)))
dest = dest.premul_alpha()
print("After premul:", dest.get_at((dest.get_width() - 1, dest.get_height() - 1)))
```

```
Surface 1: pitch = 8
Before premul: Color(120, 50, 70, 200)
After premul: Color(94, 39, 55, 200)
Surface 2: pitch = 10
Before premul: Color(120, 50, 70, 200)
After premul: Color(0, 0, 0, 0)
```
The final pixels of the 10-byte wide surface are never reached by the premultiply algorithm.

MyreMylar:
fixes #2750

From what I understand, these unusual, non-pixel aligned, surface pitches won't practically come up on any modern desktop systems (possibly no modern systems) so I prioritised fixing it for the least performance sensitive versions of this function - sse2 & the non-SIMD fallback. These are also the versions I originally wrote so I had a better idea of how they were supposed to work.

The basic fix is to add in the standard 'skip' value that is used in all the blitters to handle pitch issues between two different surfaces - usually these are pixel aligned. In the SSE2 case, to deal with the .5 of a pixel overlap in the pitch case we have to cast the skip value down to Uint8 to get to 'channel', or single byte, level of pointer incrementing as we only want to skip 2 channels worth of a pixel (2 bytes) rather than a whole pixel (4 bytes).

I think that makes sense anyway.

This probably needs feedback from @itzpr3d4t0r and @Starbuck5 to see if they think what I've changed makes sense and if we need to do anything else here.

### DIFF
--- a/src_c/alphablit.c
+++ b/src_c/alphablit.c
@@ -2841,7 +2841,8 @@ premul_surf_color_by_alpha(SDL_Surface *src, SDL_Surface *dst)
         // since we know dst is a copy of src we can simplify the normal checks
 #if !defined(__EMSCRIPTEN__)
 #if SDL_BYTEORDER == SDL_LIL_ENDIAN
-    if ((PG_SURF_BytesPerPixel(src) == 4) && pg_has_avx2()) {
+    if ((PG_SURF_BytesPerPixel(src) == 4) &&
+        (src->pitch % PG_SURF_BytesPerPixel(src) == 0) && pg_has_avx2()) {
         premul_surf_color_by_alpha_avx2(src, dst);
         return 0;
     }
@@ -2873,6 +2874,8 @@ premul_surf_color_by_alpha_non_simd(SDL_Surface *src, SDL_Surface *dst)
     int srcbpp = PG_FORMAT_BytesPerPixel(srcfmt);
     int dstbpp = PG_FORMAT_BytesPerPixel(dstfmt);
     Uint8 *src_pixels = (Uint8 *)src->pixels;
+    int srcskip = src->pitch - (width * srcbpp);
+    int dstskip = dst->pitch - (width * dstbpp);
     Uint8 *dst_pixels = (Uint8 *)dst->pixels;
 
     int srcpxskip = PG_SURF_BytesPerPixel(src);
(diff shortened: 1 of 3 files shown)


ankith26:
IG the same approach should probably be taken on all simd surface manipulations not just premul

itzpr3d4t0r (approved):
LGTM!

src_c/alphablit.c
```diff
@@ -2841,7 +2841,8 @@ premul_surf_color_by_alpha(SDL_Surface *src, SDL_Surface *dst)
         // since we know dst is a copy of src we can simplify the normal checks
 #if !defined(__EMSCRIPTEN__)
 #if SDL_BYTEORDER == SDL_LIL_ENDIAN
-    if ((PG_SURF_BytesPerPixel(src) == 4) && pg_has_avx2()) {
+    if ((PG_SURF_BytesPerPixel(src) == 4) &&
+        (src->pitch % PG_SURF_BytesPerPixel(src) == 0) && pg_has_avx2()) {
```
ankith26, line 2845: I think this test should be more robust. It should check for `src->pitch == (src->w * PG_SURF_BytesPerPixel(src))` because I can imagine pitch may still be a multiple of of bpp and not be equal to the minimal possible pitch.

MyreMylar:
```suggestion
        (src->pitch == (src->w * PG_SURF_BytesPerPixel(src))) && pg_has_avx2()) {
```

PR #2884, opened 2024-05-29
Add SIMD paths to .set_alpha() blit onto opaque

itzpr3d4t0r:
The way we currently handle this code uses `SDL_BlitSurface` internally:
```Py
import pygame

pygame.init()

s = pygame.Surface((size, size))
s.fill((31, 3, 44))
s.set_alpha(123)

screen.blit(s, (0, 0))
```

Unfortunately, SDL_BlitSurface only implements a "pixel by pixel" approach, which significantly slows down these types of blits. In comparison, performing a similar blit with a natively alpha surface filled with a specific color is substantially faster. This difference is puzzling because, with a surface alpha, we can reuse the same alpha value for all pixels, whereas we cannot do this in the other case.

Moreover, all our alpha blit algorithms are maintained for compatibility reasons. Specifically, these blitters use special alpha blending formulas to conform to older SDL versions, something the set_alpha() blit mode does not address.

Currently, we have an AVX/SSE implementation for set_alpha(), but it only applies when blitting to another surface with alpha. This PR extends the implementation to handle blitting onto an opaque surface.

This PR also adjusts our AVX2 macros to be slightly more efficient and simpler (code size wise).

As expected, the results show a significant improvement:
image

image

A small program to test some sizes and times:
```Py
import pygame
from timeit import timeit

pygame.init()

data = []

for size in range(10, 500, 10):
    s = pygame.Surface((size, size))
    s.fill((31, 3, 44))
    s.set_alpha(123)

    tot_time = timeit(lambda: screen.blit(s, (0, 0)), number=10000)
    print(f"Finished size {size} in {tot_time:.8f} seconds")
    data.append(tot_time)

print(f"Total time: {sum(data):.8f} seconds")
```

### DIFF
--- a/src_c/alphablit.c
+++ b/src_c/alphablit.c
@@ -211,6 +211,30 @@ SoftBlitPyGame(SDL_Surface *src, SDL_Rect *srcrect, SDL_Surface *dst,
 #endif /* __EMSCRIPTEN__ */
                         alphablit_alpha(&info);
                     }
+                    else if (info.src_blend != SDL_BLENDMODE_NONE &&
+                             !src->format->Amask && !dst->format->Amask) {
+#if !defined(__EMSCRIPTEN__)
+#if SDL_BYTEORDER == SDL_LIL_ENDIAN
+                        if (PG_SURF_BytesPerPixel(src) == 4 &&
+                            PG_SURF_BytesPerPixel(dst) == 4 &&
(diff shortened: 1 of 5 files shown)


Starbuck5:
> This PR also adjusts our AVX2 macros to be slightly more efficient and simpler (code size wise).

This is much less controversial, could be in another PR and evaluated on its own.

PR #2886, opened 2024-05-29
Add `(F)Rect.collidecircle()`

Damus666:
This PR adds the `Rect/Frect.collidecircle()` method to test if a circle overlaps the rect. I have included docs, tests, and the C code.
To achieve this, I had to do a small change to the C Code that won't break older code. What effectively changed by this PR:
- `(F)Rect.collidecircle()`
- `pygame.geometry.(F)Rect`

`pygame.rect.(F)Rect` still exists and works perfectly. Stubs have been updated to effect the change.
It's not convoluted, the code additions are very small and 0.1% of the C code changes. You can continue using the rect functions and objects in your C PRs as nothing happened.
This is the only solution to add the method of the title as it would cause circular imports/segfaults/ not-working checks otherwise. 
I know it can look scary but if you look at the changes, you'll see it's really harmless, and it even makes sense. The majority of changes are _moving_ code, not creating new one.

## Credits
This are the people that worked on the https://github.com/pygame-community/pygame-geometry module:
Thanks also to @Starbuck5
Finally thanks to @Starbuck5 and @itzpr3d4t0r for supporting the idea of moving the `(F)Rect` objects in an harmless way.


src_c/rect_impl.h
```diff
@@ -1337,6 +1345,21 @@ RectExport_colliderect(RectObject *self, PyObject *const *args,
...
+{
+    pgCircleBase circle;
+    InnerRect srect = self->r;
+
+    if (!pgCircle_FromObjectFastcall(args, nargs, &circle)) {
+        return RAISE(PyExc_TypeError, "A CircleType object was expected");
+    }
+
+    return PyBool_FromLong(
+        pgCollision_RectCircle(srect.x, srect.y, srect.w, srect.h, &circle));
```
itzpr3d4t0r, line 1360: Better cast the rect attributes to doubles here.

Damus666: I forgot geometry uses doubles, thanks

MyreMylar:
I think a reasonable argument can be made, regardless of the technical difficulties, that it just makes sense to have all our geometry shapes (Rect, FRect & Circle) under one submodule long term.

Damus666:
> I think a reasonable argument can be made, regardless of the technical difficulties, that it just makes sense to have all our geometry shapes (Rect, FRect & Circle) under one submodule long term.

Exactly, I totally agree. geometry is not geometry if there aren't all geometries in it.

src_c/include/_pygame.h
```diff
@@ -204,35 +210,42 @@ typedef struct {
...
-    (*(SDL_FRect * (*)(PyObject *, SDL_FRect *)) PYGAMEAPI_GET_SLOT(rect, 8))
+#define pgFRect_FromObject                       \
+    (*(SDL_FRect * (*)(PyObject *, SDL_FRect *)) \
+         PYGAMEAPI_GET_SLOT(geometry, 8))
 
-#define pgFRect_Normalize (*(void (*)(SDL_FRect *))PYGAMEAPI_GET_SLOT(rect, 9))
+#define pgFRect_Normalize \
+    (*(void (*)(SDL_FRect *))PYGAMEAPI_GET_SLOT(geometry, 9))
 
-#define import_pygame_rect() IMPORT_PYGAME_MODULE(rect)
```
MyreMylar, line 235: Should we consider just replacing all calls to `import_pygame_rect()` with `import_pygame_geometry()` and removing `import_pygame_rect()` ? It is internal API only so we are likely only confusing ourselves in the future by keeping `import_pygame_rect()` around and just having it import geometry.

MyreMylar (commented):
This works in my local testing, and I actually generally approve of moving Rect & FRect under geometry with a little python shim layer for backwards compatibility.

I don't know why we are keeping `import_pygame_rect()` around but that's not essential to this PR, so I will approve anyway.

MyreMylar (approved):
👍

src_c/include/_pygame.h
Damus666: I didn't do that as I didn't know if doing that would cause too many internal changes, but I initially wanted to rename it, and I can do it if it's needed. Update me if you think I should change it, maybe based on some other contributor's opinion.

MyreMylar: It's not essential to this PR IMHO, I guess we'll see how everyone else feels about moving Rect/Frect and shimming.

buildconfig/stubs/pygame/geometry.pyi
```diff
@@ -113,3 +122,4 @@ class Circle:
     def as_frect(self) -> FRect: ...
     def __copy__(self) -> Circle: ...
     copy = __copy__
+
```
itzpr3d4t0r, line 125: missing a newline here.

itzpr3d4t0r:
Needless to say I'm on the same boat with @Damus666 on this one. I think this move makes sense and it's actually an elegant solution which requires minimal code change on our end and has 0 impact to the end user. It might even solve some people's questions like: "why can't I import Rect/Frect from geometry? Isn't it a shape?" in the future.

itzpr3d4t0r (approved):
Aside from a docs correction this LGTM on the pygame-ce side. On the geometry side i got this error:
image
I suppose we'll need to update our `pygame.h` file in geometry, so it's not an issue.
After my approval the PR could be merged but I think it's better to hear from someone else on the topic.

ankith26 (changes requested):
Essentially, this seems like the first change where something is being added to "stable API" to support something in the "experimental API" section.

These changes look somewhat risky, and personally, I would like to hold it back for a while, till we can claim `pygame.geometry` as "stable" as a whole

PR #2888, opened 2024-05-30
fix some METH_NOARGS arg missing

pmp-p:
References:
https://docs.python.org/3/c-api/structures.html#c.METH_NOARGS

why it is important: 
https://blog.pyodide.org/posts/function-pointer-cast-handling/

A parser in CI would be usefull indeed, as only WebAssembly can spot them, and only at runtime.

### DIFF
--- a/src_c/window.c
+++ b/src_c/window.c
@@ -145,7 +145,7 @@ window_destroy(pgWindowObject *self, PyObject *_null)
 }
 
 static PyObject *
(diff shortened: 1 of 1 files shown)


Damus666 (approved):

ankith26 (approved):
LGTM, thanks :tada:

zoldalma999 (approved):

PR #2890, opened 2024-05-30
Add license identifier to readme

### LINKED ISSUE #1737: Please clarify license in README (3521)
#  **Issue &numero;3521 opened by *[ruben2020](https://github.com/ruben2020)* at 2022-10-27 09:48:01**

Is pygame licensed under LGPL version 2.1 only or under LGPL version 2.1 or later?

Please clarify this in your README file.

It may also be helpful to use the [SPDX license identifier](https://spdx.org/licenses/) which is explicit about this i.e. `LGPL-2.1-only` or `LGPL-2.1-or-later`.<hr>

#  Comments

# # #  **[MyreMylar](https://github.com/MyreMylar)* commented at 2022-11-06 12:09:52*

My feeling is that we would probably land at `LGPL-2.1-or-later`

[This page](https://fossa.com/blog/open-source-software-licenses-101-lgpl-license/) is pretty good at clarifying the difference.

Essentially, with the `or-later` if somebody took the pygame source code to make pygame-plus they could license it under LGPL 2.1 or LGPL 3.0 or a theoretical LGPL 4.0. Without the `or-later` pygame-plus would have to be licensed only under LGPL 2.1.

The general trend has been to make the LGPL more 'restrictive' (as in restrictive for any sort of commercial use) over time rather than more permissive so I wouldn't expect a 4.0 license to suddenly flip around and allow a theoretical pygame-mars-edition to be used for brutal private corpo crackdowns.

MyreMylar:
fixes #1737

I know @Starbuck5 wants to rewrite the readme anyway, but I thought I'd toss this into the pot as a thing to clarify.

The debate here would be between the two variants:

- LGPL-2.1-only
- LGPL-2.1-or-later

I have no strong feelings on this topic but I can see why some users might want it clarified for legal reasons.


novialriptide (approved):
LGTM

MyreMylar:
I should note that @ankith26 expressed a preference on discord for `LGPL-2.1-only` on the basis that it is the effective status quo.

As I said before I don't really mind either way this was originally raised by @ruben2020 so perhaps they have stronger feelings on the topic.

ruben2020:
@MyreMylar If you're fine either way, I would recommend LGPL-2.1-or-later as it is more flexible for users than LGPL-2.1-only.

ankith26:
We do have a restriction of not changing the terms of the license, and I don't want this PR to be interpreted as so.

Hence I believe we are more or less forced to maintain status quo, and I believe `LGPL-2.1-only` achieves it, even though technically both `LGPL-2.1-only` and `LGPL-2.1-or-later` have identical license text (which is what creates this confusion in the first place).

I would like to hear @Starbuck5 s opinions here to settle the matter

Starbuck5:
I'm fine to go 2.1-or-later.

Thoughts:
- My theory was that Pete Shinners chose this license because it's the same one as SDL1 used. Looking at SDL1, they almost always refer to it as the "LPGL", not "LPGL v2", so the spirit of the "or later" is there.
- This is a matter up to interpretation, so it's fine for us to interpret it this way
- This is somewhat academic, will not make a difference to many people

Starbuck5 (approved):

PR #2893, opened 2024-05-30
One method to fix the NaN/INT_MAX segfault in blit

### LINKED ISSUE #2694: Fatal Python error: Segmentation fault caused by Pygame
**Environment:**
- **Operating system**: Ubuntu 20.04 LTS
- **Python version**: 3.8.10
- **SDL version**: 2.28.5
- **pygame-ce version**: 2.4.0 (just installed from pip) 
- **Relevant hardware**: -

**Current behavior:**

This issue has appeared in both original pygame and in pygame-ce.

I have a game in which some cars navigate the environment with obstacles using sensor data and they collide quite frequently with each other and with obstacles. The problem is that sometimes during collisions the program crashes. It throws `Fatal Python error: Segmentation fault` but I'm 99% sure, that it's pygame. I use `mask.overlap()` function to detect collisions and when I remove parts that use this function, the program works perfectly fine. What's interesting but very confusing is that the trace points to the `blit()` function. Maybe I should also mention that I rotate the sprites using `rotozoom()` and create masks from rotated images.

**Expected behavior:**

Pixel-perfect sprite collisions using `mask.overlap()` without getting segmentation faults.

**Steps to reproduce:**

Create a number of moving sprites and check collisions between  them using the `mask.overlap()` function.

Unfortunately, I cannot easily provide an adequate path to reproduce the issue as I am not sure what really causes it (except that it's part of pygame). If you are really interested, I can provide the repo with the code of my program, but it also uses ROS2 and some of its additional packages so you will be required to download them. Then the steps to reproduce are:

1. Launch the program
2. Use arrows to move and collide with objects
3. Try to collid with objects until it crashes

```
Fatal Python error: Segmentation fault

Current thread 0x00007f504d0a7740 (most recent call first):
  File "/home/leev/workspaces/gr-kinematic-sim/install/gr_kinematic_sim/lib/python3.8/site-packages/gr_kinematic_sim/custom_utils/object_tools.py", line 71 in draw
  File "/home/leev/workspaces/gr-kinematic-sim/install/gr_kinematic_sim/lib/python3.8/site-packages/gr_kinematic_sim/custom_utils/robots.py", line 103 in call_sensors
  File "/home/leev/workspaces/gr-kinematic-sim/install/gr_kinematic_sim/lib/python3.8/site-packages/gr_kinematic_sim/custom_utils/robots.py", line 108 in draw
  File "/home/leev/workspaces/gr-kinematic-sim/install/gr_kinematic_sim/lib/python3.8/site-packages/gr_kinematic_sim/custom_utils/gametools.py", line 141 in draw_every_sprite_in_list
  File "/home/leev/workspaces/gr-kinematic-sim/install/gr_kinematic_sim/lib/python3.8/site-packages/gr_kinematic_sim/main.py", line 94 in main
  File "/home/leev/workspaces/gr-kinematic-sim/install/gr_kinematic_sim/lib/gr_kinematic_sim/sim", line 11 in <module>
Fatal Python error: (pygame parachute) Segmentation Fault

Current thread 0x00007f504d0a7740 (most recent call first):
  File "/home/leev/workspaces/gr-kinematic-sim/install/gr_kinematic_sim/lib/python3.8/site-packages/gr_kinematic_sim/custom_utils/object_tools.py", line 71 in draw
  File "/home/leev/workspaces/gr-kinematic-sim/install/gr_kinematic_sim/lib/python3.8/site-packages/gr_kinematic_sim/custom_utils/robots.py", line 103 in call_sensors
  File "/home/leev/workspaces/gr-kinematic-sim/install/gr_kinematic_sim/lib/python3.8/site-packages/gr_kinematic_sim/custom_utils/robots.py", line 108 in draw
  File "/home/leev/workspaces/gr-kinematic-sim/install/gr_kinematic_sim/lib/python3.8/site-packages/gr_kinematic_sim/custom_utils/gametools.py", line 141 in draw_every_sprite_in_list
  File "/home/leev/workspaces/gr-kinematic-sim/install/gr_kinematic_sim/lib/python3.8/site-packages/gr_kinematic_sim/main.py", line 94 in main
  File "/home/leev/workspaces/gr-kinematic-sim/install/gr_kinematic_sim/lib/gr_kinematic_sim/sim", line 11 in <module>
free(): invalid pointer
```

The mentioned `draw()` function:

```
    def draw(self):
        rect_to_draw = copy(self.rect)
        rect_to_draw.x += self.curr_offset_x
        rect_to_draw.y += self.curr_offset_y
        rotated_image = pygame.transform.rotozoom(self._original_image, self._current_rotation, 1)
        self.screen.blit(rotated_image, rect_to_draw)
```

MyreMylar:
fixes #2694

The logic being that these values are so large that they can only realistically come from bad data, but also we don't want to error out inside blit?

Hopefully bits earlier up the chain - like Rect/Frect can catch these invalid values and error there.


Damus666 (approved):

src_c/surface.c
```diff
@@ -3828,6 +3828,18 @@ pgSurface_Blit(pgSurfaceObject *dstobj, pgSurfaceObject *srcobj,
     SDL_Rect orig_clip, sub_clip;
     Uint8 alpha;
 
+    if ((dstrect->x < -INT_MAX) || (dstrect->x > INT_MAX) ||
+        (dstrect->y < -INT_MAX) || (dstrect->y > INT_MAX)) {
+        // destination position has values that are too large
+        return 0;
+    }
+    if (srcrect != NULL &&
+        ((srcrect->x < -INT_MAX) || (srcrect->x > INT_MAX) ||
```
novialriptide, line 3837: Any reason to not using `INT_MIN`?

novialriptide (approved):
LGTM!

Starbuck5:
I'd like to hold on merging this until I understand it, somewhat selfishly.

Um, how did the original poster get nan values in the first place?

Damus666:
> I'd like to hold on merging this until I understand it, somewhat selfishly.
> 
> Um, how did the original poster get nan values in the first place?

They didn't, and their code did not produce any segfaults for me and for other contributors. The nan was added by MyreMylar and we discovered it segfaults.

MyreMylar:
That was the second person that posted a segfault in that thread. The first
person mentioned nan values from a calculation that they then resolved
themselves. The second person posting a similar segfault mentioned just
holding down the right key which I couldn't reproduce myself at the time
but it may have been because I just didn't hold it down long enough as the
fundamental issue seems to be out of range large position values sent into
blit sometimes causing a segfault.

I don't know if this is the final solution myself. I think FRect shouldn't
accept NaN values either, though we should probably guard against these
overflow int values in blit anyway in case they creep in from elsewhere.

I'd like to find out exactly where in blit the segfault originates too and
also if there is any other position values that are in range that would
cause this segfault here. It is clearly not just the INT_MIN value on its
own as you also need the y value here and the size of the blotted surface
to cause something else to fail.

MyreMylar:
Poking at my segfault test program a bit more it looks like it is related to the width of the blit itself & the INT_MAX / INT_MIN values overflowing. Try this program and mess around with the numbers a bit:

```python
import pygame

int_max = 2147483647

surf_width = 1057
int_max_buffer = surf_width + 32

for x in reversed(range(0, 700)):
    print(x)
    pygame.display.init()
    screen = pygame.display.set_mode((x, 600))
    test_surf = pygame.Surface((surf_width, 400), flags=pygame.SRCALPHA)
    pos_rect = pygame.FRect(int_max - int_max_buffer, 200.0, 100.0, 100.0)
    screen.blit(test_surf, pos_rect)
    pygame.display.quit()
```

the surface to be blitted needs to be overlapping the target vertically here to get into the 'danger zone' but then you can turn the segfault on or off by having an x position of anything greater than: `INT_MAX - surface_width - 32`

MyreMylar:
That suggests to me that is specifically something to do with clipping the source surface to the destination surface when the right hand side of the source surface position is greater than INT_MAX. I mean maybe the other sides would also crash in the clipping but that one gets hit first because it is furthest out in the pygame coordinate system.

Basically we need to keep all four corners of the surface-to-be-clipped rectangle inside INT_MIN to INT_MAX - possibly with a small safety buffer as well.

MyreMylar:
Further refinement to strip out FRect as that is not required for the segfault:

```python
import pygame

int_max = 2147483647
surf_width = 2048

for x in reversed(range(0, 400)):
    print(x, surf_width)
    int_max_buffer = (
        surf_width 
    )  # subtract values here (e.g. 32, 64) and you will get a segfault pretty quick
    pygame.display.init()
    screen = pygame.display.set_mode((x, 600))
    test_surf = pygame.Surface((surf_width, 400), flags=pygame.SRCALPHA)
    screen.blit(test_surf, (int_max - int_max_buffer, 200.0))
    pygame.display.quit()
```

I think this probably makes it about as clear as it can get debugging from the python side. If anyone can narrow it down on the C side somehow that would be useful too.

MyreMylar:
OK, I adjusted the early exits to just always account for the dimensions of the surfaces and I can't make it segfault now. We could check for the dimensions of the source surface minus the dimensions of the destination surface - but that is extra math for something that I don't believe there is any legitimate reason to be doing (surfaces can't actually get very large so these blits would  all be clipped anyway if the clipping wasn't segfaulting). Doing it this way we will early exit with stupid source rect parameters whatever the width of the destination surface.

Other things we could do:

1. Make Rect handle float("nan") and float("inf) so they don't get converted to INT_MIN when passed in via various methods. See:
```python
import pygame

rect = pygame.Rect(float("nan"), float("inf"), float("nan"), float("inf"))
print(rect)
```
Output:
```
Rect(-2147483648, -2147483648, -2147483648, -2147483648)
```

Meanwhile:
```python
int(float("NaN"))
```
Output:
```
ValueError: cannot convert float NaN to integer
```

```python
int(float("inf"))
```
Output:
```
OverflowError: cannot convert float infinity to integer
```

3. Add a note to the SDL issue on this topic here: <https://github.com/libsdl-org/SDL/issues/8879> if we think we have anything useful to add. It sounds like SDL is already aware of INT_MAX/INT_MIN overflows in the code base but hasn't come up with a definitive solution, or decided if there is one.

MyreMylar:
I think this probably won't get merged as the segfault is too obscure and originates inside SDL, so I'm going to close it.

PR #2895, opened 2024-05-31
Optimized `mask.from_surface()` when converting an alpha surface.

itzpr3d4t0r:
This PR massively speeds up `mask.from_surface()` when dealing with the threshold argument (for 24 and 32 bit surfaces, about a 12% improvement for 16 and lower).

It achieves so by directly accessing the pixel's alpha channels instead of using SDL's functionality which is not only slow but actually did more work that we actually need for this operation by also returning the rgb channels.

With this new strategy we just jump from pixel's alpha to pixel's alpha and check that against the threshold.

I've also added a fast path for 24-bit surfaces bringing an astonishing 737 times improved runtime since we now use `bitmask_fill`.

I've also added a validity check for the threshold argument which wasn't there before.

here are the results:
image

made with this test program:
```Py
from data_utils import plot_tests, run_tests
import pygame
from pygame.mask import from_surface

pygame.init()

# ===========| CONFIG |===========

DO_TEST = 0
MAX_SIZE = 1000
REPETITIONS = 100
NUM_CALLS = 1

TITLE = "from_surface"
X_LABEL = "Surface size"
DO_SCATTER = True
MODE = "MIN"
LIMIT_TO_RANGE = MAX_SIZE
COMPARE_LIST = [(-1, 0)]
kwargs_dict = {"from_surface": from_surface}

def test_setup(curr_size: int, g: dict):
    s = pygame.Surface((curr_size, curr_size), pygame.SRCALPHA, 32)
    pygame.draw.circle(s, (255, 255, 255, 128), (curr_size // 2, curr_size // 2), curr_size // 2)
    g["s"] = s

tests = [
    ("mask from_surface new", "from_surface(s)"),
]

files = [
    ("mask from_surface new", "red"),
    ("mask from_surface old", "white")
]

if DO_TEST:
    run_tests(tests, test_setup, MAX_SIZE, REPETITIONS, NUM_CALLS, **kwargs_dict)

pygame.quit()

plot_tests(TITLE, files, MODE, LIMIT_TO_RANGE, DO_SCATTER, False, COMPARE_LIST,
           X_LABEL)
```

The data utils file (as txt):
[data_utils.txt](https://github.com/user-attachments/files/15511241/data_utils.txt)


novialriptide (approved):
The work you do never ceases to amaze me. This works on my machine just fine.

damusss (dismissed):
:pg_chad:

damusss (approved):
I've tested locally with different sizes and I get a 65+% improvement for 32 bit surfaces and 10% with 16 bit surfaces, which is really impressive, good job!

src_c/mask.c
```diff
@@ -770,22 +788,57 @@ set_pixel_color(Uint8 *pixel, Uint8 bpp, Uint32 color)
...
 
-    for (y = 0; y < surf->h; ++y) {
-        pixel = (Uint8 *)surf->pixels + y * surf->pitch;
-
-        for (x = 0; x < surf->w; ++x, pixel += bpp) {
-            SDL_GetRGBA(get_pixel_color(pixel, bpp), format, rgba, rgba + 1,
-                        rgba + 2, rgba + 3);
-            if (rgba[3] > threshold) {
-                bitmask_setbit(bitmask, x, y);
+    if (bpp == 3) { /* fast path for non-alpha surfaces */
```
ankith26, line 797: Can this check be made more generic than `bpp == 3`? I mean, there are non-alpha surfaces that have some other bpp, right? Perhaps we should just use `SDL_ISPIXELFORMAT_ALPHA` here?

src_c/mask.c
```diff
@@ -770,22 +788,57 @@ set_pixel_color(Uint8 *pixel, Uint8 bpp, Uint32 color)
...
+    for (y = 0; y < surf->h; ++y) {
+        x = 0;
+        LOOP_UNROLLED4(
+            {
+                if ((*srcp) > u_threshold)
+                    bitmask_setbit(bitmask, x, y);
+                srcp += bpp;
+                x++;
+            },
+            n, surf->w);
```
ankith26, line 839: As far as I can tell, there is nothing in this function that checks for the `surf->w == 0` case.

A `LOOP_UNROLLED4` requires `surf->w` to be non-zero otherwise it can segfault or give undefined behaviour.

We should probably add a fast exit if `surf->w == 0` outside the loop

test/mask_test.py
```diff
@@ -6089,7 +6089,18 @@ def test_from_surface(self):
...
+                    (alpha - 1) % 255,
+                    alpha,
+                    (alpha + 1) % 255,
+                    255,
+                }
+
+                invalid_thresholds = (-1, 256, 1000)
+                for threshold in invalid_thresholds:
+                    with self.assertRaises(ValueError):
+                        mask = pygame.mask.from_surface(surface, threshold)
```
ankith26, line 6103: I'd like to know why the tests are being changed here.

src_c/mask.c
```diff
@@ -838,6 +891,10 @@ mask_from_surface(PyObject *self, PyObject *args, PyObject *kwargs)
         return NULL; /* Exception already set. */
     }
 
+    if (threshold < 0 || threshold > 255) {
+        return RAISE(PyExc_ValueError, "threshold must be between 0 and 255");
+    }
```
ankith26, line 896: Even though this is probably a good change, it is *technically* a behaviour/API change and I'd like to keep it out of an optimization PR

src_c/mask.c
itzpr3d4t0r: There is a check for this on line 911 that was there before my PR.

src_c/mask.c
itzpr3d4t0r: No probs I can revert that.

src_c/mask.c
ankith26: Oh okay cool. But since this is something very easy to miss, I'd like a comment at the top of this function stating the input surface must have a non-zero width or something

src_c/mask.c
itzpr3d4t0r: We have some additional tests in `mask_test.py` that check for this already, aren't they enough?

ankith26: I'm referring to `RGBX`/`BGRX`/`XBGR`/`XRGB`: formats that have `bpp == 4` but don't have alpha. these could take the same fastpath as `bpp == 3`, unless I'm missing something?

itzpr3d4t0r: Ok I think i fixed this.

src_c/mask.c
```diff
@@ -770,22 +788,57 @@ set_pixel_color(Uint8 *pixel, Uint8 bpp, Uint32 color)
...
-                        rgba + 2, rgba + 3);
-            if (rgba[3] > threshold) {
-                bitmask_setbit(bitmask, x, y);
+    /* This function expects surf to be non-zero sized. */
+    SDL_PixelFormat *fmt = surf->format;
+    const Uint8 bpp = PG_FORMAT_BytesPerPixel(fmt);
+    int x, y, n;
+    Uint8 *srcp;
+    const int src_skip = surf->pitch - surf->w * bpp;
+
```
ankith26, line 797: Should also add an early exit for `threshold >= 255`.

Technically, those threshold values are allowed (even though not practically useful). With these changes, the new code does not behave the same for these values like the old code

ankith26 (approved):
Okay I am now satisfied that this PR should work for the common/normal cases.

Thanks for contributing! :partying_face:

PR #2897, opened 2024-06-01
Add SSE2, AVX2 & NEON support to `pygame.print_debug_info()`

MyreMylar:
These are the most common instruction sets we support paths for and could assist us in detecting bugs. There are more instruction sets listed in `system.get_cpu_instuction_sets()` but currently we don't make much use of them in pygame-ce.

Suggested by @Starbuck5

### DIFF
--- a/src_py/_debug.py
+++ b/src_py/_debug.py
@@ -8,6 +8,7 @@ from typing import Tuple, Optional, Callable
 from os import environ
 
 from pygame.version import ver
+from pygame.system import get_cpu_instruction_sets
 
 ImportResult = Tuple[str, bool, Optional[Callable]]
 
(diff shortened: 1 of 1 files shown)


Damus666 (approved):
Great idea, will help detect bugs like the recent segfaults more easily

oddbookworm (approved):
LGTM!

Starbuck5 (approved):
Thanks!

Starbuck5 (changes requested):
Ankith wanted to review this so slapping down a request changes to hold it until he can take a look.

ankith26 (changes requested):
I don't have a problem with this PR per se, but my opinion is that we should probably have a centralized way to get and set the SIMD level at runtime.

Everytime someone has to test a lower SIMD level they have to mess with either the buildconfig or the source code itself, and this is not ideal. Especially if we want a user to test out something.

One approach is to have env variables `PG_ENABLE_[AVX2|SSE2|NEON]` that can be configured before pygame init. On init, these variables are read, and along with runtime detection, we set the corresponding runtime check values.

Another approach is to have some API, probably in `pygame.system`, to pythonically get/set this

Starbuck5:
> I don't have a problem with this PR per se, but my opinion is that we should probably have a centralized way to get and set the SIMD level at runtime.

You should open an issue about that. It shouldn't block this PR.

We will get this for free when SDL3 comes out, btw. SDL_HINT_CPU_FEATURE_MASK.

ankith26 (approved):
True, I'm opening a separate issue for it, as it is independent of this PR.

Thanks for the PR :tada:

PR #2901, opened 2024-06-02
Update docs of `Surface.get_(f)rect()` and `(F)Rect.move_to()`

Damus666:
As you can see from the changed docs I think I made them clearer. I didn't know you could pass size kwargs myself for a long time. The docs technically didn't hide them, but it's very unclear that they do that. Why?
- Surface.get_rect -> it says the rect will have the surface dimentions, so _intuitively_ you'd think you can't resize it
- Rect.move_to -> the name and the docs only mentions that the rect will be moved to a new position, not that it will also resize it.

Yes, they both say that the kwargs should match the attributes and they don't say those are not allowed, but I think this small update makes it more clear.

### DIFF
--- a/docs/reST/ref/rect.rst
+++ b/docs/reST/ref/rect.rst
@@ -127,12 +127,12 @@
 
       | :sl:`moves the rectangle to the specified position`
       | :sg:`move_to(**kwargs) -> Rect`
-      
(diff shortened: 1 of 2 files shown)


ankith26 (approved):

MyreMylar (approved):
LGTM 👍

PR #2902, opened 2024-06-02
Minor docs/stubs improvements in `invlerp`/`remap`

ankith26:
Just a follow up PR to #2654

The main goal of doing this PR was to fix the `ZeroDivisionError` mentioned in the docs to `ValueError`, which is what is actually raised in the code. Also made one argument name in stubs consistent with the docs.

While I was editing the docs, I also polished it up a bit up according to my tastes.


Damus666 (approved):
Indeed clearer, nice

MyreMylar (approved):
Yep, looks nicer I agree. I think `lerp()` has the same ambiguity between weight/value.

buildconfig/stubs/pygame/math.pyi
```diff
@@ -334,7 +334,7 @@ class Vector3(_GenericVector):
 
 
 def lerp(a: float, b: float, weight: float, do_clamp: bool = True, /) -> float: ...
```
MyreMylar, line 336:
```suggestion
def lerp(a: float, b: float, value: float, do_clamp: bool = True, /) -> float: ...
```

Starbuck5 (approved):
OK

PR #2911, opened 2024-06-05
pygame.Surface docs improvements

REX2626:
When reading through the pygame.Surface docs I noticed the following issues/inconsistencies:

1. One line descriptions of Surface.blits() and Surface.fblits() seemed inconsistent with other descriptions.
2. Surfaces were sometimes referred to as "images", which didn't make sense to me.
3. Some of the one line method descriptions started with a capital letter, which was inconsistent with the other descriptions.
4. Sometimes Surface was captilalized and sometimes it wasn't. I changed these all to capitalized versions.
5. In the pygame.Surface.convert() description, there was an 'as' when I think there should be an 'a'.

I'm open to any suggestions or thoughts about the changes, whether the Surfaces should be capitalized or not etc..


REX2626:
Yep, makes sense. I've reverted the surface capitalization back to what is previously was.

ankith26 (approved):
LGTM, thanks for contributing! :tada:

Matiiss:
I think we need to decide on the capitalization now, because there are tons of places in this file where it's called a "Surface" and where it's called a "surface", so we can be consistent and lowercase it everywhere (besides type annotations and if they begin a sentence obviously).

docs/reST/ref/surface.rst
```diff
@@ -196,7 +196,7 @@
 
    .. method:: fblits
 
-      | :sl:`draw many surfaces onto the calling surface at their corresponding location and the same special_flags`
+      | :sl:`draw many surfaces onto this surface at their corresponding location and with the same special_flags`
```
Matiiss, line 199: Personally I find "the calling surface" to be clearer than "this surface" because it might not be clear which surface "this" is referring to exactly. This goes for all such occurrences in this file.

MyreMylar (approved):
LGTM 👍

docs/reST/ref/surface.rst
ankith26: TBH I prefer the current change. To me "calling surface" is not exactly clear, whereas the "this" keyword is used in many programming languages to mean "self"

PR #2912, opened 2024-06-06
Fix aalines overlap

mzivic7:
As seen on the picture bellow, there is a problem with draw.aalines.
bad
This is happening because draw.aalines require each aaline to be drawn 1px shorter (in some cases 2px - see bellow), so two connected draw_aaline are not overlapping.

<details>
  <summary>What is changed and how it works:</summary>

draw_aaline has 2 endpoints that needs to be disabled for draw.aalines only:
First endpoint is disabled when it is first line when aalines is open.
And when `from_x` value is not decimal (if it is, then line is 2px shorter).
Second endpoint is disabled when it is last line when aalines is open.
 
As a result of disabling first endpoint, now there is special case where 1px is missing between two lines, where one is steep and other line is not.
This is detected in draw.aalines.
And resolved in draw_aaline by drawing one px at place where end point should be.
At what end point will be drawn pixel: first point when line is not inverted, second point when line is inverted.
</details>

Here are some results (might need to zoom to see the difference better):
results
Top line is modified.
Middle line is original.
Bottom line is difference.
<details>
  <summary>Code used to draw this:</summary>

  ```py
  import pygame
  import numpy as np

  def curve_points(a, b, pea, t):
      sin_p = np.sin(pea)
      cos_p = np.cos(pea)
      x = a * np.cos(t)
      y = b * np.sin(t)
      x_rot = x * cos_p - y * sin_p
      y_rot = y * cos_p + x * sin_p
      return np.stack((x_rot, y_rot), axis=1)

  points1 = [[200.1, 300],
             [220.1, 300.1],
             [240.1, 300.2],
             [260.1, 300.4],
             [280.1, 300.6],
             [300.1, 300.8]]

  points2 = [[400.7, 300.1],
             [420.3, 300.5],
             [440.6, 305.1],
             [460.4, 315.2],
             [480.2, 330.5],
             [500.1, 360.7],
             [550.4, 360.5],
             [550.8, 400.2]]

  points_num = 150
  t = np.linspace(-np.pi, np.pi, points_num)
  points3 = curve_points(250, 150, 0.3, t) + np.array([350, 350])

  pygame.init()
  screen = pygame.display.set_mode((700, 700))
  clock = pygame.time.Clock()

  run = True
  while run:
          if event.type == pygame.QUIT:
              run = False
      screen.fill("black")
      pygame.draw.aalines(screen, (150, 150, 150), False, points1)
      pygame.draw.aalines(screen, (150, 150, 150), False, points2)
      pygame.draw.aalines(screen, (150, 150, 150), False, points3)
      pygame.display.flip()
      clock.tick(60)
  pygame.quit()
  ```
Results were merged for comparison in gimp.
</details>


src_c/draw.c
```diff
@@ -1361,66 +1432,123 @@ draw_aaline(SDL_Surface *surf, Uint32 color, float from_x, float from_y,
...
+    else {
+        /* Handle extra pixel for aalines.
+         * It is drawn only when one line is steep and other is not.*/
+        if (extra_pixel_for_aalines && line_inverted) {
+            if (from_x < clip_right - 1.0f) {
+                y_endpoint = to_y + gradient * (x_pixel_end - to_x);
+                x_gap = 1 - x_pixel_end + to_x;
+                brightness = y_endpoint - (int)y_endpoint;
+                if (steep) {
+                    x = (int)y_endpoint - 1;
```
oddbookworm, line 1539: Here and on other lines around here, I'm kinda sketched out by the cast from float to int. I get that it's pre-existing, but it's screaming at me that the range a float can represent is just so much larger than what an int can represent. Personally, I'd like to see some range checking to ensure that the cast won't cause us to overflow/underflow the bounds of int.

oddbookworm (changes requested):
I know this is asking for something that won't be fun, but I'd like to see some regression test(s) put in place that demonstrate this change, and will flag if the change gets broken by a future commit

src_c/draw.c
mzivic7: y_endpoint is calculated as `from_y + something_small` (something_small < 1)
Take a look after line [1385](https://github.com/mzivic7/pygame-ce/blob/fix_aalines_overlap/src_c/draw.c#L1385), it is culling and cutting lines not on surface, `from_y` and `to_y` are there.
Now, if I try to create surface larger than int size, I get `Invalid resolution for Surface`.
And if I try to create: `pygame.Surface((2147483647, 2147483647))`, I get `pygame.error: Out of memory`.
I don't know what memory is actually needed, but surely nobody will ever need that huge surface, and operate with it in reasonable time.

test/draw_test.py
```diff
@@ -3646,6 +3646,60 @@ class DrawAALinesTest(AALinesMixin, DrawTestCase):
...
+                f"aalines are overlapping, pos={(25, y)}",
+            )
+
+    def test_aalines__steep_missing_pixel(self):
+        """Ensures there are no missing pixels between steep and non-steep lines.
+
+        Draws two adjacent lines: first is not steep and second is, then
+        checks if there is missing pixel at shared point between those two lines.
+
+        See: #2912
```
oddbookworm, line 3678: nitpick, but I'd rather see a full url here instead of just a reference number, that way it's easier just to ctrl-click from an IDE or other rich text environment

oddbookworm (approved):
One small nitpick (that applies to both tests), but otherwise this looks good to me!

test/draw_test.py
mzivic7: Oh, I saw some other tests have reference numbers, so I used it too.

oddbookworm (approved):
This still looks good to me

bilhox:
So I tried to understand how the algorithm works, but even after a lot of doliprane, I couldn't understand the algorithm because of lack of graphical explanations. I believe it would be better if @mzivic7 or anyone who understood the algorithm, make a video that explains step by step the algorithm for the sake of others mental health. Like that we can push this PR faster to its merge.

MyreMylar (approved):
OK, LGTM 👍 

Output looks a lot better (Once I zoomed in, my old eyes couldn't see shit at normal resolution) and the changes make sense to me to reduce the overlapping in the original algorithm.

PR #2913, opened 2024-06-06
Allow degenerate circles (radius = 0)

damusss:
### note: I've modified the circle init error message to include the radius, as it was _very_ unclear before. If you don't want degenerate circles, this modification should still be added IMHO, so let me know.

I've slightly modified the code, the tests and the docs to allow degenerate circles to exist.
I've run through every line where the radius attribute is used and I checked for divisions with the radius but they were not present.
Why?
- You can create 'degenerate' rects by passing 0 as width and height, so for consistency I think you should be able to create degenerate circles aswell
- It will unintentionally be treated like a point when checking for collisions, and virtual attributes will be automatically affected
- degenerate circles are still considered circles in analitic geometry so I would consider them colliders
- If degenerate rects are valid colliders/shapes, degenerate circles should aswell

Tell me if you'd like more tests or if I missed something.
Of course you can not agree with allowing degenerate circles, I made this pull request so that if you think they should be there, you'll find a PR ready for it.

### DIFF
--- a/src_c/circle.c
+++ b/src_c/circle.c
@@ -47,8 +47,8 @@ pg_circle_init(pgCircleObject *self, PyObject *args, PyObject *kwds)
         PyErr_SetString(
             PyExc_TypeError,
             "Arguments must be a Circle, a sequence of length 3 or 2, or an "
-            "object with an attribute called 'circle' and the radius must be "
-            "nonnegative");
+            "object with an attribute called 'circle', all with corresponding "
+            "nonnegative radius argument");
         return -1;
     }
     return 0;
--- a/src_c/circle.c
+++ b/src_c/circle.c
@@ -48,7 +48,7 @@ pg_circle_init(pgCircleObject *self, PyObject *args, PyObject *kwds)
             PyExc_TypeError,
             "Arguments must be a Circle, a sequence of length 3 or 2, or an "
             "object with an attribute called 'circle' and the radius must be "
-            ">= 0");
+            "nonnegative");
         return -1;
     }
     return 0;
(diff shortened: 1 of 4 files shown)


oddbookworm:
I'm requesting a review from @itzpr3d4t0r as the resident pygame-geometry expert

src_c/circle.c
```diff
@@ -494,8 +495,8 @@ pg_circle_setr(pgCircleObject *self, PyObject *value, void *closure)
         return -1;
     }
 
-    if (radius <= 0) {
-        PyErr_SetString(PyExc_ValueError, "Radius must be positive");
+    if (radius < 0) {
+        PyErr_SetString(PyExc_ValueError, "Radius must be >= 0");
```
oddbookworm, line 499: I would prefer the phrasing "Radius must be nonnegative" or "Radius must be 0 or larger" instead of the inequality. It's just easier to parse IMO

oddbookworm:
I tested this locally with this test script, and it seems to work exactly as I was expecting it to

src_c/circle.c
```diff
@@ -47,7 +47,8 @@ pg_circle_init(pgCircleObject *self, PyObject *args, PyObject *kwds)
         PyErr_SetString(
             PyExc_TypeError,
             "Arguments must be a Circle, a sequence of length 3 or 2, or an "
-            "object with an attribute called 'circle'");
+            "object with an attribute called 'circle' and the radius must be "
+            "nonnegative");
```
itzpr3d4t0r, line 51:
```suggestion
            "object with an attribute called 'circle', all with corresponding nonnegative radius argument");
```

itzpr3d4t0r (approved):
I'm fine with this, still can't see where it would be useful but i guess it's no harm.

oddbookworm (approved):
Alright, with the clarified error messages, I'm happy with this. I don't see much use here that isn't already covered by a point, but we allow 0-sized Rects, so I'm not going to object here

ankith26 (approved):

Issue #2877, opened 2024-05-26
Enhance window with OpenGL with viewport and documentation


damusss:
edit: if the title wasn't clear, I'm referring to `pygame.window.Window`, not `pygame.display`
Pull request https://github.com/pygame-community/pygame-ce/pull/2659 was finally merged so it won't appear on the pull requests list and some important comments might get lost, but I think they need to be addressed before 2.5.0 official release.
Once they are, this issue can be closed.

- Update docs of `Window.flip()` with latest changes (by MyreMylar)
- Resize the OpenGL viewport on the event watch when the size changes (for full details check https://github.com/pygame-community/pygame-ce/pull/2659#issuecomment-2076500703 )

About the second one, a contributor also said they had even more enhancements in mind for the event watch, and this can be the right place to put them in.

PR #2915, opened 2024-06-07
Update OpenGL Viewport when Window resizes

damusss:
Fixes https://github.com/pygame-community/pygame-ce/issues/2877
The documentation has been updated by MyreMylar already.

As stated in the issue, the newly functional OpenGL `Window`s, unlike the display windows, would not resize the OpenGL viewport causing bugs and inconsistencies.

With the addition of a function, used both in the event watch and in the `set_size` function, the OpenGl viewport is now correctly updated. I tested locally, it works for fullscreen/maximization/autoresize/manualresize.

If it works for you too (should, it's identical to the display's one), I think this should be added not later than 2.5.1

### DIFF
--- a/src_c/window.c
+++ b/src_c/window.c
@@ -605,7 +605,7 @@ window_set_size(pgWindowObject *self, PyObject *arg, void *v)
         self->surf->surf = SDL_GetWindowSurface(self->_win);
     }
     if (self->context) {
-        /* Update the OpenGl viewport immediately instead of relying on the
+        /* Update the OpenGL viewport immediately instead of relying on the
          * event callback */
         _window_opengl_set_viewport(self->_win, self->context, w, h);
     }
--- a/src_c/window.c
+++ b/src_c/window.c
@@ -223,7 +223,7 @@ _window_opengl_set_viewport(SDL_Window *window, SDL_GLContext context,
         return -1;
     }
     if (p_glViewport == NULL) {
-        PyErr_SetString(pgExc_SDLError, "glViewport is unavailable");
+        PyErr_SetString(pgExc_SDLError, "glViewport function is unavailable");
         return -1;
     }
     p_glViewport(0, 0, wnew, hnew);
@@ -255,7 +255,8 @@ _resize_event_watch(void *userdata, SDL_Event *event)
         if (_window_opengl_set_viewport(event_window, event_window_pg->context,
                                         event->window.data1,
                                         event->window.data2) < 0) {
-            PyErr_Clear();
+            return PyErr_WarnEx(PyExc_RuntimeWarning,
+                                "Failed to set OpenGL viewport", 0);
         }
     }
 
--- a/src_c/window.c
+++ b/src_c/window.c
@@ -3,6 +3,7 @@
 #include "pygame.h"
 
 #include "pgcompat.h"
+#include "pgopengl.h"
 
 #include "doc/sdl2_video_doc.h"
 #include "doc/window_doc.h"
@@ -211,7 +212,17 @@ window_flip(pgWindowObject *self, PyObject *_null)
     Py_RETURN_NONE;
 }
 
-// Callback function for surface auto resize
+static void
+_window_opengl_set_viewport(SDL_Window *window, SDL_GLContext context,
+                            int wnew, int hnew)
+{
+    GL_glViewport_Func p_glViewport =
+        (GL_glViewport_Func)SDL_GL_GetProcAddress("glViewport");
+    SDL_GL_MakeCurrent(window, context);
+    p_glViewport(0, 0, wnew, hnew);
+}
+
+// Callback function for surface auto resize or OpenGL viewport update
 static int SDLCALL
 _resize_event_watch(void *userdata, SDL_Event *event)
 {
@@ -230,6 +241,11 @@ _resize_event_watch(void *userdata, SDL_Event *event)
     if (event_window_pg->_is_borrowed) {
         // have been handled by event watch in display.c
         return 0;
+    }
+
+    if (event_window_pg->context != NULL) {
+        _window_opengl_set_viewport(event_window, event_window_pg->context,
+                                    event->window.data1, event->window.data2);
     }
 
     if (!event_window_pg->surf)
@@ -587,6 +603,11 @@ window_set_size(pgWindowObject *self, PyObject *arg, void *v)
         /* Ensure that the underlying surf is immediately updated, instead of
          * relying on the event callback */
         self->surf->surf = SDL_GetWindowSurface(self->_win);
+    }
+    if (self->context) {
+        /* Update the OpenGl viewport immediately instead of relying on the
+         * event callback */
+        _window_opengl_set_viewport(self->_win, self->context, w, h);
     }
 
     return 0;
--- a/src_c/window.c
+++ b/src_c/window.c
@@ -96,6 +96,7 @@ display_resource_end:
 }
 
 static PyTypeObject pgWindow_Type;
+static GL_glViewport_Func p_glViewport = NULL;
 
 #define pgWindow_Check(x) \
     (PyObject_IsInstance((x), (PyObject *)&pgWindow_Type))
@@ -212,14 +213,21 @@ window_flip(pgWindowObject *self, PyObject *_null)
     Py_RETURN_NONE;
 }
 
-static void
+/* Exception already set */
+static int
 _window_opengl_set_viewport(SDL_Window *window, SDL_GLContext context,
                             int wnew, int hnew)
 {
-    GL_glViewport_Func p_glViewport =
-        (GL_glViewport_Func)SDL_GL_GetProcAddress("glViewport");
-    SDL_GL_MakeCurrent(window, context);
+    if (SDL_GL_MakeCurrent(window, context) < 0) {
+        PyErr_SetString(pgExc_SDLError, SDL_GetError());
+        return -1;
+    }
+    if (p_glViewport == NULL) {
+        PyErr_SetString(pgExc_SDLError, "glViewport is unavailable");
+        return -1;
+    }
     p_glViewport(0, 0, wnew, hnew);
+    return 0;
 }
 
 // Callback function for surface auto resize or OpenGL viewport update
@@ -244,8 +252,11 @@ _resize_event_watch(void *userdata, SDL_Event *event)
     }
 
     if (event_window_pg->context != NULL) {
-        _window_opengl_set_viewport(event_window, event_window_pg->context,
-                                    event->window.data1, event->window.data2);
+        if (_window_opengl_set_viewport(event_window, event_window_pg->context,
+                                        event->window.data1,
+                                        event->window.data2) < 0) {
+            PyErr_Clear();
+        }
     }
 
     if (!event_window_pg->surf)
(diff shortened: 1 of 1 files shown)


src_c/window.c
```diff
@@ -211,7 +212,17 @@ window_flip(pgWindowObject *self, PyObject *_null)
     Py_RETURN_NONE;
 }
 
-// Callback function for surface auto resize
+static void
+_window_opengl_set_viewport(SDL_Window *window, SDL_GLContext context,
+                            int wnew, int hnew)
+{
+    GL_glViewport_Func p_glViewport =
+        (GL_glViewport_Func)SDL_GL_GetProcAddress("glViewport");
```
oddbookworm, line 220: I'm not a fan of making this a local on the stack every time this function gets called. Can this either be made static or otherwise moved out of here?

oddbookworm, line 220: In the [docs](https://wiki.libsdl.org/SDL2/SDL_GL_GetProcAddress), this function ***can*** return `NULL`, so please check for that before trying to call the resulting function to be defensive and not cause segfaults on some weird system with a potentially bugged driver

src_c/window.c
```diff
@@ -588,6 +604,11 @@ window_set_size(pgWindowObject *self, PyObject *arg, void *v)
          * relying on the event callback */
         self->surf->surf = SDL_GetWindowSurface(self->_win);
     }
+    if (self->context) {
```
oddbookworm, line 607: On line 246, you compare to `NULL`, but not here. Personally, I prefer the explicit comparison, but I don't care which way you go. I would like to see it be consistent though

src_c/window.c
```diff
@@ -211,7 +212,17 @@ window_flip(pgWindowObject *self, PyObject *_null)
...
 }
 
-// Callback function for surface auto resize
+static void
+_window_opengl_set_viewport(SDL_Window *window, SDL_GLContext context,
+                            int wnew, int hnew)
+{
+    GL_glViewport_Func p_glViewport =
+        (GL_glViewport_Func)SDL_GL_GetProcAddress("glViewport");
+    SDL_GL_MakeCurrent(window, context);
```
oddbookworm, line 221: I suggest checking the return here and getting the error from SDL when this function fails, and throwing an exception (`pygame.error` probably)

[docs](https://wiki.libsdl.org/SDL2/SDL_GL_MakeCurrent)

oddbookworm (changes requested):

oddbookworm:
I suspect the circleCI failure is unrelated to your changes, looks like it's unable to download the SDL2 dependency. I'm going to wait a while before rerunning it

src_c/window.c
Starbuck5: Static references to things are sort of discouraged now that the Python ecosystem is trying to move towards free threading.

Damus666:
@oddbookworm Alright, I managed to make `glViewport` static and I added exceptions to my new function. To note that the window is still resized even if there are problems with opengl, so you could do something like
image
On the event watch, I cleared the errors as I don't know how to stop the python execution with an exception from there. It also seems that another check (for the surface) is returning early without an exception aswell, so I followed the same trend.
Tell me what you think :)

src_c/window.c
Damus666: > Static references to things are sort of discouraged now that the Python ecosystem is trying to move towards free threading.

So should I revert that specific change and get the `glViewport` within the function?

src_c/window.c
```diff
@@ -232,6 +251,14 @@ _resize_event_watch(void *userdata, SDL_Event *event)
         return 0;
     }
 
+    if (event_window_pg->context != NULL) {
+        if (_window_opengl_set_viewport(event_window, event_window_pg->context,
+                                        event->window.data1,
+                                        event->window.data2) < 0) {
+            PyErr_Clear();
```
oddbookworm, line 258: All you'll do here is clear the error you set inside the function, which isn't what you want to do, I'll look into this in a little while to see how it could be handled better

oddbookworm (changes requested):

src_c/window.c
Damus666: > All you'll do here is clear the error you set inside the function, which isn't what you want to do, I'll look into this in a little while to see how it could be handled better

As I said in the latest comment, I don't know how to hook up some sort of exception inside a function which is called by SDL. I've handled the exception correctly on the `set_size` function as that's called by CPython.

oddbookworm: I think the best we can do here is issue a warning. Ideally, it'll never get to do that and an actual exception will be raised, but at least there will be something in the event that doesn't happen. So I'm suggesting replacing `PyErr_Clear()` here with
```
return PyErr_WarnEx(PyExc_RuntimeWarning, "Failed to set opengl viewport", 0);
```
I chose runtime warning because it made the most sense to me of the builtin warnings, but if you can justify something else, I'm fine with that too

oddbookworm (approved):
LGTM!

MyreMylar:
Hi,

What test code are you using for this? When I try it with a resizable Window using the code from the last PR the context doesn't appear to resize and I get some refcount errors popping up. Here is the original PR code:

<details>
<summary>Open GL triangle code</summary>

```python
import pygame
import zengl

window_size = (1280, 720)

pygame.init()

window = pygame.Window(size=window_size, opengl=True)
window.resizable = True

ctx = zengl.context()

image = ctx.image(window_size, "rgba8unorm", samples=4)

pipeline = ctx.pipeline(
    vertex_shader="""
        #version 330 core

        out vec3 v_color;

        vec2 vertices[3] = vec2[](
            vec2(0.0, 0.8),
            vec2(-0.6, -0.8),
            vec2(0.6, -0.8)
        );

        vec3 colors[3] = vec3[](
            vec3(1.0, 0.0, 0.0),
            vec3(0.0, 1.0, 0.0),
            vec3(0.0, 0.0, 1.0)
        );

        void main() {
            gl_Position = vec4(vertices[gl_VertexID], 0.0, 1.0);
            v_color = colors[gl_VertexID];
        }
    """,
    fragment_shader="""
        #version 330 core

        in vec3 v_color;

        layout (location = 0) out vec4 out_color;

        void main() {
            out_color = vec4(v_color, 1.0);
            out_color.rgb = pow(out_color.rgb, vec3(1.0 / 2.2));
        }
    """,
    framebuffer=[image],
    topology="triangles",
    vertex_count=3,
)

clock = pygame.Clock()

while True:
        if event.type == pygame.QUIT:
            pygame.quit()
            quit()

    ctx.new_frame()
    image.clear()
    pipeline.render()
    image.blit()
    ctx.end_frame()

    window.flip()
    clock.tick(60)
```

</details>

After resizing a few times and hitting the close button I have received this on more than one occasion:

```
Fatal Python error: none_dealloc: deallocating None: bug likely caused by a refcount error in a C extension
Python runtime state: finalizing (tstate=0x00007ff9364f6960)

Current thread 0x0000e614 (most recent call first):
  Garbage-collecting
  <no Python frame>
```

After resizing the window smaller I sort of expected the triangle to resize to the size of the window:

image

This could be a zenGL issue though I'm not sure.

MyreMylar:
After some fiddling this is what I came up with as test code:

```python
import pygame
import zengl

window_size = (1280, 720)

pygame.init()

window = pygame.Window(size=window_size, opengl=True)
window.resizable = True

ctx = zengl.context()

image = ctx.image(window_size, "rgba8unorm", samples=4)
vert_shader = """
        #version 330 core

        out vec3 v_color;

        vec2 vertices[3] = vec2[](
            vec2(0.0, 0.8),
            vec2(-0.6, -0.8),
            vec2(0.6, -0.8)
        );

        vec3 colors[3] = vec3[](
            vec3(1.0, 0.0, 0.0),
            vec3(0.0, 1.0, 0.0),
            vec3(0.0, 0.0, 1.0)
        );

        void main() {
            gl_Position = vec4(vertices[gl_VertexID], 0.0, 1.0);
            v_color = colors[gl_VertexID];
        }
    """
frag_shader = """
        #version 330 core

        in vec3 v_color;

        layout (location = 0) out vec4 out_color;

        void main() {
            out_color = vec4(v_color, 1.0);
            out_color.rgb = pow(out_color.rgb, vec3(1.0 / 2.2));
        }
    """
pipeline = ctx.pipeline(
    vertex_shader=vert_shader,
    fragment_shader=frag_shader,
    framebuffer=[image],
    topology="triangles",
    vertex_count=3,
)

clock = pygame.Clock()

while True:
        if event.type == pygame.QUIT:
            pygame.quit()
            quit()

        if event.type == pygame.WINDOWRESIZED:
            image = ctx.image((event.x, event.y), "rgba8unorm", samples=4)
            pipeline = ctx.pipeline(
                vertex_shader=vert_shader,
                fragment_shader=frag_shader,
                framebuffer=[image],
                topology="triangles",
                vertex_count=3,
            )

    ctx.new_frame()
    image.clear()
    pipeline.render()
    image.blit()
    ctx.end_frame()

    window.flip()
    clock.tick(60)
```

Which seems to work. Does that seem correct for testing this issue?

damusss:
@MyreMylar zengl is different from moderngl, so when the window resizes you are required to remake the framebuffer image and threfore the pipeline that uses it. The "correct" version of your code is this:
(failed to make collapsable header)
NOTE: this makes this code work with OR without my PR. You should try with moderngl. hold on while I provide a moderngl example that is impacted by this PR.

```python
import pygame
  import zengl
  
  window_size = (1280, 720)
  
  pygame.init()
  
  window = pygame.Window(size=window_size, opengl=True)
  window.resizable = True
  
  ctx = zengl.context()
  
  image = ctx.image(window_size, "rgba8unorm", samples=4)
  
  def make_pipeline():
      global pipeline
      pipeline = ctx.pipeline(
          vertex_shader="""
              #version 330 core
  
              out vec3 v_color;
  
              vec2 vertices[3] = vec2[](
                  vec2(0.0, 0.8),
                  vec2(-0.6, -0.8),
                  vec2(0.6, -0.8)
              );
  
              vec3 colors[3] = vec3[](
                  vec3(1.0, 0.0, 0.0),
                  vec3(0.0, 1.0, 0.0),
                  vec3(0.0, 0.0, 1.0)
              );
  
              void main() {
                  gl_Position = vec4(vertices[gl_VertexID], 0.0, 1.0);
                  v_color = colors[gl_VertexID];
              }
          """,
          fragment_shader="""
              #version 330 core
  
              in vec3 v_color;
  
              layout (location = 0) out vec4 out_color;
  
              void main() {
                  out_color = vec4(v_color, 1.0);
                  out_color.rgb = pow(out_color.rgb, vec3(1.0 / 2.2));
              }
          """,
          framebuffer=[image],
          topology="triangles",
          vertex_count=3,
      )
  make_pipeline()
  
  clock = pygame.Clock()
  
  while True:
          if event.type == pygame.QUIT:
              pygame.quit()
              quit()
          if event.type == pygame.VIDEORESIZE:
              window_size = event.size
              image = ctx.image(window_size, "rgba8unorm", samples=4)
              make_pipeline()
  
      ctx.new_frame()
      image.clear()
      pipeline.render()
      image.blit()
      ctx.end_frame()
  
      window.flip()
      clock.tick(60)
```

MyreMylar:
OK, looks like the refcount error on window close is also fixed by updating zenGL - so I guess that was on their side not us.

MyreMylar (approved):
Approving, assuming I'm testing this correctly - please let me know if I am not.

PR #2918, opened 2024-06-09
Update transform.get/set_smoothscale_backend typehint from str to Literal

Damus666:
It's like this for other functions (for example the ones requiring an image format) and since the valid options are just a few it makes sense to be clear in the typehint about them, so one can skip going to the docs.
I omitted the deprecated ones, as other stubs do that too.


MyreMylar (approved):
LGTM 👍

ankith26 (approved):

PR #2922, opened 2024-06-12
More SDL3 compat: Surface things

Starbuck5:
Adds surface BPP macros to new code that slipped around https://github.com/pygame-community/pygame-ce/pull/2804

Makes PG_SurfaceHasRLE work everywhere, not just in rotozoom, and uses it to replace a nearly identical hack in surface.c, but with a strategy that instead calls an SDL function in SDL 2.0.14 and above (which works in SDL3, where the original hack doesn't).

### DIFF
--- a/src_c/_pygame.h
+++ b/src_c/_pygame.h
@@ -191,7 +191,9 @@ struct SDL_BlitMap {
 #define SDL_COPY_RLE_DESIRED 0x00001000
 
 #define PG_SurfaceHasRLE(surface) \
-    ((surface == NULL) ? 0 : (surface->map->info.flags & SDL_COPY_RLE_DESIRED))
+    (((surface) == NULL)          \
+         ? 0                      \
+         : ((surface)->map->info.flags & SDL_COPY_RLE_DESIRED))
 
 #endif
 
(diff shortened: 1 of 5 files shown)


src_c/_pygame.h
```diff
@@ -190,8 +190,9 @@ struct SDL_BlitMap {
 };
 #define SDL_COPY_RLE_DESIRED 0x00001000
 
-SDL_bool
-PG_SurfaceHasRLE(SDL_Surface *surface);
+#define PG_SurfaceHasRLE(surface) \
+    ((surface == NULL) ? 0 : (surface->map->info.flags & SDL_COPY_RLE_DESIRED))
```
ankith26, line 194:
```suggestion
    (((surface) == NULL) ? 0 : ((surface)->map->info.flags & SDL_COPY_RLE_DESIRED))
```
Helps avoid some macro pitfalls.

If you want to further simplify the expression you can also do
```c
    ((surface) && ((surface)->map->info.flags & SDL_COPY_RLE_DESIRED))
```

Starbuck5: So the surface can be expression itself instead of a value? Okay.

I prefer the `==NULL` checks for explicitness. Like doing `is None` instead of a boolean check in Python.

ankith26 (approved):
LGTM, thanks :+1:

zoldalma999 (approved):

Issue #2571, opened 2023-11-19
Calling a second time ``pygame.display.set_mode`` with ``vsync`` and ``vsync`` already enabled doesn't work correctly


bilhox:
**Environment:**

- **Operating system** : Windows 11
- **Python version** : 3.11.4 / 3.12.0
- **SDL version** : 2.26.5
- **PyGame version** : 2.3.2 / 2.4.0.dev2

**Current behavior:**

If you call ``pygame.display.set_mode`` with ``vsync`` enabled and then again with ``vsync`` enabled, the screen is black (like you lost the hand on the screen) and when you resize, the rest of the window is transparent.

**Expected behavior:**

You don't lost the hand on the screen, and it correctly resizes.
*Note : If you quit pygame between the two calls, the problem is fixed, could you tell me if it's what we should do ?*

**Screenshots**

![image](https://github.com/pygame-community/pygame-ce/assets/69472620/b84ae40c-fcf4-4045-b248-215cdf7bf853)
![Capture d'écran 2023-11-19 142026](https://github.com/pygame-community/pygame-ce/assets/69472620/9c745f11-8498-4d89-9c2d-db6f32689b94)

**Test program:**

```py
import pygame
import math

pygame.init()

screen = pygame.display.set_mode([800, 800], flags=pygame.RESIZABLE+pygame.SCALED, vsync=1)
screen2 = pygame.display.set_mode([900, 800], flags=pygame.RESIZABLE+pygame.SCALED, vsync=1)

clock = pygame.Clock()
pos = []
timer = 0

running = True

while running:

    dt = clock.tick() / 1000
    timer += dt

    screen.fill("black")

    pos = [400 + 40*math.cos(5*timer) + 20*math.cos(10*timer) + 200*math.cos(timer), 400 + 20*math.sin(10*timer) + 40*math.sin(5*timer) + 200*math.sin(timer)]

    pygame.draw.circle(screen, "red", pos, 20)

    pygame.display.flip()

    for event in pygame.event.get():
        if event.type == pygame.QUIT:
            pygame.quit()
            running = False
```

Starbuck5:
Is this similar to https://github.com/pygame-community/pygame-ce/issues/2376?

bilhox:
> Is this similar to https://github.com/pygame-community/pygame-ce/issues/2376?

No 🤔

Matiiss:
[...]

Issue #2190, opened 2023-05-22
pygame._sdl2.Renderer is broken on my system.


yolowex:
**Environment:**

- **Operating system**:
* OS: Kali GNU/Linux Rolling x86_64
* Kernel: 6.1.0-kali7-amd64 

- **Python version** 3.11.2
- **SDL version** 2.26.4
- **PyGame version** 2.2.1
- **Relevant hardware** 
* CPU: 11th Gen Intel i7-11700 (16) @ 4.800GHz 
* GPU: Intel RocketLake-S GT1 [UHD Graphics 750]

**Current behavior:**
_sdl2.renderer is broken on my system.
it says that Renderer is already associated with window

**Expected behavior:**
I want to render things using _sdl2

**Steps to reproduce:**

I just copied the code that @Starbuck5  made in discord _sdl2 channel, and pasted it into my problem, running it.

**Test code**
```
import pygame
from pygame._sdl2 import Renderer, Texture, Window
import os
here = os.path.abspath(".")+"/"
pygame.init()

screen = pygame.display.set_mode((0,0), flags=pygame.FULLSCREEN)
window = Window.from_display_module()
renderer = Renderer(window)

texture = Texture.from_surface(renderer, pygame.image.load(here+"Capture.PNG"))

print(list(pygame._sdl2.get_drivers()))

clock = pygame.time.Clock()

running = True
while running:
    for event in pygame.event.get():
        if event.type == pygame.QUIT:
            running = False

        if event.type == pygame.KEYDOWN and event.key == pygame.K_ESCAPE:
            running = False

    renderer.draw_color = (255, 255, 255, 255)
    renderer.clear()

    texture.draw(None, None)

    renderer.present()

    clock.tick()
    print(clock.get_fps(), end="\r")

pygame.quit()
```

* Traceback, Stacktrace, error log ( or whatever it's called )

```
Traceback (most recent call last):
  File "/home/yolo/Workstation/Python/pygame/android-unit-test/main.py", line 10, in <module>
    import sdl2_test
  File "/home/yolo/Workstation/Python/pygame/android-unit-test/sdl2_test.py", line 9, in <module>
    renderer = Renderer(window)
               ^^^^^^^^^^^^^^^^
  File "src_c\cython\pygame\_sdl2\video.pyx", line 1107, in pygame._sdl2.video.Renderer.__init__
pygame._sdl2.sdl2.error: Renderer already associated with window
```

Starbuck5:
[...]

PR #2925, opened 2024-06-14
Fix scaled-after-non-scaled usage

### LINKED ISSUE #2376: Mac - After creating a window using `display.set_mode()`, cannot use `display.set_mode()` again to change to SCALED
**Environment:**
- Operating system:  macOS Monterey 12.6.8
- Hardware:  MacBook Pro (early 2015)
- Processer:  2.7GHz Dual Core Intel Core i5
- Python version:  3.11.4
- SDL version:  SDL 2.26.5
- pygame-ce version:  2.3.1.dev4 (also reproduced on 2.3.0 and 2.3.1.dev2)

**Bug behaviour:**
When initializing a window using `display.set_mode()` and NOT using the SCALED flag, if you later attempt to use `display.set_mode()` again to change the window to be SCALED it fails with a pygame error.

**Expected behavior:**
Calling `display.set_mode()` again using SCALED should result in the window being resized with appropriate scale. 

**Screenshots**
image

**Steps to reproduce:**
1. create a window that is not SCALED using the standard set_mode method (i.e. `pygame.display.set_mode((width, height))`)
2. use set_mode again to change the display to be SCALED (i.e. `pygame.display.set_mode((width, height), pygame.SCALED)`)

**Test code:**
Pressing the 1 key re-sets the window, pressing 2 key crashes the program instead of producing a SCALED window.

```
import pygame

pygame.init()

screen = pygame.display.set_mode((800, 600))

run = True
while run:
        if event.type == pygame.QUIT:
            run = False
            break
            if event.key == pygame.K_1:
                screen = pygame.display.set_mode((800, 600))
            elif event.key == pygame.K_2:
                screen = pygame.display.set_mode((800, 600), pygame.SCALED)
    screen.fill("blue")
    pygame.display.flip()

pygame.quit()
```

**Error details:**
pygame.error: failed to create renderer

**Notes:**
I have tested the code on my Windows 11 machine and it works fine with the same Python and pygame-ce version.
Also, if the window is created with the SCALED flag, using `display.set_mode()` again without the SCALED flag correctly resizes to the window size specified. However, the same issue occurs after this, where an error is seen if you try to change back to SCALED again.

### LINKED ISSUE #2737: Calling `pygame.display.set_mode` a second time crashes with `SCALED` flag if it hasn't been called with it the first time.
**Environment:**

- **Operating system**: Windows 10
- **Python version**: 3.11.4
- **SDL version**: 2.28.5
- **pygame-ce version**: 2.4.1

**Current behavior:**

When the method `pygame.display.set_mode` is called a second time with `SCALED` flag and hasn't been called the first time with `SCALED` flag, pygame crash with the following error:

```
pygame.error: failed to create renderer
```

The error happens only when `set_mode` is called with `SCALED` flag and if it hasn't been set with it before.

**Expected behavior:**

This seems to happen only since 2.4.0, I tested it with 2.3.2 and it works properly.

**Steps to reproduce:**

1. Call `set_mode` without `SCALED` flag.
2. Call `set_mode` a second time with `SCALED` flag makes pygame crash.

Or:

1. Call `set_mode` with `SCALED` flag.
2. Call `set_mode` a second time without it, this will work as intended.
3. Call `set_mode` a third time with `SCALED` flag, then it crashes.

**Test code**

Just press SPACE key to switch mode.

```python
import pygame

pygame.init()

screen = pygame.display.set_mode((500, 500), pygame.RESIZABLE)

running = True
fullscreen = False

while running:
                fullscreen = not fullscreen
                if fullscreen is True:
                    screen = pygame.display.set_mode((500, 500), pygame.FULLSCREEN | pygame.SCALED)
                else:
                    screen = pygame.display.set_mode((500, 500), pygame.RESIZABLE)

        if event.type == pygame.QUIT:
            running = False

    screen.fill((255, 255, 255))
    pygame.draw.circle(screen, (0, 255, 0), (250, 250), 75)
    pygame.display.flip()

pygame.quit()
```

ankith26:
this PR fixes #2737 and fixes #2376 (they are essentially the same issue).

Note that I couldn't reproduce the original issue on Ubuntu, however @zoldalma999 has tested this branch and reported that this fix has worked on windows. Would be nice if someone gave this a test on a mac as well.

Also worth mentioning that the fix only works for SDL 2.28.0+, but all the wheels we distribute are on 2.30.3 so this shouldn't be a big deal.

It was also mentioned in one of the linked issues that #2571 is also the same issue, but I don't think that's the case.

### DIFF
--- a/src_c/display.c
+++ b/src_c/display.c
@@ -1215,6 +1215,15 @@ pg_set_mode(PyObject *self, PyObject *arg, PyObject *kwds)
                     SDL_SetHintWithPriority(SDL_HINT_RENDER_SCALE_QUALITY,
                                             "nearest", SDL_HINT_DEFAULT);
 
+#if SDL_VERSION_ATLEAST(2, 28, 0)
+                    /* If the window has a surface associated with it already,
+                     * we need to destroy it (if possible) because now we are
+                     * associating a renderer with it. */
+                    if (SDL_HasWindowSurface(win)) {
+                        SDL_DestroyWindowSurface(win);
+                    }
+#endif
+
                     if (vsync) {
                         pg_renderer = SDL_CreateRenderer(
                             win, -1, SDL_RENDERER_PRESENTVSYNC);
@@ -1224,8 +1233,7 @@ pg_set_mode(PyObject *self, PyObject *arg, PyObject *kwds)
                     }
 
                     if (pg_renderer == NULL) {
-                        return RAISE(pgExc_SDLError,
-                                     "failed to create renderer");
+                        return RAISE(pgExc_SDLError, SDL_GetError());
                     }
 
                     if (flags & PGS_SCALED) {


src_c/display.c
```diff
@@ -1238,6 +1238,17 @@ pg_set_mode(PyObject *self, PyObject *arg, PyObject *kwds)
                     SDL_SetHintWithPriority(SDL_HINT_RENDER_SCALE_QUALITY,
                                             "nearest", SDL_HINT_DEFAULT);
 
+#if SDL_VERSION_ATLEAST(2, 28, 0)
+                    /* If the window has a surface associated with it already,
+                     * we need to destroy it (if possible) because now we are
+                     * associating a renderer with it.
+                     * I don't know how to achieve the same effect on older
+                     * SDL versions that don't support below functions */
```
zoldalma999, line 1246: I don't think we need to do this stuff for versions below 2.28 (which added these functions) as they did not error out if they already had a surface.

zoldalma999 (approved):
As Anktith said, this works for me with SDL 2.30.2, with the original reproducer.

src_c/display.c
ankith26: In my latest code I have vendored in the destroy window logic from SDL code, now we are covered for all versions

ankith26:
A bit of interesting history: these new API functions were basically added for us by the SDL folks in response to issue #2190 so we have to make use of it!

src_c/display.c
```diff
@@ -824,6 +824,64 @@ _get_display(SDL_Window *win)
...
+PG_DestroyWindowSurface(SDL_Window *win)
+{
+    if (SDL_HasWindowSurface(win)) {
+        SDL_DestroyWindowSurface(win);
+    }
+}
+#else
+/* SDL_Window struct, as taken from the SDL source (except some pointer types
+ * have been replaced with `void *`) */
+struct SDL_Window {
```
Starbuck5, line 838: This is crazyyyyy, I appreciate the effort you put in Ankith!

I mean we've done stuff like it before but still

ankith26: thanks, it wasn't much effort, but yeah, I am fairly confident this approach is robust and can cover all our supported versions

ankith26: I actually had to revert this solution because it was leading to segfaults on 32-bit ARM, and I had no idea how those could be resolved.

rethanon:
This does seem to fix #2737 but doesn't appear to fix #2376. I tested it today and the error still happens on my Mac (but is fixed on my Windows 11 machine)

ankith26:
It doesn't fix #2376 as of now, but SDL has patched the underlying issue and the fix should be available in the SDL bump, so let this PR wait for that

ankith26:
After the latest rebase, this PR should fix the issue on mac as well.

rethanon:
Tested on my Mac (reminder it's a 2015 Intel based MacBook Pro) and yes indeed the tests were successful with the latest PR, thanks!

PR #2926, opened 2024-06-14
Very small typehint fix for pygame.Mask

damusss:
That method returns a list of rects not necessarily one.


ankith26 (approved):

MyreMylar (approved):
LGTM 👍

PR #2928, opened 2024-06-15
Remove SDL v1.2.10 checks in display.c

Starbuck5:
Ran into this while working on SDL3 support, since they changed how version number getting works. Removing this is an easy win though, regardless.


ankith26 (approved):
LGTM, thanks! :balloon:

MyreMylar (approved):
LGTM 👍

PR #2933, opened 2024-06-17
Fixes for rwobject (now iostream) SDL3

Starbuck5:
In SDL3, the RWops system has been remade into the IOStream system.

So, for SDL itself they have an oldnames compatibility layer that I've enabled at compile time for my SDL3 experimentation. This means our code can still say "SDL_RWops*" and other "RW" things. But they don't have this for the satellite libraries (yet?), so I've gone ahead and implemented explicit preprocessor macros to switch the "RW" load functions to the "IO" load functions.

The system has also changed how custom RWops are created, and the signature of the callbacks, so this patch fixes that to work in both SDL2 and SDL3. It's one of the more complex patches I've submitted for SDL3 support, but the big idea of it is that it doesn't change anything in SDL2, to stay safe.

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -85,8 +85,7 @@ sdl_ttf = '@0@_ttf'.format(sdl)
 sdl_image = '@0@_image'.format(sdl)
 
 if sdl_api == 3
-    add_global_arguments('-DPG_SDL3=1', language: 'c')
-    add_global_arguments('-DSDL_ENABLE_OLD_NAMES=1', language: 'c')
+    add_global_arguments('-DPG_SDL3=1', '-DSDL_ENABLE_OLD_NAMES=1', language: 'c')
 endif
 
 pg_inc_dirs = []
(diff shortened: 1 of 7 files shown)


Starbuck5:
1 failing check due to numpy.alltrue being removed in numpy 2.0.

MyreMylar (approved):
OK, this looks good to me 👍

ankith26:
This PR was made a while back, has there been any SDL3 changes in the meantime that makes this PR need changes?

Starbuck5:
> This PR was made a while back, has there been any SDL3 changes in the meantime that makes this PR need changes?

To compile: no
to compile without warnings (/ correctly) : yes

I've updated it now and it should be good.

I'm aware the rwobject.c changes probably look scary: all my changes have been for the SDL3 code paths, SDL2 is unaffected by these changes.

ankith26 (approved):
LGTM! This PR is perfectly SDL2 safe and passes SDL3 CI. Should be good to go now, thanks! :sunglasses:

Issue #2935, opened 2024-06-18
Have (0, 0) as default dest argument for Surface.blit


JiffyRob:
**Description**
In short, I would like to just be able to call `surface.blit(other_surface)` instead of `surface.blit(other_surface, (0, 0))`.  
I have been meaning to ask for this for a while.  The major use case for this, at least in my code, is when applying effects to entire surfaces, like darkening by making a grey surface and then blitting it onto another surface with `pygame.BLEND_RGB_MULTIPLY`, or drawing a blackish 'fog-of-war' over the map on screen.  A few `ack`, `grep`, and `wc` calls tell me that in my current pygame project, 11 of 42 blit calls use `(0, 0)` as their `dest` argument, and the rest are almost all `pygame.Rect` objects being passed in instead.

PR #2936, opened 2024-06-18
Add default dest (0, 0) to Surface.blit

damusss:
Implements https://github.com/pygame-community/pygame-ce/issues/2935

I'm not sure if you want this to be added, personally I like it. I didn't discuss it because the implementation seemed pretty simple, so even if you don't want it it's not a big deal. 
The reasoning for it can be understood in the issue description.

I only modified blit and not blits/fblits, but I'm pretty sure I can add it to them too if you want, but for now I don't think it's a good idea.

### DIFF
--- a/buildconfig/stubs/pygame/surface.pyi
+++ b/buildconfig/stubs/pygame/surface.pyi
@@ -69,7 +69,7 @@ class Surface:
     def blit(
         self,
         source: Surface,
-        dest: Union[Coordinate, RectValue],
+        dest: Optional[Union[Coordinate, RectValue]] = (0, 0),
         area: Optional[RectValue] = None,
         special_flags: int = 0,
     ) -> Rect: ...
--- a/buildconfig/stubs/pygame/surface.pyi
+++ b/buildconfig/stubs/pygame/surface.pyi
@@ -69,7 +69,7 @@ class Surface:
     def blit(
         self,
         source: Surface,
-        dest: Optional[Union[Coordinate, RectValue]] = (0, 0),
+        dest: Union[Coordinate, RectValue] = (0, 0),
         area: Optional[RectValue] = None,
         special_flags: int = 0,
     ) -> Rect: ...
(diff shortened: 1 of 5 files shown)


damusss:
I want to note that the failing check is related to numpy2 not this PR

buildconfig/stubs/pygame/surface.pyi
```diff
@@ -69,7 +69,7 @@ class Surface:
     def blit(
         self,
         source: Surface,
-        dest: Union[Coordinate, RectValue],
+        dest: Optional[Union[Coordinate, RectValue]] = (0, 0),
```
ankith26, line 72:
```suggestion
        dest: Union[Coordinate, RectValue] = (0, 0),
```
This is a common python typing pitfall...

`Optional` does not mean that an argument is optional, it means that it is allowed to be `None`, which isn't the case here

docs/reST/ref/surface.rst
```diff
@@ -90,16 +90,16 @@
...
       Draws another Surface onto this Surface.
 
       **Parameters**
           - ``source``
               The ``Surface`` object to draw onto this ``Surface``.
               If it has transparency, transparent pixels will be ignored when blittting to an 8-bit ``Surface``.
-          - ``dest``
-              The ``source`` draw position onto this ``Surface``.
+          - ``dest`` *(optional)*
+              The ``source`` draw position onto this ``Surface``, defaults to (0, 0).
```
ankith26, line 102: needs a `versionchanged` note

buildconfig/stubs/pygame/surface.pyi
damusss: Right, don't know how I let it happen! (fixed now). It's a bit weird that Optional doesn't mean optional, I'm glad the | operator exists in newer versions

ankith26 (approved):

This is an API change so I would like to have more eyes on this PR before we merge it in.

oddbookworm (approved):
LGTM!

itzpr3d4t0r (approved):
LGTM, should make it a bit faster as well in theory, and allows for shorter code.

PR #2937, opened 2024-06-18
Convert meson options from .txt to .options

novialriptide:
Click here for more info regarding `.options`: https://mesonbuild.com/Build-options.html


ankith26 (changes requested):
This PR may have to wait... it is failing our debian multiarch CI which uses system meson (that isn't on a new enough version)

ankith26:
Alternatively you could make the runner use `pip` provided meson instead of the `apt` provided one

PR #2939, opened 2024-06-18
Fix `Surface.fill` with rects with negative positions that overlap the surface

### LINKED ISSUE #2938: `Surface.fill()` `rect` argument: negative coordinates incorrectly set to 0
**Environment:**

<details>
<summary><code>pygame.print_debug_info()</code></summary>

```
Platform:       Linux-6.5.0-1022-gcp-x86_64-with-glibc2.39
System:         Linux
System Version:     #24~22.04.1-Ubuntu SMP Tue May 28 16:34:13 UTC 2024
Processor:
Architecture:       Bits: 64bit Linkage: ELF

Python:         CPython 3.11.9 (main, Apr  2 2024, 08:25:04) [GCC 13.2.0]
pygame version:     2.5.0
SDL versions:       Linked: 2.30.3  Compiled: 2.30.3
SDL Mixer versions: Linked: 2.8.0   Compiled: 2.8.0
SDL Font versions:  Linked: 2.22.0  Compiled: 2.22.0
SDL Image versions: Linked: 2.8.2   Compiled: 2.8.2
Freetype versions:  Linked: 2.13.2  Compiled: 2.13.2

Display Driver:     x11 ( xwayland == False )
Mixer Driver:       Mixer Not Initialized
```
</details>

**Current behavior:**

`Surface.fill` incorrectly handles negative coordinates for the `rect` argument if the rect partially overlaps with the surface.
Each negative coordinate (x, y) will be set to 0 before drawing without changing size, which basically changes the intended position of the rect. *Unless,* the rect does not overlap the surface at all, in which case nothing is drawn which is correct.
For example, the rect `(-50, 20, 80, 80)` would be drawn as `(0, 20, 80, 80)` instead of `(0, 20, 30, 80)`, much further right than expected. This affects the top and left edges.

`pygame.draw.rect` does not suffer from this bug.
Filling a rect going off the right or bottom edges works normally.
This bug seems to be unrelated to Surface clipping area.

**Expected behavior:**

`Surface.fill` treats negative coordinates intuitively and does not 'move' the rect inside the surface. `Surface.fill` should be able to draw just a part of the rect on the left and top edges if there are negative coordinates.
A possible fix implementation would be to add the negative coordinate(s) to the size before clamping them to 0.

**Steps to reproduce:**

1. Create a basic pygame program that shows the display.
2. Using either the display surface or a second surface that is blit on the display, call `surf.fill('green', (-60,-5,80,80))` .
3. Optionally, swap out that call with `pygame.draw.rect(surf, 'green', (-60,-5,80,80))` to see the correct drawing.

**Test code**

```python
import pygame as pg

display = pg.display.set_mode((640,360))

# display.set_clip((50,50,50,50))

while True:
  for event in pg.event.get():
    if event.type == pg.QUIT:
      raise Exception
  display.fill((16,16,16))
  print(display.fill('green', (-60,-5,80,80)))
  # print(pg.draw.rect(display, 'green', (-60,-5,80,80)))
  pg.display.update()
```

**Original issue**
#138
https://www.github.com/pygame/pygame/issues/123
The issue was from bitbucket (copied to github) in 2012, got a PR fixing an unrelated bug, then got closed

oddbookworm:
Fixes #2938

### DIFF
--- a/src_c/surface.c
+++ b/src_c/surface.c
@@ -1772,9 +1772,15 @@ surf_fill(pgSurfaceObject *self, PyObject *args, PyObject *keywds)
         }
 
         if (sdlrect.x < 0) {
+            if (abs(sdlrect.w) > abs(sdlrect.x)) {
+                sdlrect.w += sdlrect.x;
+            }
             sdlrect.x = 0;
         }
         if (sdlrect.y < 0) {
+            if (abs(sdlrect.h) > abs(sdlrect.y)) {
+                sdlrect.h += sdlrect.y;
+            }
             sdlrect.y = 0;
         }
 
--- a/src_c/surface.c
+++ b/src_c/surface.c
@@ -1755,59 +1755,32 @@ surf_fill(pgSurfaceObject *self, PyObject *args, PyObject *keywds)
         rect = &temp;
     }
 
-    if (rect->w < 0 || rect->h < 0 || rect->x > surf->w || rect->y > surf->h) {
-        sdlrect.x = sdlrect.y = 0;
-        sdlrect.w = sdlrect.h = 0;
+    // In SDL3, SDL_IntersectRect is renamed to SDL_GetRectIntersection
+    SDL_Rect surfrect = {0, 0, surf->w, surf->h};
+    if (!SDL_IntersectRect(rect, &surfrect, &sdlrect)) {
+        sdlrect.x = 0;
+        sdlrect.y = 0;
+        sdlrect.w = 0;
+        sdlrect.h = 0;
+    }
+
+    if (sdlrect.w <= 0 || sdlrect.h <= 0) {
+        return pgRect_New(&sdlrect);
+    }
+
+    if (blendargs != 0) {
+        result = surface_fill_blend(surf, &sdlrect, color, blendargs);
     }
     else {
-        sdlrect.x = rect->x;
-        sdlrect.y = rect->y;
-        sdlrect.w = rect->w;
-        sdlrect.h = rect->h;
-
-        // clip the rect to be within the surface.
-        if (sdlrect.x + sdlrect.w <= 0 || sdlrect.y + sdlrect.h <= 0) {
-            sdlrect.w = 0;
-            sdlrect.h = 0;
-        }
-
-        if (sdlrect.x < 0) {
-            if (abs(sdlrect.w) > abs(sdlrect.x)) {
-                sdlrect.w += sdlrect.x;
-            }
-            sdlrect.x = 0;
-        }
-        if (sdlrect.y < 0) {
-            if (abs(sdlrect.h) > abs(sdlrect.y)) {
-                sdlrect.h += sdlrect.y;
-            }
-            sdlrect.y = 0;
-        }
-
-        if (sdlrect.x + sdlrect.w > surf->w) {
-            sdlrect.w = sdlrect.w + (surf->w - (sdlrect.x + sdlrect.w));
-        }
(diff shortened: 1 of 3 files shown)


test/surface_test.py
```diff
@@ -4280,13 +4280,17 @@ def test_fill_negative_rectpos(self):
...
-        other_surface.fill("red", negative_x_rect_drawn)
-        other_surface.fill("blue", negative_y_rect_drawn)
-        other_surface.fill("green", negative_both_drawn)
+        red_rect_2 = other_surface.fill("red", negative_x_rect_drawn)
+        blue_rect_2 = other_surface.fill("blue", negative_y_rect_drawn)
+        green_rect_2 = other_surface.fill("green", negative_both_drawn)
+
+        self.assertEqual(red_rect_1, red_rect_2)
+        self.assertEqual(blue_rect_1, blue_rect_2)
+        self.assertEqual(green_rect_1, green_rect_2)
```
Matiiss, line 4293: What about using `assert rect_1 == rect_2`? The pixel test uses a simple `assert` and I think we have adopted `pytest` now?

oddbookworm: Oops, I didn't mean to use bare asserts lol, I missed that I did that originally

Starbuck5:
Since this is such an old bug, it could probably get a note in the docs?

Also is there really not a standard way of implementing clip rects, like SDL_IntersectRect maybe? What do the blitters do?

src_c/surface.c
```diff
@@ -1772,9 +1772,15 @@ surf_fill(pgSurfaceObject *self, PyObject *args, PyObject *keywds)
         }
 
         if (sdlrect.x < 0) {
+            if (abs(sdlrect.w) > abs(sdlrect.x)) {
```
damusss, line 1775: I have a very small note that is probably irrelevant/nitpick but I'm gonna say it anyways.
I am noticing that this code path is inside an else block that is executed only when the width and height are >= zero, so I think using abs() is redundant, and it applies to line 1781 aswell. We also know that x or y are negative so instead of abs(x|y) you can do -x|y, but it can stay like that for clarity.

oddbookworm:
> Since this is such an old bug, it could probably get a note in the docs?
> 
> Also is there really not a standard way of implementing clip rects, like SDL_IntersectRect maybe? What do the blitters do?

Note added

I replaced all the custom clipping with `SDL_IntersectRect`, but in SDL3, that function is renamed (left a comment in the code)

src_c/surface.c
oddbookworm: No longer needed with the most recent commit

ankith26 (approved):

I made a little interactive script to test this:
```py
import pygame

SPEED = 2

pygame.init()

screen = pygame.display.set_mode((200, 200))

box = pygame.Rect((50, 50, 50, 50))

clock = pygame.Clock()

run = True
while run:
    clock.tick(60)
        if event.type == pygame.QUIT:
            run = False
            break

    pressed = pygame.key.get_pressed()
    if pressed[pygame.K_UP]:
        box.y -= SPEED
    if pressed[pygame.K_DOWN]:
        box.y += SPEED
    if pressed[pygame.K_LEFT]:
        box.x -= SPEED
    if pressed[pygame.K_RIGHT]:
        box.x += SPEED

    screen.fill("black")
    screen.fill("red", box)
    pygame.display.update()
```
A simple box you can move around, after this PR the box can smoothly enter and exit all 4 edges of the window, whereas before this PR the top and left edge would have the weird behaviour.

ankith26:
Okay, seems like no one has opposed this fix so far, so let's get this in now.

PR #2940, opened 2024-06-19
EventState SDL3 compat

Starbuck5:
Continuing my long series of bite sized SDL3 patches.

Currently, a minimal compile requires patches on base, constants, display, event, ft_wrap.c, joystick, mouse, rwobject (fixed in #2933), and window.

This PR makes gets us a bit closer to taking event and joystick off that list.


ankith26 (approved):
LGTM, thanks :partying_face:

zoldalma999 (approved):

PR #2942, opened 2024-06-19
Docs generation is now `python -m buildconfig docs` and `python -m buildconfig docs full_generation`

oddbookworm:
I didn't want to have to invoke a specific python file every time I generate docs, so here's a useful helper


damusss (commented):
When running it from setup.py it says `Please use the following replacement: D:\Programs\Python312\python.exe buildconfig docs full_generation`  but running that command raises `ImportError: attempted relative import with no known parent package` as the suggestion is missing '`-m`'. Other than that, I like this helper

oddbookworm:
> When running it from setup.py it says `Please use the following replacement: D:\Programs\Python312\python.exe buildconfig docs full_generation` but running that command raises `ImportError: attempted relative import with no known parent package` as the suggestion is missing '`-m`'. Other than that, I like this helper

Oops, it's fixed now

damusss (approved):
LGTM, I also found it a bit annoying before

buildconfig/config.py
```diff
@@ -165,6 +166,10 @@ def main(auto=False):
 Only SDL2 is supported now.""")
 
     kwds = {}
+
+    if "docs" in sys.argv:
```
ankith26, line 170:
```suggestion
    if sys.argv[1] == "docs":
```
I would like this to be more stricter

oddbookworm: Committed

ankith26 (approved):

PS: the fails seem unrelated so I will merge this PR in now

PR #2944, opened 2024-06-19
Implementing warnings system and pygame.debug

### LINKED ISSUE #2894: `import pygame.debug`
**Description**

The idea here is to have a simple change that even a beginning user can make to their pygame script to enable a range of 'live' debugging information and warnings. It would replace `import pygame` and add debugging, diagnostic and warning information.

Example script:

```python
import pygame.debug
import pygame.gfxdraw

pygame.init()


display_surf = pygame.display.set_mode(
    (800, 600), flags=pygame.SCALED | pygame.HWSURFACE, vsync=1
)
pygame.display.set_caption("Test")

clock = pygame.time.Clock()

while True:
    clock.tick(60)

        if event.type == pygame.QUIT:
            pygame.quit()

    display_surf.fill("black")

    pygame.gfxdraw.aacircle(display_surf, 100, 100, 50, pygame.Color("white"))
    pygame.display.flip()

```

Example output (exact message format not relevant):
```
pygame-ce 2.5.0.dev3 (SDL 2.30.3, Python 3.11.1) - try pygame.print_debug_info() for more environment details

WARNING: flag pygame.HWSURFACE is deprecated and no longer does anything.
WARNING: using vsync together with a frame limit in clock.tick() is not recommended.
WARNING: pygame.gfxdraw.aacircle() is now superseded by pygame.draw.aacircle().
```

Whereas with the standard `import pygame` at the top we would default to a 'release' mode with nothing printed to the console by pygame-ce.

I hope this idea could resolve a bunch of stalled issues and longstanding debates regarding what warnings are too much and what we should or should not be printing to the console. The slight caveat is reducing user awareness of things like deprecation warnings so perhaps there could be some debate about what should remain in 'standard/release' mode and what would be hidden under `pgyame.debug` 

This simple change could be added to our beginner tutorials and some documentation example code snippets (perhaps with an explanatory comment). We could also advise it to users asking questions on reddit, discord and in issues here (a quick change to the bug report template) as a good first step when helping to diagnose problems.

What is incorporated into `import pygame.debug` could evolve over time and I think a basic implementation that moved, say the support prompt over to it would be enough to get started and resolve this issue.

gresm:
Fixes #2894

Notes (mostly to self):

- Should ``pygame.base.warn``/``pygame.base.get_warnings_filter``/``pygame.base.set_warnings_filter`` be public API?
- Current idea of warning priorities (can be changed):
  - 0 is the most impactful - if a warning of its type is triggered, probably the code won't work
  - 1 is less severe - like deprecation warnings
  - 2 is least important - warnings might be pedantic or about conventions or detection of common pitfalls that can contain false positives, so is suppressed by default
  - This should be documented somewhere.
-  Verifying that calls to ``pygame.base.warn``/``pgWarn`` have correct priority is necessary.
- There is no documentation for ``pgWarn`` yet.


docs/reST/ref/pygame.rst
```diff
@@ -224,6 +224,26 @@ object instead of the module, which can be used to test for availability.
 
    .. ## pygame.print_debug_info ##
 
+.. function:: warn
+
+   | :sl:`throws a warning`
+   | :sg:`warn(message, urgency, level=2, category=RuntimeWarning)`
```
bilhox, line 230: More informations have to be given on what is ``category`` argument and what it does.

bilhox (changes requested):

bilhox (approved):
Everything in the side of documentation seems fine. Maybe if you add a code example that makes full use of the function argument, it would be more helpful to users.

Starbuck5:
Hello, thanks for looking at this.

This is a complex patch, and I don't really see what adding a pygame-ce specific paradigm of filtering our existing warnings does for us.

I did not anticipate pygame.debug changing any existing warnings. It instead would add a whole new category of more opiniated ones and extra information. Maybe it could use SDL_Log and change the SDL log level so we get SDL debug information too? But then again maybe the juice isn't worth the squeeze for this.

gresm:
Well, this is a large patch, but it isn't complex - the logic behind it is quite simple, there are just a lot of places that needed to be updated to use the new warnings system (just renaming the ``PyErr_WarnEx`` to ``pgWarn`` and adding the urgency value).

Why I'm using what I'm using and not something like SDL warnings which have urgency? Because python warnings system was already used in the code base and I think that benefits that come from using python warnings system outweighs those of SDL warnings system (seamless integration with python code, advanced filtering system supporting regex, unit testing, etc...). An alternative implementation could create multiple filtering rules to ignore given warnings, but it would be slower (warnings that wouldn't be shown, would still need to go through the python warnings system) and harder to maintain (the necessity to keep the filter lists and updating them when a warning is added/removed/changed), so I think this is near optimal solution. That's my opinion on this matters.

Note that this PR still needs some finishing touches - ``pgWarn`` should behave the same as ``pygame.warn`` and ``pygame.debug`` isn't documented yet, I just didn't have time to do this.

Starbuck5:
> Well, this is a large patch, but it isn't complex - the logic behind it is quite simple, there are just a lot of places that needed to be updated to use the new warnings system (just renaming the PyErr_WarnEx to pgWarn and adding the urgency value).

Sure, but why do we need a new warnings system?

gresm:
This PR warnings system can be summarized as (omitting code for formatting warning message to include warning level):
```c
static int
pgWarn(PyObject *category, const char *message, Py_ssize_t stack_level,
       int urgency)
{
    if (pg_warn_filter < urgency)
        return 0;

    return PyErr_WarnEx(category, message, stack_level);
}
```
I used incorrect wording there - this is not much "a new warnings system", but rather a thin wrapper around an existing one (and heavily used before).
This is something like @MyreMylar propesed in https://github.com/pygame-community/pygame-ce/issues/2894#issuecomment-2143820154, but using a slightly different approach at how the warnings are filtered out.

Starbuck5:
But Myre is proposing adding a new warning there, not changing existing warnings to not trigger.

gresm:
Fair point, I'd like to hear your opinion on the warnings level I set for them (I've split warnings between 0, 1 and 2, where 0 and 1 are by default enabled with the intention of possibility introducing something like ``pygame.quiet`` if ever needed in the future).

PR #2945, opened 2024-06-20
Centralize CI config and make it more uniform (Adds macOS arm64 pypy CI)

ankith26:
The goal of this PR is to centralize all common cibuildwheel config, and make our CI more uniform.

We currently take different strategies on different workflows. manylinux takes the "build for all pythons in one workflow per arch" strategy. Mac takes an (ever so slightly harder to maintain) strategy of handwritten "split jobs evenly" while windows takes the "every run on its own" thing. 
I figured that the strategy manylinux takes is actually the most efficient, even if it's slower than the rest. Why?
- We can re-use more stuff across runs - cibuildwheel dependencies, buildtime dependencies, SDL prebuilt downloads (in windows) and doc building.
- By using less concurrent workflows per PR we are actually making room for more across-PR CI concurrency.
- Simpler config to maintain. Want to support a new python version on CI? It's now a one line change in `pyproject.toml`
- There is no real advantage from a PR maker POV if all windows builds can finish in 8 minutes, and the linux one is going to take like 15 minutes anyways which the PR maker has to wait for.
- It is also wasteful if there is say, a windows specific CI fail that just causes multiple separate runs to fail when one fail would be enough to convey the message.

Also switches build backend to `build[uv]` which should again save some CI time.

I may also explore more in the caching area in future PRs and see how that improves things.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -60,6 +60,14 @@ install = ['--tags=runtime,python-runtime,pg-tag']
 # uncomment to get werror locally
 # setup = ["-Derror_on_warns=true"]
 
+[tool.cibuildwheel]
+build = "cp3{8,9,10,11,12}-* pp3{8,9,10}-*"
+skip = "*-musllinux_*"
+build-verbosity = 3
+
+environment = { SDL_VIDEODRIVER="dummy", SDL_AUDIODRIVER="disk" }
+test-command = "python -m pygame.tests -v --exclude opengl,music,timing --time_out 300"
+
 [tool.cibuildwheel.config-settings]
 setup-args = [
     "-Derror_on_warns=true",
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -74,6 +74,7 @@ skip = "*-musllinux_*"
 
 environment = { SDL_VIDEODRIVER="dummy", SDL_AUDIODRIVER="disk" }
 test-command = "python -m pygame.tests -v --exclude opengl,music,timing --time_out 300"
+test-requires = ["numpy"]
 
 [tool.cibuildwheel.config-settings]
 setup-args = [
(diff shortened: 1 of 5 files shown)


ankith26:
While doing this PR I also happened to
- Add CI for pypy arm64 on macOS which we completely missed somehow
- Add numpy-based tests on linux

Here is a table of cumulated-time comparisons, along with per config notes

| Platform        | Before (main)| After (this PR) | Extra Notes |
| ------------------- |-------------------| ---------|---------|
| Windows x86_64| 37m 33s (3m 37s, 3m 10s, 3m 52s, 4m 23s, 4m 13s, 4m 30s, 4m 50s, 8m 58s)|15m 19s|Numpy tests removed on pypy 3.10|
| Windows x86| 19m 22s (4m 16s, 3m 5s, 4m 6s, 4m 45s, 3m 10s)|9m 10s|-|
| macOS arm64  | 15m 0s (10m 23s, 4m 37s)|10m 15s|Added 3 new pypy runs|
| macOS x86_64  |27m 7s (7m 49s, 8m 30s, 10m 48s)|15m 8s|Numpy tests removed on pypy 3.10|
| Linux arm64 | 13m 43s |11m 48s|Added numpy for tests where possible|
| Linux x86_64 | 11m 43s |10m 44s|Added numpy for tests where possible|
| Linux x86 | 12m 58s |11m 1s|-|

pyproject.toml
```diff
@@ -59,6 +59,17 @@ install = ['--tags=runtime,python-runtime,pg-tag']
 # uncomment to get werror locally
 # setup = ["-Derror_on_warns=true"]
 
+[tool.cibuildwheel]
+build-frontend = "build[uv]"
```
Starbuck5, line 63: I'd like a comment here explaining why exactly we're using uv / what this line of code does / why it makes it faster.

pyproject.toml
```diff
@@ -75,3 +86,17 @@ exclude = [
 
 [tool.ruff.format]
 quote-style = "preserve"
+
+# numpy is a test dependency, but we build for systems that numpy doesn't have
+# binary wheels for. In such cases, we do not want to waste CI time building
+# numpy from source. So, we are gonna force numpy to be "only-binary" and skip
+# numpy on platforms that it doesn't have wheels for
+[tool.uv.pip]
```
Starbuck5, line 94: This impacts when something in our project tries to pip install something with uv, and doesn't impact when something tries to install us with uv?

ankith26: This only impacts someone running the equivalent of `uv pip install numpy` in our source directory, and the impact is that `uv pip` will refuse to build `numpy` from source.

pyproject.toml
ankith26: Done

Starbuck5 (approved):
Thanks for working on this Ankith.

oddbookworm (approved):
LGTM!

PR #2950, opened 2024-06-23
Make window subclassable

### LINKED ISSUE #2625: Window should be subclassable

zoldalma999:
Fixes #2625

### DIFF
--- a/src_c/window.c
+++ b/src_c/window.c
@@ -1141,6 +1141,7 @@ static PyTypeObject pgWindow_Type = {
     PyVarObject_HEAD_INIT(NULL, 0).tp_name = "pygame.window.Window",
     .tp_basicsize = sizeof(pgWindowObject),
     .tp_dealloc = (destructor)window_dealloc,
+    .tp_flags = Py_TPFLAGS_DEFAULT | Py_TPFLAGS_BASETYPE,
     .tp_doc = DOC_WINDOW,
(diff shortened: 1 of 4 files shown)


src_c/window.c
```diff
@@ -96,8 +96,7 @@ pg_display_resource(char *filename)
 
 static PyTypeObject pgWindow_Type;
 
-#define pgWindow_Check(x) \
-    (PyObject_IsInstance((x), (PyObject *)&pgWindow_Type))
+#define pgWindow_Check(x) (PyType_IsSubtype(Py_TYPE(x), &pgWindow_Type))
```
ankith26, line 99: Does this line need changing?

According to the [`PyObject_IsInstance`](https://docs.python.org/3/c-api/object.html#c.PyObject_IsInstance) docs:
> Return 1 if inst is an instance of the class cls or a subclass of cls, or 0 if not. On error, returns -1 and sets an exception.

zoldalma999: You are right, that should work. Changed it back

ankith26 (approved):

PR #2952, opened 2024-06-23
Use static inline instead of inline

zoldalma999:
Note by @ankith26 as for why this PR is needed: This fixes gcc/mingw compilation failures when pygame is compiled with debug flags


ankith26 (approved):

damusss (approved):
LGTM :)

PR #2959, opened 2024-06-29
Get rid of docscomments.js from docs

Notenlish:
In `docs/reST/classic/elements.html` this element is included in the file: `<script type="text/javascript" src="https://www.pygame.org/comment/jquery.plugin.docscomments.js"></script>`. 

However this file isn't necessary and it is also fetched from the original pygame website. This was probably forgotten when the pygame-ce fork was made.

I ran `py -m setup.py docs` and can confirm that the documentation still works when the script is removed.

Here are some screenshots with the file removed:
image
image
image

Credits to @mzivic7  from pygame community server for pointing this out and @oddbookworm for realizing that it's not needed.

### DIFF
--- a/docs/reST/themes/classic/elements.html
+++ b/docs/reST/themes/classic/elements.html
@@ -107,5 +107,5 @@ We render three sets of items based on how useful to most apps.
 
 
 {%- block relbaritems %}
-    <script type="text/javascript" src="https://www.pygame.org/comment/jquery.plugin.docscomments.js"></script>
+
 {% endblock %}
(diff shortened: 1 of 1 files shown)


docs/reST/themes/classic/elements.html
```diff
@@ -107,5 +107,5 @@ <h5>pygame-ce documentation</h5>
 
 
 {%- block relbaritems %}
```
damusss, line 109: Is the block around the link necessary after the script is removed?

Notenlish: it raises an error if its not there

Notenlish: but if we get rid of both `{%- block relbaritems %}` and the end block it works

damusss (approved):
Alright, LGTM then!

ankith26 (approved):
LGTM, tested locally and nothing seems broken.

Thanks for contributing! :tada:

PR #2960, opened 2024-06-30
Changed dark theme slightly to make code more readable in some cases

oddbookworm:
Current version of https://pyga.me/docs/tutorials/en/chimp-explanation.html
image

This PR's version
image

Someone reported that code blocks can be hard to read because of the dark blue function names sometimes, so I fixed it. As far as I know, I got all the cases, but please lmk if I missed any

### DIFF
--- a/docs/reST/themes/classic/static/pygame.css_t
+++ b/docs/reST/themes/classic/static/pygame.css_t
@@ -606,6 +606,10 @@ dt code {
     color: {{ theme_dark_functiondefinecolor }};
 }
 
+.dark-theme .highlight .nc {
+    color: {{ theme_dark_classdefinecolor }}; font-weight: bold
+}
+
 code.download span.pre {
     font-family: inherit;
     font-weight: normal;
--- a/docs/reST/themes/classic/static/pygame.css_t
+++ b/docs/reST/themes/classic/static/pygame.css_t
@@ -601,6 +601,11 @@ dt code {
     color: {{ theme_dark_codecomment }};
 }
 
+.dark-theme .highlight .nf,
+.dark-theme .highlight .fm {
+    color: {{ theme_dark_functiondefinecolor }};
+}
+
 code.download span.pre {
     font-family: inherit;
     font-weight: normal;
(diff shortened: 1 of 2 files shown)


itzpr3d4t0r:
Not a real review but why must everything be green? Like it's a bit sickening.

oddbookworm:
I chose it semi-arbitrarily. Seemed like it would fit in with the rest of the theme. Alternative color suggestions welcome

damusss:
> I chose it semi-arbitrarily. Seemed like it would fit in with the rest of the theme. Alternative color suggestions welcome

[copied from discord]
I think some kind of light blue is fine, like #50b0fa, but i got that from a color picker quickly so maybe you can do better

oddbookworm:
@damusss @itzpr3d4t0r here's what it looks like with the color suggested
image

damusss:
> @damusss @itzpr3d4t0r here's what it looks like with the color suggested
> image
> 

The color I suggested is very similar to strings and class names. While it's not a problem to me, what about yellow? My vs code theme has it and it's great (we could use my color for the class name imo, it's brighter)

MyreMylar:
Looks like this:

image

docs/reST/themes/classic/static/pygame.css_t
```diff
@@ -601,6 +601,11 @@ dt code {
     color: {{ theme_dark_codecomment }};
 }
 
+.dark-theme .highlight .nf,
+.dark-theme .highlight .fm {
+    color: {{ theme_dark_functiondefinecolor }};
+}
+
```
MyreMylar, line 608:
```suggestion

.dark-theme .highlight .nc {
    color: {{ theme_dark_classdefinecolor }}; font-weight: bold
}

```

docs/reST/themes/classic/theme.conf
```diff
@@ -56,6 +56,7 @@ dark_notebgcolor            = #555349
 dark_notebordercolor        = #6e6b5e
 dark_cautionbgcolor         = rgb(255 25 25 / 30%)
 dark_cautionbordercolor     = rgb(255 60 60)
+dark_functiondefinecolor    = #50b0fa
```
MyreMylar, line 59:
```suggestion
dark_functiondefinecolor    = #fee32d
dark_classdefinecolor       = #50b0fa
```

MyreMylar (approved):
LGTM. More readable.

itzpr3d4t0r (approved):
LGTM

PR #2966, opened 2024-07-02
Adjust tobytes premul formula

ankith26:
While I'm working on my `tobytes` improvement branch, I ported the code to use `Surface.premul_alpha` to handle `RGBA_PREMULT` and `ARGB_PREMULT`.

The problem is, `Surface.premul_alpha` uses a slightly different formula: `((pixel + 1) * alpha) >> 8` (which is equivalent to `((pixel + 1) * alpha) / 256`) instead of the classic `(pixel * alpha) / 255` formula, probably for performance reasons. I believe these two formulas are close enough to result in very similar looking outputs.

Hence this PR: I figured I should separate out the slight behaviour change into it's own PR and keep the refactor PR for later.

### DIFF
--- a/src_c/image.c
+++ b/src_c/image.c
@@ -495,6 +495,8 @@ tobytes_surf_32bpp(SDL_Surface *surf, int flipped, int hascolorkey,
     }
 }
 
+#define PREMUL_PIXEL_ALPHA(pixel, alpha) (char)((((pixel) + 1) * (alpha)) >> 8)
+
 PyObject *
 image_tobytes(PyObject *self, PyObject *arg, PyObject *kwarg)
 {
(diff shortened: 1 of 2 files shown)


damusss (approved):
Makes sense, LGTM

src_c/image.c
```diff
@@ -992,15 +991,12 @@ image_tobytes(PyObject *self, PyObject *arg, PyObject *kwarg)
...
-                            (char)((((color & Bmask) >> Bshift) << Bloss) *
-                                   alpha / 255);
+                        data[0] = PREMUL_PIXEL_ALPHA(
+                            ((color & Rmask) >> Rshift) << Rloss, alpha);
+                        data[1] = PREMUL_PIXEL_ALPHA(
+                            ((color & Gmask) >> Gshift) << Gloss, alpha);
+                        data[2] = PREMUL_PIXEL_ALPHA(
+                            ((color & Bmask) >> Bshift) << Bloss, alpha);
                         data[3] = (char)alpha;
                         data += 4;
```
itzpr3d4t0r, line 1001: Afaik there's no format supporting per-pixel alpha on 3Bpp surfaces, like you can't do:
```Py
import pygame

pygame.init()

screen = pygame.display.set_mode((100, 100))

s = pygame.Surface((10, 10), pygame.SRCALPHA, depth=24)
```
As this is the result:
`ValueError: no standard masks exist for given bitdepth with alpha`

Moreover, premultiplying a surface's RGB by alpha=255 (such as in 3Bpp surfaces) will always be equal to the original RGB, so it wouldn't make a difference even if the tranformation was applied.
So this whole case could just be:
- a single `memcpy` call if not padding and the source format being the exact same as the expected output format.
- `memcpy` loop + `memset(0)` if padding and the source format being the exact same as the expected output format.

itzpr3d4t0r (commented):
The current setup is fine for now, though could be expanded in te future for better performance, with both simd and better scalar code.

src_c/image.c
ankith26: The goal of this PR is to get the change in as easily as possible. Optimizations and cleanups coming up later.

itzpr3d4t0r (approved):
LGTM!

MyreMylar (approved):
LGTM 👍 

The formula is different for SIMD efficiency purposes basically. in SIMD land divides suck but shifts are great,

PR #2968, opened 2024-07-03
Add Surface.pixel_format

damusss:
I implemented what the titles says including documentation, stubs and tests. Feel free to suggest different format names or anything else.

### DIFF
--- a/src_c/surface.c
+++ b/src_c/surface.c
@@ -3173,7 +3173,7 @@ surf_get_pixel_format(pgSurfaceObject *self, PyObject *_null)
             pixel_format = "RGB";
             break;
 
-#if SDL_VERSION_ATLEAST(1, 0, 12)
+#if SDL_VERSION_ATLEAST(2, 0, 12)
         case SDL_PIXELFORMAT_BGR444:
 #endif
         case SDL_PIXELFORMAT_BGR555:
--- a/src_c/surface.c
+++ b/src_c/surface.c
@@ -3173,7 +3173,9 @@ surf_get_pixel_format(pgSurfaceObject *self, PyObject *_null)
             pixel_format = "RGB";
             break;
 
+#if SDL_VERSION_ATLEAST(1, 0, 12)
         case SDL_PIXELFORMAT_BGR444:
+#endif
         case SDL_PIXELFORMAT_BGR555:
         case SDL_PIXELFORMAT_BGR565:
         case SDL_PIXELFORMAT_BGR24:
--- a/src_c/surface.c
+++ b/src_c/surface.c
@@ -196,6 +196,8 @@ static PyObject *
 surf_premul_alpha(pgSurfaceObject *self, PyObject *args);
 static PyObject *
 surf_premul_alpha_ip(pgSurfaceObject *self, PyObject *args);
+static PyObject *
+surf_get_pixel_format(pgSurfaceObject *self, PyObject *args);
 static int
 _view_kind(PyObject *obj, void *view_kind_vptr);
 static int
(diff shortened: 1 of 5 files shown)


src_c/surface.c
```diff
@@ -3151,6 +3155,113 @@ surf_premul_alpha_ip(pgSurfaceObject *self, PyObject *_null)
...
+        case SDL_PIXELFORMAT_NV21:
+            pixel_format = "NV21";
+            break;
+
+        default:
+            /* unknown format already set */
+            break;
+    }
+
+    return Py_BuildValue("s", pixel_format);
```
itzpr3d4t0r, line 3262: I think you can just use `PyUnicode_FromString` here.

src_c/surface.c
```diff
@@ -3151,6 +3155,114 @@ surf_premul_alpha_ip(pgSurfaceObject *self, PyObject *_null)
...
+            break;
+
+        case SDL_PIXELFORMAT_RGB888:
+        case SDL_PIXELFORMAT_RGB565:
+        case SDL_PIXELFORMAT_RGB555:
+        case SDL_PIXELFORMAT_RGB444:
+        case SDL_PIXELFORMAT_RGB332:
+        case SDL_PIXELFORMAT_RGB24:
+            pixel_format = "RGB";
+            break;
```
ankith26, line 3179: We shouldn't be clubbing all these formats under the "RGB" section.

`SDL_PIXELFORMAT_RGB888` is actually `SDL_PIXELFORMAT_XRGB8888`, a 32-bit format whereas `SDL_PIXELFORMAT_RGB565`, is a 16-bit format. The only true 24-bit RGB format here is `SDL_PIXELFORMAT_RGB24`, which is equivalent to `"RGB"` in `image.tobytes`

ankith26 (commented):
I have left only one review, but the same point applies to other blocks here as well.

There are a couple of complexities to consider, some of these pixel formats are actually aliases of others that have different values based on the system byteorder and stuff

Starbuck5 (changes requested):
I would like to not merge this PR in its current state.

Adding pixel format support is a good initiative, but I would like a clear plan for how this will interact with other things.

In SDL, some functions have variants that take pixelformatenum values like these, and some take the whole format struct. Should we expose the whole format struct as an object?

Surface.format (the format structure object)
Surface.format.format (pixelformat enum value)

What needs to be added to the pygame API to make pixelformat valuable? An optional kwarg in convert seems like it would be needed at least. Textures need a format.

Other concerns for pixel formats: SDL3 is adding colorspaces, how do those work. Endianness? YUV formats?

damusss:
@Starbuck5 The proposal is interesting but I'm not sure how I feel about this. Every entry in the format struct is already covered by a Surface method (get_shifts, get_bytesizs, get_pitch, get_masks, get_losses, get_palette) so it seems a bit redundant unless all of those are deprecated or something (even tho I'm aware there are already useful attributes that copy some functions so it might not be too bad). The only thing missing is the pixel format. If it was an int enum like in sdl we'd need many pygame.PIXEL_FORMAT_X constants. Ankith seemed to have a slightly different idea, so when you decide you can come back to me.

Starbuck5:
Yes all the other data is already exposed, but it's not grouped together like it is in SDL.

In SDL when you're converting a surface to another surface's format, you can use the other surface's format. Here you won't be able to in a complete way, because it doesn't contain palette information.

Also how will these strings be mapped back into pixelformat enum values. My thought is that int constants might be better.

We need to plan to export the data in a way that can be directly ingested into other functions, like pygame.Surface.convert().

damusss:
@Starbuck5 Alright then, it makes sense. What will happen to the idea of being able to pass the format to tobytes() using a surface attribute? following your proposal, I think str(SurfaceFormat) should return what this PR is currently implementing, or something similar, or maybe have an attribute on the format with its stringified version. If this decision is finalized this PR could be closed in favor of another one implementing your proposal. Might be good to hear what @ankith26 thinks too IMO.

Starbuck5:
> What will happen to the idea of being able to pass the format to tobytes() using a surface attribute? 

What's the point of doing that? People convert surfaces to bytes to pass to other applications/libraries, the format of the surface doesn't matter, it's the output format that matters.

damusss:
> > What will happen to the idea of being able to pass the format to tobytes() using a surface attribute?
> 
> What's the point of doing that? People convert surfaces to bytes to pass to other applications/libraries, the format of the surface doesn't matter, it's the output format that matters.

yeah that's true idk why it was fixed in my mind. alright then

damusss:
Closed this as the PixelFormat/SurfaceFormat object proposal is cleaner and easier to expand, which will probably be added post SDL3 (where this PR idea could be implemented). For more check [this](https://discord.com/channels/772505616680878080/772940667231928360/1259858159116357692) messages.

PR #2972, opened 2024-07-05
Fix `premul_alpha_ip` doc indentation issue

Matiiss:


ankith26 (approved):
LGTM, thanks! :tada:

oddbookworm (approved):
I suppose we can make **space** for this change, LGTM and thanks lol

bilhox:
I just noticed that the new attributes of Surface were placed at the end of the docs, they should be at the beginning just like other docs pages. Since you're "updating ``surface.rst``", it might be interesting to change it, so people use properties rather than getters.

Starbuck5:
> I just noticed that the new attributes of Surface were placed at the end of the docs, they should be at the beginning just like other docs pages. Since you're "updating surface.rst" (name updated by ankith), it might be interesting to change it, so people use properties rather than getters.

Regardless of whether this a problem or not, it should be a separate issue/PR, this one is small and focused already. Also some other docs pages don't show the properties in the list at all (Rect).

ankith26:
agree with starbuck, so that this PR can be merged in now

PR #2973, opened 2024-07-06
Remove obsolete surflock functions

Starbuck5:
In 2018, llindstrom wrote of pgSurface_LockLifetime in surflock.rst that "This function is not called anywhere within pygame. It and pgLifetimeLock_Type are candidates for removal."

In the intervening 6 years, the functions and documentation have not been meaningfully edited, and these systems are still not used anywhere.

So let's remove them.

### DIFF
--- a/docs/reST/c_api/surflock.rst
+++ b/docs/reST/c_api/surflock.rst
@@ -15,33 +15,6 @@ This extension module implements SDL surface locking for the
 Header file: src_c/include/pygame.h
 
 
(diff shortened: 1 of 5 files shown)


ankith26 (approved):
No use of keeping dead code around, nice cleanups! :+1:

damusss (approved):
I agree with Starbuck and Ankit, LGTM :)
I'm also curios if the functions were added and never used (why?) or just stopped being used.

PR #2974, opened 2024-07-06
Use PyObject_CallMethod in resource loaders

Starbuck5:
This is a slight simplification from the existing strategy of getting the functions then calling them. The new way has fewer lines of code, and doesn't need to manage the refcount and existence of references to functions, so it is simpler and likely faster.

The resource loaders have way bigger problems-- they are using a deprecated setuptools API, they likely should be reworked to use the replacement API from importlib. I also think they're being over-cautious with handling of file objects, I think they should just hand off directly to rwobject.c instead of faffing around with the .name and trying to close them if they're streams.

This PR doesn't touch that since that would be complex to research and to review. The current change is low hanging fruit but it's worthwhile too.

### DIFF
--- a/src_c/_freetype.c
+++ b/src_c/_freetype.c
@@ -231,7 +231,6 @@ load_font_res(const char *filename)
 {
     PyObject *load_basicfunc = 0;
     PyObject *pkgdatamodule = 0;
-    PyObject *resourcefunc = 0;
     PyObject *result = 0;
     PyObject *tmp;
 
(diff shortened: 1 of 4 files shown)


damusss (approved):
Basically replaced `func = self.function;func()` with `self.function()`,  makes sense to me 👍

ankith26 (approved):
LGTM, thanks! :tada:

PR #2975, opened 2024-07-07
Version-getters SDL2/SDL3

Starbuck5:
Continuing my SDL3 patch series.

This takes a look at all "version getters" from SDL and satellite libraries in one go, since they've all consistently updated.

### DIFF
--- a/src_c/_pygame.h
+++ b/src_c/_pygame.h
@@ -106,6 +106,10 @@ PG_UnlockMutex(SDL_mutex *mutex)
 #define PG_SetJoystickEventsEnabled(enabled) \
     SDL_SetJoystickEventsEnabled(enabled)
 
+#define PG_FIND_VNUM_MAJOR(ver) SDL_VERSIONNUM_MAJOR(ver)
+#define PG_FIND_VNUM_MINOR(ver) SDL_VERSIONNUM_MINOR(ver)
+#define PG_FIND_VNUM_MICRO(ver) SDL_VERSIONNUM_MICRO(ver)
+
 #else /* ~SDL_VERSION_ATLEAST(3, 0, 0)*/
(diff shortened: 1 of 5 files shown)


src_c/_pygame.h
```diff
@@ -106,6 +106,10 @@ PG_UnlockMutex(SDL_mutex *mutex)
 #define PG_SetJoystickEventsEnabled(enabled) \
     SDL_SetJoystickEventsEnabled(enabled)
 
+#define PG_FIND_VNUM_MAJOR(ver) SDL_VERSIONNUM_MAJOR(ver)
+#define PG_FIND_VNUM_MINOR(ver) SDL_VERSIONNUM_MINOR(ver)
+#define PG_FIND_VNUM_MICRO(ver) SDL_VERSIONNUM_MICRO(ver)
```
ankith26, line 111: the macros usually copy SDL3 naming, why is this an exception?

Starbuck5: I didn’t want the macro to look like our version number macros, that express our versions.

Starbuck5: Because I don’t want us to be defining SDL symbols or things that look like SDL symbols. We should define our own symbols.

Starbuck5: I've never used PGS_ in any of these SDL3 patches, I was unaware of what that stood for.

I don't want to change the name of all of our compat symbols in this PR.

ankith26: Fair enough then, we can address it later

ankith26 (approved):

MyreMylar (approved):
OK, all looks good to me and the relevant tests and functions for version checking all still work. 👍

PR #2679, opened 2024-01-19
Reduced redundancy in `surface.fblits` code and handled generator exception

Matiiss:
Somewhat of a continuation of https://github.com/pygame-community/pygame-ce/pull/2667

PR #2978, opened 2024-07-07
Restore inlining for `.fblits()` internal functionality

itzpr3d4t0r:
A previous PR (#2679) improved the fblits function by condensing the core loop into a separate function. However, this change slightly impacted performance, particularly with small surfaces, which are the primary focus of the fblits function. This new PR aims to restore the original performance without significantly increasing the code size.

Interestingly, the results are as follows:

New: 234,496 bytes
Old: 235,008 bytes

This minor discrepancy of 512 bytes represents a 0.2% increase in code size, which is negligible.
image
The test demonstrates a 10x10 pixel surface being drawn onto a 1000x1000 screen, with the x-axis representing the number of surfaces drawn.

### DIFF
--- a/src_c/surface.c
+++ b/src_c/surface.c
@@ -2107,7 +2107,7 @@ bliterror:
 #define FBLITS_ERR_INCORRECT_ARGS_NUM 12
 #define FBLITS_ERR_FLAG_NOT_NUMERIC 13
 
-int
+static int PG_FORCEINLINE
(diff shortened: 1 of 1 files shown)


oddbookworm (approved):
The minor filesize change is fine with me, I support it in favor of the speed diffs you show

ankith26 (approved):
LGTM, 512 bytes are not a big deal.

PR #2979, opened 2024-07-07
Condense `SDL_PixelFormat` copy code in `pgSurface_Blit()`

itzpr3d4t0r:
This PR tries to make code more compact, legible and not dependant on SDL versions. Not completely sure this will work though.

### DIFF
--- a/src_c/surface.c
+++ b/src_c/surface.c
@@ -476,7 +476,7 @@ surface_str(PyObject *self)
 }
 
 static intptr_t
-surface_init(pgSurfaceObject *self, PyObject *args, PyObject *kwds) c
+surface_init(pgSurfaceObject *self, PyObject *args, PyObject *kwds)
 {
     Uint32 flags = 0;
     int width, height;
--- a/src_c/surface.c
+++ b/src_c/surface.c
@@ -3841,29 +3841,15 @@ pgSurface_Blit(pgSurfaceObject *dstobj, pgSurfaceObject *srcobj,
             result = pygame_Blit(src, srcrect, dst, dstrect, 0);
         }
         else {
-            SDL_PixelFormat *fmt = src->format;
             SDL_PixelFormat newfmt;
 
+            memcpy(&newfmt, src->format, sizeof(SDL_PixelFormat));
+
             newfmt.palette = 0; /* Set NULL (or SDL gets confused) */
-#if SDL_VERSION_ATLEAST(3, 0, 0)
-            newfmt.bits_per_pixel = fmt->bits_per_pixel;
-            newfmt.bytes_per_pixel = fmt->bytes_per_pixel;
-#else
-            newfmt.BitsPerPixel = fmt->BitsPerPixel;
-            newfmt.BytesPerPixel = fmt->BytesPerPixel;
-#endif
             newfmt.Amask = 0;
-            newfmt.Rmask = fmt->Rmask;
-            newfmt.Gmask = fmt->Gmask;
-            newfmt.Bmask = fmt->Bmask;
             newfmt.Ashift = 0;
-            newfmt.Rshift = fmt->Rshift;
-            newfmt.Gshift = fmt->Gshift;
-            newfmt.Bshift = fmt->Bshift;
             newfmt.Aloss = 0;
-            newfmt.Rloss = fmt->Rloss;
-            newfmt.Gloss = fmt->Gloss;
-            newfmt.Bloss = fmt->Bloss;
+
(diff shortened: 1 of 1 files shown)


src_c/surface.c
```diff
@@ -3841,29 +3841,15 @@ pgSurface_Blit(pgSurfaceObject *dstobj, pgSurfaceObject *srcobj,
             result = pygame_Blit(src, srcrect, dst, dstrect, 0);
         }
         else {
-            SDL_PixelFormat *fmt = src->format;
             SDL_PixelFormat newfmt;
 
+            memcpy(&newfmt, src->format, sizeof(SDL_PixelFormat));
```
oddbookworm, line 3846: I'm not 100% sold that `memcpy` is more readable than what's already there. Especially since we still have to go set specific fields after this. There are actually other fields too that may need to be specially handled. According to https://wiki.libsdl.org/SDL2/SDL_PixelFormat, shouldn't we also ensure that
* `refcount` gets zeroed out on the new one
* `next` gets set to `NULL` (is that right?)
* Currently we don't handle the `padding` or `format` fields. What should we actually do with them?

src_c/surface.c
```diff
@@ -3841,29 +3841,15 @@ pgSurface_Blit(pgSurfaceObject *dstobj, pgSurfaceObject *srcobj,
...
+
             newfmt.palette = 0; /* Set NULL (or SDL gets confused) */
-#if SDL_VERSION_ATLEAST(3, 0, 0)
-            newfmt.bits_per_pixel = fmt->bits_per_pixel;
-            newfmt.bytes_per_pixel = fmt->bytes_per_pixel;
-#else
-            newfmt.BitsPerPixel = fmt->BitsPerPixel;
-            newfmt.BytesPerPixel = fmt->BytesPerPixel;
-#endif
```
oddbookworm, line 3854: If we end up not going with this PR and keeping things the way they currently are, looks like in SDL3, the `*loss* fields get renamed to `*bits` https://wiki.libsdl.org/SDL3/SDL_PixelFormat

src_c/surface.c
ankith26: I believe in the current code, those values are garbage anyways.

oddbookworm: > I believe in the current code, those values are garbage anyways.

And that spooks me somewhat. We really should be ensuring that all fields are set to at least a default value if not copied from the source IMO. Seems like a good way to get weird segfaults because of something internal to SDL

itzpr3d4t0r: Yeah i concour with that, actually funnily enough those attributes were uninitialized before, so yeah unless we check SDL's internals we can't assure weird things won't happen.

oddbookworm:
@itzpr3d4t0r that recent commit confuses me in context of all the rest of the changes here. Am I missing something, or is this incomplete for now?

itzpr3d4t0r:
> @itzpr3d4t0r that recent commit confuses me in context of all the rest of the changes here. Am I missing something, or is this incomplete for now?

You can shallow copy a struct without memcpy, so i did just that, it's not incomplete.

src_c/surface.c
```diff
@@ -3841,29 +3841,13 @@ pgSurface_Blit(pgSurfaceObject *dstobj, pgSurfaceObject *srcobj,
             result = pygame_Blit(src, srcrect, dst, dstrect, 0);
         }
         else {
-            SDL_PixelFormat *fmt = src->format;
-            SDL_PixelFormat newfmt;
+            SDL_PixelFormat newfmt = *src->format;
 
             newfmt.palette = 0; /* Set NULL (or SDL gets confused) */
```
oddbookworm, line 3846: I would like to see two changes here:
- `newfmt.palette = NULL` for clarity that it's a pointer type
- Addressing `newfmt.next`, which is also a pointer. I don't want to end up copying a pointer to memory that ends up being garbage later because we copied a pointer to memory this object doesn't own

src_c/surface.c
oddbookworm: Update:
[this entire struct got renamed in SDL3 as of like a week ago](https://wiki.libsdl.org/SDL3/SDL_PixelFormatDetails)

oddbookworm (changes requested):

MyreMylar (approved):
LGTM 👍  Though, as the comments point out, a bunch of this will need to be changed again in the near future to manage the renaming and internal changes for SDL 3.

Still I think it'll be easier to change this code chunk once it is fewer lines.

PR #2980, opened 2024-07-07
Ensure `.premul_alpha_ip()` works with zero sized Surfaces

itzpr3d4t0r:
Before this change zero sized surfaces would return `None` instead of self. Now it correctly returns self. I also added a test to check for this.


oddbookworm (approved):
LGTM!

ankith26 (approved):
LGTM, I wonder how we reviewers missed it in the original PR :sweat_smile: 

But no issues, this function hasn't seen a release yet. Thanks for the fix.

PR #2981, opened 2024-07-08
**DO NOT MERGE** Testing the removal of the cython shim for portmidi

oddbookworm:
https://github.com/matusvalo could not reproduce the cython error we were seeing, so I'm chucking this into review to get the CI going and hopefully get some build errors


oddbookworm:
Seems like it's only a problem if you use gcc >= 14...

PR #2985, opened 2024-07-11
Add 9-patch to examples

damusss:
After a short discussion with @Starbuck5 9-patch scaling is a recipe like process that doesn't involve algorithmic pixel manipulation so an example is probably good enough for it.

This is a good place to come every time you need your own 9-patch function and good to teach for GUI purposes. It's also existing in every other game engine so at least an example is needed.

I made the 9-patch code behave like a transform function would, adding comments and type hinting.


damusss:
Attaching a preview of the example for ease of review
image

examples/ninepatch.py
```diff
@@ -0,0 +1,149 @@
...
+            (ret_w - c, c),
+        )
+
+    if src_x_side > 0 and src_y_side > 0:
+        ret_surface.blit(  # central area
+            scale_func(
+                subsurface(c, c, src_x_side, src_y_side), (ret_x_side, ret_y_side)
+            ),
+            (c, c),
+        )
```
bilhox, line 99: Can this part be simplified using ``pygame.Vector2`` operations ?

bilhox (changes requested):
Overall the example is really cool. I would apprecate since you mentionned it works for any custom surface to actually use a custom surface, like an image with the necessary properties to make it work.

Anyways, good job :)

examples/ninepatch.py
damusss: No, I don't think so. It's the most compact and understandable I could. I can't even see where to insert vector2s, all I'm doing is subsurfacing, scaling and blitting.

damusss:
With my las commit I
- used pygame.Vector2.elementwise
- changed the while True
- added a real image on top of the rect

examples/ninepatch.py
```diff
@@ -0,0 +1,164 @@
...
+
+    if corner_size < 0 or corner_size > min(ret_w, ret_h):
+        raise ValueError(
+            "Corner size must be nonnegative and not greater than the smaller between width and height"
+        )
+    if corner_size == 0:  # default to normal scaling
+        return scale_func(surface, size)
+
+    src_x_side, ret_x_side = src_w - c * 2, ret_w - c * 2
+    src_y_side, ret_y_side = src_h - c * 2, ret_h - c * 2
```
bilhox, line 62:
```suggestion
    dest_size = size  # non-sequence argument catched by python
    src_size = surface.size
    c = corner_size
    scale_func = pygame.transform.smoothscale if smooth else pygame.transform.scale
    subsurface = surface.subsurface  # alias for performance/clarity

    if not 0 < corner_size < min(dest_size):
        raise ValueError(
            "Corner size must be positive and less than the smallest between destination width and destination height"
        )
    if corner_size == 0:  # default to normal scaling
        return scale_func(surface, size)
        
    ret_x_side, ret_y_side = src_size - pygame.Vector2(c) * 2
    src_x_side, src_y_side = ret_size - pygame.Vector2(c) * 2    
```
For these 2 lines : 
```py
    ret_x_side, ret_y_side = src_size - pygame.Vector2(c) * 2
    src_x_side, src_y_side = ret_size - pygame.Vector2(c) * 2  
```
It might be not beautiful, but this is because you are not currently supporting custom x and y corners. I would like to recommend this additional feature.

For beginners mental health, you should comment what the variable you created are used for.

examples/ninepatch.py
```diff
@@ -0,0 +1,164 @@
...
+    normal_scale = pygame.transform.scale(original_surface, SCALE_SIZE)
+    normal_center = SCREEN_SIZE.elementwise() / (2, 4)
+    normal_rect = normal_scale.get_rect(center=normal_center)
+    example_normal_scale = pygame.transform.scale(example_image, INNER_SCALE_SIZE)
+    example_normal_rect = example_normal_scale.get_rect(center=normal_center)
+
+    ninepatch = ninepatch_scale(
+        original_surface, SCALE_SIZE, CORNER_SIZE, True, False, None
+    )
+    ninepatch_center = SCREEN_SIZE.elementwise() / (2, (4 / 3))
```
bilhox, line 130: What about turning ``(2, (4, 3))`` into ``pygame.Vector2(2, 4 / 3)`` ?

bilhox (changes requested):
LGTM, I left some requested changes.

examples/ninepatch.py
damusss: bro wants me to suffer even more 😭 I'm about to take back the idea of adding this in the examples 😭

examples/ninepatch.py
damusss: OK

examples/ninepatch.py
bilhox: After a small discussion on discord, I agree it's not necessary.

damusss: I will improve what you said to improve in this review comment tho

MyreMylar (approved):
OK, LGTM 👍

bilhox (approved):
LGTM!

PR #2987, opened 2024-07-12
Support iterable in `display.update`

### LINKED ISSUE #2986: About Pygame's "Sequence" object and display.update
**Sometimes I want to use "display.update" to update my screen.  I would use dict.values() or chain() to improve performance , but it only receives an object which has "__len__" and "__getitem__".  Through the test,  I know some behavior the function will do. 
  For example,  I use this py-code.**

```
from typing import Generator, Generic, Iterable, Iterator, Protocol, Sequence, SupportsIndex,TypeVar, ValuesView
_T = TypeVar("_T",covariant=True)
class CanBeChained(Protocol[_T]):
    def __iter__(self)->Iterator[_T]:...
    def __len__(self) -> int: ...
class H_Chain(Generic[_T]):
    content:Sequence[CanBeChained[_T]]
    where:int=-1
    _len:int
    _generator:Generator[_T,None,None]
    def __init__(self,*canBeChained:CanBeChained[_T]) -> None:
        self.content=canBeChained
        self._len=sum(len(i) for i in canBeChained)
        self._generator=(elem for iterable in canBeChained for elem in iterable)
    def __getitem__(self,__i:SupportsIndex):
        print("GetItem",__i)
        return next(self._generator)
    def __len__(self):
        return self._len
    def __getattribute__(self, name: str):
        print("getattribute",name)
        return super().__getattribute__(name)
def test():
    display.update
    init();size=(300,300)
    sur=display.set_mode(size,RESIZABLE)
    l=H_Chain([None])#Pay attention here
    display.update(l)
if __name__=="__main__":
    test()

```
**It will print something like this:**
```
getattribute __class__
getattribute __class__
getattribute _len
getattribute rect
getattribute _len
GetItem 0
getattribute _generator
```
**I can't understand why you can't make it generator compatible and why the "__class__" and "__len__" attributes were gotten twice, while there is one "__getitem__" between two "__len__" sometimes, which causes an error.**
```
Fatal Python error: pygame_parachute: (pygame parachute) Segmentation Fault

Current thread 0x0000346c (most recent call first):
  File "C:\Users\HDS\Desktop\py\Task\temp.py", line 31 in test
  File "C:\Users\HDS\Desktop\py\Task\temp.py", line 33 in <module>

Extension modules: win32api, _win32sysloader, pygame.base, pygame.constants, pygame.rect, pygame.rwobject, pygame.surflock, pygame.bufferproxy, pygame.math, pygame.surface, pygame.window, pygame.display, pygame.draw, pygame.joystick, pygame.event, pygame.imageext, pygame.image, pygame.key, pygame.mouse, pygame.time, pygame.mask, pygame.pixelcopy, pygame.transform, pygame.font, pygame.mixer_music, pygame.mixer, pygame.scrap, numpy.core._multiarray_umath, numpy.core._multiarray_tests, numpy.linalg._umath_linalg, numpy.fft._pocketfft_internal, numpy.random._common, numpy.random.bit_generator, numpy.random._bounded_integers, numpy.random._mt19937, numpy.random.mtrand, numpy.random._philox, numpy.random._pcg64, numpy.random._sfc64, numpy.random._generator, pygame.system, pygame._freetype (total: 42)
```
**So, how to deal with this situation? I have only one botched solution.**
```
class H_Chain(Generic[_T]):
    content:Sequence[CanBeChained[_T]]
    where:int=-1
    _len:int
    _generator:Generator[_T,None,None]
    def __init__(self,*canBeChained:CanBeChained[_T]) -> None:
        self.content=canBeChained
        self._len=sum(len(i) for i in canBeChained)
    def __getitem__(self,__i:SupportsIndex):
        if __i==0:self._generator=(elem for iterable in self.content for elem in iterable)
        return next(self._generator)
    def __len__(self):return self._len
```
**That's not a method I'm happy with.(This is my first time filing an issue, and there will be many mistakes. Please forgive me.)
Looking forward to your early reply!**

ankith26:
fixes #2986 by adding `Iterable` support in `display.update`

Before someone points it out, yes it could potentially slightly lower performance, though I have done no benchmarking. There is scope for optimizing this function with our usual tricks though, i.e fastcall and list/tuple "fast sequence" special path, though I will leave that for future PRs.

Also this PR takes the opportunity to improve the tests for this function.

### DIFF
--- a/docs/reST/ref/display.rst
+++ b/docs/reST/ref/display.rst
@@ -273,7 +273,7 @@ required).
 
    | :sl:`Update all, or a portion, of the display. For non-OpenGL displays.`
    | :sg:`update(rectangle=None, /) -> None`
-   | :sg:`update(rectangle_list, /) -> None`
+   | :sg:`update(rectangle_iterable, /) -> None`
 
    For non OpenGL display Surfaces, this function is very similar to
    ``pygame.display.flip()`` with an optional parameter that allows only
@@ -285,8 +285,8 @@ required).
              updated. Whereas ``display.update()`` means the whole window is
              updated.
 
-   You can pass the function a single rectangle, or a sequence of rectangles.
-   Generally you do not want to pass a sequence of rectangles as there is a
+   You can pass the function a single rectangle, or an iterable of rectangles.
+   Generally you do not want to pass an iterable of rectangles as there is a
    performance cost per rectangle passed to the function. On modern hardware,
    after a very small number of rectangles passed in, the per-rectangle cost
    will exceed the saving of updating less pixels. In most applications it is
@@ -294,11 +294,13 @@ required).
    means  you do not need to keep track of a list of rectangles for each call
    to update.
 
-   If passing a sequence of rectangles it is safe to include None
+   If passing an iterable of rectangles it is safe to include None
    values in the list, which will be skipped.
 
    This call cannot be used on ``pygame.OPENGL`` displays and will generate an
    exception.
+
+   .. versionchanged:: 2.5.1 Added support for passing an iterable, previously only sequence was allowed
 
    .. ## pygame.display.update ##
 
--- a/src_c/display.c
+++ b/src_c/display.c
@@ -1694,31 +1694,48 @@ pg_update(PyObject *self, PyObject *arg)
             SDL_UpdateWindowSurfaceRects(win, &sdlr, 1);
     }
     else {
-        PyObject *seq;
-        PyObject *r;
-        Py_ssize_t loop, num;
+        PyObject *iterable, *single_arg, *r;
+        Py_ssize_t num;
         int count;
         SDL_Rect *rects;
         if (PyTuple_Size(arg) != 1)
             return RAISE(
                 PyExc_ValueError,
-                "update requires a rectstyle or sequence of rectstyles");
-        seq = PyTuple_GET_ITEM(arg, 0);
-        if (!seq || !PySequence_Check(seq))
+                "update requires a rectstyle or an iterable of rectstyles");
+
+        single_arg = PyTuple_GET_ITEM(arg, 0);
+        num = PyObject_Size(single_arg);
+        if (num == -1) {
+            /* Either __len__ errored, or object doesn't have __len__.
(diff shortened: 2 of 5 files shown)


docs/reST/ref/display.rst
```diff
@@ -273,7 +273,7 @@ required).
 
    | :sl:`Update all, or a portion, of the display. For non-OpenGL displays.`
    | :sg:`update(rectangle=None, /) -> None`
-   | :sg:`update(rectangle_list, /) -> None`
+   | :sg:`update(rectangle_iterable, /) -> None`
```
oddbookworm, line 276: This is what the lint CI is mad about. This would generate a change in the display doc header when you do a full generation on the docs

src_c/display.c
```diff
@@ -1694,48 +1694,78 @@ pg_update(PyObject *self, PyObject *arg)
...
+                "update requires a rectstyle or an iterable of rectstyles");
 
-        num = PySequence_Length(seq);
         rects = PyMem_New(SDL_Rect, num);
-        if (!rects)
+        if (!rects) {
+            Py_DECREF(iterable);
             return NULL;
+        }
         count = 0;
```
oddbookworm, line 1726: nitpick, but why not make this an `unsigned int` and initialize it to `0U`? Since it can never be negative, might as well give ourselves as much flex as possible right?

oddbookworm (approved):
I think this is a good change. Given the rare use I've seen of this, I don't expect too many people to take advantage though lol. Just got a couple small comments

src_c/display.c
ankith26: The reason it's an `int` is because it gets passed to `SDL_UpdateWindowSurfaceRects` and this function expects an `int`

docs/reST/ref/display.rst
ankith26: fixed

damusss (approved):
I tested locally, and the new code supports both sequences and iterables while the old one only sequences. I also did some dirty benchmark, and the performance is the same.

LGTM :)

ankith26:
Rebased to fix a merge conflict

PR #2988, opened 2024-07-12
Add `display.get_always_on_top()` and `display.set_always_on_top(bool)`

u8slvn:
Hello,

This is my first PR, I hope I did everything correctly.

### Aims

Add `get_always_on_top` and `set_always_on_top` within the `display` module to allow the user to use `SDL_SetWindowAlwaysOnTop` on the default window.

### Why

All started when I wanted to make the main window of a small game appears always above all the other windows.
After some research, I found a working solution using the lines of code above:

```python
import pygame

pyagme.init()
pygame.display.set_mode((800, 600))

window = pygame.Window.from_display_module()
window.always_on_top = True
```

The feature already exists in the `Window` class, but isn't available in the `display` module. Unfortunately, the `from_display_module` method is flagged as deprecated and I didn't find any other way to access the SDL window object initialized by `pygame.display`.

The only thing that bothers me is that `SDL_SetWindowAlwaysOnTop` is available since SDL 2.0.16. I've seen that the current version of Pygame-ce depends on SDL 2.0.10+, so it makes sens to have this in the `Window` class since the `window` module is experimentalI but I wonder if it's a good idea to add this feature into the `display` module. Anyway, I learned a lot about how Pygame works and had a lot of fun while working on this PR. I hope it could be useful!

### Example

It opens a window with a red circle in the middle. When you click on the window, it will change the color of the circle to green and set the window always on top. If you click again, it will change the color back to red and set the window not always on top.

```python
import pygame

RED = (255, 0, 0)
GREEN = (0, 255, 0)

pygame.init()
screen = pygame.display.set_mode([500, 500])

running = True
circle_color = RED

while running:
        if event.type == pygame.QUIT:
            running = False
        if event.type == pygame.MOUSEBUTTONUP:
            is_always_on_top = pygame.display.get_always_on_top()
            pygame.display.set_always_on_top(not is_always_on_top)
            circle_color = [RED, GREEN][not is_always_on_top]

    screen.fill((255, 255, 255))
    pygame.draw.circle(screen, circle_color, (250, 250), 75)
    pygame.display.flip()

pygame.quit()
```


ankith26:
> Unfortunately, the `from_display_module` method is flagged as deprecated and I didn't find any other way to access the SDL window object initialized by `pygame.display`

This is done because we want to nudge people towards using the `Window` API completely.

For example, the "older" code given below
```py
import pygame

pygame.init()

window_surface = pygame.display.set_mode((800, 600))
window = pygame.Window.from_display_module()
window.always_on_top = True

while True:
    # other game loop code here
    pygame.display.flip()  # or pygame.display.update()
```
Can be rewritten as 
```py
import pygame

pygame.init()

window = pygame.Window((800, 600))
window.always_on_top = True
window_surface = window.get_surface()

while True:
    # other game loop code here
    window.flip()
```

u8slvn:
> This is done because we want to nudge people towards using the `Window` API completely.

Oh I didn't know. So maybe it's not a good idea to expose this feature in the display module because it could motivate people later to use the window module...

> Can be rewritten as
> 
> ```python
> import pygame
> 
> pygame.init()
> 
> window = pygame.Window((800, 600))
> window.always_on_top = True
> window_surface = window.get_surface()
> 
> while True:
>     # other game loop code here
>     window.flip()
> ```

Thank you! Makes much more sens.

oddbookworm:
@u8slvn so, you see no reason to keep this pull open?

u8slvn:
> @u8slvn so, you see no reason to keep this pull open?

The only thing it brings is to avoid the use of `from_display_module`. But indeed, if the whole display module is doomed to be deprecated in the long run, I close the PR.

PR #2992, opened 2024-07-14
Fix break when the OS resizes our window in `set_mode`

### LINKED ISSUE #2991: Linux is broke, no idea cause
[Too long to explain in the time I have, so here's a discord link](https://discord.com/channels/772505616680878080/772847140992712734/1262100174680100895)

[SDL issue I opened to see if they know what's going on](https://github.com/libsdl-org/SDL/issues/10270)

Fix I found is to put an internal flip after we pump events in `set_mode` to force an update
image

This updates the screen to be whatever minimum size is being pushed on us, but now tests are failing because we can't guarantee that a small screen size is honored

oddbookworm:
Hopefully fixes #2991. I suspect that there's a lot I missed, so @robertpfeiffer @ankith26 @Starbuck5 please take a long look at these changes.

I haven't found a reliable way we can predict what the size the OS will force small windows to is, so I resorted to seeing what SDL says it is. Maybe I could just manually go through the event queue instead of pumping and grabbing the size from there?

### DIFF
--- a/src_c/display.c
+++ b/src_c/display.c
@@ -827,7 +827,7 @@ pg_set_mode(PyObject *self, PyObject *arg, PyObject *kwds)
     SDL_Surface *newownedsurf = NULL;
     int depth = 0;
     int flags = 0;
-    int w, h;
+    int w, h, w_actual, h_actual;
     PyObject *size = NULL;
     int vsync = SDL_FALSE;
     intptr_t hwnd = 0;
@@ -1096,6 +1096,8 @@ pg_set_mode(PyObject *self, PyObject *arg, PyObject *kwds)
                 }
                 else {
                     win = SDL_CreateWindow(title, x, y, w_1, h_1, sdl_flags);
+                    w_actual = w_1;
+                    h_actual = h_1;
                 }
                 if (!win)
                     return RAISE(pgExc_SDLError, SDL_GetError());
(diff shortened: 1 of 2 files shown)


src_c/display.c
```diff
@@ -1340,8 +1345,38 @@ pg_set_mode(PyObject *self, PyObject *arg, PyObject *kwds)
...
 
-    /*probably won't do much, but can't hurt, and might help*/
+    /*
+     * Can potentially yield a window resize event that forcibly changes
+     * the window size. This would invalidate the current surface we store,
+     * which can cause us to segfault in the event that we reference that
+     * surface later. So we need to flip, which forces us to update that
+     * surface as needed.
+     */
     SDL_PumpEvents();
```
ankith26, line 1355: This call was rather arbitrary anyways, removing it probably fixes the issue instead of needing the special-cased flip internal call?

By fixes the issue, I mean defers the handling to the next point of time that pump events is called.

ankith26: though it's just a comment, your approach is probably better as it raises a warning and lets the user know right at the function call.

src_c/display.c
```diff
@@ -1340,8 +1345,38 @@ pg_set_mode(PyObject *self, PyObject *arg, PyObject *kwds)
...
+    pg_flip_internal(state);
```
ankith26, line 1356: This is the second `pg_flip_internal` call in this function. The first one is called just after setting the screen to black. I reckon, that part must be removed and handled here instead

src_c/display.c
oddbookworm: Yeah, I'd rather the user know immediately, rather than some nebulous point in the future when the event queue is finally pumped

src_c/display.c
oddbookworm: removed the other one

src_c/display.c
```diff
@@ -1340,8 +1344,38 @@ pg_set_mode(PyObject *self, PyObject *arg, PyObject *kwds)
...
+     * such a way that it can, at least our internal stuff, can
+     * be respected enough that we don't need to issue a warning
+     */
+    if (!state->using_gl && ((flags & (PGS_SCALED | PGS_FULLSCREEN)) == 0) &&
+        !vsync) {
+        SDL_Surface *sdlSurf = SDL_GetWindowSurface(win);
+        if (((sdlSurf->w != w_actual) || (sdlSurf->h != h_actual)) &&
+            ((sdlSurf->flags & SDL_WINDOW_FULLSCREEN_DESKTOP) != 0)) {
+            char buffer[150];
+            char *formatString =
```
ankith26, line 1369: Just a minor nitpick, could you make use of snake case naming instead here?

src_c/display.c
```diff
@@ -1340,8 +1344,38 @@ pg_set_mode(PyObject *self, PyObject *arg, PyObject *kwds)
...
+
+    /*
+     * Tell user if their requested screen size is ignored
+     * OpenGL, SCALED, and FULLSCREEN mess with the screensize in
+     * such a way that it can, at least our internal stuff, can
+     * be respected enough that we don't need to issue a warning
+     */
+    if (!state->using_gl && ((flags & (PGS_SCALED | PGS_FULLSCREEN)) == 0) &&
+        !vsync) {
+        SDL_Surface *sdlSurf = SDL_GetWindowSurface(win);
```
ankith26, line 1365: Now that the flip is called, we probably don't need this extra call. We just need to use `surface->surf` instead

src_c/display.c
oddbookworm: done

src_c/display.c
oddbookworm: done

ankith26 (approved):

MyreMylar (approved):
OK, LGTM 👍

PR #2994, opened 2024-07-15
3.13 source support (removes deprecated functions)

Starbuck5:
This gets 3.13 building on my system with no warnings, which is an important prereq to getting 3.13 on CI.

It seems like 3.13 on CI should wait for https://github.com/pygame-community/pygame-ce/pull/2945.

Weakref changes: https://docs.python.org/3.13/c-api/weakref.html#c.PyWeakref_GetObject . You can read about in the 3.13 whatsnew.

These changes are really annoying for us, the old behavior was fine for our uses, we could handle the Nones coming out, now we have to deal with 3 separate states (there was no error report state before). The old code was ancient but worked fine. The old function is in the stable ABI, and the new function is new in the release! They are not giving us much of a window here. 

So, I had to bring in the python c api compat header to deal with this, the patch would be even more complex if I was trying to make this work on python versions without the new function too. I figure it will be nice for us to have this header around for other features, so we can write our C API code like we only target the latest version.

### DIFF
--- a/src_c/_pygame.h
+++ b/src_c/_pygame.h
@@ -33,6 +33,7 @@
 */
 #define PY_SSIZE_T_CLEAN
 #include <Python.h>
+#include "include/pythoncapi_compat.h"
 
 /* Ensure PyPy-specific code is not in use when running on GraalPython (PR
  * #2580) */
@@ -457,19 +458,6 @@ typedef enum {
 #define PG_CHECK_THREADS() \
     (RAISE(PyExc_NotImplementedError, "Python built without thread support"))
 #endif /* ~WITH_THREAD */
-
-#define PyType_Init(x) (((x).ob_type) = &PyType_Type)
-
-/* Python macro for comparing to Py_None
- * Py_IsNone is naturally supported by
- * Python 3.10 or higher
- * so this macro can be removed after the minimum
- * supported
- * Python version reaches 3.10
- */
-#ifndef Py_IsNone
-#define Py_IsNone(x) (x == Py_None)
-#endif
 
 /* Update this function if new sequences are added to the fast sequence
  * type. */
--- a/src_c/surface.c
+++ b/src_c/surface.c
@@ -993,6 +993,7 @@ surf_get_locks(PyObject *self, PyObject *_null)
 {
     pgSurfaceObject *surf = (pgSurfaceObject *)self;
     Py_ssize_t len, i = 0;
+    int weakref_getref_result;
     PyObject *tuple, *tmp;
     SURF_INIT_CHECK(pgSurface_AsSurface(self))
     if (!surf->locklist)
@@ -1004,8 +1005,16 @@ surf_get_locks(PyObject *self, PyObject *_null)
         return NULL;
 
     for (i = 0; i < len; i++) {
-        tmp = PyWeakref_GetObject(PyList_GetItem(surf->locklist, i));
-        Py_INCREF(tmp);
+        weakref_getref_result =
+            PyWeakref_GetRef(PyList_GetItem(surf->locklist, i), &tmp);
+        if (weakref_getref_result == -1) {  // exception already set
+            Py_DECREF(tuple);
+            return NULL;
+        }
+        if (weakref_getref_result == 0) {
+            tmp = Py_None;
+            Py_INCREF(tmp);
+        }
         PyTuple_SetItem(tuple, i, tmp);
     }
     return tuple;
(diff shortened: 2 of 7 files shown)


src_c/_pygame.h
```diff
@@ -460,17 +461,6 @@ typedef enum {
 
 #define PyType_Init(x) (((x).ob_type) = &PyType_Type)
```
ankith26, line 462: I looked into what this macro is out of curiosity (due to it having the `Py` prefix), and I believe this can be removed as it isn't used anywhere

src_c/surface.c
```diff
@@ -1004,8 +1005,16 @@ surf_get_locks(PyObject *self, PyObject *_null)
...
-        Py_INCREF(tmp);
+        weakref_getref_result =
+            PyWeakref_GetRef(PyList_GetItem(surf->locklist, i), &tmp);
+        if (weakref_getref_result == -1) {  // exception already set
+            Py_DECREF(tuple);
+            return NULL;
+        }
+        if (weakref_getref_result == 0) {
+            tmp = Py_None;
+            Py_IncRef(tmp);
```
ankith26, line 1016: isn't this supposed to be `Py_INCREF`? even though the docs say it's equivalent, I prefer the macro variant for consistency.

src_c/surflock.c
```diff
@@ -115,19 +115,27 @@ pgSurface_UnlockBy(pgSurfaceObject *surfobj, PyObject *lockobj)
...
-                }
-                else {
-                    found = 1;
+
+            weakref_getref_result = PyWeakref_GetRef(item, &ref);
+            if (weakref_getref_result == -1) {
+                noerror = 0;
+            }
+            if (weakref_getref_result == 1) {
+                Py_DECREF(ref);
```
ankith26, line 131: Why is this decref called before `ref` is being checked?

ankith26:
Vendoring that header in is definitely a step in the right direction

Starbuck5:
> Vendoring that header in is definitely a step in the right direction

I'm just annoyed the deprecation cycle is so sharp that I feel forced to use it.

src_c/surface.c
Starbuck5: Good spot, just typed the wrong thing.

src_c/_pygame.h
Starbuck5: Removed it.

src_c/surflock.c
Starbuck5: I did it because there are multiple paths below it would need to be decref-ed in. It shouldn't make a difference, this is converting the strong ref back into a borrowed ref.

However just to be more typical with the API usage I split it so it gets decref-ed after use. (new commit).

src_c/surflock.c
```diff
@@ -115,36 +115,52 @@ pgSurface_UnlockBy(pgSurfaceObject *surfobj, PyObject *lockobj)
...
         while (--len >= 0) {
             item = PyList_GetItem(surf->locklist, len);
-            ref = PyWeakref_GetObject(item);
-            if (ref == Py_None) {
+
+            weakref_getref_result = PyWeakref_GetRef(item, &ref);
+            if (weakref_getref_result == -1) {
+                noerror = 0;
+            }
+            if (ref == NULL) {
```
ankith26, line 153:
```suggestion
            if (weakref_getref_result == 0) {
```
`ref == NULL` would catch the error case too, right?

Starbuck5: Well the docs don't say it will write NULL on error, that detail is undefined. But it could grab a null from a previous iteration of the loop, assuming nothing is written. Adjusted the logic.

ankith26 (approved):
LGTM, thanks! :+1:

test/font_test.py
```diff
@@ -1,5 +1,4 @@
 # -*- coding: utf-8 -*-
-from re import T
```
oddbookworm, line 2: such a weird, random thing lol

oddbookworm (approved):
LGTM! I don't like having to vendor in that header, but I agree it's our best option for now.  I do have concerns about maintaining it because we have to re-vendor it every time python gets a minor revision update (and maybe in rare cases where a patch update might break something?). On top of us having to remember to re-vendor it, there's the always-present concern of us supported python revs longer than python itself, so we may still end up having to do our own shimming from time to time if they remove something from the compat header that we still need to support. But, yeah, for now I think this is the best we're gonna get

test/font_test.py
Starbuck5: I know right.

oddbookworm: I suspect someone was typing something and their linter got too enthusiastic and auto-imported what it thought they wanted

Starbuck5:
> I do have concerns about maintaining it because we have to re-vendor it every time python gets a minor revision update.

Well to get new functions from new versions we would need to update this, but new versions won't break old functions.

> On top of us having to remember to re-vendor it, there's the always-present concern of us supported python revs longer than python itself, so we may still end up having to do our own shimming from time to time if they remove something from the compat header that we still need to support.

Looking in the header it seems to still support Python 2, so I'm concerned about that.

PR #2995, opened 2024-07-15
Use python 3.8 isms in the type stubs

ankith26:
`Protocol`, `SupportsIndex`, `Literal` and `TypedDict` were all added in Python 3.8

We can make use of these directly with a `typing` import as our minimum supported python version is 3.8


oddbookworm (approved):
LGTM! Some of these cleanups look really satisfying

damusss (approved):
LGTM!

PR #2990, opened 2024-07-13
Coverage CI created, hopefully it works first time lol

oddbookworm:
I wanted to see what would happen if we generate coverage in our CI, so let's see what happens lol

PR #2996, opened 2024-07-15
Fix editable install test fixtures logic

ankith26:
Simple PR to fix a small issue in our test runner script. With editable installs, the older path resolution logic didn't give correct paths anymore, resulting in a fail (issue discovered while working on #2990)

### DIFF
--- a/test/test_utils/__init__.py
+++ b/test/test_utils/__init__.py
@@ -3,8 +3,6 @@ import pygame
 import sys
 import tempfile
 
-is_pygame_pkg = __name__.startswith("pygame.tests.")
-
 ###############################################################################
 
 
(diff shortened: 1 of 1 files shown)


oddbookworm (commented):
It would be nice to see a CI run that demonstrates that these changes do, in fact, solve the problem you've set out to solve here. Since none of our current CI uses editable installs to the best of my knowledge, maybe that would come with a modification to an existing build or a new build? Or maybe passing CI in #2990 after pulling in these changes?

ankith26:
My intention is to merge this PR and then rebase #2990 to get this fix

oddbookworm (approved):
LGTM

zoldalma999 (approved):
Tested locally, works. Thanks for the PR!

PR #2997, opened 2024-07-15
Fix some surface refcounting issues in display

ankith26:
In `get_surface`, a strong reference was being incref-ed before being returned (double incref issue)

In `set_mode` there was a decref followed by an incref, which is technically wrong (I suppose in practice it doesn't create an issue, but I fixed it anyways)

### DIFF
--- a/src_c/display.c
+++ b/src_c/display.c
@@ -588,7 +588,6 @@ pg_get_surface(PyObject *self, PyObject *_null)
             if (!new_surface)
                 return NULL;
             pg_SetDefaultWindowSurface(new_surface);
-            Py_INCREF((PyObject *)new_surface);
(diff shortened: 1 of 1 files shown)


damusss (approved):
Alright now that I understand it the bugfix makes sense, so it LGTM :)

MyreMylar:
> In `set_mode` there was a decref followed by an incref, which is technically wrong (I suppose in practice it doesn't create an issue, but I fixed it anyways)

What about the return here:

https://github.com/pygame-community/pygame-ce/blob/f5f850ac07a6aab001fb2d687879d4ab434e77cd/src_c/display.c#L1333-L1339

Won't removing the decref cause a leak if it returns  NULL?

PR #2998, opened 2024-07-16
Fix `pygame.sprite.Group.has()` behavior for no sprites given

### LINKED ISSUE #2993: `pygame.sprite.Group.has()` method behavior for no sprites given
<details>
<summary>Environment:</summary>

```
Processor:              Intel64 Family 6 Model 167 Stepping 1, GenuineIntel

Python:                 CPython 3.12.4 (tags/v3.12.4:8e8a4ba, Jun  6 2024, 19:30:16) [MSC v.1940 64 bit (AMD64)]
pygame version:         2.5.0
SDL versions:           Linked: 2.30.3  Compiled: 2.30.3

```
</details>

**Current behavior:**

According to the documentation, `pygame.sprite.Group.has()` (and `__contains__`) will return True if *all* given sprites are contained in the group, False otherwise.
However, for the edge case of no sprites given (or equivalently, empty iterable), this method incorrectly returns `False`. This is probably what was intended, but is wrong.

In `has()` method:
https://github.com/pygame-community/pygame-ce/blob/cdc27561f973bd062172c2e4b765e91090f6a8d9/src_py/sprite.py#L521-L522

**Expected behavior:**

The `has()` method should return `True` if no sprites (no arguments, empty group, etc.) are given. This is similar to the built-in `all()` function.
The technical term is "vacuous truth". If you need more explanation, ask me.

**Steps to reproduce:**
```python
import pygame
group = pygame.sprite.Group()
print('has():', group.has(), group.has([]), group.has(group))
```

aatle:
Make `Group.has()` return `True` instead of `False` when no sprites (empty list, no arguments, etc.) are given.

Closes #2993

This is *potentially* a breaking change. Might not get approved. See issue for discussion.

## Summary by CodeRabbit

- **Behavior Changes**
  - Calling the group membership check without sprites now returns `True`.
  - Empty list inputs are also treated as a successful membership check.

- **Tests**
  - Updated coverage to reflect the new results for empty argument and list cases.


oddbookworm:
Can you motivate this change with an example of when this change would actually matter? Offhand, I can't come up with a valid use-case where it should matter what this returns if no sprites are passed

aatle:
Sure.
I'll post it in the issue.

ankith26:
I agree that this change is logical, but also, as pointed out before, it could be a backwards incompatible change that could break someones usage.

MyreMylar (approved):
I agree this makes sense, but I think this should wait for compatibility breaking release before being merged as the current behaviour has probably been around since the sprite group was created.

ankith26:
Thanks for the PR!

Essentially this PR has my approval as well, I'm just drafting it for now because it cannot be merged right now for reasons Myre stated.

aatle:
This PR may now finally be merged.

PR #3000, opened 2024-07-16
Simplify c-side of events sub-module.

gresm:
WIP


damusss:
First of all, I'd like to say that I love where this is going :)
I have a few modifications that I (personally) would prefer:
- is there a reason why `_NAMES_MAPPING` has raw integers instead of `pygame.X` event constants? I tried doing that, and it worked perfectly fine. In `cursors.py`, the dictionary also has pygame constants
- I think your stub situation can be improved: IMO there should not be `_event.pyi`, rather, there should be only `event.pyi` containing both from `_event.c` and from `event.py` (there is a stub file for `cursors.py`, even tho it's made in python!). The stubs are fake code, we don't need to tell the user that event is actually split between C and py, what matters is what is accessible in it. Again I tried that myself, added `init`, `quit` and modified a signature and the `stubcheck.py` with mypy passes for me, that's why I'm suggesting it.

Tell me what you think about this

gresm:
@damusss Thanks for the comment - I'll update ``_NAMES_MAPPING`` shortly. As for ``_event.pyi``, I'm using it as a roadmap of what's left to do, but I'm not going to completely remove ``pygame._event``, just change some of the functionality, so probably this file will be included.

damusss:
> @damusss Thanks for the comment - I'll update ``_NAMES_MAPPING`` shortly. As for ``_event.pyi``, I'm using it as a roadmap of what's left to do, but I'm not going to completely remove ``pygame._event``, just change some of the functionality, so probably this file will be included.

Alright, I'll check the stub situation when you are closer to finish then. Nice work so far :>

gresm:
(Why the commits duplicated after I resolved merge conflict?)

damusss:
No actually why are part of the commits not yours?? what even happened? It looks like you are adding the deprecator decorator

gresm:
I merged commits from upstream using ``git pull origin main --rebase`` and likely this ``--rebase`` is the cause of inclusion of these commits... I'll try to discard this merge and re-attempt doing it.

gresm:
Done, even though they still appear in the conversation, I removed them from the commits tab (I guess that conversation side just logs every change and not exactly those that truly will be merged?).

gresm:
I think I'll split this PR in two parts:
1. Migrating stuff related to ``Event`` class. (nearly done, c-api documentation still needed to be updated)
2. Migrating the monstrosity of events filtering. (not touched yet)

PR #3002, opened 2024-07-16
Add pygame.typing module

damusss:
This PR would potentially implement https://github.com/pygame-community/pygame-ce/issues/1682

As the title says, this PR adds the `typing` module to pygame. I've made some libraries myself and I often wanted to annotate properly things as colors and rects, but the types they accept are a wide range, especially the rect, so it would be really nice to put `rect: pygame.typing.RectLike` in an argument for example. I see no reason not to have this.

Details about the PR:
- only the contents of `_common.pyi` have been ported. This can be easily expanded in the future.
- `_common.pyi` has been removed
- `src_py/typing.py` has been added, being similar to `_common.pyi`
- `gen_stubs.py` has the functionality of copying `typing.py` to `typing.pyi`, to make type checkers happy
- `__init__.py` was modified to correctly import the typing module
- Since the API will be public, a consistent pattern has been chosen with @ankith26 , which is to add "Like" to the types, to annotate that it's not the actual object but rather a set of types that can replace it (if one needs the object to access the attributes, `pygame.Rect` is preferred over `pygame.typing.RectLike`, while if the object is going to be passed to a pygame function that accepts `RectLike`, there's no point in forcing a `pygame.Rect`. You get the idea.)

Tests and docs have been added. Docs are currently temporary, but might not be when you reader read this message.


buildconfig/stubs/gen_stubs.py
```diff
@@ -151,3 +153,6 @@ def get_all(mod: Any):
     for element in get_all(pygame.locals):
         constant_type = getattr(pygame.locals, element).__class__.__name__
         f.write(f"{element}: {constant_type}\n")
+        
+# copy typing.py to typing.pyi for type checkers
+shutil.copyfile("src_py/typing.py", "buildconfig/stubs/pygame/typing.pyi")
```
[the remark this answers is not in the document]
damusss: Alright I updated that and I added typing to __init__ otherwise you could not actually use it :kekw:

src_py/typing.py
Matiiss, line 1: There should be a
```py
del sys, IO, Callable, Tuple, Union, TypeVar, Protocol, SupportsIndex
```
at the end of the file so that this module doesn't contain any of those names

Matiiss (changes requested):

Starbuck5:
So before we add a whole new module for this, we should think about some things.

How do other libraries do this?

Does this need to be a module or can type variables be scattered where they are relevant? Like pygame.rect.RectLike?

How can we keep this stable? This exposes more stuff that we need to keep stable, are we okay with that?

damusss:
> So before we add a whole new module for this, we should think about some things.
> 
> How do other libraries do this?
> 
> Does this need to be a module or can type variables be scattered where they are relevant? Like pygame.rect.RectLike?
> 
> How can we keep this stable? This exposes more stuff that we need to keep stable, are we okay with that?

2) you could realistically only scatter rect, color and maybe rgba, maybe path and file to rwobject, but those are all implemented in C and they would require some injection.

3) what non stable things are we exposing?

buildconfig/stubs/typing_sample_app.py
```diff
@@ -0,0 +1,92 @@
...
+        return "file.py"
+    
+def validator_PathLike(path: typing.PathLike) -> int:
+    return 0
+
+# must pass
+validator_PathLike("file.py")
+validator_PathLike(b"file.py")
+validator_PathLike(MyPath())
+
```
ankith26, line 42: I would like to see `pathlib` stuff tested here.

buildconfig/stubs/typing_sample_app.py
```diff
@@ -0,0 +1,92 @@
...
+def validator_SequenceLike(sequence: typing.SequenceLike) -> int:
+    return 0
+
+# must pass
+validator_SequenceLike(MySequence())
+validator_SequenceLike([0, 1, 2, 3])
+validator_SequenceLike((0, 1, 2, 3))
+validator_SequenceLike(pygame.Rect(-10, 10, 40, 40))
+validator_SequenceLike(pygame.Vector2())
+validator_SequenceLike("1234567890")
```
ankith26, line 28: I would also like to have type specific tests here. For example `SequenceLike[float]`

ankith26 (approved):
Overall, I really like where this PR is going! I am approving the code changes from my side. I left a few reviews for improving the test file. 

As this is a significant change, I would like for this PR to have 3 approving reviews.

buildconfig/stubs/stubcheck.py
```diff
@@ -11,9 +11,33 @@
...
+        print("ERROR: could not validate typing.py, make sure you have mypy installed")
+        return
+    
+    mypy_args = [*mypy_version_args, "typing_sample_app.py"]
+    cmd = " ".join(mypy_args)
+    print(f"Using mypy invocation: `{cmd}` (version: {version})")
+    prev_dir = os.getcwd()
+    try:
+        os.chdir(STUBS_BASE_DIR)
+        returncode = subprocess.run([*mypy_args]).returncode
```
ankith26, line 33: If this is anything other than 0, an error must be raised here/

src_py/typing.py
```diff
@@ -0,0 +1,54 @@
+import sys
```
ankith26, line 1: I would like a header docstring here, and also a comment explaining that this file and typing.pyi are meant to be kept duplicated

ankith26:
Addressing some of Starbuck's points (in favour of this PR) from my POV

> How do other libraries do this?
> Does this need to be a module or can type variables be scattered where they are relevant? Like pygame.rect.RectLike?

As Matt already pointed out, `numpy` already has a dedicated submodule for typing stuff. In the future, if needed, we can make aliases in scattered places, but I'd like all the implementation to be in one file.

> How can we keep this stable? This exposes more stuff that we need to keep stable, are we okay with that?

This PR already adds a test file, that should help with maintaining stability. OFC, the file could be expanded in future iterations.
If this makes approval more comfortable: we can always slap an "experimental" notice on the docs, and allow for more drastic changes here for a few releases.

buildconfig/stubs/pygame/typing.pyi
```diff
@@ -0,0 +1,59 @@
...
+# Used for functions that return an RGBA tuple
+RGBATuple = Tuple[int, int, int, int]
+ColorLike = Union[int, str, SequenceLike[int]]
+
+_CanBeRect = SequenceLike[Union[float, Coordinate]]
+
+
+class _HasRectAttribute(Protocol):
+    # An object that has a rect attribute that is either a rect, or a function
+    # that returns a rect confirms to the rect protocol
```
Matiiss, line 52: confirms -> conforms
lol

damusss: On my defense, this typo has been there since forever lol, I just copy pasted

Matiiss (approved):
LGTM, I sense this will be a great addition to `pygame-ce`

Maybe the type aliases could be hinted as `TypeAlias`es, but if type checkers don't complain then it's probably that important.

I suppose sometime in the future we could restructure this a bit so that there's no need to literally copy-paste code from `typing.py` to `typing.pyi`, but I suppose it should be fine for now.

Thanks! 🚀

bilhox (changes requested):
Overall LGTM, I left some requested changes for the documentation part.
I feel like for people reading the docs, a list of structure for each type in ``pygame.type`` is a lot better for the user. Even though it's abstract and can technically handles an infinite number of type as long as it respects the structure, making a list of the most common types used feels essential for the user.

docs/reST/ref/typing.rst
```diff
@@ -0,0 +1,85 @@
...
+        * ``(r, g, b, a)``
+        * ``[r, g, b, a]``
+        * ``"green"``
+        * ``0`` (mapped color)
+
+   .. data:: RectLike
+      
+      An object representing a rect such as a sequence of numbers or coordinates
+      or an object with a rect attribute or a method returning a rect. This types
+      are supported by every function accepting a rect as argument. i.e.:
```
bilhox, line 77:
```suggestion
      or an object with a rect attribute or a method returning a rect. This type
      is supported by every function accepting a rect as argument. i.e.:
```

bilhox (changes requested):
``pygame.gfxdraw`` does not handle string colors and mapped color, only sequences of 3 or 4 values are accepted (RGB and RGBA).

docs/reST/ref/typing.rst
damusss: why should it be singular? i listed multiple types in the sentence.

bilhox: Then it would be ``These`` and not ``This`` . I thought you were talking about ``Rectlike``.

damusss: ah that's just some silly grammar mistake of mine

damusss:
> ``pygame.gfxdraw`` does not handle string colors and mapped color, only sequences of 3 or 4 values are accepted (RGB and RGBA).

so should i say supported by "almost" all pygame functions or do you want me to edit gfxdraw.pyi? the typehints were there already and they were wrong before this PR.

docs/reST/ref/typing.rst
```diff
@@ -0,0 +1,85 @@
...
+
+:mod:`pygame.typing`
+====================
+
+.. module:: pygame.typing
+   :synopsis: pygame module providing common typehints
+
+| :sl:`pygame module providing common typehints`
+
+.. versionadded:: 2.5.1
```
ankith26, line 11: needs updating now

Matiiss (commented):
None of the `sprite.pyi` types seem to be exposed, is this intended? I find it would be greatly beneficial to expose those types as well. It would help with some testing I plan to do regarding those types and thus it would require access to concrete objects.

damusss:
> None of the `sprite.pyi` types seem to be exposed, is this intended? I find it would be greatly beneficial to expose those types as well. It would help with some testing I plan to do regarding those types and thus it would require access to concrete objects.

Do you mean types such as _SuportsSprite and _SupportsDirtySprite? It is kinda intended. This PR only implements the contents of common.pyi to keep the scope simple. extra typings can be added later with other PRs based on their need (at the start of the PR I added a lot more types but keeping it small is better). @ankith26 might have a say in the matter.

Matiiss:
> > None of the `sprite.pyi` types seem to be exposed, is this intended? I find it would be greatly beneficial to expose those types as well. It would help with some testing I plan to do regarding those types and thus it would require access to concrete objects.
> 
> Do you mean types such as _SuportsSprite and _SupportsDirtySprite? It is kinda intended. This PR only implements the contents of common.pyi to keep the scope simple. extra typings can be added later with other PRs based on their need (at the start of the PR I added a lot more types but keeping it small is better). @ankith26 might have a say in the matter.

Fair enough, it can be handled in a later PR then 👍

ankith26 (approved):
This PR still has my approval. This has been some nice work, thanks again for working on this! :sunglasses: 

I think we can just merge it in for now (as it is a pretty safe change that wouldn't break any code) and see how it goes. If needed we can also mark this as experimental before the next release.

PR #3003, opened 2024-07-16
Use `SDL_TTF_VERSION_ATLEAST`

zoldalma999:
Since #2464 our SDL_ttf minimum version is 2.0.15, so we can use `SDL_TTF_VERSION_ATLEAST`

### DIFF
--- a/src_c/font.c
+++ b/src_c/font.c
@@ -919,10 +919,7 @@ font_set_script(PyObject *self, PyObject *arg)
         return RAISE_FONT_QUIT_ERROR();
     }
 
-/*Sadly, SDL_TTF_VERSION_ATLEAST is new in SDL_ttf 2.0.15, still too
- * new to use */
-#if SDL_VERSIONNUM(SDL_TTF_MAJOR_VERSION, SDL_TTF_MINOR_VERSION, \
-                   SDL_TTF_PATCHLEVEL) >= SDL_VERSIONNUM(2, 20, 0)
+#if SDL_TTF_VERSION_ATLEAST(2, 20, 0)
     TTF_Font *font = PyFont_AsFont(self);
     Py_ssize_t size;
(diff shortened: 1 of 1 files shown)


src_c/font.c
```diff
@@ -989,8 +983,7 @@ font_set_direction(PyObject *self, PyObject *arg, PyObject *kwarg)
    writing this) This bug flips the top-to-bottom and bottom-to-top rendering.
    So, this is a compat patch for that behavior
  */
-#if SDL_VERSIONNUM(SDL_TTF_MAJOR_VERSION, SDL_TTF_MINOR_VERSION, \
-                   SDL_TTF_PATCHLEVEL) < SDL_VERSIONNUM(2, 22, 0)
+#if !SDL_TTF_VERSION_ATLEAST(2, 22, 0)
```
ankith26, line 986: Now that you made this PR, it got me thinking of whether this should be a compiled-version check or a linked-version check.

It isn't a big deal as probably over 99% of our users have matching compiled and linked versions, but it could be something to look into if you'd like.

ankith26 (approved):
Apart from a minor review comment this PR LGTM. Thanks for the PR :tada:

src_c/font.c
oddbookworm: I think for the purposes of compat with the original code, it should be kept as-is in this pull. If we want to change that extreme edge-case behavior, that should be a separate pull IMO (maybe one that addresses it in a bunch of places that I'm sure have a similar problem)

oddbookworm (approved):
LGTM! Leaving unmerged just in case others disagree with me about the compiled-vs-linked discussion

src_c/font.c
Starbuck5: I don't think it would dynamically link if it was lower.

PR #3006, opened 2024-07-18
Shim and deprecate pygame.gfxdraw using pygame.draw

damusss:
Functions that need to be ported to complete it:
- aaellipse
- aapolygon
- bezier

PR will be ready to review when the deprecation is 100% complete

[WARNING: the diff looks huge, but it's really just removing useless things! Don't be scared :(]

This could potentially fix this issues: https://github.com/pygame-community/pygame-ce/issues/1487, https://github.com/pygame-community/pygame-ce/issues/2450, https://github.com/pygame-community/pygame-ce/issues/2463 and start fixing https://github.com/pygame-community/pygame-ce/issues/172 .

`gfxdraw` is a dying module, `draw` is just so much better, and this creates a lot of duplicates. `gfxdraw` should become deprecated, and until the functions are removed (maybe) they should be implemented with the draw module in a python file. This is not possible 100% right now as there are functions with no equivalent, but this is the start of the shim. Everytime something is added, more shim is added until it's 100% python.

In this PR I renamed old `gfxdraw` to `_gfxdraw` (will be removed at the end of the deprecation), created `gfxdraw.py` and impllemented most of the functions (the stub is almost intact, the user shouldn't know about shims). I then went in `_gfxdraw.c` and removed all non necessary functions. Finally, I removed a LOT of code from the `src_c/SDL_gfx/` folder, everything that is not needed in the current state. I also removed a useless file which we won't need since gfxdraw is not going to grow.

[REVIEW WARNING: it seems like I added code to _gfxdraw.c and the SDL_gfx files but I did not. I only removed code!]

I am working to fix the tests.

NOTE: the signature in the `@deprecated` decorator for stubs and py files must be different otherwise the raised message will seem incomplete.


buildconfig/stubs/pygame/gfxdraw.pyi
Starbuck5, line 1: Why does adding a module shim change all these signatures?

A shim should be compatible.

Starbuck5:
> [Not Complete]

Marking draft.

bilhox:
Hello,

You did a good work ! I have extra things to say :

- If you're moving every functions that should be deprecated in python side, you should add the deprecated decorator.
- If I'm not wrong you kept ``textured_polygon`` in C side, but this can easily be done with other functions ! You draw an ellipse on a surface with alpha, and then you blit it on the image with a ``BLEND_RGBA`` flag.

Starbuck5:
My take on this is that SDL_gfx is a vendored library, so deleting parts of its source code isn't what we should be after. If that's true, then I don't see why a partial shim helps us much. It would need to be a full replacement to delete the gfxdraw source. Maybe there's some benefit to just removing the Python->C layer for the function we no longer need with a partial shim?

That being said, maybe I'm wrong. Deleting parts of their code does reduce the amount of code that needs to get compiled with SDL3. - Although, I have had barely any issues getting it to compile. Like all the patches it needed (last time I checked with SDL3) have already been added.

This is definitely a good exercise to see what draw needs. I just don't know if there's a good reason to merge a partial shim.

buildconfig/stubs/pygame/gfxdraw.pyi
damusss: as far as I know and as far as I could research, `, /` means the parameters are positional only. If the function is implemented in python, then by default parameters are both positional and keyword, so that signature is outdated. I could change it back, or mark the python signature positional only if it's possible.

damusss:
@Starbuck5 alright, maybe a partial shim isn't a good option. Draw is only missing bezier, aaellipse and aapolygon, so with a bit of effort we can soon enough shim it all. This pr could stay draft while me and other members port the missing functions and at that point all vendored code is removed and only the python shim remains.

Textured polygon can already be added to the shim.

@bilhox proposed to deprecate pie as it can be achieved with other draw functions. What do you think?
- port it to C draw
- deprecate it
- create a python version similar to textured polygon that emulates it.

Thanks for the review.

buildconfig/stubs/pygame/gfxdraw.pyi
Starbuck5:
```py
>>> def func1(a, b):
...     print(a, b)
...
>>>
>>> def func2(a, b, /):
...     print(a, b)
...
>>>
>>> func1("1", "2")
1 2
>>> func2("1", "2")
1 2
>>> func1(a="1", b="2")
1 2
>>> func2(a="1", b="2")
TypeError: func2() got some positional-only arguments passed as keyword arguments: 'a, b'
```

damusss:
After some discussion, pull request https://github.com/pygame-community/pygame-ce/pull/3010 closes this one as not planned.

PR #3008, opened 2024-07-19
Fix segfault with antialiased draw functions with a depth different than 32bits

### LINKED ISSUE #586: SDL2: gfxdraw.aa* functions segfault with 8-bit surfaces (849)
# **Issue &numero;849 opened by *[dlon](https://github.com/dlon)* at 2019-02-24 22:01:42**

An 8-bit surface without a palette (masks (0xe0, 0x1c, 0x3, 0x0)) can cause a segfault for aapolygon, aacircle, aatrigon, and aaellipse.

```python3

from pygame import Surface, Color, init
from pygame.gfxdraw import aacircle

init()

s = Surface(size=(512, 512), flags=0, depth=8, masks=(0xE0, 0x1C, 0x3, 0x0))

aacircle(
    s,
    256,
    256,
    64,
    Color("red"),
)

```

System: pygame 1.9.5 / SDL2 / Python 3.7.2 (64-bit). Doesn't happen with SDL1.2.

**Related Docs**: https://www.pygame.org/docs/ref/gfxdraw.html<hr>

# Comments

bilhox:
Closes #3008

This segfault also happens with antialiased functions from ``pygame.draw``.

*Needs some tests from other users*


bilhox:
edit : It doesn't fix ``gfxdraw`` side as it's meant to be deprecated and be redirected to ``pygame.draw`` for all antialiasing functions.

damusss (approved):
I tested locally. The segfault existed before, now it's gone. Visually it was broken before, now it's perfect.

LGTM, thanks for the fix! :)

test/draw_test.py
```diff
@@ -7176,6 +7176,20 @@ def test_color_validation(self):
...
+        """
+        for depth in (8, 16, 24, 32):
+            # all values must stay so to reproduce the segfault
+            surf = pygame.Surface(size=(512, 512), flags=0, depth=depth)
+
+            draw.aacircle(surf, pygame.Color("red"), (256, 256), 64)
+            draw.aaline(surf, pygame.Color("red"), (256, 256), (512, 512))
+            draw.aalines(
+                surf, pygame.Color("green"), False, [(256, 256), (512, 512), (400, 400)]
+            )
```
ankith26, line 7191: I'd like this test to take a few random pixels (atleast 5 or so) on the surface and assert that the color matches exactly.

src_c/draw.c
```diff
@@ -1126,15 +1126,43 @@ get_antialiased_color(SDL_Surface *surf, int x, int y, Uint32 original_color,
...
+                       &background_color[1], &background_color[2]);
+        } break;
+
+        default:
+            SDL_GetRGBA(((Uint32 *)surf->pixels)[(y * surf->w) + x],
+                        surf->format, &background_color[0],
+                        &background_color[1], &background_color[2],
+                        &background_color[3]);
+            break;
+    }
```
ankith26, line 1164: This block can be refactored to reduce some duplication. I would suggest having a `Uint32 pixel` variable declared first, then set on the case-wise basis, so that the below part can be done outside the switch statement.
```c
SDL_GetRGB(pixel, surf->format, &background_color[0],
    &background_color[1], &background_color[2]);
```

ankith26 (changes requested):
I left a few reviews. I'm a bit sceptical of the 24-bit codepath, which is also why I'm suggesting the tests to be stricter here.

src_c/draw.c
bilhox: Should be fixed now.

test/draw_test.py
bilhox: Should be fixed. I took pixels at edges since the problem concerns antialiased mode.

Starbuck5:
This whole thing is a mess.

Firstly, I would expect in a PR like this to see an explanation of what is actually being fixed, why it is segfaulting. Not just that it's fixed and needs tests from other users.

You want to use SDL_GetRGBA correctly? See https://github.com/libsdl-org/SDL/issues/8320 and https://github.com/libsdl-org/SDL/commit/f8dfee01bb4fdca3d7f77d4d2a30ed58c63e869c#diff-ec00a797fe34913f3e553761506783ef6b4cd0050704665615c93392ac7360e6R1557-R1618

They mention the performance is terrible "meant only for unittests", so that gets me thinking about performance:

So get_antialiased_color is a performance sensitive function, why is it being called over and over decoding the value of original_color every single time?

And so then this function produces a result to use in set_at, but set_at also calls SDL_GetRGB, but only on 24 bit! So in set_at it's taking a mapped color, unmapping it with SDL_GetRGB, then manually remapping it with bit shifting. Am I reading that right?

bilhox:
> This whole thing is a mess.
> 
> Firstly, I would expect in a PR like this to see an explanation of what is actually being fixed, why it is segfaulting. Not just that it's fixed and needs tests from other users.
> 
> You want to use SDL_GetRGBA correctly? See [libsdl-org/SDL#8320](https://github.com/libsdl-org/SDL/issues/8320) and [libsdl-org/SDL@f8dfee0#diff-ec00a797fe34913f3e553761506783ef6b4cd0050704665615c93392ac7360e6R1557-R1618](https://github.com/libsdl-org/SDL/commit/f8dfee01bb4fdca3d7f77d4d2a30ed58c63e869c#diff-ec00a797fe34913f3e553761506783ef6b4cd0050704665615c93392ac7360e6R1557-R1618)
> 
> They mention the performance is terrible "meant only for unittests", so that gets me thinking about performance:
> 
> So get_antialiased_color is a performance sensitive function, why is it being called over and over decoding the value of original_color every single time?
> 
> And so then this function produces a result to use in set_at, but set_at also calls SDL_GetRGB, but only on 24 bit! So in set_at it's taking a mapped color, unmapping it with SDL_GetRGB, then manually remapping it with bit shifting. Am I reading that right? (EDIT: PR to fix set_at to not call SDL_GetRGB every pixel: #3021)
> 
> P.S. maybe we can do our version of SDL_GetRGBA "manually inlined" into our source by grabbing it from the SDL source.

Hello, first, thank you for the constructive review !

Sorry, I forgot to mention what was the culprit behind this segfault. As you guessed, the problem resided on how we were extracting pixels in ``get_antialiased_color`` in ``draw.c`` (extracted as a ``Uint32`` array), leading to read at pixels out of the range of the array.

You suggested me to look at how SDL does the same work with ``SDL_ReadSurfacePixel``, are you specifically talking about 24bit path ?

Starbuck5:
> You suggested me to look at how SDL does the same work with SDL_ReadSurfacePixel, are you specifically talking about 24bit path ?

So I see you checked out the SDL implementation and put it in, nice! The big endian build is failing though since you have a discrepancy in variable names.

I found another way to get around this problem that may be better. When I looked through draw.c I saw that this is the only location that does any pixel reading, everything else is just doing pixel writing. But there are places in pygame-ce doing pixel reading, how are they doing it? I found in surface.h there's a macro `GET_PIXEL` that works for all bit depths, that could be a good solution here. Might even be faster, as it doesn't need a memcpy-- although the memcpy might be compiler-optimized-away.

Anyways, I'd be interested in seeing you check it out.

bilhox:
I checked how it's done on other files, and I saw that they all used a switch case, and manually shift bits for 3 bytes case, and ultimately call ``GET_PIXELVAL`` which is a hidden ``SDL_GetRGBA``.

Starbuck5:
> I checked how it's done on other files, and I saw that they all used a switch case, and manually shift bits for 3 bytes case, and ultimately call GET_PIXELVAL which is a hidden SDL_GetRGBA.

Yes, I wasn't saying it avoided calling that function? Just saying that we already have an implementation strategy for calling that function properly, maybe that's the way to go here as well.

bilhox:
What about adding a new C API function called ``pg_getPixelValue`` ? Everywhere they have their own macros.

bilhox:
To move this fix as close as possible to being merged. I'm doing some performance tests (just a little skeptical about the perf behind a lot of ``memcpy`` calls) which I'll be posting here tomorrow. This will determine if we keep the switch case impl or the memcpy impl.

bilhox:
My tests on ``aaline`` showed me this : 
graph_get_pixel_value

red is the switch case implementation and green is the ``memcpy`` implementation.
As you can see switch case implementation is faster than ``memcpy``. So I'll be reverting to switch case. (The results for other bpp show the same difference of performance)

src_c/draw.c
```diff
@@ -31,6 +31,8 @@
 
 #include <math.h>
 
+#include <string.h>
```
MightyJosip, line 34: Is there a reason for this?

MightyJosip (commented):
Just a question

src_c/draw.c
bilhox: I have no idea when I included this, maybe I wrote this in the wrong branch, I'll remove it.

Starbuck5: string.h is needed for memcpy.

src_c/draw.c
```diff
@@ -1126,15 +1126,42 @@ get_antialiased_color(SDL_Surface *surf, int x, int y, Uint32 original_color,
...
-    Uint32 *pixels = (Uint32 *)surf->pixels;
     SDL_GetRGBA(original_color, surf->format, &color_part[0], &color_part[1],
                 &color_part[2], &color_part[3]);
     if (x < surf->clip_rect.x || x >= surf->clip_rect.x + surf->clip_rect.w ||
         y < surf->clip_rect.y || y >= surf->clip_rect.y + surf->clip_rect.h)
         return original_color;
-    SDL_GetRGBA(pixels[(y * surf->w) + x], surf->format, &background_color[0],
+
+    Uint32 pixel = 0;
+    size_t bpp = surf->format->BytesPerPixel;
```
Starbuck5, line 1136: `bpp = PG_SURF_BytesPerPixel(surf);` for SDL3 compatibility.

Why is this a size_t?

Starbuck5: Nice work changing surf->w to surf->pitch.

bilhox: > Why is this a size_t?

Good question, I can't remember what i was thinking about when I used ``size_t``. Switching to ``int``.

Starbuck5 (approved):
Thanks!

src_c/draw.c
Starbuck5: I approved and all, but just for the record you didn't switch to int.

bilhox: Sorry I forgot, now It's updated.

ankith26 (approved):
LGTM, thanks for the PT :sunglasses:

Issue #3005, opened 2024-07-18
Deprecating gfxdraw module


mzivic7:
These functions must be implemented to `pygame.draw` module in order to deprecate `pygame.gfxdraw` module
- [X] `draw.aacircle`
    - [X] main function
    - [X] thickness
    - [X] filled
- [ ] `draw.aaline`
    - [X] main function
    - [ ] thickness
- [ ] `draw.aalines`
    - [X] main function
    - [ ] thickness
- [ ] `draw.aaellipse`
    - [ ] main function
    - [ ] thickness
    - [ ] filled
- [ ] `draw.aapolygon`
    - [ ] main function
    - [ ] thickness
    - [ ] filled
- [ ] `draw.pie`
    - can be copied to draw module
- [ ] `draw.bezier`
    - can be copied to draw module

Each `aa-` function should be interchangeable with regular function, in terms of arguments and capabilities.

bilhox:
Hello, 

First of all, thank you for the list. IMO, ``gfxdraw.pie`` shouldn't be ported to ``draw`` as it can be reproduced easily with existing ``draw`` functions.

mzivic7:
Ok, removing `draw.pie` from the list, because it can be drawn with `draw.arc` and 2 `draw.lines`.

mzivic7:
Idea for draw.aapolygon:

```
If width == 1:
    1. Draw closed aalines with width=1 
Else:
    1. Draw closed aalines with width=1
    2. Now for each point:
        3. calculate bisecting vector of angle at that point such that it points toward inside of polygon
        4. Get point moved by 1px along that vector in its direction
        This will result in 1px smaller polygon
    5. With new points draw another aalines
    If filled:
        6. Draw regular filled polygon with same points from 5.
    Else:
        6. Repeat 2.-5. with draw.lines until layer == width-2, in ecah iteration with each layer polygon will get smaller
        7. Draw another aalines with 1px smaller polygon than previous
        8. Repeat 7.
```

Why this:
Closed shapes must expand width towards inside, and it is expected for polygon to have sharp edges.
1. alone is enough to deprecate gfxdraw.aapolygon. Other points are for width and filled which is IMO more important than just aapolygon.

damusss:
> Ok, removing `draw.pie` from the list, because it can be drawn with `draw.arc` and 2 `draw.lines`.

that's how I replaced it in my shim

mzivic7:
[...]

PR #3009, opened 2024-07-21
Port bezier from ``gfxdraw`` to ``draw`` + bring the same support like every ``draw`` function

bilhox:
Hello, as part of #3005 plan, this PR moves ``pygame.gfxdraw.bezier`` to ``pygame.draw.bezier``.
Things remaining : 

- [ ] Deprecate ``pygame.gfxdraw.bezier``.
- [ ] Make ``pygame.draw.bezier`` returning the bounding rect like his fellow friends.

### DIFF
--- a/buildconfig/stubs/pygame/draw.pyi
+++ b/buildconfig/stubs/pygame/draw.pyi
@@ -89,3 +89,9 @@ def aalines(
     closed: bool,
     points: Sequence[Coordinate],
 ) -> Rect: ...
+def bezier(
+    surface: Surface,
+    points: Sequence[Coordinate],
+    steps: int,
+    color: ColorValue,
+) -> None: ...
--- a/docs/reST/ref/draw.rst
+++ b/docs/reST/ref/draw.rst
@@ -598,6 +598,37 @@ object around the draw calls (see :func:`pygame.Surface.lock` and
 
    .. ## pygame.draw.aalines ##
 
+.. function:: bezier
+
+   | :sl:`draw a Bezier curve`
+   | :sg:`bezier(surface, points, steps, color) -> None`
+
+   Draws a Bézier curve on the given surface.
+
+   :param Surface surface: surface to draw on
+   :param points: a sequence of 3 or more (x, y) coordinates used to form a
+      curve, where each *coordinate* in the sequence must be a
+      tuple/list/:class:`pygame.math.Vector2` of 2 ints/floats (float values
+      will be truncated)
+   :type points: tuple(coordinate) or list(coordinate)
+   :param int steps: number of steps for the interpolation, the minimum is 2
+   :param color: color to draw with, the alpha value is optional if using a
+      tuple ``(RGB[A])``
+   :type color: Color or string (for :doc:`color_list`) or int or tuple(int, int, int, [int])
+
+   :returns: ``None``
+   :rtype: NoneType
+
+   :raises ValueError: if ``steps < 2``
+   :raises ValueError: if ``len(points) < 3`` (must have at least 3 points)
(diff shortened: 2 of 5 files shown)


damusss:
Adding this as a note, when all of gfxdraw is ported to draw (if my PR is approved ofc), the deprecation warning will be offloaded to the python shim and the C function will be erased. (the `@deprecated` decorator actually raises the warning in py file, not in pyi files).
If you want to add the deprecation to the stubs, please use the `@deprecated("since GFX_DEPRECATED_VERSION. Use `pygame.draw.bezier` instead")` notation, to be consistent with the rest of the stubs, otherwise it will be included in my PR.

docs/reST/ref/gfxdraw.rst
```diff
@@ -602,6 +602,8 @@ For example:
    | :sl:`draw a Bezier curve`
    | :sg:`bezier(surface, points, steps, color, /) -> None`
 
+   DEPRECATED : use :func:`pygame.draw.bezier` instead.
```
damusss, line 605: [non blocking nitpick]
pwetty pwease attach the : to the left word. Anywhere else if you see it too

bilhox: I just removed this line and placed the deprecation content in front of the deprecation version.

damusss: Sounds like the best move, 💯

MyreMylar (approved):
OK, LGTM. Bezier works as expected under the draw module instead of gfxdraw in my testing. the tests all pass locally as well.

Test program:

```
import pygame

pygame.init()

sf = pygame.display.set_mode((640, 480))

font = pygame.font.Font(size=36)

points = []

STEPS = 3

print("Click on the screen to put points.")

while 1:
            points.append(event.pos)
    sf.fill((0, 0, 0))

    for p in points:
        pygame.draw.circle(sf, (255, 0, 0), p, 5)

    if len(points) > 2:
        pygame.draw.lines(sf, (255, 0, 0), False, points)
        pygame.draw.bezier(sf, points, STEPS, (255, 255, 255))

    test = f"Points: {len(points)}\nSteps: {STEPS}"
    text_sf = font.render(test, True, (255, 255, 255))
    sf.blit(text_sf, (10, 400))

    pygame.display.update()
```

bilhox:
I would like to know your opinion about the ``step`` argument before potentially merging it. I think personally the docs are not clear at all about ``step`` argument, and when I tested it with different values, I couldn't really figure out how it was used. Furthermore the code is not that great to read (I first look the variable names are not understandable), which is why I support the idea of a new implementation of Bézier curves so I can also document the terrible code.

MyreMylar:
> I would like to know your opinion about the `step` argument before potentially merging it. I think personally the docs are not clear at all about `step` argument, and when I tested it with different values, I couldn't really figure out how it was used. Furthermore the code is not that great to read (I first look the variable names are not understandable), which is why I support the idea of a new implementation of Bézier curves so I can also document the terrible code.

My main feeling is that we should try and get this merged so we can shim & steadily get rid of `gfxdraw` from docs and examples and the like. 

However, if you think the `step` argument is the least reliable argument in a future implementation of bezier line drawing then what I would do is move it to the end of the function signature instead of before the color argument, and make sure it has a default value specified so it is optional. You could also make it a keyword argument only parameter too.

That way in the future it will be much easier to remove or alter the step argument while still allowing us to handle it as it is right now and make a convenient gfxdraw shim for it.

So, to be clear the function signature (in python) would be something like:

```
def bezier(Surface surface, Sequence[Point] points, ColorLike color, *,  int step=3)
```

PR #3012, opened 2024-07-21
Optimize draw.aacircle and fix typo

mzivic7:
In `draw_circle_xiaolinwu_thin`, `pow(radius, 2)` was inside loop, but `radius` isn't changing.
So, `radius` is now moved outside of the loop.
Same thing happened in `draw_circle_xiaolinwu`.
Performance gain: 5% faster execution with r=100, and 6% with r=1000, for 1px wide line.
Also fixed typo: Xaolin -> Xiaolin


ankith26 (approved):
LGTM, thanks! :partying_face:

damusss (approved):
LGTM, nice :3

PR #3014, opened 2024-07-22
Remove unnecessary pixeloffset member of subsurface data

Starbuck5:
I was digging around our structs and found this member barely used anywhere. All it did was set the subsurf pixel pointer to the correct location when pgSurface_Prep is called. But the subsurf pixel pointer is set up to point at the right location in surface.subsurface already, and never changes from that position. So storing the offset and re-setting it occasionally is unnecessary effort.

How did I verify it was unnecessary?
- No other bits of code set `pixels` to something else on a surface. (So if nothing is messing it up, no need to reset it)
- I put in an assert in the previous prep code to fail if the current pixel pointer is different to the one it was being set to, and through the entire test suite that assertion held true.

This makes the subsurface code a tiny bit simpler and more memory efficient.

### DIFF
--- a/src_c/surflock.c
+++ b/src_c/surflock.c
@@ -44,10 +44,7 @@ pgSurface_Prep(pgSurfaceObject *surfobj)
 {
     struct pgSubSurface_Data *data = ((pgSurfaceObject *)surfobj)->subsurface;
     if (data != NULL) {
-        SDL_Surface *surf = pgSurface_AsSurface(surfobj);
-        SDL_Surface *owner = pgSurface_AsSurface(data->owner);
         pgSurface_LockBy((pgSurfaceObject *)data->owner, (PyObject *)surfobj);
-        surf->pixels = ((char *)owner->pixels) + data->pixeloffset;
     }
 }
 
--- a/src_c/_pygame.h
+++ b/src_c/_pygame.h
@@ -478,7 +478,6 @@ struct pgEventObject {
  */
 struct pgSubSurface_Data {
     PyObject *owner;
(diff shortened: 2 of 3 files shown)


MyreMylar (approved):
LGTM 👍 
I can find no other usage of the `pixeloffset` member, as you say it just seems to be setting the `surface->pixels` pointer to the place it is already pointing at.

src_c/surflock.c
```diff
@@ -44,10 +44,7 @@ pgSurface_Prep(pgSurfaceObject *surfobj)
 {
     struct pgSubSurface_Data *data = ((pgSurfaceObject *)surfobj)->subsurface;
     if (data != NULL) {
-        SDL_Surface *surf = pgSurface_AsSurface(surfobj);
-        SDL_Surface *owner = pgSurface_AsSurface(data->owner);
         pgSurface_LockBy((pgSurfaceObject *)data->owner, (PyObject *)surfobj);
-        surf->pixels = ((char *)owner->pixels) + data->pixeloffset;
```
ankith26, line 50: The code being removed here is so weird, I wonder if there was any intention behind this... IG we will never know

ankith26 (approved):
LGTM, thanks! :balloon:

src_c/surflock.c
Starbuck5: I investigated a little further, it's been like this since surflock.c was created 24 years ago! https://github.com/pygame/pygame/commit/45784b329a51062d111ddd030082dc17b17ba6b8

PR #3016, opened 2024-07-22
Draw.aaellipse()

mzivic7:
This PR adds `draw.aaellipse` function, with width and thickness options.
It is completely interchangeable with `draw.ellipse` in terms of arguments.
Algorithm used is adapted from Xiaolin Wu's general fast antialiasing algorithm, and is very similar to `draw.aacircle`.
Only notable difference: in aacircle, 8 pixels are drawn at once, but here, only 4, because ellipse has 2 symmetries.
It is added to docs too.


src_c/draw.c
```diff
@@ -2856,6 +2964,177 @@ draw_ellipse_thickness(SDL_Surface *surf, int x0, int y0, int width,
     }
 }
 
+/* Xiaolin Wu Ellipse Algorithm
+ * adapted from: https://cgg.mff.cuni.cz/~pepca/ref/WU.pdf
+ * with additional line width parameter
+ */
+static void
+draw_ellipse_xiaolinwu(SDL_Surface *surf, int x0, int y0, int width,
```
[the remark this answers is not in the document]
mzivic7: Should i then rename draw_circle_xiaolinwu functions too?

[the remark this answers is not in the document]
MightyJosip: Yea it would be good 👍

MyreMylar (approved):
Well this looks good to me. The new algorithm performs better than the one in gfxdraw (which has gaps in it) and adds the filling functionality. Tested a variety of parameters locally and all looked nice and behaved how I would expect.

Can't find any problems with the code, not going to pretend I understand the intricacies of the algorithm for drawing, but it looks sensible from a look over and produces an output I respect. 👍

mzivic7:
Note: before merging, correct versionadded in `docs/reST/ref/draw.rst`, if needed.

src_c/draw.c
```diff
@@ -730,6 +739,91 @@ ellipse(PyObject *self, PyObject *arg, PyObject *kwargs)
...
+    }
+    else if (rect->w == 1) {
+        draw_line(surf, rect->x, rect->y, rect->x, rect->y + rect->h - 1,
+                  color, drawn_area);
+    }
+    else if (rect->h == 1) {
+        draw_line(surf, rect->x, rect->y, rect->x + rect->w - 1, rect->y,
+                  color, drawn_area);
+    }
+    else {
```
Starbuck5, line 797: For the line cases above I don't see why there's a `-1` in the formulas. Maybe a good place for a comment?

For the else cases below I don't see why it's down-nested into a else with its own if / else if/ else, couldn't those all exist as part of the if / else if one level up?

Starbuck5: Also this is more of nitpick than anything, but I feel like `if (width == 0)` is easer to read than `if (!width)`.

src_c/draw.c
```diff
@@ -2624,6 +2718,22 @@ draw_circle_filled(SDL_Surface *surf, int x0, int y0, int radius, Uint32 color,
     }
 }
 
+static void
+draw_four_symetric_pixels(SDL_Surface *surf, int x0, int y0, Uint32 color,
```
Starbuck5, line 2722: symetric -> symmetric

Also this function could be made a lot more efficient by dealing with all the pixels at once instead of pixel by pixel going through all the things... but then this function would also be 10x more complicated so I'm glad it's this way.

mzivic7:
I moved `ellipse` and `aaellipse` functions bellow `circle` and `aacircle`, because `aacircle` is called in `aaellipse` and `aa*` functions should always be bellow regular one.

Starbuck5:
Other things I see.
- [x] `pygame.draw.aaellipse(screen, "red", [100,200,30,200000], width=0)` crashes the app
- [x] putting in a rect with negative dimensions leads to weird stuff being drawn
- [x] if the width and the height are equal the surface being drawn on cannot be blitted in the future (you forgot to unlock it in this case)

PR #3017, opened 2024-07-23
Skip flaky window tests on pypy

ankith26:
Until we figure out the cause of the fails, it is better to skip these fails.

LMK if I have missed any other test in this file that's also flaky on pypy

### DIFF
--- a/test/window_test.py
+++ b/test/window_test.py
@@ -1,11 +1,15 @@
 import unittest
 import pygame
 import os
+import platform
 
 from pygame import Window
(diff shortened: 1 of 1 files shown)


damusss (approved):
I don't remember any other tests. They fail pretty often, so this is a good idea for the time being IMO.
LGTM :)

oddbookworm (approved):
I think I’ve seen these occasionally fail on non-puppy runs, but I couldn’t guarantee that right now. Let’s see what happens

PR #3020, opened 2024-07-26
Use PyObject_CallNoArgs where applicable

Starbuck5:
* PyObject_CallNoArgs was added in 3.9, pythoncapi_compat.h enables support for 3.8.
* Docs say it's the "most efficient way to call a callable Python object without any argument"

This is not a performance PR, but a "use the modern thing" PR. The Python API now has this function we can use, lets use it. It's in the stable ABI, and it's a bit more explicit than the previous uses.

### DIFF
--- a/src_c/base.c
+++ b/src_c/base.c
@@ -270,7 +270,7 @@ pg_mod_autoinit(const char *modname)
     }
 
     if (funcobj) {
-        temp = PyObject_CallObject(funcobj, NULL);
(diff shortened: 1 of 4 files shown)


damusss (approved):
Makes sense, LGTM :)

gresm:
LGTM :+1:.

gresm (approved):
LGTM :+1:.

ankith26 (approved):

PR #3021, opened 2024-07-26
draw.c: better strategy for 24bpp writes

Starbuck5:
This is simpler code and better performance for `set_at` and `unsafe_set_at`.

We don't want to have to go out and call an SDL function in any sort of hot loop.

`drawhorzline` and `drawvertline` already do their writes this way, so it's already proven out in the world.

In this performance test (a width=1 line uses set_at), I saw a 20% performance increase:
```py
import random
import time

import pygame

random.seed(36)

WIDTH, HEIGHT = 200, 200

NUM_LINES = 50000
NUM_REPS = 10

lines = [
    (
        (random.randint(0, 255), random.randint(0, 255), random.randint(0, 255)),
        (random.randint(0, WIDTH), random.randint(0, HEIGHT)),
        (random.randint(0, WIDTH), random.randint(0, HEIGHT)),
    )
    for _ in range(NUM_LINES)
]

surf = pygame.Surface((100, 100), depth=24)

start = time.time()
for _ in range(NUM_REPS):
    for line in lines:
        pygame.draw.line(surf, line[0], line[1], line[2], 1)
print(time.time() - start)
```

### DIFF
--- a/src_c/draw.c
+++ b/src_c/draw.c
@@ -1213,7 +1213,6 @@ set_at(SDL_Surface *surf, int x, int y, Uint32 color)
 {
     SDL_PixelFormat *format = surf->format;
     Uint8 *pixels = (Uint8 *)surf->pixels;
-    Uint8 *byte_buf, rgb[4];
 
     if (x < surf->clip_rect.x || x >= surf->clip_rect.x + surf->clip_rect.w ||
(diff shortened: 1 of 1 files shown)


ankith26 (approved):
LGTM, thanks for contributing! :100:

damusss (approved):
LGTM, I saw the improvement locally as you said :)

PR #3023, opened 2024-07-28
Use PyObject_CallOneArg where applicable

ankith26:
A PR that follows what #3020 did

### DIFF
--- a/src_c/base.c
+++ b/src_c/base.c
@@ -2349,7 +2349,7 @@ MODINIT_DEFINE(base)
     if (!quit) { /* assertion */
         goto error;
     }
-    rval = PyObject_CallFunctionObjArgs(atexit_register, quit, NULL);
+    rval = PyObject_CallOneArg(atexit_register, quit);
     Py_DECREF(atexit_register);
     Py_DECREF(quit);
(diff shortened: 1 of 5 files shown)


Starbuck5:
This was actually part of my commit for 3020 before I pushed it up to GitHub. The reason I didn’t want to use this is because it’s not in the limited API, and I would still like to move towards compiling with the limited API.

We could have a shim that converts this to the old thing in that theoretical case, but I just felt like this was a less clear win than NoArgs, also because NoArgs docs claim speed and OneArgs docs don’t mention it.

ankith26:
I think this probably has the same level of "minor performance gain" as the other PR, but I didn't notice the limited API detail while making the PR

Starbuck5 (approved):
Alright, I'm okay with this since it's using the fancy vectorcall underneath.

We'll get some tiny unmeasurable speedups when these functions are utilized. But you're doing this more for code modernization, I get that.

I checked and the Rect performance increases I got from the vectorcall PR are preserved here.

MyreMylar (approved):
Alright LGTM 👍 Tests all pass locally, code replacement looks straightforward & correct to me

PR #3026, opened 2024-07-31
``pygame.mixer.get(set ?)_audio_device(s)``

bilhox:
Hello,

The purpose of my PR is to be able to get the list of available audio devices names. I saw we could chose an audio device in ``pygame.mixer.init``, but we couldn't get a full list of available audio devices directly in Python.
I also checked if anyone had a PR about it, and found this issue #1285 , since I see no reference to a PR, I hope it's not a duplicate.

Currently, only ``get_audio_devices`` is available, and might need a full rename. I also don't like the way we have to select another audio device by calling ``pygame.mixer.init`` each time, this is why I'm thinking about a function called ``pygame.mixer.set_audio_device`` .


buildconfig/stubs/pygame/mixer.pyi
```diff
@@ -41,6 +41,7 @@ def set_soundfont(paths: Optional[str] = None, /) -> None: ...
 def get_soundfont() -> Optional[str]: ...
 def get_busy() -> bool: ...
 def get_sdl_mixer_version(linked: bool = True) -> Tuple[int, int, int]: ...
+def get_audio_devices() -> list: ...
```
damusss, line 44: the return type should be list[str] as stated in the docs in the stub too

damusss (changes requested):
I only glanced at the PR, I'll check it in detail tomorrow. It makes sense to have it, and I might be in favor of the set function aswell. The name lgtm for now. The requested changes are about the stubs.

Starbuck5:
Have you seen `pygame._sdl2.audio.get_audio_device_names()` ?

bilhox:
> Have you seen `pygame._sdl2.audio.get_audio_device_names()` ?

As discussed on discord, I didn't see that this function existed as it was undocumented. Furthermore @Starbuck5 said "The problem is with the ergonomics", as we have to init ``pygame.mixer`` first.

I still support the implementation of this feature directly in ``pygame.mixer``, we just should have ``set_audio_device`` which would be a hidden ``quit`` and ``init``.

bilhox:
Closed because the way we get audio device after initializing pygame.mixer is not really good. Right now, we need to quit and init again the module with the specific device name. I would personally come back to this the day we can get audio devices without initialising pygame.mixer. That said, if you really want to get audio devices names, you should turn yourself to ``pygame._sdl2.get_audio_device_names`` .

PR #3030, opened 2024-08-04
Add python 3.13 wheel support

ankith26:
Python 3.13rc1 is out, we should be testing on it and making wheels for it.

This is a pretty simple PR, the only thing that could look slightly weird is the docker file updates happening in this PR. This is needed for github actions to regen the base images and therefore get 3.13rc1 in the first place.


Starbuck5 (approved):
Tested out a 3.13 wheel on Windows and it worked as expected. Thanks!

oddbookworm (approved):
LGTM!

PR #3033, opened 2024-08-04
`Sprite.frect` is now supported as an alias of `Sprite.rect`

oddbookworm:
Also improves error messaging when something goes wrong in `Group.draw`


ankith26:
In many places where pygame expects a "rect", an `FRect` is already acceptable. I don't see how adding an alias is very useful here if our goal is to promote the interchangeability/interuse of both (if this is a real word lol)

oddbookworm:
> In many places where pygame expects a "rect", an `FRect` is already acceptable. I don't see how adding an alias is very useful here if our goal is to promote the interchangeability/interuse of both (if this is a real word lol)

I've seen several times that someone has tried to do `self.frect = ...` and expected it to work. I agree that on principle, everyone should just use `.rect`, but I also don't want to keep answering the same question over and over again about why `self.frect` doesn't work. Which is why I decided to go with an alias, but also make it such a small note in the docs that I hope almost nobody would ever know it's there

Starbuck5:
I agree with Ankith here.

JiffyRob:
If it's a common enough issue that it needs to be addressed automatically, why not print a warning when someone first uses it in their code?  

Is it in the docs?  If people don't read them then you're going to have to be dealing with them needing help a lot anyway.

src_py/sprite.py
```diff
@@ -562,6 +571,14 @@ def draw(self, surface):
...
         """
         sprites = self.sprites()
+        for spr in sprites:
+            if not hasattr(spr, "rect"):
+                raise AttributeError(
+                    "Sprite must have a rect attribute of type Rect or FRect"
+                )
+            if not isinstance(spr.rect, (pygame.rect.FRect, pygame.rect.Rect)):
+                raise TypeError("Sprite.rect must be either Rect or FRect")
+
```
Matiiss, line 581: This adds an insane performance overhead, it's two (expensive) calls per sprite before iterating over all the sprites again in the `blit(s)` calls below. Also interestingly it doesn't check for `.frect` which this PR aims to implement.

src_py/sprite.py
```diff
@@ -562,6 +571,14 @@ def draw(self, surface):
 
         """
         sprites = self.sprites()
+        for spr in sprites:
+            if not hasattr(spr, "rect"):
+                raise AttributeError(
+                    "Sprite must have a rect attribute of type Rect or FRect"
+                )
```
Matiiss, line 578: `spr` is very likely (almost certain) to have this attribute since it's a property of `Sprite`. If it hasn't been set by the user, it will fail below on the `isinstance` check because it defaults to `None`.

Matiiss (commented):
Aside from the two changes that stuck out to me immediately, I also agree with Ankith. This aliasing feels like a bandage. I would side with Jiffyrob in adding a warning under the `frect` property or even raise an exception, that way, if the user wants to use `frect`, they have to override it themselves. We could even `raise NotImplemented` there with a descriptive error message.

As for the changes in `draw`, they are going to be detrimental to performance for rather little benefit. It's rather something that maybe should be sought to be improved with the `blit(s)` methods themselves.

oddbookworm:
Yeah, thinking back on this, I agree with the opinions already here. Closing this pull request

PR #3043, opened 2024-08-07
Improved geometry docs

itzpr3d4t0r:
This PR tries to simplify and improve geometry docs a bit. 

**What I tried to achieve:**
- Lessen word repetition
- Remove obvious statements
- Remove self references when describing methods (like saying: "The collidecircle method does ..." inside the collidecircle 's docs)
- Uniform collision / overlap / intersect / tests / checks word usage
- Moved `Circle.move()` docs from between collision methods to under collision methods
- Shorten phrases a bit
- Use of actual math formulas syntax


ankith26 (approved):
LGTM on a quick skim, thanks! :tada:

Issue #3049, opened 2024-08-11
Port SDL_FlashWindow


Starbuck5:
This one of the few SDL Windows features we don't have ported in pygame.Window.

https://wiki.libsdl.org/SDL2/SDL_FlashWindow (seems to be unchanged in SDL3)

This seems a fun and relatively straightforward addition.

API design wise, I think this could be window.flash(operation), and operation could be a flag representing https://wiki.libsdl.org/SDL2/SDL_FlashOperation ?

damusss:
Do you prefer "verbose" constants names such as "WINDOW_FLASH_CANCEL" or "FLASH_CANCEL"?

PR #3054, opened 2024-08-13
Add `Window.flash(operation)`

damusss:
Implements https://github.com/pygame-community/pygame-ce/issues/3049.
Stubs, docs, tests included.
I chose the non-verbose constant names following SDL.


docs/reST/ref/window.rst
```diff
@@ -388,4 +388,20 @@
...
+      | :sg:`flash(operation, /) -> None`
+
+      :param int operation: The flash operation.
+      
+      Supported flash operations are:
+         * ``pygame.FLASH_CANCEL``: Cancel the current flash state if present
+         * ``pygame.FLASH_BRIEFLY``: Flash for a short amount of time to get attention
+         * ``pygame.FLASH_UNTIL_FOCUSED``: Keep flashing until the window is focused
+
+      .. note:: This function is not supported on all platforms
```
Starbuck5, line 403: What happens if it is not supported? What platforms are supported?

damusss: > What happens if it is not supported? What platforms are supported?

When it's not supported pygame.error is raised with message "That operation is not supported". I've tried to inspect SDL source code and I could only find the function defined for windows, x11, cocoa and wayland, but I could not find an article or message specifying the platforms really supported, it's not even mentioned in the documentation. I'll try to use vscode search inside the SDL lib as it may be smarter than github search function. Annoyingly, if FlashWindow doesn't support a specfic operation or if FlashWindow is completely absent for the platform the same SDL error is raised.

Starbuck5: These are common problems with a lot of our window functions, like I wonder if it would be better to return a success/fail bool instead of raising an exception.

Because if you call this function do you really want your code to crash if it doesn't work? Or have to put it in a try except?

What you have implemented is consistent with the rest of the window module though, which is good. But the documentation should at least be explicit about what happens when it fails, and how that can be mitigated.

Starbuck5: And similarly we might as well put in the documentation that this is supported on Windows, Mac, x11 / Wayland (Linux), if that's what the SDL source says.

Also nice job working this up! I just gave it a test locally and it looks good.

damusss: thanks :)
The boolean idea is interesting but as you said the exception is consistent with `set_modal_for` and other API and in theory holds more information. I'll be more specific in the docs about the platforms and the exception.

Starbuck5: Maybe it would even benefit from an inline example w/ the try/except if that's how we expect it to be used.

Starbuck5 (approved):

src_c/window.c
```diff
@@ -1066,6 +1066,35 @@ window_from_display_module(PyTypeObject *cls, PyObject *_null)
...
+    }
+
+    if (SDL_FlashWindow(self->_win, operation) < 0) {
+        return RAISE(pgExc_SDLError, SDL_GetError());
+    }
+#else
+    if (PyErr_WarnEx(PyExc_Warning, "'Window.flash' requires SDL 2.0.16+",
+                     1) == -1) {
+        return NULL;
+    }
```
ankith26, line 1093:
```suggestion
    return RAISE(pgExc_SDLError, "'Window.flash' requires SDL 2.0.16+");
```
I would prefer a `pygame.error` here instead

damusss: The rest of the window API only sets warnings (so this would be incosistent), but maybe that's because they are attributes and not functions, should I raise the error anyways?

damusss: @ankith26 I feel like raising `pygame.error` would mix up the "unsupported because platform" and "unsupported because low SDL" problems when catching it. I can still change it if needed.

ankith26: I feel like the mix up is fine because the user would want to handle both cases in the same way?

damusss: Good point you got there. I'll ignore the inconsistency then

ankith26 (approved):

PR #3057, opened 2024-08-16
Add `pygame.mixer_music.Queue`

damusss:
Docs and stubs are already included. Missing tests and docstrings.

It is not guaranteed to pass the checks as the C code gave me headaches locally and will probably fail remotely (hope not).

The stubs pass the stubcheck.py.

The implementation uses a little bit of help from C but it is mainly pythonic. Except for the Queue class no API is modified.

The idea was proposed by @bilhox and it looks like there is already someone that would like it.

API changes or renames are welcome. I can name stop() and resume() to pause() and unpause() if you prefer.


bilhox:
I think this need a very explicative issue on how it works otherwise some people can be confused on what it does exactly.

bilhox:
reference to #3058

damusss:
We have to rework mixer_music and put more thought into it, and that won't happen very soon, this PR doesn't make much sense existing IMO. I will only keep the branch as a backup.

PR #3060, opened 2024-08-17
make use of ``Mix_GetMusicPosition``

bilhox:
This PR offers a cleaner way to get music position of the current music playing using ``Mix_GetMusicPosition``. It also fixes several problems, like when a music ends the position jumps to another not obvious position, or when you quit pygame and re-init, you can still get a position when it's supposed to return -1.

This PR uses SDL_Image 2.6.0, is there a guarantee that every users using the latest pygame version have SDL_Mixer >=2.6.0 ?

I'll keep that as draft for the moment, I need to find a solution for the old version.


bilhox:
It looks like the current implementation does not even follow what its name says about it which should be "getting the position of the music" which definitely does not mean : "getting the time the music has been playing for". And the problem with this is that ``SDL_GetMusicPosition`` actually get the correct "music position", supporting playing offset and ``set_pos``.

We discussed on discord about a ``pygame.Music`` object which could handle a music more easily, so I'll close this PR in a few days (if anyone has an objection) and let the old behaviour, and I will start an implementation of ``pygame.Music`` which actually get the "music position".

damusss:
@bilhox  (which I think should reside in pygame.music module and not pygame.mixer_music) this creates the candidate of `.elapsed_time` attribute (retaining the old behaviour) in the new API

PR #3061, opened 2024-08-18
migrate importlib to resolve warning about pkg_resources

raspofabs:
I came across this warning when running pytest. I wondered if this fix would be appropriate.
At least on my machine the warning goes away and satisfies the deprecation warning from https://setuptools.pypa.io/en/latest/pkg_resources.html

I think I'm only using this feature for Fonts, so don't know how to check if I have broken anything with this change.

### DIFF
--- a/src_py/pkgdata.py
+++ b/src_py/pkgdata.py
@@ -22,23 +22,29 @@ import sys
 import os
 
 try:
-    import importlib_resources
+    if sys.version_info[:2] > (3, 8):
+        import importlib_resources
 
-    def resource_exists(_package_or_requirement, _resource_name):
-        return (
-            importlib_resources.files(_package_or_requirement)
-            .joinpath(_resource_name)
-            .is_file()
-        )
+        def resource_exists(_package_or_requirement, _resource_name):
+            return (
+                importlib_resources.files(_package_or_requirement)
+                .joinpath(_resource_name)
+                .is_file()
+            )
 
-    def resource_stream(_package_or_requirement, _resource_name):
-        ref = (
-            importlib_resources.files(_package_or_requirement)
-            .joinpath(_resource_name)
-            .is_file()
-        )
-        return ref.open('rb')
+        def resource_stream(_package_or_requirement, _resource_name):
+            ref = importlib_resources.files(_package_or_requirement).joinpath(
+                _resource_name
+            )
+            return ref.open('rb')
+    else:
+        from importlib import resources
 
+        def resource_exists(_package_or_requirement, _resource_name):
+            return resources.is_resource(_package_or_requirement, _resource_name)
+
+        def resource_stream(_package_of_requirement, _resource_name):
+            return resources.read_binary(_package_of_requirement, _resource_name)
 
 except ImportError:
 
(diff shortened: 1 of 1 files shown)


src_py/pkgdata.py
```diff
@@ -22,7 +22,24 @@ def getResource(identifier, pkgname=__name__):
 import os
 
 try:
-    from pkg_resources import resource_stream, resource_exists
+    import importlib_resources
```
ankith26, line 25:
```suggestion
    import importlib.resources
```
This is a third party backport library, and I think it would be better to use the stdlib one directly (and handle the python 3.8 case in code)

src_py/pkgdata.py
```diff
@@ -22,7 +22,24 @@ def getResource(identifier, pkgname=__name__):
...
+            importlib_resources.files(_package_or_requirement)
+            .joinpath(_resource_name)
+            .is_file()
+        )
+
+    def resource_stream(_package_or_requirement, _resource_name):
+        ref = (
+            importlib_resources.files(_package_or_requirement)
+            .joinpath(_resource_name)
+            .is_file()
```
ankith26, line 38: this line should be removed

ankith26 (commented):
Hi. First of all, thanks for contributing! :tada: 

We just need to handle the python 3.8 separately in an `if` condition, using `importlib.resources.open_binary` and `importlib.resources.is_resource` for that case.

ankith26:
Also, we could potentially be looking into dropping python 3.8 support soon now that it's approaching EOL. That should make the implementation of this PR simpler

src_py/pkgdata.py
raspofabs: oops 😄

src_py/pkgdata.py
raspofabs: I'm not sure I understand the problem you're describing. I've provided a new solution given how I understand the problem. I have a feeling I have got this all backwards.

src_py/pkgdata.py
```diff
@@ -22,7 +22,24 @@ def getResource(identifier, pkgname=__name__):
...
+            ref = files(_package_or_requirement).joinpath(_resource_name)
+            return ref.open('rb')
+    else:
+        from importlib import resources
+
+        def resource_exists(_package_or_requirement, _resource_name):
+            return resources.is_resource(_package_or_requirement, _resource_name)
+
+        def resource_stream(_package_of_requirement, _resource_name):
+            return resources.read_binary(_package_of_requirement, _resource_name)
```
ankith26, line 41:
```suggestion
            return resources.open_binary(_package_of_requirement, _resource_name)
```
`read_binary` returns a `bytes` instance, but we need `open_binary` which returns a filelike object

ankith26 (commented):
Yes, this is the approach I'm suggesting.

Except the CI seems to be failing, IG we have to do `_package_or_requirement = _package_or_requirement.split(".")[0]` in all the functions because we need it to be relative to the main `pygame` package and not `pygame.pkgdata` submodule.

ankith26 (commented):
Almost there... Just gotta add `# pylint: disable=deprecated-method;` comment in front of the deprecated functions to make pylint happy

MyreMylar (approved):
LGTM, would be good to clear this warning up somewhat from the output (even if we are not really using this file much anymore). I'm familiar with the replacement package and use it for the same purposes in pygame GUI.

ankith26 (approved):

PR #3063, opened 2024-08-18
Add alias `pygame.Sound`

damusss:
It has been enough time for the Sound class to get its alias!
I don't recall the last time I used a Channel object but somehow that has the alias and not Sound, which is more common.

I also slightly changed the examples to use the alias, if you prefer how it was previously just say so and I'll revert at the speed of light.

### DIFF
--- a/buildconfig/stubs/gen_stubs.py
+++ b/buildconfig/stubs/gen_stubs.py
@@ -68,7 +68,7 @@ PG_AUTOIMPORT_CLASSES = {
     "_debug": ["print_debug_info"],
     "event": ["Event"],
     "font": ["Font"],
-    "mixer": ["Channel"],
+    "mixer": ["Sound", "Channel"],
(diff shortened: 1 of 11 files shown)


Starbuck5:
I think it would be good to hold off on examples and usage changes for now, since the docs go up to the website before the next release. And people sometimes open issues when examples fail to run on the latest release version (as they should).

damusss:
> I think it would be good to hold off on examples and usage changes for now, since the docs go up to the website before the next release. And people sometimes open issues when examples fail to run on the latest release version (as they should).

I didn't think about that! reverted :) (this will probably have to be squashed for merging because reverting took me 4 commits for no reason)

ankith26 (approved):
LGTM, thanks for the PR :+1:

bilhox (approved):
LGTM, this will be a good feature!

PR #3064, opened 2024-08-19
Font constructor now raises an error when it fails to load a valid font file

oddbookworm:
Sample piece of code:
```py
import pygame

pygame.init()

win = pygame.Window()
screen = win.get_surface()

font = pygame.font.Font("MODERN.fnt", 20)
text = font.render("Hello World!", True, "white")

running = True
while running:
        if event.type == pygame.QUIT:
            running = False

    screen.fill("black")
    screen.blit(text, (0, 0))
    win.flip()
```

Sample font file (same file is added in the examples dir in this pull):
[MODERN.zip](https://github.com/user-attachments/files/16655315/MODERN.zip)
[Source for the sample file](http://justsolve.archiveteam.org/wiki/FNT_(Windows_Font))


ankith26 (approved):
Before this PR I get a `pygame.error: Passed a NULL pointer` error, after this PR I get `pygame.error: Couldn't load font file` which is much better OFC

Thanks! :100:

test/font_test.py
```diff
@@ -743,6 +743,16 @@ def test_load_default_font_filename_default(self):
...
         f = pygame_font.Font(pygame_font.get_default_font())
 
+    def test_load_file_valid_but_fails(self):
+        # .fnt files don't load (for example), but they're valid fonts
+        # this tests that a sample one throws a pygame.error on load
+        # (if not a FileNotFoundError due to the font format not being gucci
+        # for that version of SDL_ttf)
+        font_path = os.path.join("examples", "data", "MODERN.fnt")
+        self.assertRaises(
+            (pygame.error, FileNotFoundError), lambda: pygame_font.Font(font_path)
```
Starbuck5, line 753: How could this raise a FileNotFoundError?

pygame_font here could be ftfont based on freetype?
freetype.font segfaults on my system when I try to load this path, but maybe on some other systems it raises filenotfound.

oddbookworm: Yeah, I tracked down to where the segfault happens. Didn’t figure out exactly what went wrong, but the function call is here:
https://github.com/pygame-community/pygame-ce/blob/main/src_c/freetype/ft_wrap.c#L343

Starbuck5: So pygame_font is either Font or ftfont, and this PR introduces a difference between the Font and ftfont APIs (which are supposed to match).

This is why I bring it up. The test shouldn't need to test for either behavior, it should test for the one correct behavior which should be shared.

PR #3069, opened 2024-08-20
`Vector{2,3}.__delattr__` messaging fixed when deleting `x, y, {z}` and removed `Vector4` zombies

### LINKED ISSUE #3066: Incorrect error message for vectors
I was digging through the math.c source and found this:

```py
>>> import pygame
pygame-ce 2.5.2.dev1 (SDL 2.30.6, Python 3.9.5)
>>> a = pygame.Vector2(30,40)
>>> del a.y
TypeError: Cannot delete the x attribute
```

I didn't try to delete the x attribute, I tried to delete the y attribute!

Implementation code (unchanged for 8 years):
```c
static int
vector_set_component(pgVector *self, PyObject *value, int component)
{
    if (value == NULL) {
        PyErr_SetString(PyExc_TypeError, "Cannot delete the x attribute");
        return -1;
    }
    if (component >= self->dim) {
        PyErr_BadInternalCall();
        return -1;
    }

    self->coords[component] = PyFloat_AsDouble(value);
    if (PyErr_Occurred())
        return -1;
    return 0;
}
```

oddbookworm:
Fixes #3066 and removes the remnants of whatever Vector4 used to be there

### DIFF
--- a/src_c/math.c
+++ b/src_c/math.c
@@ -195,20 +195,12 @@ static PyObject *
 vector_gety(pgVector *self, void *closure);
 static PyObject *
 vector_getz(pgVector *self, void *closure);
-#ifdef PYGAME_MATH_VECTOR_HAVE_W
-static PyObject *
-vector_getw(pgVector *self, void *closure);
-#endif
 static int
 vector_setx(pgVector *self, PyObject *value, void *closure);
 static int
 vector_sety(pgVector *self, PyObject *value, void *closure);
 static int
 vector_setz(pgVector *self, PyObject *value, void *closure);
-#ifdef PYGAME_MATH_VECTOR_HAVE_W
-static int
-vector_setw(pgVector *self, PyObject *value, void *closure);
-#endif
 static PyObject *
 vector_richcompare(PyObject *o1, PyObject *o2, int op);
 static PyObject *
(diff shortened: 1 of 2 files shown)


src_c/math.c
```diff
@@ -1219,11 +1200,28 @@ static int
...
+                                "Cannot delete the x attribute");
+                break;
+            }
+            case 1: {
+                PyErr_SetString(PyExc_TypeError,
+                                "Cannot delete the y attribute");
+                break;
+            }
+            case 2: {
+                if (self->dim >= 3) {
```
ankith26, line 1215: Why is this check needed? I think in this path an error should be raised either way regardless of dim value

oddbookworm: Differentiate between vector2 and vector3. This also guarantees the old behavior of going to badinternalcall if somehow someone manages to get a vec2 with a z property

ankith26: The old code checked for 
```c
if (component >= self->dim) {
    PyErr_BadInternalCall();
}
```
Irrespective of the status of `value` but the changes do not preserve this behaviour.

oddbookworm: removed the check

Starbuck5 (approved):
Looks good to me, thanks!

src_c/math.c
ankith26: Your latest commit didn't address what I was trying to say.

Consider a call where `value != NULL && component >= self->dim`. In this case the old code correctly raises an error but the new code doesn't. The `PyErr_BadInternalCall` check should happen independently of the `NULL` check outside the block and above it.

MyreMylar (approved):
OK, LGTM. The code makes sense and I can verify that the new tests fail before this PR is compiled and pass afterwards (tested locally).

Issue #2757, opened 2024-03-19
Do we need to add docstrings to the pyi files?


Starbuck5:
I feel like I used to see docstrings in my dev environment (VS code), but I don't see them now.

All I see right now is the type information, for example:
![image](https://github.com/pygame-community/pygame-ce/assets/46412508/b3486aa8-6109-4160-be5d-bc7586f0f75f)

If I go into event.pyi as a test and replace the `...` with a docstring, it shows up:
```py
def event_name(type: int, /) -> str: ...
def set_blocked(type: Optional[_EventTypes], /) -> None:
    """blocked blocked blocked blocked..."""
def set_allowed(type: Optional[_EventTypes], /) -> None: ...
def get_blocked(type: _EventTypes, /) -> bool: ...
def set_grab(grab: bool, /) -> None: ...
```

If I hover over pygame.event.set_blocked I now see this very insightful information:
![image](https://github.com/pygame-community/pygame-ce/assets/46412508/6e77e8df-3673-4328-847b-df4470780cf5)

This seems like a big win if we can solve this in a nice manner.

ankith26:
I remember discussing this in previous issues (I'm not able to find any links RN)

Anyways, my take was that we can either
1) Make or use a tool that can copy the docs from code into the stubs
2) Update the tool being used by VS code to somehow pick up the docs from the implementation?

Basically, I don't want it to be a manually updated thing that adds developer burden

Starbuck5:
There are lots of possibilities here.

- Our runtime docstrings are generated from the docs sl/sg attribute
- Docstrings could be more detailed by taking in information from the rest of the docs
- Do most editors render docstrings as markdown?
- We can use markdown instead of reST in our docs, should we?
- Different function signatures could be annotated differently to explain the args that are actually being used.
- We could generate docs from stubs and have the stubs be the universal source of truth, or vice versa. We could put a universal source of truth for stubs and docs in the code and have everything be generated off that.
- We could try to unify the method overloading in the stubs with the method overloading expressed in the documentation.
- Manual might not be so bad, the stubs are already manually written.

ankith26:
> We could generate docs from stubs and have the stubs be the universal source of truth, or vice versa. We could put a universal source of truth for stubs and docs in the code and have everything be generated off that.

[...]

PR #3076, opened 2024-08-22
Fix average_color docs and stubs to match implementation

Starbuck5:
Reported by damus on discord. https://discord.com/channels/772505616680878080/772940667231928360/1275929591793582081


buildconfig/stubs/pygame/transform.pyi
```diff
@@ -42,7 +41,7 @@ def average_surfaces(
 ) -> Surface: ...
 def average_color(
     surface: Surface, rect: Optional[RectLike] = None, consider_alpha: bool = False
-) -> Color: ...
+) -> Tuple[int, int, int, int]: ...
```
ankith26, line 44: better to use `RGBATuple` from `pygame.typing` here

Starbuck5: Sure, I guess?

But looking at the existing uses of RGBATuple every single one of them is wrong. masks, shifts, and losses are not RGBA, they're values for bits.

Starbuck5: Ok I missed get_colorkey, that use is correct. Every other use is wrong.

Starbuck5: image

Honestly this seems worse to me. Instead of seeing what it actually is I only see this random name for it.

ankith26: Good catch with the shifts/losses/masks, that was definitely incorrect typing.

For this case I think the name conveys more information. I immediately know the return value is a tuple representation of a color that has all 4 components when I see this name.

damusss: yeah I agree that masks and company should not return RGBATuple as the docs state the values are clamped 0-255

Starbuck5: > yeah I agree that masks and company should not return RGBATuple as the docs state the values are clamped 0-255

That's not my problem with them, my problem with them is that they aren't RGBA Tuples!

> For this case I think the name conveys more information. I immediately know the return value is a tuple representation of a color that has all 4 components when I see this name.

I think it's unnecessary. If I see a 4 tuple coming from a function with "color" in the name I think most people already know what to expect. Additional information about what the return value means should be in the docstring, which should show up here, see #2757

I feel like this is overloading docstring information into the type system, which is not where it belongs, and just obscures the actual types being returned.

damusss: Since the module isn't released yet we could easily remove RGBATuple and replace it with a tuple of 4 ints. About the docstrings, we could make some automation script, but if we were to copy over the whole documentation paragraph there would be a big amount of duplicate code, so maybe we can start with the function summary, since that's what's currently exported bu C functions (example DOC_MATH_LERP).

Starbuck5: image

This is what it could look like.

damusss: That's something I remember "complaining" of not being there in 2023. How can we achieve it without duplicating the whole docs "codebase"?

Starbuck5: > Since the module isn't released yet we could easily remove RGBATuple and replace it with a tuple of 4 ints.

When I reviewed your PR I asked about stability implications! This is what I mean. Once this module is in use we could never remove or change any of the type members.

Starbuck5: > That's something I remember "complaining" of not being there in 2023. How can we achieve it without duplicating the whole docs "codebase"?

It would be such a nice usability win it would be really cool to get into people's hands.

Pick your poison. Would you rather generate stubs from docs, generate docs from stubs, generate docs and stubs from our own proprietary format, stitch docs into stubs at build time, ... ?

damusss: _but that time will only come with 2.5.2..._ I mean apart from rgba tuple the rest is pretty stable

Starbuck5: Ok I've decided to leave this returning "RGBATuple" in this PR, to advance the more important changes in this PR without drawing out this debate. I will plan on a follow up PR removing RGBATuple because I don't think it's how we should type our returns.

ankith26 (approved):
LGTM, thanks for the PR :+1: 

I agree with `RGBATuple` removal ideas being pushed to future issues/PRs

damusss (approved):
LGTM, thanks for the fix :)

PR #3078, opened 2024-08-22
Fix `(F)Rect.scale_by(_ip)` handling of the `scale_by` parameter

ankith26:
1. `scale_by` was incorrectly returning a valid object even when the underlying `scale_by_ip` operation was failing. This PR fixes this and also the memleak resulting from it.
2. The `scale_by` argument was keyword only so far, and the docs didn't make a mention of this. So I updated the implementation to handle a sequence scale if it is passed positionally

### DIFF
--- a/src_c/rect_impl.h
+++ b/src_c/rect_impl.h
@@ -1084,31 +1084,40 @@ static PyObject *
 RectExport_scalebyIp(RectObject *self, PyObject *args, PyObject *kwargs)
 {
     float factor_x, factor_y = 0;
-
-    static char *keywords[] = {"x", "y", NULL};
-
-    if (kwargs) {
-        PyObject *scale_by = PyDict_GetItemString(kwargs, "scale_by");
-        float temp_x, temp_y = 0;
-
-        if (scale_by) {
-            if (PyDict_Size(kwargs) > 1) {
-                return RAISE(PyExc_TypeError,
-                             "The 'scale_by' keyword cannot be combined with "
-                             "other arguments.");
-            }
-            if (!pg_TwoFloatsFromObj(scale_by, &temp_x, &temp_y)) {
-                return RAISE(PyExc_TypeError, "number pair expected");
-            }
-            PyDict_SetItemString(kwargs, "x", PyFloat_FromDouble(temp_x));
-            PyDict_SetItemString(kwargs, "y", PyFloat_FromDouble(temp_y));
-            PyDict_DelItemString(kwargs, "scale_by");
+    PyObject *scale_by =
+        kwargs ? PyDict_GetItemString(kwargs, "scale_by") : NULL;
+    if (scale_by) {
+        if (PyDict_Size(kwargs) > 1) {
+            return RAISE(PyExc_TypeError,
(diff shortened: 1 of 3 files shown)


src_c/rect_impl.h
```diff
@@ -1084,31 +1084,40 @@ static PyObject *
...
+                         "The 'scale_by' argument must be a number pair");
+        }
+    }
+    else {
+        static char *keywords[] = {"x", "y", NULL};
+        PyObject *arg_x, *arg_y = NULL;
+        if (!PyArg_ParseTupleAndKeywords(args, kwargs, "O|O", keywords, &arg_x,
+                                         &arg_y)) {
+            return NULL;
+        }
```
itzpr3d4t0r, line 1106: I mean if we fall into this 'else' block it's because `scale_by` is `NULL`, but that only happens where there are no kwargs so I guess we shouldn't be using `PyArg_ParseTupleAndKeywords` here.

ankith26: `scale_by` is NULL if `scale_by` is not in kwargs. `PyArg_ParseTupleAndKeywords` is still needed to handle the `x` and `y` arguments.

PS: While experimenting I did make a `METH_FASTCALL | METH_KEYWORDS` variant of this function, but ultimately decided against PRing it because it was bigger code that could be harder to review

src_c/rect_impl.h
```diff
@@ -1084,31 +1084,40 @@ static PyObject *
...
+        if (!pg_TwoFloatsFromObj(scale_by, &factor_x, &factor_y)) {
+            return RAISE(PyExc_TypeError,
+                         "The 'scale_by' argument must be a number pair");
+        }
+    }
+    else {
+        static char *keywords[] = {"x", "y", NULL};
+        PyObject *arg_x, *arg_y = NULL;
+        if (!PyArg_ParseTupleAndKeywords(args, kwargs, "O|O", keywords, &arg_x,
```
Matiiss, line 1103: There shouldn't be a case where passing in `x=5` suddenly scales the rect 5 times in both directions. That behavior just doesn't make sense.

ankith26: That is not behaviour decided by us. The goals of this PR are simple: to improve the code and extend the API to allow an argument to be passed positionally.

Changing behaviour is out of scope of this particular PR, though if enough people wanted it changed I suppose one could make a new PR

Matiiss: > That is not behaviour decided by us.

Yes, it is. It is bizarre to maintain such ridiculous design choices in the name of cross compatibility.
"A foolish consistency is the hobgoblin of little minds (..)"

> The goals of this PR are simple: to improve the code

While I feel like fixing this is very much in line with both improving the code and fixing argument handling (the title of the PR), I'm fine with this being fixed in another PR (along with raising an exception on negative values).
I will rename this PR to a more precise title though.

Matiiss (approved):
LGTM

damusss (approved):
LGTM, thanks :)

MyreMylar (approved):
OK, LGTM 👍

PR #3079, opened 2024-08-22
Add `Window.focused`

damusss:
stubs, docs, test included.

## Can it be done with events?
Yes. but you need to keep an extra variable and it's more annoying. If there is a simpler way, why not have it?

## Is this different from the other focus-related api?
Yes. Other functions such as `event.get_grab`, `mouse/key.get_focused` return True if the state is true for _any_ open window. This attribute only returns true for the current `Window` object, which is important if you need to redirect interactions to one window only (for example, `get_pos` returns the relative pos of the current window, so you need a way to decide which window can use the `get_pos` or `get_pressed` results. I have a real life example if needed)

## Why Window.focused?
I am only following the attribute convention of Window such as `mouse_grabbed`, `keyboard_grabbed`.

## Test code
```py
import pygame

pygame.init()
win = pygame.Window("focused test", (500, 500))

while True:
    for e in pygame.event.get():
        if e.type == pygame.WINDOWCLOSE and e.window == win:
            pygame.quit()
            raise SystemExit

    print(win.focused)
    win.flip()

    # play around with the window, focus other windows, check the terminal

```


ankith26 (approved):
LGTM, thanks :tada:

docs/reST/ref/window.rst
```diff
@@ -128,6 +128,18 @@
 
       .. versionadded:: 2.4.0
 
+   .. attribute:: focused
+
+      | :sl:`Get if the window is focused (**read-only**)`
+      | :sg:`focused -> bool`
+
+      Get if the window is currently focused. The same result can be achieved using
+      the ``WINDOWFOCUSGAINED`` and ``WINDOWFOCUSGLOST`` events.
```
Starbuck5, line 137: Window Focus Glost ? 😄

damusss: yeah it's the hidden and secret window event (omw to fix it)

PR #3084, opened 2024-08-27
Enhance `pygame.mouse.get_pressed` docs

damusss:
@pgvalle made me notice that the documentation for the function `pygame.mouse.get_pressed` could be improved to say something about the button constants. Since I was enhancing, I also removed some trailing whitespaces, added some quotes around pygame constants and reordered the notions inside the get_pressed function body (I made one note an actual .. note:: and converted 2 notes to description.)

I find it to be better, but it is subjective, so let me know.

### DIFF
--- a/docs/reST/ref/mouse.rst
+++ b/docs/reST/ref/mouse.rst
@@ -15,13 +15,13 @@ When the display mode is set, the event queue will start receiving mouse
 events. The mouse buttons generate ``pygame.MOUSEBUTTONDOWN`` and
 ``pygame.MOUSEBUTTONUP`` events when they are pressed and released. These
 events contain a button attribute representing which button was pressed. The
-mouse wheel will generate ``pygame.MOUSEBUTTONDOWN`` and 
-``pygame.MOUSEBUTTONUP`` events when rolled. The button will be set to 4 
-when the wheel is rolled up, and to button 5 when the wheel is rolled down. 
-Whenever the mouse is moved it generates a ``pygame.MOUSEMOTION`` event. The 
-mouse movement is broken into small and accurate motion events. As the mouse 
-is moving many motion events will be placed on the queue. Mouse motion events 
-that are not properly cleaned from the event queue are the primary reason the 
+mouse wheel will generate ``pygame.MOUSEBUTTONDOWN`` and
+``pygame.MOUSEBUTTONUP`` events when rolled. The button will be set to 4
+when the wheel is rolled up, and to button 5 when the wheel is rolled down.
+Whenever the mouse is moved it generates a ``pygame.MOUSEMOTION`` event. The
+mouse movement is broken into small and accurate motion events. As the mouse
+is moving many motion events will be placed on the queue. Mouse motion events
+that are not properly cleaned from the event queue are the primary reason the
 event queue fills up.
 
 If the mouse cursor is hidden, and input is grabbed to the current display the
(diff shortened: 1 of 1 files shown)


docs/reST/ref/mouse.rst
```diff
@@ -86,23 +86,30 @@ scroll, such as ``which`` (it will tell you what exact mouse device trigger the
...
    of the call.
 
-   Note, to get all of the mouse events it is better to use either 
-   ``pygame.event.wait()`` or ``pygame.event.get()`` and check all of those 
+   To get all of the mouse events it is better to use either
+   ``pygame.event.wait()`` or ``pygame.event.get()`` and check all of those
    events to see if they are ``MOUSEBUTTONDOWN``, ``MOUSEBUTTONUP``, or
-   ``MOUSEMOTION``.
+   ``MOUSEMOTION``. Remember to call ``pygame.event.get()`` before this function,
+   otherwise it will not work as expected.
```
ankith26, line 93: Technically, only a pump is required here I think.

damusss: > Yes, pump updates the information, but this is a technicality and if the queue gets full, pump will be giving a not up-to-date information, I think. So maybe something like
> 
> ```
> Remember to call ``pygame.event.get()`` or ``pygame.event.pump()`` before this function, otherwise it will not work as expected.
> ```
> 
> ?

Yeah I'll modify it to this

docs/reST/ref/mouse.rst
```diff
@@ -86,23 +86,30 @@ scroll, such as ``which`` (it will tell you what exact mouse device trigger the
...
-   can be emitted.
+   .. note:: To use the pygame button constants such as ``pygame.BUTTON_LEFT``
+      with this function, 1 must be subtracted from the constant to obtain the
+      correct index of the tuple.
 
-   Note, remember to call ``pygame.event.get()`` before this function.
-   Otherwise it will not work as expected.
+      ::
+
+         if pygame.mouse.get_pressed()[pygame.BUTTON_LEFT-1]:
```
Starbuck5, line 106: I don't see why you would use a constant for this.

The documentation specifies what each index of the returned tuple means, as far as I'm concerned it's a complete coincidence this works.

damusss: I mean, you use key constants _both_ in key events and in the result of `get_pressed()`, but this is not the case for the mouse, which is a bit strange. I used the BUTTON_ index in my own code, it's quite nice, but you need to somehow get to know the relation between constant and index.

Starbuck5: I don't think it's nice to use constants where they are not meant to be used. If pygame.BUTTON_LEFT changes to 36423 tomorrow it shouldn't mean your code crashes. This is not completely theoretical, pygame 1 -> 2 changed some constants (it would've been really complex not to) and this negatively impacted people using constants this way.

I hadn't realized that consistent structure and inconsistent convention between key.get_pressed and mouse.get_pressed.

damusss: Should I then discourage the use of the BUTTON_ constants? Also you mentioned they are not documented anywhere, they should probably be documented in event.rst as it's the only place you are supposed to use them.

ankith26: I think the constants should only be used for testing equality in the corresponding event attribute, not here for reasons starbuck pointed out.

If this block of example is removed then this PR is in a mergeable state IMO

damusss: alright, done :)

ankith26 (approved):

bilhox (approved):
Checked this PR a while ago, but forgot to leave feedbacks. LGTM now, thanks for the PR.

PR #3086, opened 2024-08-30
Make `pygame.typing.PathLike` internal/private

### LINKED ISSUE #3085: Proposal: Remove `pygame.typing.PathLike` type from the API
From the discussion in #3080, I think it may be a good idea to remove [`typing.PathLike`](https://github.com/pygame-community/pygame-ce/blob/e01a44e1481af73ccd1fca0dc0ab78e9257cd47e/src_py/typing.py#L32) union type alias.
Here are some arguments for removing it *or* making it internal:
- The pygame typing API has not been released yet (2.5.2); it is easier to add `PathLike` back in if it is useful than it is to remove `PathLike` if it is unnecessary
- Only used in one place in pygame other than in `typing.FileLike`: inside [`rwobject.pyi`](https://github.com/pygame-community/pygame-ce/blob/e01a44e1481af73ccd1fca0dc0ab78e9257cd47e/buildconfig/stubs/pygame/rwobject.pyi) as the parameter type for `encode_string()` and `encode_file_path()`. Though there could be more undiscovered uses
- This specific union is too general; not very specific to pygame (unlike `FileLike`)
  - Was considered for addition to standard typing module in [PEP 519](https://peps.python.org/pep-0519/#provide-specific-type-hinting-support), but ultimately not added
  - Currently exists as one of the stable `_typeshed` types: [`StrOrBytesPath`](https://github.com/python/typeshed/blob/f58dac1d62d33bb2255b762783d06463c40f5065/stdlib/_typeshed/__init__.pyi#L176)
- Slight name conflict with `os.PathLike`, which is used in the definition of `pygame.typing.PathLike` (confusing?)

___
Also proposing a separate but related *internal* change: Still use `os.PathLike` for python <=3.8 (where it is not generic), by defining the other type in the version conditional, as so:
```python
if sys.version_info >= (3, 9):
    PathLike = Union[str, bytes, _PathProtocol[str], _PathProtocol[bytes]]
else:
    PathLike = Union[str, bytes, _PathProtocol]
```
This is so we don't have to define a custom internal protocol class with the same behavior as `os.PathLike`.

aatle:
Closes #3085

Renames all occurrences from `PathLike` to `_PathLike`, and removes it from docs and `__all__` list.

### DIFF
--- a/docs/reST/ref/typing.rst
+++ b/docs/reST/ref/typing.rst
@@ -16,9 +16,15 @@ type aliases for proper typehint annotations.
 
    .. data:: FileLike
 
-      An object representing a file. This includes any path-like objects such as
-      ``str`` or ``pathlib.Path()`` and file buffers (``IO`` of strings or bytes)
-      such as the return value of ``open()``.
+      An object representing a file. This includes both path-like
+      objects and file-like objects, i.e.:
+
+        * ``"my/string/path.txt"``
+        * ``open("my/file/path.txt")``
+        * ``pathlib.Path("my/pathlib/path.txt")``
+        * ``io.BytesIO(b"my data: \x00\x01")``
+        * ``b"my/bytes/path.txt"``
+        * Any object implementing the path protocol or file protocol.
 
    .. data:: SequenceLike
 
--- a/docs/reST/ref/typing.rst
+++ b/docs/reST/ref/typing.rst
@@ -14,20 +14,11 @@ A lot of pygame functions and methods allow the user to provide different types
 for the same value like colors or coordinates. This module exports the most common
 type aliases for proper typehint annotations.
 
-   .. data:: PathLike
-
-      An object representing a file path, i.e.:
-
-        * ``"my/string/path.txt"``
-        * ``pathlib.Path("my/pathlib/path.txt")``
-        * ``b"my/bytes/path.txt"``
-        * Any object implementing the path protocol (with a ``__fspath__`` magic method)
-
    .. data:: FileLike
 
-      An object representing a file. Same as :mod:`pygame.typing.PathLike` with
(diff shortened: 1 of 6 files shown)


docs/reST/ref/typing.rst
```diff
@@ -14,20 +14,11 @@ A lot of pygame functions and methods allow the user to provide different types
...
-        * ``pathlib.Path("my/pathlib/path.txt")``
-        * ``b"my/bytes/path.txt"``
-        * Any object implementing the path protocol (with a ``__fspath__`` magic method)
-
    .. data:: FileLike
 
-      An object representing a file. Same as :mod:`pygame.typing.PathLike` with
-      the addition of file buffers (``IO`` of strings or bytes) such as the
-      return value of ``open()``.
+      An object representing a file. This includes any path-like objects such as
```
damusss, line 19: I think it would be best if you used the same docs style as PathLike for FileLike, listing each possible type 
- "mypath"
- pathlib.Path
- b"mybytespath"
- anything supporting the file protocol 
- file buffers (IO of str and bytes) such as the return value of open

something like this (this is just for the idea, the list item text should be better, I'm on phone now)

damusss (changes requested):
This isn't particularly important, the rest looks good!

MyreMylar (approved):
This seems reasonable I agree that the the `os.PathLike` vs `pygame.typing.PathLike` is likely to be confusing.

docs/reST/ref/typing.rst
aatle: Okay, I listed 6 possible types. Let me know if you'd like any changes to the items I put in.

ankith26 (approved):

damusss (approved):
LGTM, Thank you!

PR #3087, opened 2024-08-30
Fix component bounds checking

ankith26:
minor fix to address the unresolved review comment I had in #3069

### DIFF
--- a/src_c/math.c
+++ b/src_c/math.c
@@ -1199,6 +1199,10 @@ static PyMappingMethods vector_as_mapping = {
 static int
 vector_set_component(pgVector *self, PyObject *value, int component)
 {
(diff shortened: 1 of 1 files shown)


aatle:
Question, would it be any better to manually inline `vector_set_component` at its 3 places to get rid of bad internal call possibilities and the error switch statement?

Starbuck5:
Yeah, that could potentially be more performant. Who knows, maybe the compiler is smart enough to do the inlining and remove conditionals anyway.

But changing the way that works can happen another time, it shouldn’t stop this PR from happening imo.

damusss (approved):
Makes sense 👍

zoldalma999 (approved):
LGTM, Thanks!

PR #3089, opened 2024-08-31
Implementation of ``pygame.(F)Rect.rel_center``

bilhox:
Implementation of #1242 according to MyreMylar name proposal.

While I'm bringing this feature to a potential release, what's your opinion on ``rel_centerx`` and ``rel_centery`` ?


damusss:
I'm not sure about the x and y variants, as
`(r.w/2+N, r.h/2+M)` is shorter than `(r.rel_centerx+N, r.rel_centery+M)` (where N and M are arbitrary numbers not to make the expression useless), but I like `rel_center`, it makes sense.

src_c/rect_impl.h
```diff
@@ -2804,6 +2814,33 @@ RectExport_getcenter(RectObject *self, void *closure)
                                   self->r.y + (self->r.h / 2));
 }
 
+/*center*/
```
MightyJosip, line 2817: relcenter

damusss: you brought up a point, should it be `rel_center` or `relcenter`

aatle:
`rel_centerx` and `rel_centery` are probably less readable than `width/2`, and less useful compared to `centerx`.

It seems to me like most uses would be related to `Surface`s.
I think more concrete use cases for this feature need to be listed.

bilhox:
> `rel_centerx` and `rel_centery` are probably less readable than `width/2`, and less useful compared to `centerx`.
> 
> It seems to me like most uses would be related to `Surface`s. I think more concrete use cases for this feature need to be listed.

With argument you gave with damus, indeed ``rel_centerx`` and ``rel_centery`` is not needed.
Except that, I think the person who opened the issue gave a fairly good usecase, and I personally support the feature for the same reason.

aatle:
I agree that the feature is logical and should probably be implemented, but I wonder if there are more use cases than the one given:
> I often find myself evaluating (rect.w / 2, rect.h / 2) to find the middle of a rect, relative to itself. This is useful when drawing a circle onto a surface for a sprite, for example.

If I understand this case correctly, one could use `surface.get_rect().center` instead, right? For rects from surfaces, the position is (0,0) since only size is relevant, so `rel_center` is not needed.
I personally don't see any use cases in my own project codebase, but that's probably because I use `Vector2` for this stuff.

docs/reST/ref/rect.rst
```diff
@@ -92,6 +92,25 @@
...
+
+         > my_rect = pygame.Rect(0, 0, 2, 2)
+         > my_rect.rel_center
+         > (1, 1)
+         > my_rect.rel_center = (128, 128)
+         > my_rect.rel_center, my_rect.size
+         > ((128, 128), (256, 256))
+      
+      Beware of non integer relative centers ! Using a Rect instead of FRect will round down
+      the values of the returned ``Coordinate``.
```
MyreMylar, line 112:
```suggestion
   .. versionadded:: 2.5.2
      ``rel_center`` added to Rect / FRect. This will return you a ``Point`` of
      the center relative to the topleft of the Rect. Setting a ``Point`` to it will 
      modify the size of the rect to 2 times the ``Point`` used. Below you can find a 
      code example of how it should work :

      .. code-block:: python

         > my_rect = pygame.Rect(0, 0, 2, 2)
         > my_rect.rel_center
         > (1, 1)
         > my_rect.rel_center = (128, 128)
         > my_rect.rel_center, my_rect.size
         > ((128, 128), (256, 256))
      
      Beware of non integer relative centers ! Using a Rect instead of FRect will round down
      the values of the returned ``Point``.
```

buildconfig/stubs/pygame/rect.pyi
```diff
@@ -93,6 +93,10 @@ class _GenericRect(Collection[_N]):
     @center.setter
     def center(self, value: Coordinate) -> None: ...
     @property
+    def rel_center(self) -> Tuple[_N, _N]: ...
+    @rel_center.setter
+    def rel_center(self, value: Coordinate) -> None: ...
```
MyreMylar, line 98:
```suggestion
    def rel_center(self, value: Point) -> None: ...
```

src_c/rect_impl.h
MyreMylar: I kind of hate it, but probably `relcenter` for consistency with the other pre-existing rect attributes.

MyreMylar (changes requested):
Can we rename it `relcenter` ? Then I think we can get this in.

MyreMylar (approved):

buildconfig/stubs/pygame/rect.pyi
```diff
@@ -7,6 +7,7 @@ from typing import (
     Union,
     overload,
     Optional,
+    Tuple,
```
aatle, line 10: Use `tuple` instead of `Tuple`.

docs/reST/ref/rect.rst
```diff
@@ -93,6 +93,25 @@
       and ``__new__()`` is assumed to take no arguments. So these methods should be
       overridden if any extra attributes need to be copied.
 
+   .. versionadded:: 2.5.2
```
MyreMylar:
```suggestion
   .. versionadded:: 2.5.6
```

aatle (approved):
LGTM

Issue #3092, opened 2024-09-06
SysFont on Windows clobbers base fonts with alternate styled versions


rmccampbell:
**Environment:**

```
Platform:               Windows-11-10.0.22631-SP0
System:                 Windows
System Version:         10.0.22631
Processor:              Intel64 Family 6 Model 154 Stepping 3, GenuineIntel     SSE2: Yes       AVX2: Yes    NEON: No
Architecture:           Bits: 64bit     Linkage: WindowsPE

Python:                 CPython 3.12.5 (tags/v3.12.5:ff3bc82, Aug  6 2024, 20:45:27) [MSC v.1940 64 bit (AMD64)]
pygame version:         2.5.1
SDL versions:           Linked: 2.30.6  Compiled: 2.30.6
SDL Mixer versions:     Linked: 2.8.0   Compiled: 2.8.0
SDL Font versions:      Linked: 2.22.0  Compiled: 2.22.0
SDL Image versions:     Linked: 2.8.2   Compiled: 2.8.2
Freetype versions:      Linked: 2.11.1  Compiled: 2.11.1

Display Driver:         windows
Mixer Driver:           wasapi
```
**Current behavior:**

Fonts which have multiple variants (see [mods](https://github.com/pygame/pygame/blob/88475d861931ce515baba1115261a980ce332ece/src_py/sysfont.py#L104)) are normalized on Windows by stripping certain keywords, and if multiple variants have the same name after normalization, only the last version will be kept. This may clobber the base version of fonts. For example, on my computer requesting the font "Arial" results in the actual font "Arial Narrow". Interestingly, this does *not* strip the word "Regular", so you end up with some odd cases such as "lucidasans" mapping to Lucida Sans Unicode or "lucidafax" to Lucida Fax Demibold while "lucidasansregular" and "lucidafaxregular" map to the plain fonts.

**Expected behavior:**

1. The regular fonts should be preserved. I.e. "Arial" should be "Arial"
2. Alternate font style variants should be selectable individually. So "Arial Narrow" should give the narrow font.

To preserve some backwards compatibility, I think the best solution is to register fonts with multiple aliases: the full name which only strips bold/italic, and the stripped name which does the complete normalization. If there are collisions between a full name and a stripped name then the full name font should always be preferred. There also should probably be some special casing for "Regular".

**Test code**

```
import pygame as pg
pg.init()
s = pg.display.set_mode((640, 480))
s.fill((255, 255, 255))

[...]

Issue #2760, opened 2024-03-20
***DISCUSSION*** pygame-ce 3.0 API changes


oddbookworm:
This is a tracker and discussion forum for API changes that we want to make when we are working on 3.x.x now that SDL3 is nearing prerelease (and thus release in the not-so-distant future). Want a piece of API gone or modified? Or maybe new API? Suggest it!

oddbookworm:
Old scrap is deprecated and *new* scrap only supports text. SDL3 supports a range of MIME types. IMO we should wrap the SDL3 clipboard API to replace scrap completely with fancy new code that isn't as bad as *old* scrap, and is more reliable than I remember old scrap being

bilhox:
With pygame 3.0, I believe dropping the support of queuing a ``pygame.MOUSEBUTTONDOWN`` when the mouse wheel is rolled would be good. It's almost not used / it should not be used, because we have ``pygame.MOUSEWHEEL`` event now. I believe according to oddbookworm it was a feature in SDL1, because no ``pygame.MOUSEWHEEL`` event.

Starbuck5:
Things off the top of my head:

Lots of deprecated could stuff could get removed.
We need to get weird of that weird font thing where the default font's size is shrunk 66% in some cases
Sysfont should be changed to be "font weight" aware, so it can give normal fonts instead of the first thing it sees that isn't bold or italic.
Mouse and time apis should return floats.

robertpfeiffer:
We should really think long and hard about locking and maybe deprecate a bunch of manual locking/unlocking in favour of enforcing with-statements.

itzpr3d4t0r:
I think we should also drop returning rects/ rects lists in functions like blit/blits since it was a thing only as an optimization for display.update(), but as @MyreMylar found out, it's no longer relevant and isn't applicable in most if not all cases anymore.

itzpr3d4t0r:
Possibly we could also make all alphablit algorithms no longer behave as SDL1 (which to be fair it's kinda weird). This should be investigated further but i suspect it could allow us to further optimize them. This could slightly change visual results but not by a huge margin, or at least not in common usecases.

robertpfeiffer:
Consider a "quirks mode" or "pygame 1.9 compatibility flag" to support old textbooks.

robertpfeiffer:
We could consider deprecating the whole sprite system and write a new one.

snowfruit:
[...]

PR #3093, opened 2024-09-07
Improve SysFont name resolution on Windows

rmccampbell:
This solves the problem in #3092 with font style variants being stripped and an arbitrary style being picked. It also handles some additional cases such as "Regular" and "Oblique" in font names, with Oblique being treated the same as Italic.

I tried to maintain backwards compatibility by generating two name variants for each font: a full name that only strips Regular, Bold, and Italic/Oblique, which should be unambiguous; and a legacy name that strips the somewhat arbitrary set of suffixes: "demibold", "narrow", "light", "unicode", "bt", and "mt". The legacy name doesn't strip Regular or Oblique. If a full name and legacy name collide then the full name will be used, and for multiple legacy names only the first one will be used.

Note that I didn't attempt to create any explicit precedence between legacy names, so they may still resolve to an unexpected style if they don't match the full regular font name.

Some other slight changes:
- I moved the "(TrueType)" suffix removal before the "&" splitting, since it is only at the end of each registry name.
- I removed a mysterious ";" that I found after this suffix in some fonts which prevented it from being stripped.
- I excluded "Demi Bold" from Bold, since some of these also have a true Bold version.

### DIFF
--- a/src_py/sysfont.py
+++ b/src_py/sysfont.py
@@ -43,11 +43,12 @@ def _simplename(name):
     return "".join(c.lower() for c in name if c.isalnum())
 
 
-def _addfont(name, bold, italic, font, fontdict):
+def _addfont(name, bold, italic, font, fontdict, replace=False):
     """insert a font and style into the font dictionary"""
     if name not in fontdict:
         fontdict[name] = {}
-    fontdict[name][bold, italic] = font
+    if replace or (bold, italic) not in fontdict[name]:
+        fontdict[name][bold, italic] = font
 
 
 def initsysfonts_win32():
@@ -83,44 +84,65 @@ def initsysfonts_win32():
                 if not dirname(font):
                     font = join(fontdir, font)
 
-                # Some are named A & B, both names should be processed separately
-                # Ex: the main Cambria file is marked as "Cambria & Cambria Math"
-                for name in name.split("&"):
-                    if os.path.exists(font):  # check if the file actually exists
-                        _parse_font_entry_win(name, font, fonts)
+                if os.path.exists(font):  # check if the file actually exists
+                    _parse_font_entry_win(name, font, fonts)
 
     return fonts
 
 
 def _parse_font_entry_win(name, font, fonts):
     """
-    Parse out a simpler name and the font style from the initial file name.
+    Parse out simpler name(s) and the font style from the initial file name.
+
+    This only does minimal processing of the font name, however it also creates
+    aliases with certain keywords stripped for backwards compatibility, if they
+    don't collide with another full font name.
 
     :param name: The font name
     :param font: The font file path
     :param fonts: The pygame font dictionary
-
-    :return: Tuple of (bold, italic, name)
     """
+    # Strip off TrueType suffix and possible semicolon
     true_type_suffix = "(TrueType)"
-    mods = ("demibold", "narrow", "light", "unicode", "bt", "mt")
+    name = name.rstrip(";")
     if name.endswith(true_type_suffix):
-        name = name.rstrip(true_type_suffix).rstrip()
-    name = name.lower().split()
-    bold = italic = False
-    for mod in mods:
-        if mod in name:
-            name.remove(mod)
-    if "bold" in name:
-        name.remove("bold")
-        bold = True
-    if "italic" in name:
-        name.remove("italic")
-        italic = True
-    name = "".join(name)
-    name = _simplename(name)
+        name = name[:-len(true_type_suffix)].rstrip()
 
-    _addfont(name, bold, italic, font, fonts)
+    # Some fonts are named A & B, both names should be processed separately
+    # Ex: the main Cambria file is marked as "Cambria & Cambria Math"
+    for subname in name.split(" & "):
+        parts = subname.lower().split()
+        # Common handling for legacy and non-legacy
+        bold = italic = oblique = False
+        # Ignore "Demi Bold"
+        if "bold" in parts and "demi" not in parts:
+            parts.remove("bold")
+            bold = True
+        if "italic" in parts:
+            parts.remove("italic")
+            italic = True
+        # Copy for legacy name
+        legacy_parts = parts.copy()
+
+        # New name handling
+        if "regular" in parts:
+            parts.remove("regular")
+        if "oblique" in parts:
+            parts.remove("oblique")
+            oblique = True
+        # The full name takes presedence over legacy stripped names
+        full_name = _simplename("".join(parts))
+        _addfont(full_name, bold, italic or oblique, font, fonts, replace=True)
+
+        # Legacy name handling
+        # Strip common style variants/suffixes for backwards compatibility
+        legacy_mods = {"demibold", "narrow", "light", "unicode", "bt", "mt"}
+        legacy_parts = [p for p in legacy_parts if p not in legacy_mods]
+        legacy_name = _simplename("".join(legacy_parts))
+        if legacy_name != full_name:
+            # Keep only the first matching legacy name, if it doesn't collide
+            # Note that this may have an arbitrary style
+            _addfont(legacy_name, bold, italic, font, fonts, replace=False)
 
 
 def _parse_font_entry_darwin(name, filepath, fonts):


MyreMylar:
Is it possible to create a test that fails on our current Windows CI testing, but then passes after this PR? I.e. it tries to load a default font with a specific style and currently returns the wrong one.

I think that would help ensure that we don't backslide in this area and improve the 'what is this for?' aspect of this PR.

MyreMylar:
I'm not as bothered about exact historical backwards compatibility here because this function has always been 'fuzzy' in that it is entirely environment dependant what you get back when using SysFont. Two different versions of windows/Mac/Linux don't have the same set of system fonts. 

If you wanted fine precision in your font choice in pygame you were already shipping and loading your own precisely selected font.

MyreMylar (commented):
See above comments about test. I think we should also add a note to the documentation here if we are going to change the behaviour of this function.

rmccampbell:
I added a test case which tries to create "Arial" and "Arial Narrow" sysfonts and checks their name/styles. Though I'm not sure if this is reliable enough across all Windows systems; maybe the test should be ignored/skipped if the expected font files don't exist? And IDK if hard coding the name/style_name is a good idea.

rmccampbell:
Hmm Arial Narrow isn't listed on the official list of Windows fonts: https://learn.microsoft.com/en-us/typography/fonts/windows_10_font_list
So maybe we should skip testing that and instead just test that Arial has Regular style. WDYT?

rmccampbell:
Alternatively, "Calibri Light" is also affected and on that list so we can test that instead.

rmccampbell:
As for documenting the change: if it's documented then we should also probably address the related Unix issue too (https://github.com/pygame-community/pygame-ce/issues/3092#issuecomment-2335042370). But I'm not sure quite how we would want to handle that.

Starbuck5:
Thank you for your PR, I've now looked it over (took me a bit)

I wanted a way to generate comparisons, so I wrote this script that reaches into the sysfont internals and outputs a json serialized dictionary.
```py
import json

import pygame

pygame.init()

f = pygame.font.SysFont("Arial", 32)

adjusted = {name: [[list(designation), path] for designation, path in content.items()] for name, content in pygame.sysfont.Sysfonts.items()}

adjusted_sorted = dict()
for key in sorted(list(adjusted.keys())):
   adjusted_sorted[key] = adjusted[key]

print(json.dumps(adjusted_sorted))
```

I then used VS code to format it and quickdiff.net to compare the two line by line. Here is the comparison: https://quickdiff.net/?unique_id=D8C69C31-12B1-0700-8F27-1554E0E065AA (will expire after 14 days)


Going through the changes there are some that are good, and some that are questionable:
- There are lots of new "mt" fonts in the system that are identical to other fonts
- berlinsans went from "berlinsansfbdemi" to "berlinsansfbdemibold". (see thoughts about font weight below)
- lots of new "light" fonts in the system (ditto)

On the bright side I also notice:
- Lots of things are now normal instead of light variants on their fonts
- Got rid of a weird "truetype" suffix on bizudpgothic

On another note, I'd like to think about approach. The sysfont api allows a way of requesting bold and a way of requesting italic. Bold is a font "weight", see https://developer.mozilla.org/en-US/docs/Web/CSS/font-weight#common_weight_name_mapping. So bold = font weight 700, black = font weight 900. Do we really want Arial Black to be seen in the system as its own font. If we do that, why not give Arial Bold it's own font entry too? I think the system should be more font weight aware, perhaps even exposing that as an API to the user in the future. A thought experiment: if this runs on a system that has Arial Demi Bold (font weight 600) but no normal bold and the user asks for "Arial" and bold, should the system give them demi bold as a close enough?

My suggestion is for you to split some of the smaller changes you've made here into another smaller PR, that will be easier to get merged and into the codebase. Then this PR can be more focused on the light -> normal issue.

rmccampbell:
All of those changes were intentional. The "MT" is part of the font's name, as it is listed in MS Word and recognized in a web browser, and we don't remove other similar suffixes like "MS" or "FB". Ideally we would only have one version of those font names to avoid confusion, but I decided to keep the existing normalized names where possible to avoid breaking existing users. As for "demibold" and "light", yes these are font weights and not separate font families, but since they can't otherwise be selected by sysfont I think it makes most sense to treat them as independent. (Interestingly, Word does list it as "Berlin Sans FB Demi". But the font's name in the registry contains the two words "Demi Bold", whereas other fonts have a single word "Demibold" or "Semibold", and it seems that these should be treated identically and different from plain "Bold".)

Starbuck5:
Alright, I'm receptive to the MT change with your reasoning. From googling MT means monotype. MS presumably means Microsoft. Do you know what FB means?

> As for "demibold" and "light", yes these are font weights and not separate font families, but since they can't otherwise be selected by sysfont I think it makes most sense to treat them as independent

Well, we could change that in sysfont. I don't like a situation where normal and bold are shunted into one bucket and everything else is shunted into another bucket with no plan to remedy the disparity.

rmccampbell:
Apparently FB is a company Font Bureau: https://en.wikipedia.org/wiki/Font_Bureau.

As for the different weights, yeah but it would probably require a much more extensive redesign which I think is out of scope here. Also both MS Word and Chrome seem to recognize most of the font weight variations as separate fonts (although Chrome for some reason doesn't seem to recognize any "Semilight" fonts; Light, Demibold, etc. are mostly recognized). I think making them available separately doesn't necessarily preclude also having them as weight variants of the base font in the future.

Starbuck5:
> I think making them available separately doesn't necessarily preclude also having them as weight variants of the base font in the future.

I don't want to start adding something to our API surface if we don't plan on offering it in the future.

rmccampbell:
I mean we could offer both. There's no reason we couldn't have e.g. Calibri Light and Calibri (font_weight= 50%) or whatever map to the same thing.

Starbuck5:
> I mean we could offer both. There's no reason we couldn't have e.g. Calibri Light and Calibri (font_weight= 50%) or whatever map to the same thing.

Yes, we could. But I see this as a paradigm altering change to sysfont, but it doesn't have a consistent plan for full realization (what happens with bold?), and more importantly if it shifts how things work on Windows, what about Mac and Linux?

I'm not comfortable approving this PR in this state because I'm unsure about some of the changes, and I feel that it will make sysfont less consistent between platforms.

rmccampbell:
I mentioned in the bug that Linux probably also needs a similar change, due to similar collision issues with multiple font styles. Regarding platform consistency however, the normalization done on Windows seems much more arbitrary and ad-hoc then on the other platforms and isn't really consistent at all.

An alternate strategy might be instead of keeping all font variants, we attempt to only keep the "regular" variant of each font. But this seems more iffy, as 1) there's no obviously complete set of what should be treated as "weights" or variants verses separate fonts: what about Black, Medium, Thin, Hairline, Condensed, Book, Roman... (all labels l found in the "style" field of fonts in Linux); 2) if we did this more consistently in Windows it would remove fonts that are currently available, including some commonly known ones like Arial Black; and 3) what would we do if there is no "regular" version of a family? Just fall back to a random style like we do now?

Starbuck5:
Well, I think there's potential to separate out the "font discovery" and "font name parsing" components so each platform feeds into the same font name parsing machinery.

I've also floated the idea of making the underlying system font weight aware, and then having the existing sysfont API select on top of that. https://github.com/pygame-community/pygame-ce/issues/2760#issuecomment-2016724409. But I was thinking this would be a pygame-ce 3.0 thing to minimize compatibility worries.

> But this seems more iffy: 1) there's no obviously complete set of what should be treated as "weights" or variants verses separate fonts; what about Black, Medium, Thin, Hairline, Condensed, Book, Roman... (all labels l found in the font "style" fields in Linux);

There's no complete set of what we should do for any of this stuff!

> 2) if we did this more consistently in Windows it would remove fonts that are currently available, including some commonly known ones like Arial Black; 

Your PR right now eliminates fonts: `berlinsansfbdemi`, `bizudpgothictruetype`, `bizudpminchomediumtruetype`. Maybe Arial Black is more common, but I've never seen anyone use it in sysfont. I've seen people use arial, comic sans, times new roman, some of the chinese/japanese ones for the extended character set.

> 3) what would we do if there is no "regular" version of a family? Just fall back to a random style like we do now?

I guess so, or not give the user it entirely?

rmccampbell:
>There's no complete set of what we should do for any of this stuff!

Which seems to me to point in the direction of doing as little as possible and leaving it more up to the user. Someone could even implement their own font weight system if they want, if we provided all the font variants.

>Your PR right now eliminates fonts: berlinsansfbdemi, bizudpgothictruetype, bizudpminchomediumtruetype.

Yeah it does eliminate a few names, but only very obscure and mostly broken cases, and they are still available under a different name.

It's worth pointing out that in the current status quo, Arial itself (probably the most common font) gives the wrong font, at least when Arial Narrow is installed (I'm guessing that comes bundled with MS Office or some other MS software, not sure). So any change that fixes that would be at least a good start.

Starbuck5:
We will revisit this for pygame-ce 3.0. So I will mark this PR as a draft.

PR #3098, opened 2024-09-11
Fix test_inflate_ip__larger.

### LINKED ISSUE #3096: Incorrect tests & missing test in ``rect_test``
I'm opening this issue as I don't want to work on it for the moment.

I have noticed many errors when I tried to implement ``rel_center`` :  
- ``test_inflate_ip__larger`` does not check what it is supposed to check (duplicate code of ``test_inflate_ip__smaller``)
- ``test_inflate_ip__smaller``'s docstring is incorrect
- each attributes are not checked with literal values if they have the right values. You might think calculating it based on other attributes garantie that the attributes return the right value but this shouldn't be the case ! Not for tests at least ! This is why I would like to see a test that checks if the attributes return the right value based on literal values.

avpai-dinosaur:
This test case was not checking what it was supposed to check (duplicate code of test_inflate_ip__smaller).

Fixes #3096

### DIFF
--- a/test/rect_test.py
+++ b/test/rect_test.py
@@ -801,18 +801,18 @@ class RectTypeTest(unittest.TestCase):
         """The inflate_ip method inflates around the center of the rectangle"""
         r = Rect(2, 4, 6, 8)
         r2 = Rect(r)
-        r2.inflate_ip(-4, -6)
+        r2.inflate_ip(4, 6)
 
         self.assertEqual(r.center, r2.center)
-        self.assertEqual(r.left + 2, r2.left)
-        self.assertEqual(r.top + 3, r2.top)
-        self.assertEqual(r.right - 2, r2.right)
-        self.assertEqual(r.bottom - 3, r2.bottom)
-        self.assertEqual(r.width - 4, r2.width)
-        self.assertEqual(r.height - 6, r2.height)
+        self.assertEqual(r.left - 2, r2.left)
+        self.assertEqual(r.top - 3, r2.top)
+        self.assertEqual(r.right + 2, r2.right)
+        self.assertEqual(r.bottom + 3, r2.bottom)
+        self.assertEqual(r.width + 4, r2.width)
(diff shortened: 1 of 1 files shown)


test/rect_test.py
```diff
@@ -801,18 +801,18 @@ def test_inflate_ip__larger(self):
...
+        self.assertEqual(r.left - 2, r2.left)
+        self.assertEqual(r.top - 3, r2.top)
+        self.assertEqual(r.right + 2, r2.right)
+        self.assertEqual(r.bottom + 3, r2.bottom)
+        self.assertEqual(r.width + 4, r2.width)
+        self.assertEqual(r.height + 6, r2.height)
 
     def test_inflate_ip__smaller(self):
-        """The inflate method inflates around the center of the rectangle"""
+        """The inflate_ip method inflates around the center of the rectangle"""
```
bilhox, line 815: While you are editing the docstring, it's better if you actually say what does the test do brievly. Same for ``test_inflate_ip__larger``

bilhox (changes requested):
LGTM, thanks for the contribution ! I left a comment for you.

test/rect_test.py
avpai-dinosaur: Took a stab at a more descriptive docstring, lmk what you think

bilhox (approved):
LGTM 🥳

gresm (approved):
LGTM. The docstrings for tests of ``pygame.Rect.scale_by`` could also benefit from updating them, but it's probably out of scope of this PR.

bilhox:
> LGTM. The docstrings for tests of `pygame.Rect.scale_by` could also benefit from updating them, but it's probably out of scope of this PR.

Indeed I can see that for certain tests. I think it's something a bit linked to the topic of this PR. So I would suggest a change there as gresm mentionned it. Otherwise it will not be fixed in a near future.

MyreMylar (approved):
LGTM 👍

PR #3100, opened 2024-09-15
Bump build dependencies

ankith26:
Simple PR to build a couple of build dependencies.

Also gets us using numpy on more of our builds since numpy recently added support for cpython 3.13 and pypy 3.10


Starbuck5 (approved):
Seems legit!

zoldalma999 (approved):
Good stuff DependaBo... wait its Ankith!

PR #3102, opened 2024-09-16
Remove RGBATuple

Starbuck5:
My contention is that we don't need a type alias for this, we should show what functions actually return instead of adding an additional level of indirection. It is the documentation's job to say how the returned values should be interpreted, not the type system's.

Previously discussed here: https://github.com/pygame-community/pygame-ce/pull/3076/files/cdf6e919b7aa397122e394c7b38d12c7bb48d002#r1726390905

I don't think we've reached consensus on this yet, but this PR needs to be resolved before 2.5.2 unless RGBATuple will "never" be able to be removed. Normally I'd err on the side of the status quo, but for this issue we can always add it back, but once it is in a full release it will be hard to remove.


ankith26 (commented):
I don't feel strongly about it either way. But I think we can go with this PR because as you said, we can always add it later if we see the need.

ankith26 (approved):
LGTM, thanks! :partying_face:

damusss (approved):
I Agree, LGTM 👍

PR #3107, opened 2024-09-17
Bump mac target to 10.13 on py >= 3.12

ankith26:
for #3103


ankith26:
I just realised that this PR is technically not required, cibuildwheel automatically bumps up the version to whatever it considers minimum for the given wheel tags.

PR #3110, opened 2024-09-22
``pygame.Music``

bilhox:
Hello everyone,

Proud to propose you a first version of the ``pygame.Music`` object after days of work.

Tasks not finished yet :
- [ ] Reduce the need for SDL_Mixer +2.6.0 functions
- [ ] Add macros so github tests can breath
- [ ] Return -1.0 for unsupported formats (wav, midi ...) when getting the duration and other things
- [ ] Add doc
- [ ] Add tests
- [ ] ``fadein`` and ``fadeout`` properties
- [ ] ``playing`` property 
- [ ] ``MUSICENDED`` event when the music is stopped / ends.
*... and many bugs hidden under the carpet to fix ...*

Current test code (Use your own musics, who knows, there might be some issues):
```py
import pygame

def main():
    pygame.init()
    music = pygame.Music("music1.mp3")
    music2 = pygame.Music("music2.mp3")

    musics = [music, music2]
    index = 0

    screen = pygame.display.set_mode([800, 800])
    font = pygame.Font(None)

    musics[index].play()

    clock = pygame.Clock()

    running = True

    while running:

        dt = clock.tick() / 1000
        screen.fill("white")

        text = f"""
        fps = {round(clock.get_fps(), 2)}fps

        Title : {musics[index].title}
        Artist : {musics[index].artist}
        Album : {musics[index].album}
        Copyright : {musics[index].copyright}

        music position = {round(musics[index].position, 2)}s
        music duration = {round(musics[index].duration, 2)}s
        music volume = {round(musics[index].volume * 100)}%
        music paused = {musics[index].paused}
        music ended = {musics[index].ended}
        """
        screen.blit(font.render(text, True, "black"))

        pygame.display.flip()

            if event.type == pygame.QUIT:
                running = False
                if event.key == pygame.K_s:
                    musics[index].stop()
                elif event.key == pygame.K_r:
                    musics[index].rewind()
                elif event.key == pygame.K_p:
                    musics[index].play(fade_in=3)
                elif event.key == pygame.K_a:
                    musics[index].position -= 5
                elif event.key == pygame.K_z:
                    musics[index].position += 5
                elif event.key == pygame.K_o:
                    musics[index].paused = not musics[index].paused
                
                if event.key == pygame.K_UP:
                    musics[index].volume += 0.05
                elif event.key == pygame.K_DOWN:
                    musics[index].volume -= 0.05
                elif event.key == pygame.K_LEFT:
                    index = (index - 1) % len(musics)
                elif event.key == pygame.K_RIGHT:
                    index = (index + 1) % len(musics)

            elif event.type == pygame.TEXTINPUT:
                try:
                    music_index = int(event.text)
                    index = pygame.math.clamp(music_index - 1, 0, 1)
                except:
                    continue

    pygame.quit()

if __name__ == "__main__":
    main()
```


buildconfig/stubs/pygame/mixer_music.pyi
```diff
@@ -19,3 +19,34 @@ def queue(filename: FileLike, namehint: str = "", loops: int = 0) -> None: ...
 def set_endevent(event_type: int, /) -> None: ...
 def get_endevent() -> int: ...
 def get_metadata(filename: Optional[FileLike] = None, namehint: str = "") -> Dict[str, str]: ...
+
+class Music:
+    def __init__(self, filename: FileLike, namehint: Optional[str] = "") -> None: ...
```
aatle, line 24: `namehint` cannot be type `None`: it should be just `str`, not `Optional[str]`.
In fact, the type annotation for `pygame.mixer.music.load` should be fixed in this way too because it also cannot take `None`.

Starbuck5:
So this is some syntactic sugar around the existing music system. And unlike pygame.Window it's not an OOP-ization of a system because we need to support multiple when before we only supported one, we'll still only support one playing at once.

If we add this I'd like it to be easy to maintain, so in that vein, why is this written in C?

buildconfig/stubs/pygame/mixer_music.pyi
```diff
@@ -19,3 +19,34 @@ def queue(filename: FileLike, namehint: str = "", loops: int = 0) -> None: ...
...
+    def copyright(self) -> str: ...
+
+    @property
+    def position(self) -> float: ...
+    @position.setter
+    def position(self, value : float) -> None: ...
+    @property
+    def duration(self) -> float: ...
+    @property
+    def paused(self) -> bool: ...
```
aatle, line 46: Missing `paused` setter in stub.

bilhox:
> So this is some syntactic sugar around the existing music system. And unlike pygame.Window it's not an OOP-ization of a system because we need to support multiple when before we only supported one, we'll still only support one playing at once.
> 
> If we add this I'd like it to be easy to maintain, so in that vein, why is this written in C? 

For several reasons :
- music pre-loading (it's not about performances, but for playing a music in just 2 lines)
- fixes the position problem
- Uses new mixer functions (You suggested to access this via python functions, I don't support idea)
- Needs access to the music playing via Music obj or old system so only one music is set to playback.

aatle:
Is there a specific reason why there is an `ended` property instead of a `playing` property, because I think `playing` would make much more sense.

PR #3114, opened 2024-09-23
Add japanese readme

hisa10:
Added Japanese version of README, also added links to zh-tw to avoid conflict with commit from @noahcse

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -8,7 +8,7 @@
 |PyPiVersion| |PyPiLicense|
 |Python3| |GithubCommits| |BlackFormatBadge|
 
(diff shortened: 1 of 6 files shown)


ankith26 (approved):
I don't know Japanese but @ maya on discord has given this a read.

Thanks for contributing! :tada: though we should probably merge the other PR first because there's definitely going to be merge conflicts for one of you both to fix :)

bilhox (approved):
Same as ankith said : 

I don't know Japanese but @ maya on discord has given this a read.

Thanks for contributing! 🎉 :)

Like we did for french and other translations, we can always wait for japanese people to open an issue concerning this!

PR #3116, opened 2024-09-23
Drop MSYS2 clang32 from our CI

ankith26:
ref: https://github.com/msys2/msys2.github.io/commit/89521b9f8d43b6ebf4a4f4a51c912066ab5f5b50

basically MSYS2 is removing clang32 (i686) support, so one of our builds has started to fail. We can safely remove this on our end as well, because this build doesn't do release wheels.


gresm (approved):
LGTM :+1:
Thanks for a quick response.

Starbuck5 (approved):

PR #3117, opened 2024-09-23
Fix editable install on windows

ankith26:
An (untested) attempt to making editable installs work on windows.

Basically the idea is to special case the DLL handling code which already exists, to handle editable installs differently. With an editable install no DLL moving happens, so we just have to hunt for the DLLs in the source tree and track the folders.

### DIFF
--- a/src_py/__init__.py
+++ b/src_py/__init__.py
@@ -30,16 +30,32 @@ import platform
 # Choose Windows display driver
 if os.name == "nt":
     pygame_dir = os.path.split(__file__)[0]
+    dll_parents = {pygame_dir}
+    try:
+        # For editable support, add some more folders where DLLs are available.
+        # This block only executes under an editable install. In a "normal"
+        # install, the json file will not be installed at the supplied path.
+        with open(
+            os.path.join(
+                os.path.dirname(pygame_dir), "buildconfig", "win_dll_dirs.json"
+            ),
+            encoding="utf-8",
+        ) as f:
+            import json
 
-    # pypy does not find the dlls, so we add package folder to PATH.
-    os.environ["PATH"] = os.environ["PATH"] + ";" + pygame_dir
+            dll_parents.update(json.load(f))
+            del json
+    except (FileNotFoundError, ValueError):
+        pass
 
-    # Windows store python does not find the dlls, so we run this
-    if sys.version_info > (3, 8):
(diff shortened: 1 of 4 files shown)


ankith26:
I just tested this on a fresh Windows 11 VM and it seems to work okay

src_py/__init__.py
```diff
@@ -30,16 +30,32 @@
...
-    if sys.version_info > (3, 8):
-        os.add_dll_directory(pygame_dir)  # only available in 3.8+
+    pygame_dir_parent, pygame_dir_name = os.path.split(pygame_dir)
+    if pygame_dir_name == "src_py":
+        import glob
+
+        dll_parents = {
+            os.path.dirname(file)
+            for file in glob.iglob(
+                os.path.join(pygame_dir_parent, "**", "*.dll"), recursive=True
```
ankith26, line 40: Note to self: should maybe handle only the `prebuilt-*` style folders here

src_py/__init__.py
```diff
@@ -30,16 +30,32 @@
...
+        del glob
+    else:
+        dll_parents = {pygame_dir}
+
+    for d in dll_parents:
+        # pypy does not find the dlls, so we add package folder to PATH.
+        os.environ["PATH"] = os.environ["PATH"] + ";" + d
+
+        # Windows store python does not find the dlls, so we run this
+        if sys.version_info > (3, 8):
```
gresm, line 52: Are there any python versions before 3.8 that we support?

damusss: > Are there any python versions before 3.8 that we support?

No. And we will drop support of 3.8 some time in the future since it's stopping to get security updates from october of this year.

ankith26: I have removed the conditional check in my updated commit

ankith26:
While this strategy works, I believe it can be improved further. I'll experiment with this a bit more.

ankith26:
Okay, I think this PR should be done from my side.

The new strategy basically makes the meson buildconfig keep track of the DLL dirs in a JSON file, and then at runtime this file can be handled if it is available (in the editable case only)

bilhox (approved):
Tested on Windows, and it works fine, thanks for the PR!

oddbookworm (approved):
LGTM!

PR #3118, opened 2024-09-24
mac/manylinux deps updates 2024

ankith26:
It's that time of the year again, this is the deps update PR of 2024.

Most of the dependency updates are straightforward, the only non-trivial changes here are change in buildsystem for
- SDL and friends: SDL3 is going to exclusively support cmake, so I figured we should make the jump right away. I believe it is better maintained these days anyways.
- harfbuzz now supports meson only
- flac/ogg/vorbis had some issues, so jump to CMake on that one as well
- Also fixed a couple of other misc issues with libxmp and `LD_LIBRARY_PATH` handling


ankith26:
This PR slightly increases wheel size, but that is because so far SDL and friends were probably not being compiled with optimization flags, but now they are. So that is a plus of this PR.

Also, I am hopeful that this PR should help with reviving my win deps PR from last year (https://github.com/pygame-community/pygame-ce/pull/2505)

buildconfig/manylinux-build/docker_base/mesa/libdrm/build-libdrm.sh
Starbuck5, line 1: The fact that we're building some of these libraries ourselves at all is crazy.

MyreMylar (approved):
Alright, all looks good to me, but holy heck that is a lot of version bumps. Still the CI is happy, so I'm happy?

I can also reveal that non of these many linux build changes had any effect whatsoever on the windows build (which I'm sure you were wondering).

Starbuck5:
Going through a wheel produced on this branch and a wheel produced on main recently, I notice the one on main has libxmp in the pygame_ce.libs folder, and the wheel from this branch does not.

Is that a problem?

ankith26:
I noticed that too. But then I realised that it is somewhat intentional with the cmake build: libxmp gets statically linked against SDL_image. You can also see that SDL_image size goes up the same amount as the old libxmp shared lib.

We have a libxmp unittest, that passed on this branch too, so we should be all good here

Starbuck5:
You mean SDL_mixer.

Now 75% (at least) of SDL_mixer's file size is this one library! And they're bigger together than they were separately!*

Also a test run of the dep on the machine that had all the dev libraries to build the thing in the first place is less convincing.


*assuming the switch to position independent code in libxmp compilation didn't grab the 10 additional kb

ankith26:
> Also a test run of the dep on the machine that had all the dev libraries to build the thing in the first place is less convincing.

We never need to worry about this on macs and linuxs because
- mac builds delete all the dev files on the CI before running the tests
- I tested the linux wheels locally and saw no issues

Also like I mentioned above
> This PR slightly increases wheel size, but that is because so far SDL and friends were probably not being compiled with optimization flags, but now they are. So that is a plus of this PR.

I attribute binary size increases to this

Starbuck5:
> I tested the linux wheels locally and saw no issues

You also have the libraries installed to build a bunch of things from source, so that's not convincing either.

I believe you that it works, but because of the filesize, not either of those tests.

ankith26:
trust auditwheel and delocate devs :smile: 

So far the issue you have talked about has only ever bitten us on windows (some random missing DLL), atleast in my memory

Starbuck5:
Also would it be easy to not have libxmp statically in SDL_mixer? It's the only one that is, this seems like it could confuse us in the future.

ankith26:
I can open an issue/PR to SDL_mixer about it, I feel like they would also see it as an inconsistency

Starbuck5:
CMake doesn't give us control of that?

Starbuck5 (approved):
Thanks for working this up!

PR #3119, opened 2024-09-25
Update the dependency section of zh-cn README

yunline:
Update the dependency section of zh-cn README.


ankith26 (approved):
Can't read Simplified Chinese but with google translate it looks file. Thanks for the PR :+1:

bilhox (approved):
Can't read chinese but you can natively, it seems fine for me. Thanks for the PR!

PR #3121, opened 2024-09-25
Events as types in python

### LINKED ISSUE #2759: Idea: event types are actual types

gresm:
Redo of #2866, but as requested here: https://github.com/pygame-community/pygame-ce/pull/2866#pullrequestreview-2079777435 in python.

This is the third and hopefully last pull request from the chain of: #3000 and #3045, this patch by itself isn't very large, but it relies on the work of these pull requests, so it might be easier reviewing them one by one.

Closes #2759


gresm:
Per https://github.com/pygame-community/pygame-ce/pull/3122#issuecomment-2380714982, the ``pygame.event.peek(None)`` was removed, even though this PR seems to fix the crash.

PR #3122, opened 2024-09-25
Remove event.peek() behavior, minor doc and stub fixes

### LINKED ISSUE #1196: pygame.event.peek docs missing a detail (2329)
#  **Issue &numero;2329 opened by *[ankith26](https://github.com/ankith26)* at 2020-11-15 09:27:16**

According to pygame event.peek docs, the function is supposed to return a Boolean on whether a particular event is present in the queue or not. What it misses out, is that if you pass None or do not pass any argument, it returns the peeked event itself (I don’t know why, but that’s what the code is doing) https://www.pygame.org/docs/ref/event.html# pygame.event.peek

We must either mention this behaviour in the docs, or have this weird behaviour removed

Side note: this is the last event function I’m working on. Hopefully after this, event refactor PR coming soon :) <hr>

#  Comments

# # #  **[MyreMylar](https://github.com/MyreMylar)* commented at 2020-11-15 09:48:59*

I think I would expect this function to return True is there were _any_ events in the queue if the type was unspecified, and False if the queue was empty.
<hr>

# # #  **[MyreMylar](https://github.com/MyreMylar)* commented at 2020-11-15 09:50:37*

Likely worth searching github for usages though to see if changing this would break something. The current behaviour is likely based off what happens with a standard stack/queue data structure when you call `.peek()` on it.
<hr>

# # #  **[ankith26](https://github.com/ankith26)* commented at 2020-11-15 10:30:16*

Did do a github search, and it turns out that a lot of people include the local venv folder while uploading their projects to github (so most of results are from pygames unit tests)
<hr>

# # #  **[ankith26](https://github.com/ankith26)* commented at 2020-12-20 05:40:36*

@illume what’s your vote on this. Should we get rid of this weird behavior, or should we document this one.

AFAIK, The function has been in that state since years. And this has not been documented since many years too, so I don’t think many people know that this even exists. So it might be safe to remove it.

Either way, I would do this along with event-refactor PR
<hr>

# # #  **[illume](https://github.com/illume)* commented at 2020-12-20 08:35:58*

I'd vote with leaving the behaviour as it is for backwards compatibility, but leave it undocumented still.

If someone comes along later who wants to do more research on usages of this, and then fix it either way... then there is this issue here for them to work on it.
<hr>

# # #  **[ankith26](https://github.com/ankith26)* commented at 2020-12-20 13:06:34*

I did quite a good amount of research on this. By going through a github search of all usages.
And found that nobody uses this feature. IMHO, it might be a good idea to remove this behavior 

I did find a few rare cases where the function was called without arguments. But all those were testing it in an if condition to test whether any events were waiting in the queue or not. 
So if we take MyreMylars suggestion, given above, we would still be good to go

zoldalma999:
Fixes #1196 by documenting the behavior of pygame.event.peek if nothing is passed in.

Also marks `event.type` and `event.dict` readonly in the stubs, adds runtime docs to them, and makes the pump argument more accurate in the stubs.


bilhox (changes requested):
LGTM, thanks for the PR!
Worth to mention that @gresm also proposed a fix for this issue in his latest PR.
Otherwise, I left little changes.

buildconfig/stubs/pygame/event.pyi
```diff
@@ -28,13 +31,13 @@ _EventTypes = Union[int, SequenceLike[int]]
...
     eventtype: Optional[_EventTypes] = None,
-    pump: Any = True,
+    pump: bool = True,
     exclude: Optional[_EventTypes] = None,
 ) -> List[Event]: ...
 def poll() -> Event: ...
 def wait(timeout: int = 0) -> Event: ...
-def peek(eventtype: Optional[_EventTypes] = None, pump: Any = True) -> bool: ...
-def clear(eventtype: Optional[_EventTypes] = None, pump: Any = True) -> None: ...
+def peek(eventtype: Optional[_EventTypes] = None, pump: bool = True) -> bool: ...
```
ankith26, line 38: The return value probably needs to be updated here to be consistent with the docs.

Also, the reason it was typed as `Any` in the first place is because almost any object in python is acceptable as a bool, as you probably already know

ankith26 (commented):
If this is becoming public, a test or two for it would also be good

zoldalma999:
> If this is becoming public, a test or two for it would also be good

Wanted to write one for it, but got a segfault with this code:
```python
import pygame

pygame.init()
pygame.event.clear()

e = pygame.Event(pygame.KEYDOWN, key=pygame.K_SPACE)
pygame.event.post(e)

print("1")
assert e == pygame.event.peek(pump=False)
print("2")
assert e == pygame.event.peek(None, False)
print("3") # This one never gets printed
```
Debugging it a bit more, seems like the dict proxy on this line is already freed
https://github.com/pygame-community/pygame-ce/blob/ddc900edc037552763f390fdfb0cc612581a17c3/src_c/event.c#L930

ankith26:
After giving it a bit of thought, my current opinion is that we should remove this functionality all together now.
- If it is segfaulting right now and no one has reported it yet, no one knows or cares about this behaviour.
- For whatever reason if we want to document it down the line, it can be readded and fixed then.

ankith26 (commented):
I think there is a bit of confusion here, what I actually wanted to convey was "make the return value a bool regardless of the input, just remove the behaviour of returning an event type". I would still like to retain the `None`/optional-arg case

buildconfig/stubs/pygame/event.pyi
```diff
@@ -17,20 +18,21 @@ class Event:
...
     eventtype: Optional[_EventTypes] = None,
-    pump: Any = True,
+    pump: bool = True,
     exclude: Optional[_EventTypes] = None,
 ) -> list[Event]: ...
 def poll() -> Event: ...
 def wait(timeout: int = 0) -> Event: ...
-def peek(eventtype: Optional[_EventTypes] = None, pump: Any = True) -> bool: ...
-def clear(eventtype: Optional[_EventTypes] = None, pump: Any = True) -> None: ...
+def peek(eventtype: _EventTypes, pump: bool = True) -> bool: ...
```
MyreMylar, line 34:
```suggestion
def peek(eventtype: Optional[_EventTypes] = None, pump: bool = True) -> bool: ...
```

docs/reST/ref/event.rst
```diff
@@ -317,8 +317,7 @@ On Android, the following events can be generated
 .. function:: peek
 
    | :sl:`test if event types are waiting on the queue`
-   | :sg:`peek(eventtype=None) -> bool`
-   | :sg:`peek(eventtype=None, pump=True) -> bool`
+   | :sg:`peek(eventtype, pump=True) -> bool`
```
MyreMylar, line 320:
```suggestion
   | :sg:`peek(eventtype=None) -> bool`
   | :sg:`peek(eventtype=None, pump=True) -> bool`
```

src_c/doc/event_doc.h
```diff
@@ -4,7 +4,7 @@
 #define DOC_EVENT_GET "get(eventtype=None) -> Eventlist\nget(eventtype=None, pump=True) -> Eventlist\nget(eventtype=None, pump=True, exclude=None) -> Eventlist\nget events from the queue"
 #define DOC_EVENT_POLL "poll() -> Event instance\nget a single event from the queue"
 #define DOC_EVENT_WAIT "wait() -> Event instance\nwait(timeout) -> Event instance\nwait for a single event from the queue"
-#define DOC_EVENT_PEEK "peek(eventtype=None) -> bool\npeek(eventtype=None, pump=True) -> bool\ntest if event types are waiting on the queue"
+#define DOC_EVENT_PEEK "peek(eventtype, pump=True) -> bool\ntest if event types are waiting on the queue"
```
MyreMylar, line 7:
```suggestion
#define DOC_EVENT_PEEK "peek(eventtype=None) -> bool\npeek(eventtype=None, pump=True) -> bool\ntest if event types are waiting on the queue"
```

docs/reST/ref/event.rst
```diff
@@ -328,6 +327,8 @@ On Android, the following events can be generated
 
    .. versionchangedold:: 1.9.5 Added ``pump`` argument
 
+   .. versionchanged:: 2.5.3 ``eventtype`` is no longer an optional argument
+
```
MyreMylar, line 331:
```suggestion
   .. versionchanged:: 2.5.3 no longer mistakenly returns an event when ``eventtype`` is None or not passed.

```

src_c/event.c
```diff
@@ -2250,52 +2250,44 @@ pg_event_peek(PyObject *self, PyObject *args, PyObject *kwargs)
 
     static char *kwids[] = {"eventtype", "pump", NULL};
 
-    if (!PyArg_ParseTupleAndKeywords(args, kwargs, "|Op", kwids, &obj,
+    if (!PyArg_ParseTupleAndKeywords(args, kwargs, "O|p", kwids, &obj,
```
MyreMylar, line 2253:
```suggestion
    if (!PyArg_ParseTupleAndKeywords(args, kwargs, "|Op", kwids, &obj,
```

src_c/event.c
```diff
@@ -2250,52 +2250,44 @@ pg_event_peek(PyObject *self, PyObject *args, PyObject *kwargs)
...
-                return NULL;
-            }
-            res = PG_PEEP_EVENT(&event, 1, SDL_PEEKEVENT, type);
-            if (res) {
-                Py_DECREF(seq);
+    for (loop = 0; loop < len; loop++) {
+        type = _pg_eventtype_from_seq(seq, loop);
+        if (type == -1) {
+            Py_DECREF(seq);
+            return NULL;
```
MyreMylar, line 2269:
```suggestion
            return Py_RETURN_FALSE;
```

src_c/event.c
```diff
@@ -2250,52 +2250,44 @@ pg_event_peek(PyObject *self, PyObject *args, PyObject *kwargs)
...
-            return RAISE(pgExc_SDLError, SDL_GetError());
-        return pgEvent_New(res ? &event : NULL);
-    }
-    else {
-        seq = _pg_eventtype_as_seq(obj, &len);
-        if (!seq)
-            return NULL;
+    seq = _pg_eventtype_as_seq(obj, &len);
+    if (!seq)
+        return NULL;
```
MyreMylar, line 2263:
```suggestion
        return Py_RETURN_FALSE;
```

test/event_test.py
```diff
@@ -250,13 +250,6 @@ def test_clear(self):
...
 
     def test_peek(self):
-        pygame.event.peek()
-        pygame.event.peek(None)
-        pygame.event.peek(None, True)
-
-        pygame.event.peek(pump=False)
-        pygame.event.peek(pump=True)
-        pygame.event.peek(eventtype=None)
         pygame.event.peek(eventtype=[pygame.KEYUP, pygame.KEYDOWN])
```
MyreMylar, line 253:
```suggestion
        pygame.event.peek()
        pygame.event.peek(None)
        pygame.event.peek(None, True)

        pygame.event.peek(pump=False)
        pygame.event.peek(pump=True)
        pygame.event.peek(eventtype=None)
        pygame.event.peek(eventtype=[pygame.KEYUP, pygame.KEYDOWN])
```

test/event_test.py
```diff
@@ -761,11 +754,6 @@ def test_peek__empty_queue(self):
         """Ensure peek() works correctly on an empty queue."""
         pygame.event.clear()
 
-        # Ensure all events can be checked.
-        peeked = pygame.event.peek()
-
-        self.assertFalse(peeked)
-
         # Ensure events can be checked individually.
```
MyreMylar, line 757:
```suggestion
        # Ensure all events can be checked.
        peeked = pygame.event.peek()

        self.assertFalse(peeked)

        # Ensure events can be checked individually.
```

src_c/event.c
```diff
@@ -2258,44 +2258,36 @@ pg_event_peek(PyObject *self, PyObject *args, PyObject *kwargs)
...
-            return RAISE(pgExc_SDLError, SDL_GetError());
-        return pgEvent_New(res ? &event : NULL);
-    }
-    else {
-        seq = _pg_eventtype_as_seq(obj, &len);
-        if (!seq)
-            return NULL;
+    seq = _pg_eventtype_as_seq(obj, &len);
+    if (!seq)
+        return Py_RETURN_FALSE;
```
MyreMylar, line 2263:
```suggestion
        Py_RETURN_FALSE;
```

src_c/event.c
```diff
@@ -2258,44 +2258,36 @@ pg_event_peek(PyObject *self, PyObject *args, PyObject *kwargs)
...
-                return NULL;
-            }
-            res = PG_PEEP_EVENT(&event, 1, SDL_PEEKEVENT, type);
-            if (res) {
-                Py_DECREF(seq);
+    for (loop = 0; loop < len; loop++) {
+        type = _pg_eventtype_from_seq(seq, loop);
+        if (type == -1) {
+            Py_DECREF(seq);
+            return Py_RETURN_FALSE;
```
MyreMylar, line 2269:
```suggestion
            Py_RETURN_FALSE;
```

MyreMylar:
Hmm, well I tried to match what @ankith26 was asking for but it now seems to be segfaulting in the CI. No idea why.

ankith26:
I made my own attempt at resolving this PR, and it's now in a new PR here: #3283

I have taken some work done in this PR onto the new one, so thanks for working on this! So, I am closing this one.

PR #3123, opened 2024-09-26
Bump ubuntu to 24.04 on CI, fix cppcheck

ankith26:
Ubuntu 24.04 has been declared as generally available on github actions. At the same time, Ubuntu 20.04 is being phased out and will be removed soon.

This PR makes a couple of changes to accommodate for 24.04 in our code. Most notably, the `cppcheck` version change has made me do a couple of changes to the test command as well as the code in one instance (the new version did not like something in `color.c`)

This PR is also a prerequisite for dropping Python 3.8 support because Ubuntu 20.04 is on Python 3.8

### DIFF
--- a/test/mixer_music_test.py
+++ b/test/mixer_music_test.py
@@ -40,8 +40,10 @@ class MixerMusicModuleTest(unittest.TestCase):
         "|tags:music|"
         self.music_load("house_lo.flac")
 
+    # system installed SDL_mixer may not support wavpack
     @unittest.skipIf(
-        pygame.mixer.get_sdl_mixer_version() < (2, 8, 0),
+        pygame.mixer.get_sdl_mixer_version() < (2, 8, 0)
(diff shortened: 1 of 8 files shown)


damusss (approved):
LGTM :)

test/mixer_music_test.py
Starbuck5, line 1: How about we make these expected fails if running on CI and Linux, instead of bonking the tests to make a lower standard of behavior acceptable?

Or even just expected fail if there's a certain flag in the environment?

I would like these things to be tested for our release wheels with our binaries.

ankith26: I agree, I was thinking of a way to do this but couldn't do so while making this PR.

I like the idea of having an env variable set on our Ubuntu 24.04 CI to specifically skip such tests, I will implement it soon.

MyreMylar (approved):
OK, LGTM 👍

PR #3125, opened 2024-09-28
Set error when too many numbers of argument in ``pygame.Color.from_colorspace``

### LINKED ISSUE #2386: ability to call ``pygame.Color.from_colorspace`` methods with more than 4 args + python doctype wrong
**PyGame version**: pygame 2.3.1.dev4

**Current behavior:** We can pass as much arguments as we want to ``pygame.Color.from_colorspace`` methods. Furthermore, python doctype is wrong, for example with ``pygame.Color.from_hsla``, it expects 4 arguments or a tuple of 4 arguments, however we can pass 3 arguments.

**Expected behavior:** Raises an error if too much arguments were passed, and add default value to the last argument (for hsla, hsva, i1i2i3).

**Test code**

```py
import pygame

pygame.init()

screen = pygame.display.set_mode([200,100])

color = pygame.color.Color.from_hsva(130, 40, 12) # technically wrong according to python doctype
color2 = pygame.Color.from_cmy(1., 1., 1., "I like arguments")
color3 = pygame.Color.from_hsla(120, 32, 34, 1, 2234832948, "Mattiss likes fish.")
color4 = pygame.Color.from_i1i2i3([0, 0, 0, 12, 34209, "In a rectangle triangle, a² + b² = c², where 'c' is the hypotenuse length and a and b the length of the two other sides"])

while True:

    screen.fill(color)

        if event.type == pygame.QUIT or event.type == pygame.KEYDOWN and event.key == pygame.K_ESCAPE:
            pygame.quit()
            exit(0)

    pygame.display.flip()
```
This code works perfectly fine.

bilhox:
Final step of @ankith26 's tasks in #2386 .
A year after ankith deprecated the use of this implementation, we can finally set an error for these.

Closes #2386

### DIFF
--- a/test/color_test.py
+++ b/test/color_test.py
@@ -830,9 +830,10 @@ class ColorTypeTest(unittest.TestCase):
         self.assertEqual(expected_i1i2i3, i1i2i3)
         self.assertEqual(expected_i1i2i3, i1i2i3_tuple)
 
-        with self.assertRaises(ValueError):
-            pygame.Color.from_i1i2i3(0, 0, 0, "lel", "foo")
-            pygame.Color.from_i1i2i3((0, 0, 0, 0))
+        self.assertRaises(
+            ValueError, lambda: pygame.Color.from_i1i2i3(0, 0, 0, "lel", "foo")
+        )
+        self.assertRaises(ValueError, lambda: pygame.Color.from_i1i2i3((0, 0, 0, 0)))
 
     def test_from_normalized(self):
         normal = pygame.Color.from_normalized(1, 1, 1, 1)
--- a/test/color_test.py
+++ b/test/color_test.py
@@ -782,13 +782,9 @@ class ColorTypeTest(unittest.TestCase):
         self.assertEqual(expected_cmy, cmy)
         self.assertEqual(expected_cmy, cmy_tuple)
 
-        with self.assertWarns(DeprecationWarning):
-            self.assertEqual(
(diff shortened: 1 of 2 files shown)


damusss (approved):
As the deprecation guru, I can't not approve this. I guess no notice in the docs is needed because the docs never mentioned allowing more than the exact number of components.

test/color_test.py
```diff
@@ -782,13 +782,9 @@ def test_from_cmy(self):
...
-        with self.assertWarns(DeprecationWarning):
-            self.assertEqual(
-                expected_cmy, pygame.Color.from_cmy(0.5, 0.5, 0.5, "lel", "foo")
-            )
-
-        with self.assertWarns(DeprecationWarning):
-            self.assertEqual(expected_cmy, pygame.Color.from_cmy((0.5, 0.5, 0.5, 0.5)))
+        with self.assertRaises(ValueError):
+            pygame.Color.from_cmy(0.5, 0.5, 0.5, "lel", "foo")
+            pygame.Color.from_cmy((0.5, 0.5, 0.5, 0.5))
```
ankith26, line 787: these should be in separate blocks otherwise the second line is redundant and will never be tested

bilhox: Why the second one will not be tested ?

ankith26: Because the first line raises `ValueError` (as we want it to) and exits the `with` block

bilhox: fixed

ankith26 (approved):

PR #3126, opened 2024-09-28
Add draw.aapolygon

mzivic7:
This PR adds `draw.aapolygon` function, with filled option.
It was not possible to overlap `draw.poylgon` and `draw.aalines` without issues:
function `draw_fillpoly` sometimes draws polygon slightly larger, and those fully opaque pixels are hiding antialiased pixels from `aalines`
So I made separate, modified, `draw_fillpoly_for_aapolygon` that makes borders slightly thinner at those problematic places.
I didn't want to change original `draw_fillpoly`: to not add extra arguments and `if`-s inside loops, which would make it slower.
`draw.aapolygon` is added to docs, examples and tests.

Screenshot_From_2024-09-28_12-44-51
Left: `draw.polygon`+`draw.aalines`
Right: `draw.aapolygon`

<details>

  ```py
import pygame

points = [[100, 100],
          [200, 160],
          [120, 255],
          [70, 121]]
points2 = [[220, 100],
           [320, 160],
           [240, 255],
           [190, 121]]
pygame.init()
screen = pygame.display.set_mode((400, 400))
clock = pygame.time.Clock()

run = True
while run:
        if event.type == pygame.QUIT:
            run = False
    screen.fill("black")
    pygame.draw.aalines(screen, "white", True, points)
    pygame.draw.polygon(screen, "white", points)
    pygame.draw.aapolygon(screen, "white", points2)
    pygame.display.flip()
    clock.tick(60)
pygame.quit()
  ```
</details>


bilhox:
What about adding a kwarg called ``antlialiased`` in ``pygame.draw.polygon``, just like in ``Font.render``, rather than an actual different function ?

src_c/draw.c
```diff
@@ -981,6 +984,124 @@ polygon(PyObject *self, PyObject *arg, PyObject *kwargs)
...
+        PyMem_Free(ylist);
+        return RAISE(PyExc_RuntimeError, "error locking surface");
+    }
+
+    if (length != 3) {
+        draw_fillpoly_for_aapolygon(surf, xlist, ylist, length, color,
+                                    drawn_area);
+    }
+    else {
+        draw_filltri(surf, xlist, ylist, color, drawn_area);
```
bilhox, line 1085: but the filled triangle will not be antialiased then ?

src_c/draw.c
```diff
@@ -2984,6 +3100,100 @@ draw_fillpoly(SDL_Surface *surf, int *point_x, int *point_y,
     PyMem_Free(x_intersect);
 }
 
+/* This is same algorythm as draw_fillpoly but modified for filled
+ * draw.aapolygon so borders of fillpoly are not covering outer
+ * antialiased pixels from draw.aalines
+ */
+static void
+draw_fillpoly_for_aapolygon(SDL_Surface *surf, int *point_x, int *point_y,
```
bilhox, line 3108: let's make the function name more obvious with ``drawFilledAAPolygon``.

bilhox (changes requested):
good job

damusss:
> What about adding a kwarg called `antlialiased` in `pygame.draw.polygon`, just like in `Font.render`, rather than an actual different function ?

First of all, font antialiasing should have been an attribute. Secondly, while it would technically be better imo, it would surely be incosistent with the other 3 antialiased draw functions.

src_c/draw.c
damusss: sure but not `drawFilledAAPolygon` rather `draw_filled_aapolygon`. get java outta here

bilhox:
> Secondly, while it would technically be better imo, it would surely be incosistent with the other 3 antialiased draw functions.

While we are at it, let's make it a kwarg. I believe ankith or starbuck told us we should make it kwarg now. we're not in C, we're on python, so let's make it more friendly.

damusss:
> While we are at it, let's make it a kwarg. I believe ankith or starbuck told us we should make it kwarg now. we're not in C, we're on python, so let's make it more friendly.

it sounds good. what about aacircle, aaline and aalines?

damusss:
@bilhox only problem is that aapolygon has filled argument while polygon has width argument. they are not compatible

bilhox:
okay then I agree with keeping it a different function.

damusss:
I suppose if this one has no width there must be a specific reason, which I'm curious to know

bilhox:
> I suppose if this one has no width there must be a specific reason, which I'm curious to know

Currently there is no implementation with antialiased line, however, I hope the man who made the video I sent on discord will make a video for antialiased line with width.

mzivic7:
Implementing line width is much more complex, because it has to grow inwards, so I can't just use draw.aalines with width argument.
The problem is in finding points inside polygon, and this is not simple thing. 
There is a whole library for shrinking polygons: https://pypi.org/project/shapely/

I can make draw.aapolygon with draw.aalines, and width will grow in both sides, same as draw.polygon.
But that would leave gaps where endpoints are, same as draw.polygon.

Making edges sharp would need some intense math, so it's out of question. 

For draw.aalines, if I make width arg, I can put aacircle on each endpoint to make edges rounded. 
For that I would make draw.aaline with width arg, also with circles on endpoints.

draw.aalines with width will be very hard to implement, even harder with rounded edges. That's why it's better to leave aapolygon with just filled arg.

So, the decisions that needs to be made for width > 1 are:
-Should draw.aaline have rounded endpoints with aacircle?
- Should draw.line have rounded endpoints with circle?
- Should draw.aalines have rounded edges?
- Should draw.lines have rounded edges?
- Should draw.aapolygon have rounded edges?
- Should draw.polygon have rounded edges?

src_c/draw.c
mzivic7: But this is NOT drawing antialiased polygon, it is drawing regular polygon, modified so it does not interfere with `draw.aalines` which are responsible for antialiasing.

mzivic7:
Also note on merging `draw.aa<function>` into `draw.<function>`:
This would require almost all draw modules to be merged: line, lines, circle, ellipse, polygon
It will not make much problems in c functions, except they will just get bigger and less readable,
But merging tests (with `draw.aaellipse` incl)... Almost 9000 LOC, around 300 tests that each needs to be checked, merged and modified.

src_c/draw.c
mzivic7: `draw_fillpoly_for_aapolygon` just draw 'fillpoly' - filled polygon, its not antialiased.
Antialiasing is done for both `fillpoly` and `filltri` by calling `aalines` on line 1096.

src_c/draw.c
```diff
@@ -1780,6 +1783,64 @@ draw_filltri(SDL_Surface *surf, int *xlist, int *ylist, Uint32 color,
...
+        }
+    }
+
+    if ((p0y == p1y) && (p1y == p2y) && (p0x == p1x) && (p1x != p2x)) {
+        swap_coordinates(&p1x, &p1y, &p2x, &p2y);
+    }
+
+    float d1 = (float)((p2x - p0x) / ((p2y - p0y) + 1e-17));
+    float d2 = (float)((p1x - p0x) / ((p1y - p0y) + 1e-17));
+    float d3 = (float)((p2x - p1x) / ((p2y - p1y) + 1e-17));
```
bilhox, line 1821: I'm not familiar with the algorithm, but why the divisor is being added ``1e-17`` ?

src_c/draw.c
```diff
@@ -1082,7 +1085,7 @@ aapolygon(PyObject *self, PyObject *arg, PyObject *kwargs)
                                     drawn_area);
     }
     else {
-        draw_filltri(surf, xlist, ylist, color, drawn_area);
+        draw_filltri_for_aapolygon(surf, xlist, ylist, color, drawn_area);
```
bilhox, line 1088: Let's avoid duplicate code, and make a 6th argument for ``draw_filltri``. Because from what I can see, there is no big differences in its code.

bilhox (changes requested):
Now it works fine ! Good job :)
Before approving it, I would like to see some code refactoring (requested changes).

src_c/draw.c
bilhox: Same here, I didn't check correctly yesterday, but this code is 99% duplicate from ``draw_fillpoly``

bilhox (changes requested):

src_c/draw.c
mzivic7: Any ideas for 6th argument name?

damusss: Maybe `target_antialias`. Or similar.

src_c/draw.c
mzivic7: This value is added as safety against division by zero, when any of two points of triangle have same y coordinate.
Those d values are line gradients. In loop bellow is rest of formula but it is expressed for x, that is why gradient is calculated as (x2-x1)/(y2-y1). And for this case, when gradient is infinite, it is straight horizontal line.

src_c/draw.c
mzivic7: Decided to call it `aapolygon_fix`

docs/reST/ref/draw.rst
```diff
@@ -147,6 +147,46 @@ object around the draw calls (see :func:`pygame.Surface.lock` and
...
+      ``points`` parameter (float values will be truncated) and its width and
+      height will be 0
+   :rtype: Rect
+
+   :raises ValueError: if ``len(points) < 3`` (must have at least 3 points)
+   :raises TypeError: if ``points`` is not a sequence or ``points`` does not
+      contain number pairs
+
+   .. note::
+       For an aapolygon, use :func:`aalines()` with ``closed=True``.
```
MyreMylar, line 184: I think this note has been copied across by mistake from the docs above. Also, we should probably update the note above for `draw.polygon()` now to point to this function instead now that it exists.

MyreMylar (approved):
LGTM 👍 

I noticed one small issue in the docs, where they are *technically correct* - but still wrong.

I tested this locally and it all works well in my tinkering around, tests all pass locally as well. I like the interface chosen, it should neatly replace everything `gfxdraw.aapolygon()` does and adds filling as well which is great. 👍👍

damusss:
@mzivic7 I see this version has the filled argument. Do you plan on replacing it with a width, based on your testing, or do you think it's not possible/not worth it/other reason? Just curious, to make sure you intend for the PR to be ready

docs/reST/ref/draw.rst
```diff
@@ -147,6 +147,46 @@ object around the draw calls (see :func:`pygame.Surface.lock` and
...
+   | :sl:`draw an antialiased polygon`
+   | :sg:`aapolygon(surface, color, points) -> Rect`
+   | :sg:`aapolygon(surface, color, points, filled=True) -> Rect`
+
+   Draws an antialiased polygon on the given surface.
+
+   :param Surface surface: surface to draw on
+   :param color: color to draw with, the alpha value is optional if using a
+      tuple ``(RGB[A])``
+   :type color: Color or string (for :doc:`color_list`) or int or tuple(int, int, int, [int])
```
bilhox, line 161: Here it would be interesting to refer to ``ColorLike`` from ``pygame.typing``, using a link to its paragraph in the docs.

docs/reST/ref/draw.rst
```diff
@@ -147,6 +147,46 @@ object around the draw calls (see :func:`pygame.Surface.lock` and
...
+   :param points: a sequence of 3 or more (x, y) coordinates that make up the
```
bilhox, line 162: Same here for coordinates with ``pygame.typing.Coordinate``.

docs/reST/ref/draw.rst
```diff
@@ -147,6 +147,46 @@ object around the draw calls (see :func:`pygame.Surface.lock` and
...
+         | if filled == False, don't fill the polygon
+         |
+
+   :returns: a rect bounding the changed pixels, if nothing is drawn the
+      bounding rect's position will be the position of the first point in the
+      ``points`` parameter (float values will be truncated) and its width and
+      height will be 0
+   :rtype: Rect
+
+   :raises ValueError: if ``len(points) < 3`` (must have at least 3 points)
```
bilhox, line 179: why should the function (and also ``polygon``) raise a ``ValueError`` when the number of point is lower than 3 ? Like if the number is 2, it would draw a line, if the number is 1 it draws a point, and if the number is 0, it draws nothing ?

bilhox (changes requested):
Good job for C part ! :)

Just left some minor requested changes concerning docs, and a question.

docs/reST/ref/draw.rst
damusss: perhaps you are right, but regardless this isn't scope of this PR. it could be added to both functions in a separate PR if needed. And I'm not sure it makes much sense, as line and point are technically not polygons, so the function would be behaving differently from its name.

mzivic7:
> @mzivic7 I see this version has the filled argument. Do you plan on replacing it with a width, based on your testing, or do you think it's not possible/not worth it/other reason? Just curious, to make sure you intend for the PR to be ready

I finished width for `draw.aalines` but only in python code form. Unfortunately it is very complex and I'm afraid it will be slow. 
It has 150LOC, cyclomatic complexity = 14, and there is a ton of math, perhaps someone should review that python code.
For one `draw.aalines` with width, it calls: x4 `draw.aalines`, x1 `draw.lines`, xn to x2n `draw.polygon` where n is number of points.
If this is acceptable I can do same "upgrade" to `draw.lines` since it also has gaps on edges (which includes `draw.polygon`).
If this is not acceptable, then i can make it with `draw.aalines` and their width arg (which will come soon in another PR), but it will work same as `draw.lines` and have those gaps on edges.

docs/reST/ref/draw.rst
mzivic7: It would be better to add this to all functions in docs at once, in one separate PR. Like just one find-and-replace.

bilhox: Why it should be in a different PR when it concerns directly ``aapolygon`` ?

bilhox:
Maybe I didn't see correctly, where is the PR reverting the changes of ``Coordinate`` to ``Point`` ?

mzivic7:
> Maybe I didn't see correctly, why is the PR reverting the changes of `Coordinate` to `Point` ?

Forgot to sync all my branches. Sorry.

MyreMylar:
These are the two tests failing:

```
  FAIL: test_draw_diamond (pygame.tests.draw_test.DrawAAPolygonTest)
    File "C:\Users\runneradmin\AppData\Local\Temp\cibw-run-prw312qk\cp38-win_amd64\venv-test\lib\site-packages\pygame\tests\draw_test.py", line 4722, in test_draw_diamond
      self.assertEqual(self.surface.get_at((x, y)), GREEN, msg=str((x, y)))
  AssertionError: Color(255, 0, 0, 255) != Color(0, 255, 0, 255) : (5, 3)
  
  FAIL: test_illumine_shape (pygame.tests.draw_test.DrawAAPolygonTest)
  non-regression on pygame-ce issue #328
    File "C:\Users\runneradmin\AppData\Local\Temp\cibw-run-prw312qk\cp38-win_amd64\venv-test\lib\site-packages\pygame\tests\draw_test.py", line 4828, in test_illumine_shape
      self.assertEqual(self.surface.get_at((x, 0)), GREEN)  # upper border
  AssertionError: Color(255, 0, 0, 255) != Color(0, 255, 0, 255)
  
```

MyreMylar:
It looks like the change is clipping off the right hand side for this test comparing the draw polygon 'diamond' versus the 'gfxdraw' one:

image

mzivic7:
> It looks like the change is clipping off the right hand side for this test comparing the draw polygon 'diamond' versus the 'gfxdraw' one:

Actually its not `draw.aapolygon` issue. Remember this PR: #2912, I did not test it with 2px long lines. I will fix it soon and make new PR for that.
But still we should wait with merging this PR, until I implement width argument to `draw.aalines`, so `draw.aapolygon` will have width too.

mzivic7:
Ok to fix it I just have to disable my overlapping-aalines-fix when this and/or previous point coordinates are ints.
Should I: 
- Open new PR with new branch for this (its just ~15 lines), 
- Do it in upcoming PR for `draw.aalines` width, because it will get significantly modifying
- or just do it here?

MyreMylar:
New PR I think.

Starbuck5:
I was told on discord by Mzivic they are planning to modify the algorithm for this, so I am marking as a draft.

PR #3128, opened 2024-09-28
Add `dev.py` implementing dev command shorthands

ankith26:
This is an attempt at simplifying the steps a pygame-ce dev has to do to get out a complete working change that passes our CI.

Some goals
- help new devs figure stuff out faster. This script does an auto-install of all dependencies needed, so virtually no setup is needed to run this script. For people who don't like the idea of a script messing with the global package space, this script also provides a flag to do everything inside a venv.
- make the dev workflow more efficient, with shorthand commands for all the common steps one usually needs.
- get everyone onboard the editable install strategy, with one convenient command.
- serve as a self-documenting system (with help commands)


ankith26:
## Building pygame

`python dev.py build`

Builds and installs pygame in editable mode by default. Can pass a flag to do a regular install. Either way, this command is bound to be faster and efficient than a `pip install .` or an equivalent because it doesn't do build isolation and uses the same builddir every run.

Also has a flag to build in debug mode (optimizations turned off and debug symbols turned on) easily. Finally, this command also enables more warnings as errors by default to mimic CI as closely as possible. Again, there's a flag to opt out of this.

## Building docs

`python dev.py docs` builds the docs, `python dev.py build --full` does a full generation.

## Running code format

`python dev.py format` should just format the code for the peeps who don't like or can't use the `pre-commit` strategy. This still uses `pre-commit` internally, but does auto-install, and doesn't install any git hooks. The command should therefore be pretty much exactly like the legacy `setup.py format`, functionally.

## Running lint

`python dev.py lint` runs pylint and mypy tests on the codebase, mimic-ing CI.

## Running tests

`python dev.py test` is a simple shorthand for `python -m pygame.tests`, though as a little enhancement it can also take a list of modules to run tests on that subset only.

ankith26:
Converting it to a draft because there's still stuff to do.
- Most importantly, get more feedback on this PR.
- Make CI use this script directly where possible

dev.py
```diff
@@ -0,0 +1,271 @@
...
+        self.deps["all"] = set()
+        for k in self.deps.values():
+            self.deps["all"] |= k
+
+    def cmd_build(self):
+        editable = not self.args.get("install", True)
+        debug = self.args.get("debug", False)
+        lax = self.args.get("lax", False)
+
+        install_args = ["--no-build-isolation", "-Cbuild-dir=.mesonpy-build"]
```
damusss, line 112: Not sure if it's a me issue, but either the terminal or the actual command on Windows doesn't like at all the build folder starting with a dot. Not sure if adding quotes fixes it, but removed the dot did that sometimes when I ran into that error. I believe you are running on Linux, I can test further on Windows 10 if you'd like.

dev.py
```diff
@@ -0,0 +1,271 @@
...
+        pip_install(self.py, install_args)
+
+    def cmd_docs(self):
+        full = self.args.get("full", False)
+
+        pprint(f"Generating docs (with {full=})")
+        extra_args = ["full_generation"] if full else []
+        cmd_run([self.py, "-m", "buildconfig", "docs", *extra_args])
+
+    def cmd_lint(self):
```
damusss, line 135: Perhaps `gen_stubs.py` could be run aswell. Easy to miss if you add a new constant or edit typing.py.

dev.py
```diff
@@ -0,0 +1,271 @@
...
+
+    def cmd_format(self):
+        pre_commit = self.py.parent / "pre-commit"
+
+        pprint("Formatting code (with pre-commit)")
+        cmd_run(
+            [pre_commit if pre_commit.exists() else "pre-commit", "run", "--all-files"]
+        )
+
+    def cmd_test(self):
```
damusss, line 148: Might be asking too much, but basically running tests completely breaks the maximization of every open window when fullscreen is hit, so if the PR is not adding anything new the all command could benefit not running any tests. Maybe "mod none" could disable them. I hope I'm not sounding contraddictory or hypocritical.

dev.py
ankith26: It's a windows shell thing. It is perfectly legal for paths on windows to start with a dot.

I have updated the wiki with fixed quoting, so the new command should work on every shell now.

dev.py
ankith26: Good idea. But it's technically not linting. I'm not sure at what point should it be called

dev.py
ankith26: I mean, you can always hit Ctrl+C while doing the tests if you want to stop it. I think running it by default makes sense.

dev.py
damusss: Maybe while formatting?

dev.py
damusss: Fair enough then

dev.py
ankith26: hmm, how about adding a new subcommand for this? Say, `stubs`. I can also move the stub test from lint over to this subcommand.

damusss: oh yeah that sounds like a good plan

ankith26: I have implemented this

ankith26:
This PR is ready for review and suggestions. I have updated the CI to add a `dev.py all` test

MyreMylar (approved):
Lovely, these commands all work great locally on windows for me. 

Should simplify the setup, and onboarding for new developers - especially once we've documented all these on the developer wiki pages.

ankith26:
Just added one commit, for supporting the `--sdl3` option. So one can do `py dev.py build --sdl3`

damusss:
@ankith26  So the last time I tested it I noticed it was nice but a little bit slower than usual because every command checks if every single module is installed.
I think when dev.py finishes executing it should save a `.dev-cache.txt` file (gitignored ofc) that keeps track of the installed version of the needed dependencies.
This way when you start dev.py again instead of checking with pip (very slow) it only needs to check the contents of that file, and install things only if they aren't high enough or not present.
What do you think?

ankith26:
I would like to explore ways of speeding it up further, but I don't think `.dev-cache.txt` is the way to go. Basically, the user may make changes to the environment by using pip directly, and then the cache would be useless and incorrect. 

What I could try to do, is somehow add a faster way to check that the version requirements match via pip itself, and therefore eliminate the networking call which I think is the reason for most of the initial slowdown

ankith26:
In my latest commit I just implemented a solution that should speed up the startup time of the script.

Basically I was trying to unconditionally upgrade pip in the script, and this can be slow or even error out because it does networking. Now I added logic that tests for the pip version and only attempts the upgrade if it's below some version threshold set.

This way, this script now does 0 networking operations if all dependencies are in place and have expected versions.

dev.py
```diff
@@ -0,0 +1,475 @@
...
+
+        # Build command
+        build_parser = subparsers.add_parser("build", help="Build the project")
+        build_parser.add_argument(
+            "--wheel",
+            nargs="?",
+            const=DIST_DIR,  # Used if argument is provided without a value
+            default="",  # Used if argument is not provided at all
+            help=(
+                "Generate a wheel and do a regular install from it. By default, this"
```
Starbuck5, line 351: Missing space between "this" and "value".

pyproject.toml
```diff
@@ -54,11 +54,11 @@ pygame_ce = 'pygame.__briefcase.pygame_ce:PygameCEGuiBootstrap'
...
 requires = [
-  "meson-python<=0.16.0",
-  "meson<=1.5.1",
-  "ninja<=1.11.1.1",
+  "meson-python<=0.17.1",
+  "meson<=1.6.0",
+  "ninja<=1.12.1",
   "cython<=3.0.11",
-  "sphinx<=7.2.6",
+  "sphinx<=8.1.3",
```
Starbuck5, line 61: Are these just general package updates? I would like Sphinx in a different PR, usually each major version requires some docs reworking.

dev.py
```diff
@@ -0,0 +1,475 @@
...
+        )
+        build_parser.add_argument(
+            "--sdl3",
+            action="store_true",
+            help="Build against SDL3 instead of the default SDL2",
+        )
+        build_parser.add_argument(
+            "--coverage",
+            action="store_true",
+            help="Do a coverage build",
```
Starbuck5, line 375: What is a coverage build?

dev.py
```diff
@@ -0,0 +1,475 @@
...
+        )
+
+        # Docs command
+        docs_parser = subparsers.add_parser("docs", help="Generate docs")
+        docs_parser.add_argument(
+            "--full", action="store_true", help="Generate full docs"
```
Starbuck5, line 381: It always generates full docs, I would word this more as "forces them to be fully regenerated".

It's useful for developing themes or changing the build system, because sphinx doesn't pick up on those things and just keeps around the cache from last time because it sees the docs content hasn't changed.

Starbuck5:
This is great stuff! I put down a few comments, the one I care most about is the sphinx one. But once those are resolved I will approve.

dev.py
ankith26: nice catch

pyproject.toml
ankith26: I did check the docs after this change and everything seemed okay on a glance, but yeah maybe it should be a separate PR. The reason these bumps are needed is to get this PR pass msys2/multiarch tests because there we use preinstalled dependencies and if the version constraint does not match it triggers a reinstall, which can error on those exotic platforms

dev.py
ankith26: It is what andrew added to CI in https://github.com/pygame-community/pygame-ce/pull/2990

Basically it is a tool for checking our "test coverage". Compiling pygame with this flag, running the unit tests and then generating the "coverage report" is useful for analyzing how many codepaths our tests actually cover.

Starbuck5: I genuinely was asking, but I was also hinting that the help string could be more descriptive of what the command does.

pyproject.toml
Starbuck5: I don't fully understand, but a random msys2 environment having preinstalled deps should not be setting our version constraints for us.

Starbuck5 (approved):
Thanks for pushing some of the changes out of this PR.

Let's get this in!

PR #3129, opened 2024-09-28
Make `copy` methods explicit for linting

damusss:
I get annoyed by the fact `Object().copy()` is _blue_ (variable color) while `Object().__copy__()` is _yellow_ (method color). I think the public API method should be rightfully have the method color along with the magic dunder API.

Am I really making a PR because of a method color in my specific IDE theme? Yes.


ankith26 (commented):
Maybe you don't have to do an invert at all, another option is to be explicit and use the full line in both variants

damusss:
> Maybe you don't have to do an invert at all, another option is to be explicit and use the full line in both variants

Good idea, I just did that.

ankith26 (approved):

zoldalma999 (approved):
LGTM!

PR #3131, opened 2024-09-29
Added `geometry.Line` base

itzpr3d4t0r:
This pull request adds the `Line` class to the `pygame.geometry` module.

It also adds the ax, ay, bx, by, a, b attributes, str functionality (for printing), and the copy method. It also adds their associated docs and tests.

## Credits

**Geometry Project:**
For code, docs and tests:
Also thanks to @Starbuck5 for kickstarting the idea and the occasional help!

**Functionality added in this PR**
- `geometry.pyi`
https://github.com/pygame-community/pygame-geometry/blame/main/geometry.pyi
Credits to @Emc2356 @itzpr3d4t0r @novialriptide @ScriptLineStudios @avaxar

- `geometry.rst`
https://github.com/pygame-community/pygame-geometry/blame/main/docs/line.rst
Credits to @Emc2356 @itzpr3d4t0r @ScriptLineStudios @avaxar

- `line.c`
https://github.com/pygame-community/pygame-geometry/blame/main/src_c/line.c
Credits to @Emc2356 @itzpr3d4t0r @novialriptide @ScriptLineStudios @avaxar

- `geometry.h`
https://github.com/pygame-community/pygame-geometry/blame/main/src_c/include/geometry.h
Credits to @Emc2356 @itzpr3d4t0r @maqa41

- `geometry.c`
https://github.com/pygame-community/pygame-geometry/blame/main/src_c/geometry.c
Credits to @Emc2356 @itzpr3d4t0r @novialriptide @ScriptLineStudios @avaxar

- `geometry_test.py`
https://github.com/pygame-community/pygame-geometry/blame/main/test/test_line.py
Credits to @Emc2356 @itzpr3d4t0r @novialriptide @ScriptLineStudios @avaxar


damusss:
_woah_

buildconfig/stubs/pygame/geometry.pyi
```diff
@@ -13,13 +13,30 @@ from .rect import Rect, FRect
 from .math import Vector2
 
 _CanBeCircle = Union[Circle, Tuple[Coordinate, float], SequenceLike[float]]
+_CanBeLine = Union[
+    Rect,
+    Line,
+    Tuple[float, float, float, float],
+    Tuple[Coordinate, Coordinate],
+    Coordinate,
+    Sequence[Coordinate],
```
damusss, line 22: Unless you have another reasons, this should probably be `SequenceLike` from `pygame.typing`

damusss (changes requested):
In addition, is it really necessary to add a `..versionadded::` tag for every method and attribute? if the object was added in 2.5.2, it's kind of obvious that everything inside it was also added in the same version.

buildconfig/stubs/pygame/geometry.pyi
itzpr3d4t0r: > In addition, is it really necessary to add a ..versionadded:: tag for every method and attribute? if the object was added in 2.5.2, it's kind of obvious that everything inside it was also added in the same version.

I don't believe this is accurate. 
It's unclear to state the object was added in 2.5.2 without specifying when each attribute was introduced. We, as developers, know exactly what was added through changelogs, but regular users rely solely on the documentation to see what they can use. How can someone be certain the attribute they want was added in 2.5.2 if it's not specified? In that case, they would only know the object was added in that version, not the attributes.

damusss: Fair enough I guess, sure

buildconfig/stubs/pygame/geometry.pyi
```diff
@@ -115,3 +130,39 @@ class Circle:
...
+    @property
+    def b(self) -> Tuple[float, float]: ...
+    @b.setter
+    def b(self, value: Coordinate) -> None: ...
+    @overload
+    def __init__(self, line: Line) -> None: ...
+    @overload
+    def __init__(self, xa: float, ya: float, xb: float, yb: float) -> None: ...
+    @overload
+    def __init__(self, first: Coordinate, second: Coordinate) -> None: ...
```
damusss: Minor nitpick which could just be related to the porting, but it makes more sense to me to call "first" and "second" "a" and "b", which is more consistent with the other init overload

damusss: BRO I sent while you committed that exact change, no way

itzpr3d4t0r: Just did that.

damusss (approved):
Nothing else comes to mind, LGTM, thanks :)

buildconfig/stubs/pygame/geometry.pyi
```diff
@@ -8,19 +8,34 @@ from typing import (
...
+    SequenceLike[Point],
+]
 
 class _HasCirclettribute(Protocol):
     # An object that has a circle attribute that is either a circle, or a function
     # that returns a circle
     circle: Union[_CanBeCircle, Callable[[], _CanBeCircle]]
 
-_CircleValue = Union[_CanBeCircle, _HasCirclettribute]
+class _HasLineAttribute(Protocol):
```
MyreMylar, line 31:
```suggestion
_CircleValue = Union[_CanBeCircle, _HasCirclettribute]

class _HasLineAttribute(Protocol):
```

buildconfig/stubs/pygame/geometry.pyi
```diff
@@ -134,3 +149,38 @@ class Circle:
     def as_frect(self) -> FRect: ...
     def copy(self) -> Circle: ...
     def __copy__(self) -> Circle: ...
+    copy = __copy__
```
MyreMylar, line 152:
```suggestion
```

buildconfig/stubs/pygame/geometry.pyi
```diff
@@ -8,19 +8,34 @@ from typing import (
     List,
 )
 
+
```
MyreMylar, line 11:
```suggestion
```

src_c/geometry.h
```diff
@@ -15,12 +15,25 @@ typedef struct {
 
 #define pgCircle_CAST(o) ((pgCircleObject *)(o))
 #define pgCircle_AsCircle(o) (pgCircle_CAST(o)->circle)
-#define pgCircle_GETX(self) (pgCircle_CAST(self)->circle.x)
-#define pgCircle_GETY(self) (pgCircle_CAST(self)->circle.y)
-#define pgCircle_GETR(self) (pgCircle_CAST(self)->circle.r)
```
MyreMylar, line 20: are these deletions intentional?

MyreMylar: could be a merge error...

itzpr3d4t0r: Yes, they're intentional. Better keep unused stuff to a minimum.

MyreMylar (approved):
OK, LGTM 👍 

Tests, docs & code seem fine and it all works locally. Good work! 💯

buildconfig/stubs/pygame/geometry.pyi
```diff
@@ -14,13 +14,29 @@ from .rect import Rect, FRect
 from .math import Vector2
 
 _CanBeCircle = Union[Circle, Tuple[Point, float], SequenceLike[float]]
+_CanBeLine = Union[
```
damusss, line 17: This is just me being confused, would a tuple `(xa, ya, xb, yb)` be ok for a line? because it doesn't seem to get covered. also how can a `Rect` and `FRect` be considered lines? is it just because they are iterable, which will result in the line weirdly inheriting `xa=x, ya=y, xb=width, yb=height`? (if yes, perhaps it can be omitted in this union as already covered by `SequenceLike[float]` to avoid confusion).
Ok maybe it does get covered by `Point`, but I think you should actually be explicit with `SequenceLike[float]` because point is intended to be used as `(x, y)`, not a sequence with arbitrary length. If it was possible it would be annotated in the type itself but python doesn't allow it.

damusss (changes requested):

buildconfig/stubs/pygame/geometry.pyi
itzpr3d4t0r: Idk, I just wanted to be explicit about what you can pass and can be interpreted as a Line. Without that Rect/Frect people could randomly discover that and think it wasn't expected when infact it was a known thing all the time.

damusss: Alright then I'm only concerned about Point, because logically it doesn't make sense to create a line from a single Point, while SequenceLike[float] makes sense. I know Point and SequenceLike[float] are exactly the same, but their names convey different meanings, and Point isn't suited for this job.

About the degenerate thing, rather than removing it you could explicit the fact that is allowed but that some operations might not be available in that case. Not required of course, just my 2 cents about it!

aatle: It is best to remove `Rect` and `FRect` from the union (and give a warning in docs to prevent surprises).
Redundantly putting `Rect` in the union may mislead users into thinking that `Rect` is intuitively or specially supported, when this is not the case.

I agree that `SequenceLike[float]` should be used rather than `Point`. This construct is already being used for `RectLike`.

buildconfig/stubs/pygame/geometry.pyi
```diff
@@ -134,3 +150,37 @@ class Circle:
...
+    @b.setter
+    def b(self, value: Point) -> None: ...
+    @overload
+    def __init__(self, xa: float, ya: float, xb: float, yb: float) -> None: ...
+    @overload
+    def __init__(self, a: Point, b: Point) -> None: ...
+    @overload
+    def __init__(self, line: _LineValue) -> None: ...
+    def __copy__(self) -> Line: ...
+    copy = __copy__
```
aatle, line 186: In accordance with a recent PR, the `copy` method should have the full type annotation for explicitness.

aatle:
Is there any specific reason why the coordinates are named with `xa`, `yb` instead of `ax`, `by`?
The current is annoying for me to parse mentally, as it looks backwards and the 'subscript' appears in front of the point.

buildconfig/stubs/pygame/geometry.pyi
itzpr3d4t0r: Alright, fixed that I suppose.

damusss (approved):
Alright thanks, changes LGTM

PR #3137, opened 2024-10-02
wasm: controller is not multiphase init + workaround for wasm 313t

pmp-p:
also only controller.c is needed, the cython based controller_old is redundant here. Still fence possible use of cython code for future cpython versions.


MyreMylar (approved):
OK, LGTM 👍

ankith26 (approved):
LGTM, thanks! :partying_face:

PR #3140, opened 2024-10-04
Add draw.aaline width argument

mzivic7:
This PR adds width argument to `draw.aaline`.

How it works:
First draw `draw_line_width` (its not antialiased).
Calculate its corner points, but slightly move them so relevant antialiased pixels don't overlap with fully opaque pixels.
Then draw 2 thin `draw_aaline`.

I made this calculation in separate function `line_width_corners`, because it will be used again when implementing width for `draw.aalines` and `draw.aapolygon`.

Screenshot From 2024-10-04 20-53-09

<details>

  ```py
import pygame

pygame.init()
clock = pygame.time.Clock()
run = True
while run:
        if event.type == pygame.QUIT:
            run = False
    screen.fill("black")
    pygame.draw.aaline(screen, "white", (120, 100), (220, 160), 16)
    pygame.draw.aaline(screen, "red", (180, 170), (100, 260), 16)
    pygame.draw.aaline(screen, "blue", (70, 240), (150, 150), 7)
    pygame.display.flip()
    clock.tick(60)
pygame.quit()
  ```
</details>


bilhox (approved):
LGTM, big thanks for this work!

It will just need a little update with upstream and test fixes.

bilhox (changes requested):
Actually I forgot something important, the final step is also to add new unittests.

damusss:
> Actually I forgot something important, the final step is also to add new unittests.

well also perhaps it'd be nice if there weren't 15 failed checks lmao

bilhox:
Judging by how much line got updated/added, I'm moving this PR to 2.5.2 milestone.

bilhox (approved):
LGTM

MyreMylar (approved):
OK, LGTM. Line output looks good in local testing and the code makes sense.

👍 🎆 🍾

PR #3141, opened 2024-10-05
Add meson flag for compiling with SDL3

ankith26:
Add a flag to allow compilation with SDL3. But for obvious reasons we still default to compiling with SDL2.

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -84,6 +84,10 @@ sdl_mixer = '@0@_mixer'.format(sdl)
 sdl_ttf = '@0@_ttf'.format(sdl)
 sdl_image = '@0@_image'.format(sdl)
 
+if sdl_api == 3
+    add_global_arguments('-DPG_SDL3=1', language: 'c')
+endif
+
 pg_inc_dirs = []
 pg_lib_dirs = []
 if plat == 'win' and host_machine.cpu_family().startswith('x86')
--- a/meson.build
+++ b/meson.build
@@ -78,6 +78,12 @@ endif
 
 pg_dir = py.get_install_dir() / pg
 
+sdl_api = get_option('sdl_api')
+sdl = 'SDL@0@'.format(sdl_api)
+sdl_mixer = '@0@_mixer'.format(sdl)
+sdl_ttf = '@0@_ttf'.format(sdl)
+sdl_image = '@0@_image'.format(sdl)
+
 pg_inc_dirs = []
 pg_lib_dirs = []
 if plat == 'win' and host_machine.cpu_family().startswith('x86')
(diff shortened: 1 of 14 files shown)


meson.build
```diff
@@ -106,47 +111,47 @@ if plat == 'win' and host_machine.cpu_family().startswith('x86')
...
     sdl_lib_dir = sdl_dir / 'lib' / arch_suffix
     pg_inc_dirs += fs.relative_to(sdl_dir / 'include', base_dir)
     pg_lib_dirs += sdl_lib_dir
-    dlls += sdl_lib_dir / 'SDL2.dll'
+    dlls += sdl_lib_dir / '@0@.dll'.format(sdl)
 
-    # SDL2_image
-    sdl_image_dir = prebuilt_dir / 'SDL2_image-@0@'.format(sdl_image_ver)
+    # SDL_image
+    sdl_image_dir = prebuilt_dir / '@0@-@1@'.format(sdl, sdl_image_ver)
```
MyreMylar, line 122:
```suggestion
    sdl_image_dir = prebuilt_dir / '@0@-@1@'.format(sdl_image, sdl_image_ver)
```

MyreMylar (approved):
OK, looks good! 👍

ankith26:
Converting to draft because I want to do a bit more in this PR

ankith26:
- Added a CI for testing that we can build against SDL3
- Add SDL3 guards on meson code. The purpose of this is for incremental porting: we can port modules one by one and keep removing the SDL3 guards as we go
- Fixed SDL3 include. `#include <SDL.h>` no longer works (ref: https://github.com/libsdl-org/SDL/issues/11074)

ankith26:
This PR is now done from my end

Starbuck5:
With regards to the include changes, in my development I've just made sure the include path goes inside the SDL folder instead of editing every include.

Starbuck5:
Another thing I did in my branch to make this stuff compile better was `-DSDL_ENABLE_OLD_NAMES=1`, which is very handy.

Also because of my prior PRs several modules you've masked out in this compile, at least last time I checked (may have changed). Like key, image, base, etc.

ankith26:
The latest changes have broken some of your patches. For instance, `PG_SURF_BitsPerPixel`/`PG_SURF_BytesPerPixel` is now not implementable without an extra function call and a null check.
When I tested this branch locally none of the modules could compile. But I wasn't aware of `-DSDL_ENABLE_OLD_NAMES=1`.

> With regards to the include changes, in my development I've just made sure the include path goes inside the SDL folder instead of editing every include.

That was my first idea but it is virtually impossible to do from our end in a cross platform way (see my older comment with the link to the SDL issue).

Starbuck5:
> The latest changes have broken some of your patches. For instance, PG_SURF_BitsPerPixel/PG_SURF_BytesPerPixel is now not implementable without an extra function call and a null check.

My first reaction: oh boy 😢 

However-- We can use macros on the pixelformat itself, `SDL_BYTESPERPIXEL` and `SDL_BITSPERPIXEL`. Whew.

> That was my first idea but it is virtually impossible to do from our end in a cross platform way (see my older comment with the link to the SDL issue).

I don't know anything about .pc files.

It makes me annoyed at meson that we don't have the freedom to do that. It seems like we should be able to add an include path to an existing dependency object, which would solve this because we could add an include path which is the old include path + "/sdl3".

Starbuck5:
It's so easy on Windows with the old SETUP file that I didn't even bat an eye when I rewrote the include path.

Starbuck5 (approved):

PR #3142, opened 2024-10-05
New up-to-date review of french readme + note of when it was last reviewed

bilhox:
This PR is an up-to-date review of my french translation created a year ago, I mainly changed some wording I was using which wasn't natural, and more importantly, I added a note at the end of the file saying when it was last reviewed.

### DIFF
--- a/docs/readmes/README.fr.rst
+++ b/docs/readmes/README.fr.rst
@@ -176,8 +176,7 @@ au domaine public.
 
 Voir les docs/licences pour les licences des dépendances.
 
-**Last reviewed** : 05 October 2024
-**Dernière vérification** : 05 Octobre 2024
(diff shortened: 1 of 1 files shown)


damusss:
@bilhox it's actually very similar to what I did, I only added an header line before. would you not agree it would be fitting?

bilhox:
> @bilhox it's actually very similar to what I did, I only added an header line before. would you not agree it would be fitting?

yes, just add also an english version of the date like I did.

MyreMylar (approved):
LGTM 👍

TheRealGabHas (commented):
I think all the changes are good, they make the reading process more natural.

ankith26 (approved):
Thanks!

PR #3143, opened 2024-10-05
Drop python 3.8 support

ankith26:
Python 3.8 is EOL. We can now consider dropping it.

The biggest advantage of doing this is nicer type stubs.


bilhox:
Could we first release 2.5.2 then drop the support for 3.8 ?

Starbuck5:
I agree with bilhox, no rush to drop it.

ankith26:
> Could we first release 2.5.2 then drop the support for 3.8 ?

Yeah definitely. I just wanted to put this PR out there. We can wait a couple of months on this.

buildconfig/stubs/pygame/event.pyi
```diff
@@ -1,22 +1,17 @@
...
-    List,
-    Optional,
-    Union,
-    final,
-)
+from typing import Any, Optional, Union, final
 
 from pygame.typing import SequenceLike
 
+_AttrDict = dict[str, Any]
```
Starbuck5, line 5: I don't see why this was changed.

It makes it harder to understand the popups in VScode, because now it just says the parameter is supposed to be an _AttrDict.

buildconfig/stubs/pygame/geometry.pyi
Starbuck5, line 1: Changes Sequence to pygame.typing.SequenceLike.

ankith26: That is an intentional fix. Though I can see that it is out of scope of this PR, if needed I can separate it out into its own thing

buildconfig/stubs/pygame/event.pyi
ankith26: I remember hitting some weird mypy issue where using the type directly gave some errors, and doing this fixed it

Starbuck5: I would like to see that in a separate PR, along with the geometry sequencelike change.

Unless you're saying that using `dict` here leads to an issue that `Dict` doesn't have, in which case I'd say leave this one as `Dict`.

buildconfig/stubs/pygame/geometry.pyi
ankith26: I separated the fix out into its own PR, while also removing the fix from this PR entirely.

src_py/_debug.py
```diff
@@ -4,13 +4,13 @@
 import sys
 import traceback
 import importlib
-from typing import Tuple, Optional, Callable
+from typing import Optional, Callable
```
zoldalma999, line 7: I'd guess this should switch to `collection.abc.Callable` too.

Though I don't think `ImportResult` (the only thing using it) needs to be defined here? Stubs already have typing for this module.

ankith26: I made the switch to `collection.abc.Callable`. Even though I agree `ImportResult` can be removed, I will not be doing that in this PR to keep this as focused on the goal as possible

damusss (approved):
LawesomeTM

PR #3144, opened 2024-10-05
Missing pixels in `aalines` when using integer coordinates

mzivic7:
It turned out that those endpoints I disable at specific situations are needed when points coordinates are integer.
So it created gaps on edges. And those gaps are making `draw.aapolygon` PR tests to fail. (#3126)
I just made it so disabling endpoints happens only for points that have float coordinates.

Both aalines are drawn with int values, left is before, right after.
Screenshot From 2024-10-05 19-53-27

<details>

  ```py
import pygame
pygame.init()
screen = pygame.display.set_mode((100, 100))
run = True
while run:
        if event.type == pygame.QUIT:
            run = False
    screen.fill("black")
    # using ints, point of interest: (5, 3)
    pygame.draw.aalines(screen, "white", True, [(1, 3), (3, 5), (5, 3), (3, 1)])
    # using floats, point of interest: (11, 3.1)
    pygame.draw.aalines(screen, "white", True, [(7, 3), (9, 5), (11, 3.1), (9, 1)])
    pygame.display.flip()
pygame.quit()
  ```
</details>

### DIFF
--- a/src_c/draw.c
+++ b/src_c/draw.c
@@ -283,6 +283,7 @@ aalines(PyObject *self, PyObject *arg, PyObject *kwargs)
     float x, y;
     int l, t;
     int extra_px;
+    int disable_endpoints;
     int steep_prev;
     int steep_curr;
     PyObject *blend = NULL;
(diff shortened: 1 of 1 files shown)


damusss (approved):
_yes_

PR #3146, opened 2024-10-06
SDL3 support for pygame.system

Starbuck5:
A little bit more gnarly than I would like, but the code comment explains my approach here.

### DIFF
--- a/src_c/meson.build
+++ b/src_c/meson.build
@@ -321,8 +321,6 @@ newbuffer = py.extension_module(
 )
 
 # new/experimental/uncommon stuff, but built by default
-# TODO: support SDL3
-if sdl_api != 3
 system = py.extension_module(
     'system',
     'system.c',
@@ -331,7 +329,6 @@ system = py.extension_module(
     install: true,
     subdir: pg,
 )
-endif
 
 geometry = py.extension_module(
     'geometry',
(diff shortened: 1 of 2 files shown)


damusss (approved):
LGTM :)

ankith26 (changes requested):
```c
SDL_Locale **
PG_GetPreferredLocales(int *num_locales)
{
    return SDL_GetPreferredLocales(num_locales);
#elif SDL_VERSION_ATLEAST(2, 0, 14)
    SDL_Locale *locales = SDL_GetPreferredLocales();
    if (locales) {
        SDL_Locale *current_locale = locales;
        *num_locales = 0;
        /* The array is terminated when the language attribute of the last
         * struct in the array is NULL */
        while (current_locale->language) {
            (*num_locales)++;
            current_locale++;
        }
        return &locales;
    }
#endif
    return NULL;
}
```

Untested code, somewhat based on your idea but written so that it can be abstracted out in a separate function

Starbuck5:
SDL2: array of SDL_locales
SDL3: array of pointers to SDL_locales

Is a pointer to an array of SDL_locales the same thing as an array of pointers to SDL_locales ?

This is what I was sidestepping worrying about with my implementation but would need to think about for your code suggestion.

ankith26:
okay NVM my suggestion, they are probably not the same

ankith26 (approved):
LGTM, thanks :star2:

Issue #3149, opened 2024-10-06
How to fix the recent `TypeAlias` addition


damusss:
A recent PR marked types inside `typing.py` with `TypeAlias`, but that object is only available from python 3.10.
We now know how it passed the checks: inside `__init__.py` we catch import errors, and importing TypeAlias from typing before 3.10 results in an import error, so pygame.typing becomes a missing module with no exception. Once a PR fixing this is merged, the unit test for typing can be merged and issues like this won't appear again.

We can fix this in two ways:
- remove the addition of `TypeAlias`
- version check the relevant bits of the files to use `TypeAlias` on 3.10+ but use the old system otherwise

The one choosing should be @zoldalma999 as they made the original PR.
:3

aatle:
Note that `typing.py` and `typing.pyi` should exclude `TypeAlias` in either case, because it is not guaranteed that `typing_extensions` will be available at runtime.

ankith26:
The linked PR puts up what I think is the best fix for now.

PR #3150, opened 2024-10-06
Add basic pygame.typing unit test, remove problematic TypeAlias usage

ankith26:
To catch issues like #3149

### DIFF
--- a/buildconfig/stubs/pygame/typing.pyi
+++ b/buildconfig/stubs/pygame/typing.pyi
@@ -14,7 +14,7 @@ __all__ = [
 
 import sys
 from abc import abstractmethod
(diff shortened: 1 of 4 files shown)


damusss (approved):
Yeah, good idea. LGTM!

ankith26:
I have implemented a commit fixing the issue in this PR.

For now, the best fix seems to be reverting the `TypeAlias` changes entirely. And when our minimum supported python version becomes 3.12 (which should about 3 years from now), we can consider using the new `type` keyword and the other fancy new typing stuff.

damusss (approved):
I agree. Still looking good to me

Starbuck5 (approved):
Thanks!

PR #3156, opened 2024-10-08
Fix handling all color types in `pygame.transform.threshold`

damusss:
The c internal function `_color_from_obj` only allows mapped colors for some reason, but all color arguments in the stubs for `pygame.transform.threshold` pretend to allow `ColorLike` which also allows strings. My modification makes sure all the promised color types are allowed and add a test to make sure it doesn't break in the future.

`pygame.transform.threshold` should be the only function using the internal c function because all new api coming for transform is going to use the pygame C api function by default.

### DIFF
--- a/src_c/transform.c
+++ b/src_c/transform.c
@@ -1830,7 +1830,7 @@ _color_from_obj(PyObject *color_obj, SDL_PixelFormat *format,
 {
     if (color_obj) {
         if (!pg_MappedColorFromObj(color_obj, format, color,
-                                   PG_COLOR_HANDLE_INT)) {
+                                   PG_COLOR_HANDLE_ALL)) {
             return -1;
         }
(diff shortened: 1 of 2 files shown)


ankith26 (approved):

As a minor side note, maybe we could add a line in the docs about this, as it is technically new behaviour.

damusss:
It is new behaviour but it wasn't a common and long awaited bugfix like the one about filling with negative topleft area, the c code was just betraying the stubs (someone encountering a value error before this fix should have reported it). Stub-wise, nothing changed, so I think it's not too big of a deal to add a line in the docs. A note in the changelog for 2.5.2 should suffice the people that were using the function anyways.

Temmie3754 (approved):
Looks and runs fine for me. 👍

Issue #2269, opened 2023-06-22
Method to normalize zero vectors


tigeryzhang:
**Description**
When normalizing a vector, it is necessary to first check it is not of length zero, otherwise the program will crash.

The game engine bevy has a method for its `Vec`s that is `normalize_or_zero()` that allows you to normalize a vector that is potentially of length zero, where it returns a zero vector if that is the case, otherwise normalizing it.

I think this would be useful to have in pygame, as it saves having to type an extra line pretty much every time you want to normalize a vector.

bilhox:
Mathematically speaking, it's impossible to normalize a zero vector, so returning a zero vector would be weird. That's why, I don't agree with you.

Yu266426:
Well, I don't agree in the sense that just because something doesn't "make sense mathematically" means that a more convenient method shouldn't exist.

But also, that's why the method is called `normalize_OR_ZERO`, which I think helps illustrate that it isn't just normalizing it.

bigwhoopgames:
I would be in support of having this method added. In nearly ever instance of me using normalize() I use a zero vector if I can't normalize.

Matiiss:
I like the idea of having a way to conveniently get the zero vector back when normalizing, the proposed method's name is quite clear in what it would do and thus I don't see how it wouldn't make sense mathematically either, you either can normalize or you return a zero vector.

An alternative idea would be to add an optional argument to `normalize`, like `zero` that would achieve the same thing: `vec.normalize(zero=True)` (perhaps, it could be a positional-only argument for performance: `vec.normalize(True)` albeit that makes it less readable, but maybe using keyword args doesn't make it all that slow, so that should be tested)

DaNubCoding:
Perhaps `xnormalize` is also an option, meaning "exclusive normalize", where it will normalize everything excluding zero vector in which case it is handled differently. This is easier to type and doesn't clutter code with long method names. Stuff like FRect uses a similar method of naming.

Mega-JC:
I'm not sure if pygame needs this to be honest, especially if there is already a super simple solution to this problem of getting a potential `ValueError`: Simply checking it using an if statement, or short-circuting with `and` by writing the expression `vector and vector.normalize()`. In the latter case, if `vector` is a zero vector, the expression will simply  return `vector`. If `vector` is not a zero vector, `vector.normalize()` will be returned instead. This seems to be the exact behavior desired here. 

Another example:
```py
>>> from pygame import Vector2
>>> 
[...]

PR #3157, opened 2024-10-08
Add ability to normalize a Vector of length zero

Doublestuf:
In issue #2269, it's been decided that pygame-ce should follow game engines such as Unity and Godot in allowing users to normalize zero-vectors for convenience.

Update the function in `math_test.c`, tests in `math_test.py`, and documentation in `math.rst` accordingly.

### DIFF
--- a/docs/reST/ref/math.rst
+++ b/docs/reST/ref/math.rst
@@ -237,6 +237,11 @@ Multiple coordinates can be set using slices or swizzling
       Returns a new vector that has ``length`` equal to ``1`` and the same
       direction as self.
 
+      .. versionchanged:: 2.5.2 It is now possible to use ``normalize`` on a zero-vector.
+
+      .. note::
+         Before pygame-ce 2.5.2, attempting to normalize a zero vector would always raise a ``ValueError``
+
       .. ## Vector2.normalize ##
 
    .. method:: normalize_ip
@@ -246,6 +251,11 @@ Multiple coordinates can be set using slices or swizzling
 
       Normalizes the vector so that it has ``length`` equal to ``1``.
       The direction of the vector is not changed.
+
+      .. versionchanged:: 2.5.2 It is now possible to use ``normalize_ip`` on a zero-vector.
+
+      .. note::
+         Before pygame-ce 2.5.2, attempting to normalize a zero vector would always raise a ``ValueError``
 
       .. ## Vector2.normalize_ip ##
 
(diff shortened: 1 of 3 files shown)


docs/reST/ref/math.rst
```diff
@@ -237,6 +237,11 @@ Multiple coordinates can be set using slices or swizzling
       Returns a new vector that has ``length`` equal to ``1`` and the same
```
oddbookworm, line 237: You should update this if we're gonna go this route, as well as the sluglines above

docs/reST/ref/math.rst
```diff
@@ -247,6 +252,11 @@ Multiple coordinates can be set using slices or swizzling
       Normalizes the vector so that it has ``length`` equal to ``1``.
```
oddbookworm, line 252: Update this and the sluglines above

test/math_test.py
```diff
@@ -742,19 +742,20 @@ def test_normalize(self):
...
         # v has length != 1 before normalizing
         self.assertNotEqual(v.x * v.x + v.y * v.y, 1.0)
         # inplace operations should return None
         self.assertEqual(v.normalize_ip(), None)
+        self.assertEqual(self.zeroVec.normalize_ip(), None)
         # length is 1
         self.assertAlmostEqual(v.x * v.x + v.y * v.y, 1.0)
         # v2 is parallel to v1
         self.assertAlmostEqual(self.v1.x * v.y - self.v1.y * v.x, 0.0)
-        self.assertRaises(ValueError, lambda: self.zeroVec.normalize_ip())
```
oddbookworm, line 757: These changes also need to be applied to the `Vector3` tests

oddbookworm (changes requested):
I'm not sure that it was actually determined for sure that this needs to be implemented in pygame-ce. I'm still partially against it because there are several nonintuitive things that could break if you expect normalizing to return a vector of magnitude 1

docs/reST/ref/math.rst
oddbookworm: Also, you still need to update the `Vector3` docs with these changes as well

Doublestuf:
Good point -- this adds the possibility of `example_vector.normalize().is_normalized()` returning False, which could complicate things and essentially defeat the point of this change. I think [this](https://github.com/pygame-community/pygame-ce/issues/2269#issuecomment-1606507408) idea of making a new function called `xnormalize` while leaving the current `normalize` function would be more intuitive.

bilhox:
> I'm not sure that it was actually determined for sure that this needs to be implemented in pygame-ce. I'm still partially against it because there are several nonintuitive things that could break if you expect normalizing to return a vector of magnitude 1

@oddbookworm  Could you provide an example ? If we merge this, it will not break any code as the following structure : 
```py
if vector:
  vector.normalize_ip()
```
Would still work. (which is what people are currently doing / supposed to do)

bilhox:
@Doublestuf So the test can pass, you'll have to update the test for ``normalize``, because it's currently expecting an error.

Issue #2727, opened 2024-02-25
`pygame.draw.ellipse` behaves inconsistently when provided rect `width` or `height` are negative


denballakh:
**Environment:**
- `pygame-ce 2.4.1 (SDL 2.28.5, Python 3.12.0)`:

**Current behavior:**
Consider this code:
```py

import pygame as pg
s = pg.display.set_mode((250, 250))
# draw the outline where other circles will (or will not) be drawn:
pg.draw.ellipse(s, 'darkgrey', ( 50,  50,  50,  50), width=1)
pg.draw.ellipse(s, 'darkgrey', (150,  50,  50,  50), width=1)
pg.draw.ellipse(s, 'darkgrey', ( 50, 150,  50,  50), width=1)
pg.draw.ellipse(s, 'darkgrey', (150, 150,  50,  50), width=1)
# actual bug:
pg.draw.ellipse(s, 'red',      ( 50,  50,  50,  50), width=1) # normal
pg.draw.ellipse(s, 'green',    (150,  50,  50, -50), width=1) # not drawn
pg.draw.ellipse(s, 'blue',     ( 50, 150, -50,  50), width=1) # drawn, but to the left and filled
pg.draw.ellipse(s, 'yellow',   (150, 150, -50, -50), width=1) # not drawn

pg.display.update()
while True: pg.event.get()
```
When I run it I see this:
![image](https://github.com/pygame-community/pygame-ce/assets/47365157/baa292cf-dbfa-4685-90d0-94ee82911bbb)

2/3 weird circles are not draws (its ok), one of them is drawn to the left and is filled.

If I do not provide `width=1`, then same issue appears:
![image](https://github.com/pygame-community/pygame-ce/assets/47365157/eb74adef-3630-41a3-8f86-28e556aff735)

**Expected behavior:**
Either all weird-circles are draw, or none of them.

Gabryel-lima:
I'm trying to see if I can fix this, for me it gave me the same problem as you reported.

Gabryel-lima:
So, is the negative value really necessary? I think there is no need for a negative value. And in the #Normal annotation you are superimposing the red color on top of the dark gray

oddbookworm:
[...]

PR #3158, opened 2024-10-09
Add warning when using negative dimensions in draw.ellipse

mzivic7:
Using negative rect dimension values have some undesirable results as shown in #2727.
This PR adds deprecation warning to `draw.ellipse` when using negative value for rect width and height.
And prevents ellipse from being drawn at wrong position. Nothing is drawn instead.
Warning is also added to docs, and is tested for in tests.
<details>

  ```py
  import pygame
pygame.init()
clock = pygame.time.Clock()
run = True
while run:
        if event.type == pygame.QUIT:
            run = False
    screen.fill((0, 0, 0))
    pygame.draw.ellipse(screen, "white", (150, 150, -50, 30), 0)
    pygame.display.flip()
    clock.tick(60)
pygame.quit()
  ```
</details>

### DIFF
--- a/src_c/draw.c
+++ b/src_c/draw.c
@@ -706,6 +706,17 @@ ellipse(PyObject *self, PyObject *arg, PyObject *kwargs)
                             PG_SURF_BytesPerPixel(surf));
     }
 
+    if (drawn_area[0] - drawn_area[2] < 0 || drawn_area[2] < 0 ||
+        drawn_area[1] - drawn_area[3] < 0 || drawn_area[3] < 0) {
+        if (PyErr_WarnEx(PyExc_DeprecationWarning,
+                         "Negative values in position and dimension rect are "
+                         "deprecated and have no functionality. This will "
+                         "cause an error in a future version of pygame-ce.",
+                         1) == -1) {
+            return pgRect_New4(0, 0, 0, 0);
+        }
+    }
+
     CHECK_LOAD_COLOR(colorobj)
 
     if (width < 0) {
--- a/src_c/draw.c
+++ b/src_c/draw.c
@@ -706,17 +706,6 @@ ellipse(PyObject *self, PyObject *arg, PyObject *kwargs)
                             PG_SURF_BytesPerPixel(surf));
     }
 
-    if (drawn_area[0] - drawn_area[2] < 0 || drawn_area[2] < 0 ||
-        drawn_area[1] - drawn_area[3] < 0 || drawn_area[3] < 0) {
-        if (PyErr_WarnEx(PyExc_DeprecationWarning,
-                         "Negative values in position and dimension rect are "
-                         "deprecated and have no functionality. This will "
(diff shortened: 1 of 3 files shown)


src_c/draw.c
```diff
@@ -716,14 +716,25 @@ ellipse(PyObject *self, PyObject *arg, PyObject *kwargs)
...
-        width >= MIN(rect->w / 2 + rect->w % 2, rect->h / 2 + rect->h % 2)) {
-        draw_ellipse_filled(surf, rect->x, rect->y, rect->w, rect->h, color,
-                            drawn_area);
+    if (rect->w < 0 || rect->h < 0) {
+        if (PyErr_WarnEx(PyExc_DeprecationWarning,
+                         "Negative rect dimension values are deprecated and "
+                         "have no functionality. This will cause an error in "
+                         "a future version of pygame-ce.",
+                         1) == -1) {
+            return pgRect_New4(0, 0, 0, 0);
```
ankith26, line 725: This should be `return NULL;`

bilhox (approved):
It seems fair to me, LGTM !

ankith26:
while thinking about this, I realised that some other draw API also takes in rect style arguments. How are negative dims handled everywhere else. Whatever fix we do, we should try to be consistent with the approach everywhere

mzivic7:
I added same warning, docs and tests to `draw.rect` and `draw.arc`

ankith26 (commented):
Can you make all the 2.5.2 -> 2.5.3 in this PR?

mzivic7:
Sure, here it comes.

ankith26 (approved):

Starbuck5 (changes requested):
I'm not convinced this is the thing to do.

If it's a problem have the algorithm normalize the rect first?

Starbuck5:
Did some testing,

- pygame.draw.arc: Seems to not draw if the rect dimensions are negative
- pygame.draw.ellipse: Described in original issue, but my testing TLDR is that negative rect widths are drawn, negative rect heights are not, and negative rect widths are always drawn filled for some reason.
- pygame.draw.rect: With or without outline nothing is drawn with negative dimensions
- pygame.draw.rect (rounded corners): Interestingly rects with rounded corners draw with negative dimensions fine.

I still think the most straightforward way to fix this is normalize the rect arguments when they come in. (https://pyga.me/docs/ref/rect.html#pygame.Rect.normalize)

mzivic7:
Why is normalizing input rects bad:
- Arc can be non-symmetrical, so if its rect is normalized, should it be mirrored?
- Normalize function will have no purpose, because every rect input is normalized internally.
- One drags right side of rect over left side and get negative width, and he want to calculate area or perimeter, but it turns out to be wrong, so he then has to call normalize() to get correct area or use few abs().
- Enforced normalization would only add more cpu load, those few who needs this functionality just have to use normalize(), and probably already do.

So I will remove warning and only disable drawing if rect width/height is negative.

PR #3159, opened 2024-10-10
Fix thick draw.aaline with float coordinates

mzivic7:
`draw.aaline` with width argument has 2 problems:
First: I used floats to detect steepines, but `draw.line` converts values to ints before math. That's why added aalines, when width > 1, are drawn at wrong position. (left on the image)
Second: I used floats for original line coordinates to calculate positions for antialiased lines. But `draw.line` converts values to ints, so antialiased lines got a little too far and created gaps. (right on the image)
badcode

<details>

  ```py
import pygame
pygame.init()
clock = pygame.time.Clock()
run = True
while run:
        if event.type == pygame.QUIT:
            run = False
    screen.fill((0, 0, 0))
    pygame.draw.aaline(screen, "white", (140, 259.93), (150, 269.47), 20)
    pygame.draw.aaline(screen, "white", (170, 259.93), (190, 269.47), 20)
    pygame.display.flip()
    clock.tick(60)
pygame.quit()
  ```
</details>

### DIFF
--- a/src_c/draw.c
+++ b/src_c/draw.c
@@ -1875,7 +1875,8 @@ line_width_corners(float from_x, float from_y, float to_x, float to_y,
 {
     float aa_width = (float)width / 2;
     float extra_width = (1.0f - (width % 2)) / 2;
-    int steep = fabs(to_x - from_x) <= fabs(to_y - from_y);
+    // "steep" is same as "xinc" in draw_line_width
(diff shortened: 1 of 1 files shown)


bilhox (approved):

mzivic7:
Because this implementation uses Bresenham's algorithm for drawing thick lines, it does not support float values for coordinates.
New PR will be open with new implementation, using modified Xaolin Wu's algorithm.

PR #3160, opened 2024-10-10
Adjust pg_MappedColorFromObj API for SDL3-porting

Starbuck5:
This PR gets the color module compiling in SDL3.

Why so many files changed? The one problem in Color was with a C-API function, so I needed to adjust all the callers of the function.

In SDL3, surf->format is not a good enough source of information about the surface to do things like MapRGB, because it doesn't have palette information. Therefore I think it makes sense to pass around surfaces instead of SDL_PixelFormats.


ankith26 (approved):
LGTM, thanks :tada:

damusss (approved):
Alright, I agree. LGTM 😄

PR #3163, opened 2024-10-10
wasm: better wasm fencing

pmp-p:


damusss (approved):
Makes sense, LGTM :)

ankith26 (commented):
A lot of places in the codebase use `__EMSCRIPTEN__` for webasm fencing, what about those usages, should they all be updated to have wasi fencing too?

pmp-p:
each fence need invididual review wasi and web can be very different, in case of timers it was not clear how to make them work BOTH on wasi and web but it may be possible to run them from pump()

Let's assume `defined(__EMSCRIPTEN__) || defined(__wasi__)` means there is a way to do both and just __EMSCRIPTEN__ would mean unknown yet.

ankith26 (approved):
Okay, LGTM. Thanks! :partying_face:

PR #3164, opened 2024-10-10
Allow all color types mask.from_threshold and fix typing for PixelArray methods

damusss:
I understand why slice-assignment of pixelarrays limits the allowed types, but this was likely not intended for the `replace()` and `extract()` methods, as they were typehinted with `ColorValue` but don't currently allow all color values.
This PR adds a C function specifically for those methods, making the typehints true.

The same thing applies to `mask.from_threshold`.

If you think they should not accept all color types, I can change the PR to change the typehints instead, of course.


ankith26 (commented):
As a matter of personal preference, I would like to have consistent handling within `pixelarray` for color values. As it is sort of documented at the module level, it would make sense to keep that behaviour everywhere in the module. Therefore I think this case is a "stubs should be fixed" thing

The mask change on the other hand, I am fine with.

ankith26 (approved):

zoldalma999 (approved):
Looks good to me!

PR #3165, opened 2024-10-10
Improve color notation in the docs

damusss:
Notation about color values in the docs was messy and scattered. Since `pygame.typing.ColorLike` exists, the docs should point to it to be elegant. (Note: even if 2.5.2 is not out yet, clicking on the data tag will still bring a user to a list of allowed values).

color.rst was not changed as it dives deeper on the allowed values, nor gfxdraw.rst since 1) it's not supposed to be used, 2) not all color values are actually supported, but we shouldn't worry about fixing any of this since it's not supposed to be used.

I also modified referring to RGB when it should have actually said Color (pygame.Color) or color (pygame.typing.ColorLike).


bilhox (approved):
LGTM !

ankith26 (approved):
LGTM, thanks :tada:

PR #3166, opened 2024-10-11
Image module SDL3 support

Starbuck5:
- Rework argument order of PG_CreateSurfaceFrom to match newest SDL3 (they changed it)
- Deal with all the pixelformat fallout throughout the file
- Fix TGA saving routine file writes
- Switch out SDL_SetPaletteColors (all colors) for SDL_SetSurfacePalette

If desired, I could split this out into a couple smaller PRs.

### DIFF
--- a/src_c/_pygame.h
+++ b/src_c/_pygame.h
@@ -98,8 +98,8 @@ PG_UnlockMutex(SDL_mutex *mutex)
     return 0;
 }
 
-#define PG_SURF_BitsPerPixel(surf) surf->format->bits_per_pixel
-#define PG_SURF_BytesPerPixel(surf) surf->format->bytes_per_pixel
+#define PG_SURF_BitsPerPixel(surf) SDL_BITSPERPIXEL(surf->format)
+#define PG_SURF_BytesPerPixel(surf) SDL_BYTESPERPIXEL(surf->format)
 #define PG_FORMAT_BitsPerPixel(format) format->bits_per_pixel
 #define PG_FORMAT_BytesPerPixel(format) format->bytes_per_pixel
 
@@ -138,7 +138,7 @@ PG_UnlockMutex(SDL_mutex *mutex)
 
 #define PG_CreateSurface(width, height, format) \
     SDL_CreateRGBSurfaceWithFormat(0, width, height, 0, format)
-#define PG_CreateSurfaceFrom(pixels, width, height, pitch, format) \
+#define PG_CreateSurfaceFrom(width, height, format, pixels, pitch) \
     SDL_CreateRGBSurfaceWithFormatFrom(pixels, width, height, 0, pitch, format)
 #define PG_ConvertSurface(src, fmt) SDL_ConvertSurface(src, fmt, 0)
 #define PG_ConvertSurfaceFormat(src, pixel_format) \
--- a/src_c/image.c
+++ b/src_c/image.c
@@ -422,24 +422,39 @@ tobytes_surf_32bpp_sse42(SDL_Surface *surf, int flipped, char *data,
 #endif /* PG_COMPILE_SSE4_2 */
 
 static void
-tobytes_surf_32bpp(SDL_Surface *surf, int flipped, int hascolorkey,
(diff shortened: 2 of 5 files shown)


src_c/_pygame.h
```diff
@@ -138,7 +138,7 @@ PG_UnlockMutex(SDL_mutex *mutex)
 
 #define PG_CreateSurface(width, height, format) \
     SDL_CreateRGBSurfaceWithFormat(0, width, height, 0, format)
-#define PG_CreateSurfaceFrom(pixels, width, height, pitch, format) \
+#define PG_CreateSurfaceFrom(width, height, format, pixels, pitch) \
```
bilhox, line 141: I'm not sure to understand what is happening here. Why the arguments are in a whole new order ?

src_c/image.c
```diff
@@ -1630,6 +1658,21 @@ SaveTGA_RW(SDL_Surface *surface, SDL_RWops *out, int rle)
...
+    if (!surf_format) {
+        return -1;
+    }
+    SDL_PixelFormat output_format;
+
+    SDL_Palette *surf_palette = SDL_GetSurfacePalette(surface);
+#else
+    SDL_PixelFormat *surf_format = surface->format;
+    SDL_Palette *surf_palette = surface->format->palette;
+    Uint32 output_format;
```
ankith26, line 1673:
```suggestion
    SDL_PixelFormatEnum output_format;
```
Just to be a bit more specific

src_c/image.c
```diff
@@ -1671,34 +1714,41 @@ SaveTGA_RW(SDL_Surface *surface, SDL_RWops *out, int rle)
...
-                                      surface->format->palette->ncolors)) {
-            /* SDL error already set. */
-            goto error;
-        }
+#if SDL_VERSION_ATLEAST(3, 0, 0)
+        if (!SDL_SetSurfacePalette(linebuf, surf_palette))
+#else
+        if (SDL_SetSurfacePalette(linebuf, surf_palette) < 0)
+#endif
+            goto error; /* SDL error already set. */
```
ankith26, line 1751: This seems like the only somewhat concerning part of the PR to me. Why is the SDL2 codepath being changed to use different API in this case. Or rather, why does the current code use `SDL_SetPaletteColors` instead of `SDL_SetSurfacePalette`, could there be a reason for it?

src_c/_pygame.h
ankith26: SDL3 changed the order of `SDL_CreateSurfaceFrom`, so this macro is being updated to match that order

bilhox: But then it's not the macro argument order that needs to be changed, but the function arg order that's called.

Starbuck5: This line is only compiled on SDL2, it defines a macro that re-orders the arguments from SDL3 style into the SDL2 function.

src_c/image.c
Starbuck5: Yeah so my thought there is that setting a surface's palette is more simple than setting every color of a surface's palette to another surface's palette.

I know codepath changes make these harder to review so sorry about that.

In terms of reason why, I bet it's because SDL1's set palette does it color by color, it was probably ported from this. https://www.libsdl.org/release/SDL-1.2.15/docs/html/sdlsetpalette.html

src_c/image.c
Starbuck5: Sure!

src_c/image.c
Starbuck5: I added a test for TGA saving to hope to reassure you (and me!) that this works the same way, and now the tests are segfaulting on two of the Linux builds. I'm unclear if it's because of my changes or whether it would've segfaulted on the old implementation too.

Starbuck5: The TGA test and the fix for that segfault went through in https://github.com/pygame-community/pygame-ce/pull/3169

ankith26 (approved):

bilhox (approved):
LGTM !

PR #3169, opened 2024-10-12
Fix TGA save bug, add test

ankith26:
My take for the TGA save bug discovered in #31686

The test was contributed by @Starbuck5


MyreMylar (approved):
OK, LGTM 👍  

The new test passes locally with the fix on Windows.

bilhox (approved):
Approving this while I don't fully understand the bug, so I'll not merge it yet. Also hey, I didn't know #31686 existed :)

Matiiss (approved):
The algorithm does appear to skip the repetition chunk now if we're at the end of the line (x == w) and the last pixel was not repeated by the one before it.
LGTM

PR #3170, opened 2024-10-12
Publicize window docs

Starbuck5:
This PR represents the transition of the Window API to the pygame-ce public API.

I left it in the "other" category in the top bar in the docs, as it's not quite experimental not quite advanced, hopefully it's useful! It can be recategorized later of course.

I also expect lots of API additions later, as we see how people use this and what stuff we've missed.

I would also like to get some more in depth examples on how to use this in before 2.5.2-final, but that doesn't block a dev release necessarily.

I know there's still debate about features of this, and I don't want to clobber that, but I feel that getting this functionality available to our users will be valuable. We don't want to let the perfect be the enemy of the good, and this module has been in the oven for a long time. There's still lots of room for iteration after it becomes a public API.

### DIFF
--- a/docs/reST/index.rst
+++ b/docs/reST/index.rst
@@ -62,9 +62,6 @@ turn them to stable API in the future ^^.
 :doc:`ref/geometry`
   Pygame module for the Circle, Line, and Polygon objects.
 
-:doc:`ref/window`
-  Pygame object that handles a window.
-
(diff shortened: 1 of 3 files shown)


bilhox (approved):
LGTM !

ankith26 (approved):
Okay fine, lets get this in. Thanks :night_with_stars:

PR #3172, opened 2024-10-13
Minor fixes and changes to `geometry.pyi` stub file

aatle:
Improved some stuff.
- Change has-attribute protocols to use properties (to fix typing problems)
- Remove type unions `_CanBeCircle`, `_CanBeLine`. These internal types provide little benefit, and a similar union `_CanBeRect` was removed previously.
  - This has one minor effect: the circle attribute can now be another object with a circle attribute, recursively (which reflects actual implementation)
- Rename `_CircleValue` -> `_CircleLike`, `_LineValue` -> `_LineLike`, to be consistent with `RectLike`, adhere to naming scheme.
- Fix name typo `_HasCirclettribute` -> `_HasCircleAttribute`
- Remove redundant `Vector2` type from `_CanBeCollided`. `Vector2` is clearly implied by `Point`.
- Mark `Circle.contains(shape)` parameter positional-only
- Format


buildconfig/stubs/pygame/geometry.pyi
```diff
@@ -10,31 +10,26 @@ from typing import (
...
-    line: Union[_CanBeLine, Callable[[], _CanBeLine]]
-
+    @property
+    def line(self) -> Union[_LineLike, Callable[[], _LineLike]]: ...
 
-_LineValue = Union[_CanBeLine, _HasLineAttribute]
+_LineLike = Union[Line, SequenceLike[float], SequenceLike[Point], _HasLineAttribute]
 
-_CanBeCollided = Union[Circle, Rect, FRect, Point, Vector2]
+_CanBeCollided = Union[Circle, Rect, FRect, Point]
```
damusss, line 32: I am not sold on this removal. I know Vector2 is covered by Point but leaving it here even if redundant could be nice to remind users that vectors are allowed and perhaps suggest them to use them. Not too many people are aware that a Vector2 is a compatible sequence so they might convert it to a list out of habit without trying to pass it to any function if they don't see it there.

damusss (changes requested):

buildconfig/stubs/pygame/geometry.pyi
aatle: If `Vector2` would have to be included in this type union along with `Point` for clarity, then every use of `Point` should also be annotated with `Vector2` by that logic.

I think we can trust that typing users (mostly intermediate to advanced) already understand usage of `Point`/`SequenceLike[float]` (a very common type that is documented to include `Vector2`).

damusss: Alright, I'm 51% convinced so I'm going to approve it.

damusss (approved):
LGTM, thanks

buildconfig/stubs/pygame/geometry.pyi
```diff
@@ -10,31 +10,26 @@ from typing import (
...
-    SequenceLike[Point],
-]
-
-class _HasCirclettribute(Protocol):
+class _HasCircleAttribute(Protocol):
     # An object that has a circle attribute that is either a circle, or a function
     # that returns a circle
-    circle: Union[_CanBeCircle, Callable[[], _CanBeCircle]]
+    @property
+    def circle(self) -> Union[_CircleLike, Callable[[], _CircleLike]]: ...
```
ankith26, line 18: This is forward referencing `_CircleLike`. I believe this could be problematic on older python versions. But I'm not sure if this is problematic only on source files and if this won't be a problem on stubs.

buildconfig/stubs/pygame/geometry.pyi
ankith26, line 32:
```suggestion
_CanBeCollided = Union[_CircleLike, RectLike, Point]
```
I haven't really checked the code, but I think this typing is more accurate

buildconfig/stubs/pygame/geometry.pyi
aatle: I've already checked and raw forward references are used liberally in the type stubs (except `typing.pyi`).
The most common forward reference use is for methods that return an instance of their own class.

(*All* python versions would raise an error when encountering a raw forward reference at runtime.)

buildconfig/stubs/pygame/geometry.pyi
aatle: The collide methods that use this type union want only concrete types according to the docs, in order to be able to distinguish between different shape types.

buildconfig/stubs/pygame/geometry.pyi
ankith26: Yeah that makes sense

buildconfig/stubs/pygame/geometry.pyi
ankith26: Maybe the docs are wrong as well. The code uses API like `pgFRect_FromObject` and `pgCircle_FromObject` to all input which don't expect concrete types. Though it would be nice if any geometry authors review this PR.

damusss: > Maybe the docs are wrong as well. The code uses API like `pgFRect_FromObject` and `pgCircle_FromObject` to all input which don't expect concrete types. Though it would be nice if any geometry authors review this PR.

image

I have to disagree as the collision methods for circle actually use `pgRect_Check` and `pgCircle_Check`. Which makes sense, because a random sequence of 4 or 3 numbers cannot be mapped to the right shape arbitrarly, so the stubs are actually correct.

ankith26: True, I was looking at `colliderect` when I made my previous comment and that handles any rect-ey argument. And that is already typed as `RectLike` in the stubs, so that's all good.

I still think `_CanBeCollided` should be made more general but then that is definitely out of scope for this PR as it would require code, docs, test changes.

ankith26 (approved):

PR #3178, opened 2024-10-16
Better meson error on windows when libraries versions are not correct

### LINKED ISSUE #3099: Windows - Meson does not check if the version of installed libraries match with the one built with
Hello,

First of all, this is more an enhancement proposal. A fast solution is just to delete ``prebuild`` directories, and let meson (``download_win_prebuilt.py``) recreate them.

Currently this is how meson checks if libraries are downloaded and download them if it's not the case : 
```
prebuilt_dir = base_dir / 'prebuilt-' + arch_suffix

    # download prebuilts (uses legacy builconfig code)
    if not fs.is_dir(prebuilt_dir)
        run_command(
            [
                find_program('python3', 'python'),
                'buildconfig/download_win_prebuilt.py',
            ],
            check: true,
        )
    endif
```

As you can see it only checks if the directory is here. So basically if the version does not match with the downloaded version, it should delete the directories, and redownload them.

bilhox:
Closes #3099 .

All the details are in the title.

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -195,7 +195,11 @@ if plat == 'win' and host_machine.cpu_family().startswith('x86')
     )
 
     # put dlls in root of install
-    install_data(dlls, install_dir: pg_dir, install_tag: 'pg-tag')
+    try
+        install_data(dlls, install_dir: pg_dir, install_tag: 'pg-tag')
+    catch(e)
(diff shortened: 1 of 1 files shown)


ankith26 (approved):

damusss (approved):
I've actually encountered an unclear error that required this fix at least twice, thanks 👍

PR #3182, opened 2024-10-17
Remove aaline width for 2.5.2

mzivic7:
Width argument for `draw.aaline` is not yet ready (does not support float values in coordinates) and will have different implementation in future versions.
This PR disables width argument for `draw.aaline` and removes it from docs/tests/examples.


ankith26:
A couple of questions.
1. With `width=1` are you sure the issue you are talking about does not happen?
2. If yes, what kinds of `width` are problematic?
3. If no, then maybe the whole PR should be reverted instead?

mzivic7:
With `width = 1` it always uses `draw_aaline` algorithm (Xiaolin Wu) which allows drawing at float coordinates.
My new implementation does not use Bresenham's algorithm at all (used by `draw_line_width`), so `line_width_corners` is unnecessary (used to get corners of `draw_line_width`) and I would remove it in new PR. 
So almost all current code would be replaced in new PR anyway. That's why I think it would be better to revert whole #3140.

bilhox (approved):
LGTM !

mzivic7:
#3140 is completely reverted in #3192 instead.

PR #3183, opened 2024-10-18
Include the explicit type in `ColorLike` and `RectLike`

aatle:
Add the concrete classes to the `ColorLike` and `RectLike` type unions.
They are already implied by the other types, but should be added for clarity.

Since they are only type aliases to unions, their variable name is not always given, so what they represent might not be entirely obvious to users.

Bonus: Sort types in `ColorLike` by how often they are used.

Example mypy error message, before:
```
$ mypy test.py
test.py:38: error: Incompatible types in assignment (expression has type "float", variable has type "int | str | SequenceLike[int]")  [assignment]
test.py:39: error: Incompatible types in assignment (expression has type "int", variable has type "SequenceLike[float] | SequenceLike[SequenceLike[float]] | _HasRectAttribute")  [assignment]
```
After (much more clear!):
```
$ mypy test.py
test.py:38: error: Incompatible types in assignment (expression has type "float", variable has type "Color | SequenceLike[int] | str | int")  [assignment]
test.py:39: error: Incompatible types in assignment (expression has type "int", variable has type "Rect | FRect | SequenceLike[float] | SequenceLike[SequenceLike[float]] | _HasRectAttribute")  [assignment]
```
Note:
`typing.py` now imports things from pygame. I don't think there is any risk of circular imports though.

Note 2:
The `_<Shape>Like` type unions in `geometry.pyi` include their concrete classes already.

### DIFF
--- a/buildconfig/stubs/pygame/typing.pyi
+++ b/buildconfig/stubs/pygame/typing.pyi
@@ -57,7 +57,7 @@ Point = SequenceLike[float]
 # This is used where ints are strictly required
 IntPoint = SequenceLike[int]
 
-ColorLike = Union[Color, int, str, SequenceLike[int]]
+ColorLike = Union[Color, SequenceLike[int], str, int]
 
 
 class _HasRectAttribute(Protocol):
--- a/buildconfig/stubs/pygame/typing.pyi
+++ b/buildconfig/stubs/pygame/typing.pyi
@@ -16,6 +16,10 @@ import sys
 from abc import abstractmethod
 from typing import IO, Callable, Tuple, Union, TypeVar, Protocol
 
+from pygame.color import Color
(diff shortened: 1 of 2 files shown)


damusss (approved):
No, there shouldn't be circular imports.
I have nothing against it, it's redundant but it's clearer, so why not.

ankith26 (approved):
I'm not 100% onboard making verbose unions more verbose, but I can see your argument for these changes. Either way, this PR is not doing anything incorrect so I am approving anyways. Thanks! :tada:

ankith26:
I'll just leave this PR open for a bit more before merging incase anyone else wants to chime in

bilhox (approved):
LGTM ! The changes seem fair to me.

Issue #1515, opened 2023-02-12
Pygame support for python "complex" type (2993)


GalacticEmperor1:
#  **Issue &numero;2993 opened by *[gresm](https://github.com/gresm)* at 2022-01-19 09:45:42**

**Description**

Add support for complex objects to be convertible to pygame Vector2 and vice versa. 
This would allow to use "cmath" build-in python library for mathematical operations on complex numbers.<hr>

#  Comments

# # #  **[gresm](https://github.com/gresm)* commented at 2022-01-19 09:48:43*

Current result:
```
>>> pygame.Vector2(1j+1)
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
ValueError: Vector2 must be set with 2 real numbers, a sequence of 2 real numbers, or another Vector2 instance
```
Expected result:
```
>>> pygame.Vector2(1j+1)
<Vector2(1, 1)>
```
<hr>

# # #  **[ankith26](https://github.com/ankith26)* commented at 2022-01-19 11:38:37*

Hi!
This is an interesting proposal for sure, but I personally don't think it should be added to pygame.
1. This seems like a pretty rare usecase. I've never come across any games having to use complex numbers to represent vectors or do vector math, unless the app is something like a graphing calculator of sorts
2. Even if you are in need of having to do something like this, you can always pass `(z.real, z.imag)` as an argument to `Vector` (where `z` is the complex number). IMHO this is more elegant, because the function call is more clear and readable, and one immediately knows that `z.real` corresponds to x axis and `z.imag` corresponds to y axis. If we are passing a complex number directly, do we always assume that people want to follow this convention? What if people want to flip the axes? By explicitly making a tuple with the components of a complex number, users also have freedom to easily flip axes.
3. This is not really scalable to `Vector3`

I'm not super against the proposal, it would be fine to see this get added if many people think it's a good idea :)
<hr>

# # #  **[gresm](https://github.com/gresm)* commented at 2022-01-19 12:26:54*

I don't say that someone should do it, I am doing it now.
Will it be accepted? - I don't know, but worth to try.
I can't assign myself to this issue, because I created it.
<hr>

# # #  **[Starbuck5](https://github.com/Starbuck5)* commented at 2022-01-20 01:05:55*

> I don't say that someone should do it, I am doing it now.

Say that in the issue then!

> I can't assign myself to this issue, because I created it.

You can't assign yourself because you don't have member rights on the repo.

[...]

PR #3185, opened 2024-10-20
Add support for ``complex()`` using ``__complex__``

bilhox:
This PR overloads the method ``__complex__`` so you can convert a vector2 to a complex number using ``complex()``.
The usage ? Math only, definitely. 
I'm supporting this PR as I know a lot of people using pygame for a non gamedev usage.


docs/reST/ref/math.rst
```diff
@@ -171,6 +171,9 @@ Multiple coordinates can be set using slices or swizzling
       Inherited methods of vector subclasses now correctly return an instance of the
       subclass instead of the superclass
 
+   .. versionchanged:: 2.5.3
```
damusss, line 174: This should be more specific. How is a Vector2 converted to a complex? you should mention x becomes the real part and y becomes the imaginary part (at least this is what I believe is happening)

src_c/math.c
```diff
@@ -2200,6 +2200,19 @@ vector2_new(PyTypeObject *type, PyObject *args, PyObject *kwds)
...
 }
 
+static PyObject *
+vector2_complex(pgVector *self, PyObject *_null)
+{
+    PyObject *complex =
+        PyComplex_FromDoubles(self->coords[0], self->coords[1]);
+
+    if (complex == NULL) {
+        RAISE(PyExc_ValueError, "Couldn't build complex number from Vector2");
```
damusss, line 2210: I'm a bit skeptical of this. Isn't it best to leave the exception that PyComplex_FromDoubles has set? this way the user knows exactly why x and y couldn't become a complex number. Not sure, tell me if I missed something.

damusss (changes requested):

src_c/math.c
bilhox: i'm between yes and no, I can't really check what kind of error message it will throw.

docs/reST/ref/math.rst
bilhox: true

src_c/math.c
damusss: Nah I think you're right. If the user is using such a function, and they run in the error, printing the vector components should give them the reason why it failed. we can't rely on the error message of that C api function.

damusss, line 2210: This should be return RAISE

ankith26: This whole function can just be `return PyComplex_FromDoubles(self->coords[0], self->coords[1]);` because the function already sets the python exception on error, we don't need to set an error ourselves

ankith26 (commented):
I don't see the point of adding API for something that can be done in user code by typing just one more character than what this PR makes possible.
```
a = pygame.Vector2(1, 2)
complex(*a)
```
Bonus of `complex(*a)` is that it actually works on any 2-sized sequence like a tuple of list of numbers, which is also something some people use to represent coordinates instead of `Vector2`

bilhox:
> I don't see the point of adding API for something that can be done in user code by typing just one more character than what this PR makes possible.
> 
> ```
> a = pygame.Vector2(1, 2)
> complex(*a)
> ```
> 
> Bonus of `complex(*a)` is that it actually works on any 2-sized sequence like a tuple of list of numbers, which is also something some people use to represent coordinates instead of `Vector2`

It's one character less than what you proposed, which is still better. And also I think the tools ``Vector2`` provides are way better than just list or tuple of 2 elements. In another side, It's not like it's something difficult for us to support, I mean, it's not like I have added a hundred of lines to bring support to it. (And I did not discuss yet why complex in the vector constructor is not supported)

ankith26:
Also, we have had debates about it in the past and back then we had consensus that it was not needed, atleast among the active reviewers at the time: https://github.com/pygame-community/pygame-ce/issues/1515

bilhox:
With a lot of denial, I'm closing this PR.

PR #3188, opened 2024-10-22
Generate reST/ref docs from python or stub files, move time and cursors docs

zoldalma999:
Related to #2757

This PR implements autodoc style documentation generation. This allows us to have documentation in python files or (if the module is implemented in C) in stub files. It uses [autoapi](https://github.com/readthedocs/sphinx-autoapi) because stub files are not runnable python files thus we cannot use autodoc. These changes are under the commit [Generate reST/ref docs from python or stub files](https://github.com/pygame-community/pygame-ce/commit/bbedc0abd2b110cce36d3e784ad5dae6174a17ab).

This PR also ports two modules ([time](https://github.com/pygame-community/pygame-ce/commit/ba4bc00a38e83ecaea21364212dbfaaf6738fa09) which is a C based module and [cursors](https://github.com/pygame-community/pygame-ce/commit/9258b191e0db906e81a1229a37b59f3e622e9cb9) which is a python based module) as reference. Following modules can be easily ported based on these by anyone.

The generated docs should look almost the same. The only changes are to signatures and summary lines. Signatures will now use the stubs as their source, which means that argument names and default values should be more accurate. But this also means that we cannot just write anything there. Because of this return values are now types (or just "..." if it is too complex) not some descriptive names. Summary lines also change a bit to conform to the convention of having a full sentence as the first line of docstrings (this of course can be reverted if others are against it).

Future work related to this:
- Move all the of the ref/ content either to python/stub files or to guides.
- Switch to full autogeneration, and delete the ref/ folder.
- Add accurate typing info to the documentation, taken from stubs.
- (Possibly) Generate better header files (include the whole docstring not just the summary line and signatures).
- Write a guide (or extend an existing one) on how to contribute docs (conventions and standards, how to contribute, maybe small guide to rst and useful links).
- Remove version[added|changed|deprecated]old and let version[added|changed|deprecated] figure out if it needs to say pygame or pygame-ce.


docs/reST/conf.py
```diff
@@ -23,11 +23,22 @@
...
-extensions = ['sphinx.ext.autodoc', 'sphinx.ext.doctest',
-              'sphinx.ext.coverage', 'ext.headers', 'ext.boilerplate',
-              'ext.customversion', 'ext.edit_on_github']
+extensions = ['autoapi.extension', 'ext.headers', 'ext.boilerplate',
+              'ext.customversion', 'ext.edit_on_github', 'ext.documenters']
 
 
+autoapi_dirs = [
+    pathlib.Path(os.path.abspath(".")).parent.parent / 'buildconfig' / 'stubs' / 'pygame' / '',
+    pathlib.Path(os.path.abspath(".")).parent.parent / 'src_py' / '',
```
ankith26, line 32: I don't understand what the empty string at the end of the path is supposed to mean. If it's needed, can a comment be added clarifying what it does and why it is needed?

minor nitpick: using `pathlib` and `os.path` in the same line seems off to me. Can you convert the expression to use only pathlib? `pathlib.Path(".").resolve().parent.parent` or something should probably work

docs/reST/ref/cursors.rst
```diff
@@ -3,248 +3,12 @@
...
-
-   .. ## pygame.cursors.Cursor.data ##
-
-   .. versionaddedold:: 2.0.1
-
-   .. ## pygame.cursors.Cursor ##
-
-.. ## pygame.cursors ##
+.. autopgclass:: Cursor
+   :members: copy, type, data
```
ankith26, line 11: maybe I'm asking for too much, I don't know if it would be feasible to implement, but I see redundancy here. Why are we required to maintain info duplication here in the `rst` file at all, can it just not pick all the info from the stubs?

src_py/cursors.py
```diff
@@ -773,12 +911,21 @@ def compile(strings, black="X", white=".", xor="o"):
...
+    images into a single cursor.
+
+    The cursorfile and maskfile arguments can either be filenames or file-like
+    object with the readlines method.
 
-    Arguments can either be filenames or filelike objects
-    with the readlines method. Not largely tested, but
-    should work with typical XBM files.
+    The return value cursor_args can be passed directly to the
+    ``pygame.mouse.set_cursor()`` function.
```
ankith26, line 928: Is it required to document info here as well? Again, I would like to avoid duplication where possible. Also should help with release wheel size, which we would like to minimize.

ankith26: Ignore this comment, I was tripping.

Starbuck5:
I'm curious about the flexibility of this system. Can we have content in the rst docs that surrounds content pulled from stubs? I feel like it would to nice to have method docs pulled in from stubs into a manually set up page with examples and other things.

zoldalma999:
> I'm curious about the flexibility of this system. Can we have content in the rst docs that surrounds content pulled from stubs? I feel like it would to nice to have method docs pulled in from stubs into a manually set up page with examples and other things.

I don't see why we wouldn't have examples and "other things" already in the docs but yes we can, though it is a bit more verbose. Say you wanted to add some extra info under `pygame.time.delay`
```rst
.. autopgmodule:: pygame.time

.. autopgfunction:: get_ticks

.. autopgfunction:: wait

.. autopgfunction:: delay
   
   This sentence will be right under delay's docs (note the extra newline above).

.. autopgfunction:: set_timer

.. autopgclass:: Clock
   :members:
```

> Why are the cursor docstrings in the Python code for Python modules? All modules have stubs that can be used.

Because I wanted those to be *proper* docstrings that get picked up by python. If we put them in the stub files things like the builtin `help` function, `pydoc` or `object.__doc__` would not work (but Ankith asked me the same question, at least a note somewhere of how this works could be nice?).

src_c/doc/cursors_doc.h
```diff
@@ -1,8 +1,8 @@
...
-#define DOC_CURSORS_CURSOR_COPY "copy() -> Cursor\ncopy the current cursor"
-#define DOC_CURSORS_CURSOR_TYPE "type -> string\nget the cursor type"
-#define DOC_CURSORS_CURSOR_DATA "data -> tuple\nget the cursor data"
+#define DOC_CURSORS "Pygame module for cursor resources."
+#define DOC_CURSORS_COMPILE "compile(strings, black='X', white='.', xor='o') -> ...\nCreate binary cursor data from simple strings."
+#define DOC_CURSORS_LOADXBM "load_xbm(curs, mask) -> ...\nLoad cursor data from an XBM file."
+#define DOC_CURSORS_CURSOR "Cursor(constant=...) -> Cursor\nCursor(cursor) -> Cursor\nCursor(size, hotspot, xormasks, andmasks) -> Cursor\nCursor(hotspot, surface) -> Cursor\nPygame object representing a cursor."
+#define DOC_CURSORS_CURSOR_COPY "copy() -> Cursor\nCopy the current cursor."
+#define DOC_CURSORS_CURSOR_TYPE "type -> Literal['system', 'color', 'bitmap']\nGet the cursor type."
+#define DOC_CURSORS_CURSOR_DATA "data -> ...\nGet the cursor data."
```
ankith26, line 8: I think that the original typehint `tuple` was better here.
Basically, I think that we shouldn't use `...` to replace `tuple[big list of stuff here]`, in such cases it is probably better to display `tuple`.
I think the same strategy could be applied to `list`/`dict`/etc as well.

`Union` and `Optional` are also cases that may be need to be handled specially. I think it is better to leave those as is.

Starbuck5 (approved):
This is a hard PR to approve because it changes our approach to documentation significantly.

But this should lead to a really nice improvement in user experience.

I like how you did it as two example modules at first, minimizing the changes. I like your writeup. I like that you are not a drive by contributor and if this breaks I can ping you about it 😄

So with that in mind, thank you for the PR-- let's try this out!

MyreMylar (approved):
OK, LGTM. Let's try this out in 2025!

PR #3189, opened 2024-10-23
Do regular install where possible on CI

ankith26:
Earlier we were doing `--no-build-isolation` in some places where a regular install would work, for CI time optimisation reasons. But those were probably premature optimisations, and doing things the "regular way" has the advantages of not having to update the CI script for changes in build dependencies


ankith26:
This PR is epic fail

PR #3192, opened 2024-10-23
Revert 3140 draw.aaline width

ankith26:
Reverts pygame-community/pygame-ce#3140 completely. #3182 is doing a partial revert already but I believe a full revert is safer in this case, especially given the author of the original PR has come up with an entirely new implementation


MyreMylar (approved):
LGTM 👍

oddbookworm:
I'm curious why only the circleci runs ran here

oddbookworm (approved):
LGTM! Although, I'm not merging quite yet until I understand why not all actions ran here

ankith26:
github CI is only configured to run on pushes and PRs made against the `main` branch. The CI is going to do a run during the release as well, so if there is any issue we can still catch it then

oddbookworm:
🤦‍♂️ yep, I should have paid attention to what branch it was going to

PR #3194, opened 2024-10-26
Fix grammar and formatting in Window docs

gavinmorrow:


docs/reST/ref/window.rst
```diff
@@ -61,8 +61,8 @@
 
    Event behavior if multiple Windows are created: When the close button is
    pressed, a ``WINDOWCLOSE`` event is sent. You need to explicitly destroy
-   the window. Note that the event ``QUIT`` will only be sent if all Window
-   has been destroyed.
+   the window. Note that the event ``QUIT`` will only be sent if all Windows
```
bilhox, line 64: if it's plurial, then it should not be capitalized as well.

bilhox (changes requested):

docs/reST/ref/window.rst
oddbookworm: I would actually suggest doing something like this instead
```rst
\`\`Window\`\`s
```

gavinmorrow: I was trying to follow the example of the rest of the file, where Window is capitalized (including plurals). In the [`Surface`](https://pyga.me/docs/ref/surface.html#:~:text=Normal%20Surfaces%20should%20not%20require%20the%20masks%20argument) docs, the same convention is followed.

gavinmorrow: @bilhox @oddbookworm what do you think?

bilhox: @gavinmorrow I would follow what oddbookworm suggested.

ankith26 (approved):

zoldalma999 (approved):
LGTM, Thanks!

gavinmorrow:
It looks completely wrong, does anyone know why? I'm not familiar with `.rst`.

Screenshot 2024-11-02 at 19 16 25

gavinmorrow:
oh yeah didn't see #3205

PR #3201, opened 2024-10-29
Fix latest SDL3 and MSYS2 fails

ankith26:
These are actually two independent issues that somehow popped up at the exact same time, hence the common PR fixing both as the fixes are trivial.

- Added a command installing SDL3 dependencies on our SDL3 build
- mingw32 support is being gradually dropped (ref: https://github.com/msys2/MINGW-packages/issues/22327) and therefore our build has failed here. I simply removed the mingw32 run.


bilhox (approved):
LGTM !

Starbuck5 (approved):
Seems legit!

PR #3191, opened 2024-10-23
New draw.aaline width algorithm

mzivic7:
This PR adds completely new implementation for `draw.aaline` width. Previous implementation (#3140) uses Bresenham's algorithm, and because of that, line cant be drawn at float coordinates (they are converted to ints instead).
New implementation uses modified Xiaolin Wu's algorithm that allows drawing at float coordinates, and is faster.

Old width implementation is removed for pgce 2.5.2 in #3192

aaline

<details>
  <summary>How it works</summary>

  Original `draw_aaline` code is used, heavily modified:
  - 2 antialiased pixels (2 are drawn for each point) are moved, leaving gap between, the size of width
  - that gap is filled either with horizontal or vertical lines, depending on steepines
  - Same thing is done to both endpoints
  - Clipping is also modified to support width
  - Using `drawhorzlineclipbounding` and new `drawvertlineclipbounding` to handle special cases when drawing lines outside surface.
</details>

<details>
  <summary>Sample code</summary>

  ```py
  import pygame
  pygame.init()
  screen = pygame.display.set_mode((300, 300))
  clock = pygame.time.Clock()
  run = True
  while run:
      for event in pygame.event.get():
          if event.type == pygame.QUIT:
              run = False
      screen.fill("black")
      pygame.draw.aaline(screen, "white", (120, 100), (220, 160), 16)
      pygame.draw.aaline(screen, "white", (180, 170), (100, 260), 16)
      pygame.draw.aaline(screen, "white", (70.7, 240.2), (150.2, 150.6), 7)
      pygame.display.flip()
      clock.tick(60)
  pygame.quit()
  ```
</details>

PR #3204, opened 2024-11-02
Correct draw.aaline docs

Starbuck5:
I see this is addressed already in #3191, but I'd like to get it deployed without waiting on the rest of the PR to be reviewed. To keep on docs on the website as correct as possible.

### DIFF
--- a/docs/reST/ref/draw.rst
+++ b/docs/reST/ref/draw.rst
@@ -493,7 +493,7 @@ object around the draw calls (see :func:`pygame.Surface.lock` and
    .. versionchangedold:: 2.0.0 Added support for keyword arguments.
    .. versionchanged:: 2.4.0 Removed deprecated 'blend' argument
    .. versionchanged:: 2.5.0 ``blend`` argument readded for backcompat, but will always raise a deprecation exception when used
-   .. versionchanged:: 2.5.2 Added line width
+   .. versionchanged:: 2.5.3 Added line width
 
    .. ## pygame.draw.aaline ##
 


docs/reST/ref/draw.rst
```diff
@@ -493,7 +493,7 @@ object around the draw calls (see :func:`pygame.Surface.lock` and
    .. versionchangedold:: 2.0.0 Added support for keyword arguments.
    .. versionchanged:: 2.4.0 Removed deprecated 'blend' argument
    .. versionchanged:: 2.5.0 ``blend`` argument readded for backcompat, but will always raise a deprecation exception when used
-   .. versionchanged:: 2.5.2 Added line width
+   .. versionchanged:: 2.5.3 Added line width
```
ankith26, line 496: There are more references to `width` in this block, though

Starbuck5: Yes, and that's fine. The docs on the site always have the features of main.

ankith26 (approved):
Yeah, lets get this in. Thanks! :tada: 

This PR got me thinking if it's better to stick to latest-stable-release docs on the website. I think that would be less confusing for users and prevent this kind of situation entirely. What say?

Starbuck5:
Well if we're going to be fancy we could have a version selector in our docs, like Python does.

Starbuck5:
But I guess that would be fine? Prerelease and testing users could use local docs?

oddbookworm (approved):
LGTM!

PR #3205, opened 2024-11-02
Fix window docs

Starbuck5:
After https://github.com/pygame-community/pygame-ce/pull/3194 it looks like 
image

This fixes that. Had to read the docs on the docs system to figure out how to end a inline markup without a space.

### DIFF
--- a/docs/reST/ref/window.rst
+++ b/docs/reST/ref/window.rst
@@ -59,10 +59,10 @@
              pygame.quit()
              raise SystemExit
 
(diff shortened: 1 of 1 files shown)


ankith26 (approved):
LGTM, this time I actually tested the docs locally and it works as expected. Thanks! :partying_face:

oddbookworm (approved):
I feel responsible for this one because I suggested using the inline markup here on the PR that introduced this... oops lol

LGTM

PR #3206, opened 2024-11-02
Port time, surflock and constants to SDL3

ankith26:
I clubbed all these updates in one PR because these modules were mostly trivial to port.

`import pygame` works now! (although with warnings, and in a very unusable state)


src_c/time.c
```diff
@@ -316,8 +321,8 @@ accurate_delay(Sint64 ticks)
     if (ticks <= 0)
         return 0;
 
-    if (!SDL_WasInit(SDL_INIT_TIMER)) {
-        if (SDL_InitSubSystem(SDL_INIT_TIMER)) {
+    if (!SDL_WasInit(PG_INIT_TIMER)) {
```
Starbuck5, line 324: I looked into the source and I don't believe this will work as expected at runtime.

```c
Uint32 SDL_WasInit(SDL_InitFlags flags)
{
    int i;
    int num_subsystems = SDL_arraysize(SDL_SubsystemRefCount);
    Uint32 initialized = 0;

    // Fast path for checking one flag
    if (SDL_HasExactlyOneBitSet32(flags)) {
        int subsystem_index = SDL_MostSignificantBitIndex32(flags);
        return SDL_SubsystemRefCount[subsystem_index] ? flags : 0;
    }

    if (!flags) {
        flags = SDL_INIT_EVERYTHING;
    }
```
(src/SDL.c, whatever SDL3 I have locally)

Since PG_INIT_TIMER goes to 0 it will check if every subsystem is initalized.

I mean this will work fine because it will then go do an initialization no-op, but still something I noticed.

ankith26: True, I think it would be better to guard the entire block for SDL2 and do nothing on SDL3, right?

Starbuck5: That checks out to me.

Starbuck5 (approved):
Looks good, thanks!

src_c/surflock.c
```diff
@@ -99,7 +99,11 @@ pgSurface_LockBy(pgSurfaceObject *surfobj, PyObject *lockobj)
     if (surf->subsurface != NULL) {
         pgSurface_Prep(surfobj);
     }
+#if SDL_VERSION_ATLEAST(3, 0, 0)
+    if (!SDL_LockSurface(surf->surf)) {
+#else
     if (SDL_LockSurface(surf->surf) == -1) {
+#endif
```
gresm, line 106: Not really an issue, but many simpler tools don't like unmatched parenthesis. A very simple solution to this problem is to move it out of the conditional preprocess:
```suggestion
#if SDL_VERSION_ATLEAST(3, 0, 0)
    if (!SDL_LockSurface(surf->surf))
#else
    if (SDL_LockSurface(surf->surf) == -1)
#endif
    {
```

gresm (approved):
LGTM :+1:  @ankith26 Left a comment on one part, but it's not anything blocking this PR.

src_c/surflock.c
ankith26: Good idea, I have applied your suggestion as a commit

PR #3207, opened 2024-11-02
Port key and mouse to SDL3

ankith26:
Just following the footsteps of https://github.com/pygame-community/pygame-ce/pull/3206 but separated out in a different PR because these changes are not as trivial as those

### DIFF
--- a/src_c/mouse.c
+++ b/src_c/mouse.c
@@ -65,7 +65,14 @@ mouse_set_pos(PyObject *self, PyObject *args)
 static PyObject *
 mouse_get_pos(PyObject *self, PyObject *args, PyObject *kwargs)
 {
+#if SDL_VERSION_ATLEAST(3, 0, 0)
+    /* SDL3 changed the mouse API to deal with float coordinates, for now we
+     * still truncate the result to int before returning to python side.
+     * This can be changed in a breaking release in the future if needed. */
+    float x, y;
+#else
     int x, y;
+#endif
     int desktop = 0;
 
     static char *kwids[] = {"desktop", NULL};
@@ -114,7 +121,14 @@ mouse_get_pos(PyObject *self, PyObject *args, PyObject *kwargs)
 static PyObject *
 mouse_get_rel(PyObject *self, PyObject *_null)
 {
+#if SDL_VERSION_ATLEAST(3, 0, 0)
+    /* SDL3 changed the mouse API to deal with float coordinates, for now we
+     * still truncate the result to int before returning to python side.
+     * This can be changed in a breaking release in the future if needed. */
+    float x, y;
+#else
     int x, y;
+#endif
 
     VIDEO_INIT_CHECK();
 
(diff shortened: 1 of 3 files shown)


src_c/mouse.c
```diff
@@ -114,7 +121,14 @@ mouse_get_pos(PyObject *self, PyObject *args, PyObject *kwargs)
 static PyObject *
 mouse_get_rel(PyObject *self, PyObject *_null)
 {
+#if SDL_VERSION_ATLEAST(3, 0, 0)
+    /* SDL3 changed the mouse API to deal with float coordinates, for now we
+     * still truncate the result to int before returning to python side.
+     * This can be changed in a breaking release in the future if needed. */
+    float x, y;
```
Starbuck5, line 128: So it gets it as a float, how does it get converted to an int? Because it gets passed a function that takes ints.

ankith26: Yeah that just happens implicitly, I didn't bother documenting it because I didn't think it should cause any confusion

Starbuck5: If you had written `int ix = (int)x;` I wouldn't have batted an eye. Even `int ix = x;` So I assume that passing to into a parameter of a type is the same as assignment?

ankith26: Yeah, true. I can make the function call
`return pg_tuple_couple_from_values_int((int)x, (int)y);` if that helps. It becomes more explicit but verbose

oddbookworm: Tbh, I'm not a fan of downcasting without explicitly checking that the value is within range of the destination data type. That just seems like a recipe for hard-to-diagnose bugs

Starbuck5: I think we use casts to ints from floats in other places, I don't see how we could check range elegantly. For this I think the mouse coordinates reported will be reasonable numbers.

MyreMylar (approved):
LGTM 👍

Starbuck5:
I wanted to test this, so I rebased against main and ran `python dev.py build --sdl3 --wheel`

I got the following warnings-to-errors:
```c
  ../src_c/mouse.c(100): error C2220: the following warning is treated as an error
  ../src_c/mouse.c(100): warning C4244: '=': conversion from 'int' to 'float', possible loss of data
  ../src_c/mouse.c(101): warning C4244: '=': conversion from 'int' to 'float', possible loss of data
  ../src_c/mouse.c(109): warning C4244: '=': conversion from 'int' to 'float', possible loss of data
  ../src_c/mouse.c(113): warning C4244: '=': conversion from 'int' to 'float', possible loss of data
  ../src_c/mouse.c(118): warning C4244: 'function': conversion from 'float' to 'int', possible loss of data
  ../src_c/mouse.c(118): warning C4244: 'function': conversion from 'float' to 'int', possible loss of data
  ../src_c/mouse.c(149): warning C4244: 'function': conversion from 'float' to 'int', possible loss of data
  ../src_c/mouse.c(149): warning C4244: 'function': conversion from 'float' to 'int', possible loss of data
  ../src_c/mouse.c(253): warning C4244: '=': conversion from 'SDL_WindowFlags' to 'Uint32', possible loss of data
 ```

ankith26:
I have rebased this branch on main and force pushed. Previously I didn't see these warnings on CI but lets see how things go now

Starbuck5 (approved):
Let's merge it in

PR #3208, opened 2024-11-02
Port joystick to SDL3

ankith26:
This PR ports joystick ~~and scrap~~ to SDL3.

~~Porting scrap turned out to be surprisingly trivial, but it makes sense given that we are using like 3 functions of the SDL API, and the rest are custom implementations.~~ UPDATE: that has been moved to a separate PR now.
EDIT: There was some windows specific code that I initially missed when I made this comment, but that should hopefully be ported now as well.

### DIFF
--- a/src_c/joystick.c
+++ b/src_c/joystick.c
@@ -112,7 +112,17 @@ static PyObject *
 get_count(PyObject *self, PyObject *_null)
 {
     JOYSTICK_INIT_CHECK();
+#if SDL_VERSION_ATLEAST(3, 0, 0)
+    int ret;
+    SDL_JoystickID *joysticks = SDL_GetJoysticks(&ret);
+    if (!joysticks) {
+        return RAISE(pgExc_SDLError, SDL_GetError());
+    }
+    SDL_free(joysticks);
+    return PyLong_FromLong(ret);
+#else
     return PyLong_FromLong(SDL_NumJoysticks());
+#endif
 }
 
 static PyObject *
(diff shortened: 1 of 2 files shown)


src_c/scrap.c
```diff
@@ -27,9 +27,8 @@
 #include <SDL3/SDL.h>
 #else
 #include <SDL.h>
-#endif
-
 #include "SDL_syswm.h"
```
Starbuck5, line 30: Looking at this, I believe this should be moved to scrap_win.c.

Also I would accept a PR that doesn't get scrap_win.c compiling at all, and uses scrap_sdl2 on all platforms for SDL3.

Starbuck5 (approved):
Thanks!

src_c/scrap.c
ankith26: I thought of moving the include but then I decided it's better to take our usual strategy of "don't change the SDL2 codepath", as that minimizes the diff as well in this case

src_c/joystick.c
```diff
@@ -200,18 +210,53 @@ joy_get_guid(PyObject *self, PyObject *_null)
         guid = SDL_JoystickGetGUID(joy);
     }
     else {
+#if SDL_VERSION_ATLEAST(3, 0, 0)
+        guid = SDL_GetJoystickGUIDForID(pgJoystick_AsID(self));
```
ankith26, line 214: Something I didn't think about while making the PR: SDL_GetJoystickGUIDForID uses the instance ID and not the device ID returned by pgJoystick_AsID. This block needs more attention and thought

zoldalma999: In general the joystick port will need some more thought. They removed device indices completely, meaning we need to rework the initialization of joystick objects in general. There is no easy way to emulate this either (and we shouldn't anyway!).

With that said, you can use `SDL_GetJoystickID` to get the instance_id of an already open joystick.

MyreMylar: It seems like we may need to store something like `initial_instance_id` or `last_known_instance_id` on the pygame Joystick object.

at the minute it has these attributes:

```
jstick->id = id; // ID passed in by the user in range 0 to number of connected joysticks at time of creation - called 'device id'
jstick->joy = joy; // SDL Joystick Object
jstick->prev = NULL;
jstick->next = joylist_head;
```

It is a bit of an odd code path here, as you only go down it when the SDLJoystick object is invalid - which shouldn't happen, and I'm not sure why you'd want the GUID of an invalid SDL Joystick or if it would even produce any usable output. Are we testing this code path at all?

The instance id of an SDLJoystick changes to a unique number whenever a joystick is connected or disconnected. We are checking the instance ID in this case when we explicitly cannot get the SDL Joystick object - perhaps because it has been disconnected.

At the minute the joystick module has no support for 'hot plugging' or plugging & unplugging controllers and we probably shouldn't try and add it in as simple conversion PR - we also have the newer controller module for that sort of thing.

I would probably just store the initial instance id when we first create the joystick, like we do the device id, and then add another pg function to get it like we do the device ID. We could also update the `last_known_instance_id` in `get_guid()` 

I notice there was also a lot of renaming in the joystick module:
<https://github.com/libsdl-org/SDL/issues/6881>

I'm sure this has been covered somewhere, but are we intending to eventually rename our SDL function calls as well? Or just relying on the SDL 'old names' wrapper?

damusss (approved):
LGTM!

src_c/joystick.c
ankith26: I addressed this particular issue by just setting an error (and raising it), just to get this PR in a mergeable state. OFC, there is probably a lot of room for future work here, but all that need not stall this PR as this gets joystick compiling on SDL3

PR #3209, opened 2024-11-03
Rewrite sprite collision functions

aatle:
The current sprite collision functions code has some code repetition. This implementation structures it differently, which has a few effects.

- If `dokill` is true in `spritecollide` or `groupcollide`, the sprites are killed *after* all colliding sprites have been determined, instead of *during iteration* (should be unnoticeable)
- As a consequence of the previous thing, in these cases, iteration over `group.sprites()` is replaced by iteration over `group` itself, resulting in the following:
  - Avoids a list copy of the group, improving performance slightly in around 1/8 of possible cases
  - It is no longer an error to use a non-group iterable of sprites here (already no error if `dokill` is False)
    - I advocate for typing the sprite collide functions as supporting all iterables of sprites, instead of just sprite groups
  - "Old-style" groups that don't support iteration raise error now here (already errors if `dokill` is False)
- It is now easy to implement an `ignore_self` parameter, e.g., with the new structure

Performance tested in 3.13.0, no regression (similar performance).
Renamed a few local variables for clarity.


Starbuck5:
I heard somewhere (can't find a back reference) that the old "pull out a bound function into a local variable" trick is actually slower now, I'd be interested to see you look into it with this case.

aatle:
Thanks. I looked into it and found no noticeable performance increase nor penalty from either using `sprite_rect.colliderect` instead of `sprite_rect_collide` or `spritecollide` global instead of `sprite_collide_func` local, on python 3.13.0.
The most likely thing is that the impact is very minimal. The talk mentions that for pure python methods, the cached method may still be faster by 3% on 3.11.

Since I saw no performance increase from removing the cached methods/functions, I will leave it in just in case python 3.9 or 3.10 needs it.

src_py/sprite.py
```diff
@@ -1684,36 +1684,21 @@ def spritecollide(sprite, group, dokill, collided=None):
...
-            else:
-                if default_sprite_collide_func(group_sprite.rect):
-                    group_sprite.kill()
-                    append(group_sprite)
-
-        return crashed
-
     if collided is not None:
-        return [
+        crashed = [
```
MyreMylar, line 1688:
```suggestion
        collided_sprites = [
```

src_py/sprite.py
```diff
@@ -1684,36 +1684,21 @@ def spritecollide(sprite, group, dokill, collided=None):
...
         ]
-
-    return [
-        group_sprite
-        for group_sprite in group
-        if default_sprite_collide_func(group_sprite.rect)
-    ]
+    else:
+        sprite_rect_collide = sprite.rect.colliderect
+        crashed = [
```
MyreMylar, line 1693:
```suggestion
        collided_sprites = [
```

src_py/sprite.py
```diff
@@ -1738,17 +1723,13 @@ def groupcollide(groupa, groupb, dokilla, dokillb, collided=None):
     # pull the collision function in as a local variable outside
     # the loop as this makes the loop run faster
     sprite_collide_func = spritecollide
+    for group_a_sprite in groupa:
+        collisions = sprite_collide_func(group_a_sprite, groupb, dokillb, collided)
+        if collisions:
+            crashed[group_a_sprite] = collisions
```
MyreMylar, line 1729:
```suggestion
            collided_sprites[group_a_sprite] = collisions
```

src_py/sprite.py
```diff
@@ -1684,36 +1684,21 @@ def spritecollide(sprite, group, dokill, collided=None):
...
+        sprite_rect_collide = sprite.rect.colliderect
+        crashed = [
+            group_sprite
+            for group_sprite in group
+            if sprite_rect_collide(group_sprite.rect)
+        ]
+    if dokill:
+        for group_sprite in crashed:
+            group_sprite.kill()
+    return crashed
```
MyreMylar, line 1701:
```suggestion
    return collided_sprites
```

MyreMylar (approved):
OK, LGTM 👍 

Nice simplification. I renamed the 'crashed' variable because it's name didn't make much sense and this was a good opportunity.

src_py/sprite.py
```diff
@@ -1684,36 +1684,21 @@ def spritecollide(sprite, group, dokill, collided=None):
...
-    ]
+    else:
+        sprite_rect_collide = sprite.rect.colliderect
+        collided_sprites = [
+            group_sprite
+            for group_sprite in group
+            if sprite_rect_collide(group_sprite.rect)
+        ]
+    if dokill:
+        for group_sprite in crashed:
```
MyreMylar, line 1699:
```suggestion
        for group_sprite in collided_sprites:
```

ankith26 (approved):

After looking around for a bit trying to think of any case where doing a kill later would change behaviour I have not come up with anything. So I think this is safe to merge

PR #3214, opened 2024-11-04
Optimize pg_Two(Ints/Floats)FromObj

Starbuck5:
pg_IntFromObjIndex doesn't take advantage of the fact that pg_TwoIntsFromObj already knows that obj is a sequence. Since obj is known to be a sequence, and negative indices are not needed, PySequence_ITEM can be used instead of PySequence_GetItem for better performance.
(+Same for pg_FloatFromObj and pg_FloatFromObjIndex)

Information on PySequence_ITEM: https://docs.python.org/3/c-api/sequence.html#c.PySequence_ITEM

```py
"""Rough testing script"""
import time

r = pygame.Rect(1,2,3,4)
fr = pygame.FRect(1,2,3,4)
start = time.time()
for _ in range(10000000):
    #r.inflate_ip((2,3))
    r.collidepoint((2,3))
    #fr.collidepoint((2,3))

print(time.time() - start)
```

In lightweight operations that use these functions for argument parsing I am able to get a single digit percent performance improvement. Hard to pin down an exact number, as it is hard to see the performance difference sometimes given noisy tests. But when averaging runs I see the average runtime is definitely lower with these changes.

This is a micro optimization, but I think it is a worthwhile thing to really scrutinize, as these functions are used all over the place. (Channeling my inner @itzpr3d4t0r on this PR)

### DIFF
--- a/src_c/base.c
+++ b/src_c/base.c
@@ -541,10 +541,25 @@ pg_TwoIntsFromObj(PyObject *obj, int *val1, int *val2)
     if (!PySequence_Check(obj) || PySequence_Length(obj) != 2) {
         return 0;
     }
-    if (!pg_IntFromObjIndex(obj, 0, val1) ||
-        !pg_IntFromObjIndex(obj, 1, val2)) {
+
+    PyObject *item1 = PySequence_ITEM(obj, 0);
+    PyObject *item2 = PySequence_ITEM(obj, 1);
+
+    if (item1 == NULL || item2 == NULL) {
+        Py_XDECREF(item1);
+        Py_XDECREF(item2);
+        PyErr_Clear();
         return 0;
     }
+
+    if (!pg_IntFromObj(item1, val1) || !pg_IntFromObj(item2, val2)) {
+        Py_DECREF(item1);
+        Py_DECREF(item2);
+        return 0;
+    }
+
+    Py_DECREF(item1);
+    Py_DECREF(item2);
     return 1;
 }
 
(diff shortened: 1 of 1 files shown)


itzpr3d4t0r:
You didn't full channel though xD, ig this could've been expanded to be more in line with pg_TwoDoublesFromObj with the whole sequencefast treatment.

Starbuck5:
Is sequencefast actually worth it for a two element sequence?

itzpr3d4t0r:
I've seen 5-10% improvements. Fastcall methods that don't do much benefit the most out of the bunch. This might even speed up fblits or stuff that takes lots of positions since one could also pass only tuples as positions.

Starbuck5:
Alright you've pushed me to be better, so I've turned this simple speedup PR into a better (but way more complicated speedup PR). I got a lot of inspiration from pg_TwoDoublesFromObj, but I haven't copied that strategy exactly here for now.

I've updated the PR description with my findings.

itzpr3d4t0r:
Happy to report that this strategy is actually faster than the FastSequence strategy lol.

Starbuck5:
I’m surprised that you say that, I thought the sequence fast way was still faster. I just thought this way was less complex and slightly safer code (like for example in the Python source it checks for List and Tuple exactly, and we just check for list and tuple non exactly).

I’ve also been thinking about how we could optimize this in Pythons C API. Like would it be possible to do a function that returns elements and size and checks whether it’s a sequence?

MyreMylar (approved):
The PR changes run faster for me in testing under Python 3.11. Looks about 5% faster

👍

ankith26 (approved):

PR #3215, opened 2024-11-06
Reduce errors in src_c/surface.c when compiling with SDL3

gresm:
Not a full port.

Things left to do:
- [x] ~Logic for palettes.~
- [x] ~``RLEACCEL``~
- [ ] ``alphablit``
- [ ] ``surface_fill``
- [ ] ``simd_blitters_avx2``
- [ ] ``simd_blitters_sse2``
- [ ] ``simd_surface_fill_avx2``
- [ ] ``simd_surface_fill_sse2``

Unchecked points will be done in a different PR / different PRs.

### DIFF
--- a/src_c/_pygame.h
+++ b/src_c/_pygame.h
@@ -85,7 +85,7 @@
     SDL_GetPixelFormatDetails(PG_SurfaceFormatEnum(surf))
 #define PG_GetSurfacePalette SDL_GetSurfacePalette
 #define PG_SurfaceMapRGBA(surf, r, g, b, a) \
-    SDL_MapSurfaceRGBA(surface, r, g, b, a)
+    SDL_MapSurfaceRGBA(surf, r, g, b, a)
 #define PG_GetSurfaceClipRect(surf, rect) SDL_GetSurfaceClipRect(surf, rect)
 
 #define PG_SurfaceHasRLE SDL_SurfaceHasRLE
@@ -171,7 +171,7 @@ PG_FillRect(SDL_Surface *dst, const SDL_Rect *rect, Uint32 color)
 #define PG_GetSurfaceFormat(surf) (surf)->format
 #define PG_GetSurfacePalette(surf) (surf)->format->palette
 #define PG_SurfaceMapRGBA(surf, r, g, b, a) \
-    SDL_MapRGBA(PG_GetSurfaceFormat(surface), r, g, b, a)
+    SDL_MapRGBA(PG_GetSurfaceFormat(surf), r, g, b, a)
 #define PG_GetSurfaceClipRect(surf, rect) \
     {                                     \
         rect = &surf->clip_rect;          \
--- a/src_c/_pygame.h
+++ b/src_c/_pygame.h
@@ -26,6 +26,8 @@
 #ifndef _PYGAME_INTERNAL_H
 #define _PYGAME_INTERNAL_H
 
+#include <stdbool.h>
+
 #include "pgplatform.h"
 /*
     If PY_SSIZE_T_CLEAN is defined before including Python.h, length is a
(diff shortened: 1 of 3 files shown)


src_c/_pygame.h
```diff
@@ -78,6 +78,15 @@
...
 #define PG_PixelFormatEnum SDL_PixelFormat
+#define PG_PixelFormat const SDL_PixelFormatDetails
+#define PG_PixelFormatMut SDL_PixelFormatDetails
+#define PG_SurfaceFormatEnum(surf) (surf)->format
+#define PG_GetSurfaceFormat(surf) \
+    SDL_GetPixelFormatDetails(PG_SurfaceFormatEnum(surf))
+#define PG_GetSurfacePalette SDL_GetSurfacePalette
+#define PG_SurfaceMapRGBA(surf, r, g, b, a) \
+    SDL_MapSurfaceRGBA(surf, r, g, b, a)
+#define PG_GetSurfaceClipRect(surf, rect) SDL_GetSurfaceClipRect(surf, rect)
```
ankith26, line 89: I think this approach is fine for now, but eventually we may have to look into how this can be improved. For now we have no idea how cheap or how expensive these functions are, and they are called in tight loops in performance critical sections.
Many of those places probably need to be refactored to call this method once outside any tight loops and then use that result wherever possible.

src_c/_pygame.h
```diff
@@ -100,6 +109,13 @@ PG_UnlockMutex(SDL_mutex *mutex)
     return 0;
 }
 
+/* Emulating SDL2 SDL_FillRect API. In SDL3, it returns -1 on failure. */
```
Starbuck5, line 112: You mean, in SDL2 it returns -1 on failure.

With my macros I've been trying to set it up to emulate SDL3 where possible, since eventually we're going to be SDL3-only, might as well go SDL3-style now if not hard.

Starbuck5:
Hello.

Looking through this I see some parts that won't work for us. `SDL_GetPixelFormatDetails` and `SDL_GetSurfaceClipRect` need to be error checked, for one thing. Also when mapping to or from the same surface over and over again getting the format every time is going to be a large hit to performance, which doesn't come up a bunch in surface.c but does apply at least to `surf_get_bounding_rect`.

src_c/_pygame.h
gresm: Thanks for catching that. Regarding the SDL2-way usage - I just copied how ``PG_LockMutex`` and ``PG_UnlockMutex`` were done just above this.

gresm:
> ``SDL_GetPixelFormatDetails`` and ``SDL_GetSurfaceClipRect`` need to be error checked, for one thing.

Oh right... ``SDL_GetSurfaceClipRect`` errors only when invalid surface is passes, so I assumed that it would be checked, but in many cases it doesn't seem like it. And for ``SDL_GetPixelFormatDetails``, didn't notice that.

> which doesn't come up a bunch in surface.c but does apply at least to ``surf_get_bounding_rect``.

You're right, but currently seems like it's only for ``surf_get_bounding_rect``.

src_c/_pygame.h
gresm: Definitely something to remember in the future ports, but in the ``src_c/surface.c`` it seems like only ``surf_get_bounding_rect`` falls in this category.

src_c/_pygame.h
```diff
@@ -78,6 +80,15 @@
...
 #define PG_PixelFormatEnum SDL_PixelFormat
+#define PG_PixelFormat const SDL_PixelFormatDetails
+#define PG_PixelFormatMut SDL_PixelFormatDetails
+#define PG_SurfaceFormatEnum(surf) (surf)->format
+#define PG_GetSurfaceFormat(surf) \
+    SDL_GetPixelFormatDetails(PG_SurfaceFormatEnum(surf))
+#define PG_GetSurfacePalette SDL_GetSurfacePalette
+#define PG_SurfaceMapRGBA(surf, r, g, b, a) \
+    SDL_MapSurfaceRGBA(surf, r, g, b, a)
+#define PG_GetSurfaceClipRect SDL_GetSurfaceClipRect
```
Starbuck5, line 91: Technically SDL_GetSurfaceClipRect can fail and return false.

Starbuck5:
Hi Gresm, thanks for your effort in working this up in the first place, but these changes have been supplanted by other incremental Surface SDL3 PRs, so I will close this out.

PR #3217, opened 2024-11-09
Port event.c to SDL3

ankith26:
`event.c` is now compiling on SDL3.

This PR takes our usual strategy of not changing any SDL2 codepaths, and I have been super conservative with the changes.

The thing that needs discussion here is: how are we going to handle the gesture changes? For now I have just ifdef'd out the gesture parts, but maybe we should consider vendoring in SDL_gesture? Thoughts on this, chat?


Starbuck5 (approved):
Thanks Ankith!

I'm approving this because I think it is safe and ready to merge, but I also have some obvious suggestions you've probably thought about yourself.

It seems like the largest difference by lines is `event->key.keysym` vs `event->key`. Maybe there could be a compatibility macro like `PG_EV_KEYDATA(event)` or something, I believe this would shrink the diff in event.c. I also think the cdevice -> gdevice, ctouchpad -> gtouchpad stuff would be a good opportunity for a compat macro.

zoldalma999 (approved):
Looks good to me, thanks!

PR #3221, opened 2024-11-12
add new angle and angle_rad property for vector2

AntoineMamou:
This PR follows this one : https://github.com/pygame-community/pygame-ce/pull/3216 and make the suggested changes. 

I added two new properties for Vector2. 
`angle_rad` : Returns the vector’s angle in radians relative to the positive x-axis.
`angle` : Returns the angle in degrees, normalized to (180,180].


AntoineMamou:
something was wrong, sorry

Issue #3195, opened 2024-10-27
Add `Vector2.angle` and `Vector2.angle_rad` properties


aatle:
**Proposal:**
Add properties `angle` and `angle_rad` to `pygame.Vector2` with getter and possibly a setter.

```py
vec.angle_rad == math.atan2(vec.y, vec.x) == vec.as_polar()[1]  # note: no vec.angle_rad_to() method
vec.angle == -vec.angle_to([1, 0]) == degrees(vec.angle_rad)
```
(These angles are normalized: degrees within interval (-180, 180], even if set to an angle outside this range.)
The zero vector has an angle of 0.

**Reasons:**
In math, the angle of a vector is often relative to the positive x-axis (increasing angle from 0 goes towards positive y-axis).
For example, polar coordinates and trigonometry functions are connected to vectors through their angle to positive x-axis.
`vec == Vector2(r * cos(theta), r * sin(theta))`

The current ways to get these angles are indirect and less readable, for such an (IMO) important aspect of vectors. See above. In `as_polar()` result, the length is available via `vec.length`, but the angle is not directly available.

In a codebase, it's convenient to only have to store one float (angle to +x axis) to mark the direction of something, rather than another vector (especially when the distance between two points is unnecessary/inapplicable). Many physics libraries or engines use angles like this heavily, usually using radians. Even if angles are technically only relative, to store them conveniently there has to be 'absolute' angles.

<details><summary>Example uses:</summary>

Calculate the angle of one position to another position, in degrees.
```py
# Before
self.angle = (vec2 - vec1).angle_to([1, 0])
# After
self.angle = (vec2 - vec1).angle
```
Find the angle between two vectors, in radians.
```py
# Before
self.angle_rad = math.radians(vec1.angle_to(vec2))
self.angle_rad = vec2.as_polar()[1] - vec1.as_polar()[1]
# After
self.angle_rad = vec2.angle_rad - vec1.angle_rad
```
Rotate a vector in-place at the direction of a given angle (from +x axis), in degrees.
```py
# Before
vec.rotate_ip(angle - vec.angle_to([1, 0]))
# After
vec.angle = angle
```
</details>

AntoineMamou:
Hi @aatle,
Can I work on this issue ? I understand that these tow properties : `angle_rad` and `angle`  shoul be add in `pygame.Vector2`.
`angle_rad`: Returns the vector’s angle in radians relative to the positive x-axis.
[...]

PR #3222, opened 2024-11-12
new angle and angle_rad property in vector2

AntoineMamou:
This PR follows this one : https://github.com/pygame-community/pygame-ce/pull/3216

I make the changes needed and add two properties : 
`angle_rad` : Returns the vector’s angle in radians relative to the positive x-axis.
`angle` : Returns the angle in degrees, normalized to (180,180].

### DIFF
--- a/docs/reST/ref/math.rst
+++ b/docs/reST/ref/math.rst
@@ -625,7 +625,7 @@ Multiple coordinates can be set using slices or swizzling
 
       Usage: Accessing `angle` provides the current angle of the vector in degrees within the predefined range of [-180, 180].
 
-      .. versionadded:: 2.5.4
+      .. versionadded:: 2.5.5
 
    .. attribute:: angle_rad
 
@@ -636,7 +636,7 @@ Multiple coordinates can be set using slices or swizzling
 
       Usage: Accessing `angle_rad` provides the current angle of the vector in radians within the predefined range of [-π, π].
 
-      .. versionadded:: 2.5.4
+      .. versionadded:: 2.5.5
 
    .. ## pygame.math.Vector2 ##
 


src_c/math.c
```diff
@@ -1269,6 +1273,48 @@ vector_setz(pgVector *self, PyObject *value, void *closure)
     return vector_set_component(self, value, 2);
 }
 
+static PyObject *
+vector_get_angle(pgVector *self, void *closure)
+{
+    pgVector *vec = (pgVector *)self;
+
+    if (vec->coords[0] == 0.0 && vec->coords[1] == 0.0) {
+        return PyFloat_FromDouble(0.0);
```
oddbookworm, line 1282: I’d rather an exception get raised here. The zero vector is special and getting it’s angle makes no sense

src_c/math.c
```diff
@@ -1269,6 +1273,48 @@ vector_setz(pgVector *self, PyObject *value, void *closure)
     return vector_set_component(self, value, 2);
 }
 
+static PyObject *
+vector_get_angle(pgVector *self, void *closure)
+{
+    pgVector *vec = (pgVector *)self;
```
oddbookworm, line 1279: Why this cast to the type it already is?

src_c/math.c
```diff
@@ -1269,6 +1273,48 @@ vector_setz(pgVector *self, PyObject *value, void *closure)
...
+
+    if (vec->coords[0] == 0.0 && vec->coords[1] == 0.0) {
+        return PyFloat_FromDouble(0.0);
+    }
+
+    double angle = atan2(vec->coords[1], vec->coords[0]) * 180.0 / M_PI;
+
+    if (angle > 180.0) {
```
oddbookworm, line 1287: Make constants instead of reusing the same hardcoded numbers

src_c/math.c
```diff
@@ -1269,6 +1273,48 @@ vector_setz(pgVector *self, PyObject *value, void *closure)
...
+    }
+    else if (angle <= -180.0) {
+        angle += 360.0;
+    }
+
+    return PyFloat_FromDouble(angle);
+}
+
+static PyObject *
+vector_get_angle_rad(pgVector *self, void *closure)
```
oddbookworm, line 1298: Just reuse the function you already have instead of writing the same code twice. Since you already got the output in a form you like, just convert it from degrees to radians

src_c/math.c
```diff
@@ -2585,6 +2631,8 @@ static PyMethodDef vector2_methods[] = {
 static PyGetSetDef vector2_getsets[] = {
     {"x", (getter)vector_getx, (setter)vector_setx, NULL, NULL},
     {"y", (getter)vector_gety, (setter)vector_sety, NULL, NULL},
+    {"angle", (getter)vector_get_angle, NULL, NULL, NULL},
```
oddbookworm, line 2634: Missing the docs component of these structs, which should be generated in the docs header, which is generated from the documentation updates you need to make

oddbookworm (changes requested):
Needs documentation, and there are a few issues I see offhand

src_c/math.c
AntoineMamou: ok, I thought it should be 0 because of the initial issue https://github.com/pygame-community/pygame-ce/issues/3195, but I can change and raise an exception

src_c/math.c
oddbookworm: No, you need to add to math.rst

oddbookworm: Yes

AntoineMamou: Hello @oddbookworm, I have a question about the building of the docs. When I try to build the docs I add in math.rst, I get two new lines in math_doc.h, which is what I want. However I get this : 
```
#define DOC_MATH_VECTOR2_ANGLE ""
#define DOC_MATH_VECTOR2_ANGLERAD "" 
```

However, I try to add a short descriptive sentence like this : 
```
| :sl: `Gives the angle of the vector in degrees, relative to the X-axis, normalized to the interval (-180, 180].`
```
Is this normal ?

src_c/math.c
aatle: @oddbookworm 
I strongly think the angle properties should be consistent with the existing behavior of `.as_polar()`, `.angle_to()`, `.rotate()` and more importantly [python `math.atan2()`](https://github.com/python/cpython/blob/9a3ffc0492d1310ead9ce8f5ee678c26b20a338d/Modules/mathmodule.c#L591) (which is based off the a C standard) and polar coordinates.
These functions don't raise errors and (for not `nan` args) don't return `nan`, and most actually do not document this special case.

oddbookworm: I still don't agree that we should silently just say the angle of the zero vector is zero. I don't like those functions treating that special case in the way they do either. And actually, the C standard **does not** guarantee that `atan2(0, 0)` returns 0. Here's a snippet from the C standard
image

oddbookworm: One reason I hate that behavior is because it can lead to very confusing behavior. Consider the case where you're just continuously scaling down a vector without changing it's direction. At some point, you'll go from a constant angle to 0, regardless of what the starting angle is.

aatle: From what I gathered from my sources including [this](https://stackoverflow.com/questions/55037585/is-it-safe-to-assume-that-atan20-0-returns-0) and [this](https://stackoverflow.com/questions/47909048/what-will-be-atan2-output-for-both-x-and-y-as-0) and [this](https://en.wikipedia.org/wiki/Atan2#Realizations_of_the_function_in_common_computer_languages), at least some sort of standard.
Silently saying the angle of the zero vector is zero (or even `pi` or `-pi` as is in some of the functions' behavior) is not really that bad because if the magnitude is 0, then the angle practically doesn't even matter (in contrast with `scale_to_length` which actually needs it). The polar coordinates `(0.0, x)` represent the origin for all x.
I think an error (or `nan`) is worse, for reasons which I will explain.

aatle: > At some point, you'll go from a constant angle to 0, regardless of what the starting angle is.

Setting the length to 0 (or continuously scaling down, in which case using `scale_to_length` must be special cased anyway) inevitably loses the angle data, and an undocumented angle is as valid to use as an error.
Most of the time, when a zero vector is involved, the angle does not matter so it is better to *not* raise an error and instead give a valid value. Since the magnitude is 0, any finite value including 0.0 is a valid angle to represent the zero vector.
For practicality reasons and consistency, the angle getter and setter shouldn't raise an error.

oddbookworm: I disagree. If literally any answer makes as much sense as any other, we should not assert any answer IMO. As for implementing per Annex F as your sources indicate, the second one in particular actually shows the relevant content of that annex, which is more complex than just "return 0". It actually can return +0, -0, +pi, -pi technically per that annex (but of course, annexes are just suggestions, they're not a mandatory part of the spec). But it's ultimately implementation-specific. You also don't need to use `scale_to_length` either to scale a vector down. It has the `*` and `/` operators defined on it. I can create this behavior with a bit of fiddling with the boundaries of what doubles will allow (on my system with my installed python)
image

oddbookworm:
```python3
import pygame

pygame.init()

window = pygame.Window()
screen = window.get_surface()

my_vector = pygame.Vector2(30, 43)
print(f"Starting angle: {my_vector.angle}")

SCALE_FACTOR = 0.1

running = True
shrinking = True
while running:
        if event.type == pygame.QUIT:
            running = False

    if shrinking and my_vector.x > 5e-324:
        my_vector *= SCALE_FACTOR
        print(f"New angle (shrinking): {my_vector.angle}, {my_vector}")
    elif my_vector.length() < 100:
        shrinking = False
        my_vector /= SCALE_FACTOR
        print(f"New angle (growing): {my_vector.angle}, {my_vector}")
```

oddbookworm: And it's not even 100% consistent with itself. This came from the same code
image

aatle: I believe your example (of using the angle of a scaled vector instead of other way around) is uncommon and already bad to implement by itself because of floating point limits. The zero vector itself does not have to be reached for the angle to change because the design of the code encounters imprecision. If one were to actually hit the zero vector and face the special behavior, they wouldn't even be able to grow it back.

We do not have to assert any answer. Like python `atan2`, it can be left undocumented but still give a value instead of an error, even if its implementation specific (which we can prevent).
The point is, using an error is worse than an arbitrary but valid angle, for various reasons. Mathematically, this isn't even really bad.

oddbookworm: How is raising an error worse than giving an undocumented result? At least an error gives the user some kind of feedback that maybe something went wrong rather than just silently accept that the user knows what they're doing (which is never a given)

aatle: First, mathematically it is okay to give e.g. 0.0 for the angle. All vectors with a magnitude of 0 are equivalent regardless of their direction, and polar coordinates needs the origin anyway. Imagine if the vector were to be implemented with magnitude + direction internally instead of components.

Having a possible error for getting or setting an `angle` is *very* annoying and would create *rare* crashes, cases that would most likely be better off having a defined and even undocumented angle for the zero vector.
One thing is that the angle is a property and is essentially an attribute. Nobody likes it when their attribute accesses raise a `ValueError` depending on the contents of the object.

Second, using a valid angle for it is superior as default behavior than raising an error. Imagine an unknowing user, implementing something that requires the angle between two points. When the two points are equivalent which is very rare, 99% of cases you want it to give any angle instead of blowing up and necessitating extra code checks.
My pymunk code would be exploding if its `Vec2d` gave an error for the angle of the zero vector, and much more error-prone.

oddbookworm: Mathematically, ***anything*** is acceptable because it's undefined. 0 is just as valid as 180 degrees/pi radians. That doesn't mean we ***should***. All polar vectors of magnitude 0 are equivalent, regardless of their angle. If you have a vector defined with `r=0`, then `theta` does not matter at all. There's absolutely no rule in math that it ***must*** be 0, because that's nonsense. It doesn't make sense at all to even talk about the angle of the zero vector.

No, I 100% disagree that having an undocumented return in one special case is better than actually requiring the user to be responsible. We are a graphics library (plus other stuff). That's not our job. If we were an actual game engine that abstracted that away, then sure. But we're not. Our own opinions should not make it into the library. Be explicit. Make the user validate that their thing makes sense. Also, I will not support setting angle using these methods. Read-only attribute. If you need to set the angle, do it properly using the components or one of the rotate methods.

"Superior" is subjective. IMO the superior thing to do would be to check if the two points are the same before even thinking about getting the angle between them. Give me one good use-case where not checking that beforehand is overly burdensome to the user. I think in most cases, it would actually allow for short-circuits in code. Location and destination are the same? Don't do anything then. Easy. Fast. Error-free.

aatle: Continuing the discussion on discord

AntoineMamou: Hello @aatle and @oddbookworm, 
I saw on discord that there is many vision for what should be done for the zero vector. And I wanted to know if people agree on one of these in order to maybe change my code

AntoineMamou:
I see that two checks were not successful but I don't understand why. Can someone help me ?

src_c/math.c
aatle: Hi @AntoineMamou, it has been settled that we will imitate python's `math.atan2()` function behavior. (Returns certain value depending on sign of zeroes, for zero vector.)
See https://github.com/python/cpython/blob/ed81971e6b26c34445f06850192b34458b029337/Lib/test/test_math.py#L327 for the relevant test cases.
Here's the old CPython 3.7.2 implementation for special cases (I can't find the most recent implementation): https://github.com/python/cpython/blob/v3.7.2/Modules/mathmodule.c#L566-L600

AntoineMamou: Hi @aatle, so if we want to imitate math.atan2(), for the case where x=y=0, it should return 0, right ?

AntoineMamou: And I'm using atan2 to find the angle, so basically I can skip the if  statement ?

AntoineMamou: Maybe I'm missing something here

buildconfig/stubs/pygame/math.pyi
```diff
@@ -221,6 +221,8 @@ class Vector2(_GenericVector):
     xy: Vector2
     yx: Vector2
     yy: Vector2
+    angle: float
+    angle_rad: float
```
aatle, line 225:
```suggestion
    @property
    def angle(self) -> float: ...
    @property
    def angle_rad(self) -> float: ...
```
These are properties and they are read-only.

docs/reST/ref/math.rst
```diff
@@ -616,6 +616,23 @@ Multiple coordinates can be set using slices or swizzling
       find that either the margin is too large or too small, in which case changing ``epsilon`` slightly
       might help you out.
 
+   .. attribute:: angle
+
+      | :sl:`Gives the angle of the vector in degrees, relative to the X-axis, normalized to the interval (-180, 180].`
+
+      Read-only attribute representing the angle of the vector in degrees relative to the X-axis. This angle is normalized to
+      the interval (-180, 180].
```
aatle, line 624: My mistake, the interval is actually inclusive of -180. So, `[-180, 180]`

docs/reST/ref/math.rst
```diff
@@ -616,6 +616,23 @@ Multiple coordinates can be set using slices or swizzling
...
+
+      Usage: Accessing `angle` provides the current angle of the vector in degrees within the specified range.
```
aatle, line 626: Reword "specified range". The word 'specified' implies that the user provides it.

src_c/math.c
```diff
@@ -1269,6 +1275,54 @@ vector_setz(pgVector *self, PyObject *value, void *closure)
...
+    PyObject *angle_obj = vector_get_angle_rad(self, closure);
+    double angle_rad = PyFloat_AsDouble(angle_obj);
+    double angle_deg = angle_rad * RAD_TO_DEG;
+
+    if (angle_deg > 180.0) {
+        angle_deg -= 360.0;
+    }
+    else if (angle_deg <= -180.0) {
+        angle_deg += 360.0;
+    }
```
aatle, line 1290: Seems unnecessary. The radians is within `[-pi, pi]`, so the degrees will be within `[-180, 180]`.

src_c/math.c
```diff
@@ -1269,6 +1275,54 @@ vector_setz(pgVector *self, PyObject *value, void *closure)
...
+
+    return PyFloat_FromDouble(angle_deg);
+}
+
+static PyObject *
+vector_get_angle_rad(pgVector *self, void *closure)
```
aatle, line 1296: I would suggest creating a helper function `_pg_atan2` that takes doubles and returns double. That way, one doesn't have to convert back to double and then to pyobject when doing more operations on it.

buildconfig/stubs/pygame/math.pyi
AntoineMamou: `@property` doesn't say that it's a read-only attribute, right ?

src_c/math.c
AntoineMamou: Where I should make the documentations for `_pg_atan2` ?

aatle (approved):
LGTM.
Thanks for implementing my feature request.

docs/reST/ref/math.rst
```diff
@@ -616,6 +616,23 @@ Multiple coordinates can be set using slices or swizzling
...
 
+   .. attribute:: angle
+
+      | :sl:`Gives the angle of the vector in degrees, relative to the X-axis, normalized to the interval [-180, 180].`
+
+      Read-only attribute representing the angle of the vector in degrees relative to the X-axis. This angle is normalized to
+      the interval [-180, 180].
+
+      Usage: Accessing `angle` provides the current angle of the vector in degrees within the predefined range of [-180, 180].
+
```
aatle, line 627: need to document `versionadded` e.g.
```reST
.. versionadded:: 2.5.x
```

damusss: > Should this be `.. versionadded:: 2.5.3` ? I'm not sure about which version I should put.

yes 👍

LittleBread69 (approved):
Legitimate.
Had no issues running it on MacOS 15.1 with python3.12 and python3.10 .

ankith26 (commented):
Hi, sorry for taking a while to get to this proposal.

Personally I really like the overall idea. Definitely +1 for keeping it as a property. A couple of things to discuss
- Do we really need both `angle` and `angle_rad`? Yes, I can understand why it was done (consistency with other vector methods), but because this is a new property, we can consider dropping one of these. I'm just putting this idea up for discussion, personally have no opinion here
- Why restrict it to read only property? I can imagine that setting an angle without changing magnitude would be a useful feature as well. Though that could be a separate follow up PR too.

Again, thanks for taking up this proposal and contributing to pygame-ce

aatle:
To answer those questions,
Yes, we need both `angle` and `angle_rad`. Degrees is not droppable. Radians is very useful with many mathematical properties and a relation to trigonometric functions. Because of this, radians is often used instead of degrees in many places: physics engines, math/science, etc.
It has been implemented as a read-only property at @oddbookworm's [request](https://github.com/pygame-community/pygame-ce/pull/3216#pullrequestreview-2420984544), the reasoning given on Discord as:
> If you want to change the vector, you do it component-wise or use the other methods available.

I guess since it is a property, it may be appropriate to allow setting; however, there should be a non-inplace equivalent. This functionality can be always be added in a later PR though.

docs/reST/ref/math.rst
ankith26: For what case would it return -180? Shouldn't it just return 180 in that case because essentially they are the same thing?

aatle: The case of `math.atan2(-0.0, -1.0)` returns `-pi` due to special edge cases: the sign of y is copied to the result here.

ankith26: oh okay. Yeah just another float quirk IG. Not exactly ideal but if the standard atan2 behaviour is this then who are we to judge

ankith26 (approved):

ankith26:
Requesting a re-review from @oddbookworm here as they had previously reviewed it.

docs/reST/ref/math.rst
```diff
@@ -616,6 +616,27 @@ Multiple coordinates can be set using slices or swizzling
...
+   .. attribute:: angle_rad
+
+      | :sl:`Gives the angle of the vector in radians, relative to the X-axis, normalized to the interval [-π, π].`
+
+      Read-only attribute representing the angle of the vector in radians relative to the X-axis. This value is equivalent
+      to the `angle` attribute converted to radians and is normalized to the interval [-π, π].
+
+      Usage: Accessing `angle_rad` provides the current angle of the vector in radians within the predefined range of [-π, π].
+
+      .. versionadded:: 2.5.4
```
oddbookworm, line 639: Unfortunately, it took me too long to get back to this before they decided to release a dev version of 2.5.4 (I was asleep). This needs to be updated to 2.5.5
```suggestion
      .. versionadded:: 2.5.5
```

oddbookworm (approved):
LGTM! 🥳
Sorry it took me so long to get back to this, I've been super busy irl 😢

oddbookworm:
As soon as everything passes, I'll merge it

PR #3224, opened 2024-11-17
Batch and error check allocations

Starbuck5:
I was looking into optimization opportunities and found some allocation requests (mallocs and PyMem_News) right next to each other, and I thought surely calling the function once will be faster than calling it twice. It is also simpler to check if one buffer failed to allocate, and simpler to free one buffer at the end, rather than 2 or 3.

Looking through the code I also saw that the malloc statements in the blur routines weren't error checked at all, so I fixed that.

Running very favorable microbenchmarks I saw very small performance increases, like 1-3%, or no impact. I'm unsure if running a tight loop over identical calls reduces the execution time burden of malloc / PyMem_New, so maybe performance would be affected differently in a more realistic scenario?

Anyways since this is a small thing I'm not trying to pitch this as a performance PR, instead this is a code quality PR. We want our code to be the best it can be, and I think that reducing the number of allocation calls helps that goal.

### DIFF
--- a/src_c/transform.c
+++ b/src_c/transform.c
@@ -3562,7 +3562,7 @@ surf_average_color(PyObject *self, PyObject *args, PyObject *kwargs)
     return Py_BuildValue("(bbbb)", r, g, b, a);
 }
 
-static void
+static int
 box_blur(SDL_Surface *src, SDL_Surface *dst, int radius, SDL_bool repeat)
 {
     // Reference : https://blog.csdn.net/blogshinelee/article/details/80997324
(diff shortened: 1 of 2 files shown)


src_c/transform.c
```diff
@@ -3574,9 +3574,17 @@ box_blur(SDL_Surface *src, SDL_Surface *dst, int radius, SDL_bool repeat)
...
-    Uint32 *sum_h = malloc(nb * sizeof(Uint32));
+
+    // Allocate bytes for buf, sum_v, and sum_h at once to reduce allocations.
+    Uint32 *overall_buf = malloc(sizeof(Uint32) * (dst_pitch * 2 + nb));
+    Uint32 *buf = overall_buf;
+    Uint32 *sum_v = overall_buf + dst_pitch;
+    Uint32 *sum_h = overall_buf + dst_pitch * 2;
+
+    if (overall_buf == NULL) {
+        PyErr_NoMemory();
```
ankith26, line 3585: This function is called within `Py_BEGIN_ALLOW_THREADS`/`Py_END_ALLOW_THREADS` so I don't think we are allowed to call this here as it is python API.

Starbuck5: Good catch! Fixed.

ankith26 (approved):

bilhox (changes requested):
LGTM, just a little thing to fix, but otherwise it seems fair.

bilhox (approved):
LGTM

PR #3226, opened 2024-11-19
Utility Window implementation.

gresm:
Adds ``pygame.window.Window(utility=True)`` which makes the window not appear in the taskbar (+``pygame.window.Window.utility`` to get if this option is enabled for the window).
Note: SDL might not support this for all platforms (macOS?).


docs/reST/ref/window.rst
```diff
@@ -289,6 +290,15 @@
 
       .. versionadded:: 2.5.0
 
+   .. attribute:: utility
+
+      | :sl:`Get if the windos is an utility window`
+      | :sg:`utility -> bool`
+
+      ``True`` if the window doesn't appear in the task bar, ``False`` otherwise
```
damusss, line 298: If it's true it's not guaranteed to work on all platforms then it should be noted somehow I think.

damusss (changes requested):

docs/reST/ref/window.rst
gresm: On some platforms this will fail silently. It's a good idea to write about that, but I don't have a comprehensive list for which platforms will this work.

damusss: You should just say it's not guaranteed to behave as expected on all platforms. This will stop people from coming to us if it's not actually hidden in the taskbar (or if they do, we can point out that it was written all along)

gresm: Okay, this seems to work only for X11 and Windows.

docs/reST/ref/window.rst
```diff
@@ -43,6 +43,7 @@
                               (unrelated to INPUT_GRABBED).
    :param bool always_on_top: Create a window that is always presented above
                               others.
+   :param bool utility: Create a window that doesn't appear in the task-bar .
```
damusss, line 46:
```suggestion
   :param bool utility: Create a window that doesn't appear in the task-bar.
```
(it bugs me lmao)

gresm: Weird that this happened, but fixed it.

docs/reST/ref/window.rst
```diff
@@ -289,6 +290,16 @@
 
       .. versionadded:: 2.5.0
 
+   .. attribute:: utility
+
+      | :sl:`Get if the windos is an utility window`
```
damusss, line 295: Other (except for opengl, for no reason) read-only properties of Window append `(**read-only**)` to the :sl:, I believe this could do the same

yunline:
Since SDL3 uses `SDL_WINDOW_SKIP_TASKBAR` as the flag, is it better to use `skip_taskbar` instead of `utility` as the name of the argument?
I think `skip_taskbar`  is more clear than `utility`

Starbuck5:
> Since SDL3 uses `SDL_WINDOW_SKIP_TASKBAR` as the flag, is it better to use `skip_taskbar` instead of `utility` as the name of the argument? I think `skip_taskbar` is more clear than `utility`

I love the way you're thinking but I think it's the other way around. SDL3 is utility, SDL2 is skip taskbar.

gresm:
@yunline @Starbuck5 So, what's the decision?

damusss:
> So, what's the decision?

I wasn't asked, but I think it's best to follow sdl3. plus, `utility` is a bit more generic leaving "freedom of conceptual behaviour" to the OS. it just means the window is an utility, which usually results in no taskbar, but the OS can decide what utility means. I vote for utility. Btw you forgot to run `py dev.py all`

ankith26:
I also vote for leaving it as `utility` for reasons damus already mentioned, and also the fact that it the name SDL3 picked. Some setups don't have the taskbar at all, and SDL may still do something differently if `utility` is passed.

yunline:
> So, what's the decision? 

SDL docs said: "window should be treated as a utility window, not showing in the **task bar** and **window list**". 

From the docs we can see "utility window" is a broader concept than "skip taskbar window". They are not the same.

Some window environments don't have a task bar, they use window list

window list picture copied from somewhere of the internet

Some window environments use dock



Although "skip_task_bar" could be more clear for Windows users, "utility" is more universal on all platforms.

So I vote for "utility".

gresm:
Then it's ``utility``.

docs/reST/ref/window.rst
```diff
@@ -289,6 +290,16 @@
...
 
+   .. attribute:: utility
+
+      | :sl:`Get if the windos is an utility window (**read-only**)`
+      | :sg:`utility -> bool`
+
+      ``True`` if the window doesn't appear in the task bar, ``False`` otherwise.
+      This only works for X11 and Windows, for other platforms, creating ``Window(utility=True)`` won't change anything.
+
+      .. versionadded:: TBD
```
yunline, line 301:
```suggestion
      .. versionadded:: 2.5.0
```

ankith26 (approved):
Normally I would complain about adding new API that isn't cross platform, but in the case of `Window` I guess we already have enough API that isn't cross platform so it shouldn't be out of place.

Thanks for the PR! :tada:

PR #3229, opened 2024-11-20
Modification in draw.rect

Neri19:
Changed the function rect in draw.c to call the function draw_round_rect with a radius positive or 0 to resolve issues with negative radii. In fact, in the doc, the radius can't be negative. 
We can choose to raise an exception but i don't think that it's necessary.

Related to #2285
I don't know if it has been fixed by #2303


Matiiss:
A more relevant open PR addressing this issue would be https://github.com/pygame-community/pygame-ce/pull/3158

MyreMylar:
This branch (a main branch) now seems to have a selection of random changes smooshed together, and we also already have a PR tackling this subject so I'm closing it.

@Neri19 feel free to jump onto the other PR on this topic if you want to help get that one over the line.

PR #3232, opened 2024-11-21
Implement lazy loading of pygame submodules (`surfarray`, `sndarray`)

aatle:
**Motivation:**
Previously, I discovered via profiling that `numpy` was being imported or loaded in my pygame program, when there were no mandatory runtime dependencies on it, though I had it installed.
The pygame submodules `surfarray` and `sndarray` both attempt to import numpy if possible, and pygame always tries to load these submodules. (If numpy is not available, they are replaced by `MissingModule`.)

When running `import pygame`, `numpy` (if available) loading takes up at least 40% of the time, around 100 milliseconds depending on the hardware. However, the `surfarray` and `sndarray` modules are not used in many pygame programs, and some users install `numpy` globally, or, like me, have dev dependencies that use numpy.

**Implementation:**
To fix this, `numpy` shouldn't be loaded unless it is actually needed by the user. This requires lazy loading/importing. Obviously, the implementation should be backwards-compatible.

<details><summary>Considered lazy import implementations</summary>

Five ways to implement lazy loading that I know of:
1. Don't import submodule in the parent package itself
2. Put all expensive imports inside of functions
3. Use [importlib.util.LazyLoader](https://docs.python.org/3/library/importlib.html#implementing-lazy-imports)
4. Use a custom `ModuleType` subclass, as in https://pypi.org/project/lazy-imports/
5. (**Chosen**) Use the [module `__getattr__` feature](https://peps.python.org/pep-0562/) (python 3.7+), similar to as in https://snarky.ca/lazy-importing-in-python-3-7/

Only the fifth is suitable because of backwards compatibility. It also happens to be one of the simplest.
Problems with 1 - 4:
1. Cannot implement the existing `MissingModule` behavior (when `numpy` is unavailable)
2. Too much effort, massive changes to code of the submodules. Does not preserve `MissingModule` behavior
3. Cannot implement `MissingModule` upon import failure
4. Too complicated. Cannot implement `MissingModule` behavior

</details>

The pygame submodules `surfarray` and `sndarray` will be lazily imported, deferring execution/loading of them to when the user themselves access it (if they do), e.g. `pygame.surfarray.array2d` or `import pygame.sndarray`.
If they do use it, `numpy` will usually not be loaded at an inconvenient time (to cause framedrops) *unless* the first use (including imports) is of the form `pygame.surfarray.___` inside a function (easy user fix).

**Backwards compatibility / Testing:**
<details><summary>All combinations of these factors must be compatible</summary>

- numpy is or is not installed
- `sndarray`, `surfarray` is or is not accessed by user
- If accessed, the access method used:
  - `import pygame.surfarray`
  - `from pygame import surfarray`
  - `pygame.surfarray.array2d`
  - `from pygame.surfarray import array2d`

In addition:
- Accessing undefined attribute on `pygame` module (same error message)
- PyInstaller and such must still work

</details>

I did not include test cases in this PR, though I think the implementation is probably compatible.

Notes:
Since `MissingModule` class is deleted at the end of `__init__.py` and unavailable to `__getattr__`, there is a new private alias `_MissingModule` in the pygame namespace.
I don't like the `MissingModule` behavior that much because half of the access forms bypass it.
If this PR is merged, then more optional pygame submodules may be lazily imported in the future.

### DIFF
--- a/docs/reST/ref/sndarray.rst
+++ b/docs/reST/ref/sndarray.rst
@@ -23,7 +23,7 @@ Each sample is an 8-bit or 16-bit integer, depending on the data format. A
 stereo sound file has two values per sample, while a mono sound file only has
 one.
 
-.. versionchanged:: 2.5.3 sndarray module is lazily loaded
+.. versionchanged:: 2.5.3 sndarray module is lazily loaded to avoid loading NumPy needlessly
 
 .. function:: array
 
--- a/docs/reST/ref/sndarray.rst
+++ b/docs/reST/ref/sndarray.rst
@@ -23,6 +23,8 @@ Each sample is an 8-bit or 16-bit integer, depending on the data format. A
 stereo sound file has two values per sample, while a mono sound file only has
 one.
 
+.. versionchanged:: 2.5.3 sndarray module is lazily loaded
+
 .. function:: array
 
    | :sl:`copy Sound samples into an array`
--- a/src_py/__init__.py
+++ b/src_py/__init__.py
@@ -264,17 +264,32 @@ except (ImportError, OSError):
 
 # lastly, the "optional" pygame modules
 
+# Private, persisting alias for use in __getattr__
 _MissingModule = MissingModule
 
 
 def __getattr__(name):
+    """Implementation of lazy loading for some optional pygame modules.
+
+    The surfarray and sndarray submodules use numpy, so they are loaded
+    lazily to avoid a heavy numpy import if the modules are never used.
+
+    The first access of a lazily loaded submodule loads it and sets it
+    as an attribute on the pygame module. Pygame itself doesn't import these modules.
+    If the first access is an attribute access and not an import, then __getattr__ is
+    invoked (as the attribute isn't set yet), which imports the module dynamically.
+
+    All lazy submodules are directly referenced in the packager_imports function.
+    """
     from importlib import import_module
 
     LAZY_MODULES = "surfarray", "sndarray"
     if name not in LAZY_MODULES:
+        # Normal behavior for attribute accesses that aren't lazy modules
         raise AttributeError(f"module '{__name__}' has no attribute '{name}'")
     try:
         module = import_module(f"{__name__}.{name}")
+        # A successful import automatically sets the module attribute on the package
     except (ImportError, OSError):
         module = _MissingModule(name, urgent=0)
         globals()[name] = module
--- a/src_py/__init__.py
+++ b/src_py/__init__.py
@@ -72,6 +72,8 @@ class MissingModule:
     _NOT_IMPLEMENTED_ = True
 
     def __init__(self, name, urgent=0):
+        import sys
+
         self.name = name
         exc_type, exc_msg = sys.exc_info()[:2]
         self.info = str(exc_msg)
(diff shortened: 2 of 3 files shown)


src_py/__init__.py
```diff
@@ -259,7 +261,26 @@ def PixelArray(surface):  # pylint: disable=unused-argument
 except (ImportError, OSError):
     transform = MissingModule("transform", urgent=1)
 
+
 # lastly, the "optional" pygame modules
+
+_MissingModule = MissingModule
```
Starbuck5, line 267: I think it would make sense to store this as a local variable on `__getattr__`

src_py/__init__.py
```diff
@@ -259,7 +261,26 @@ def PixelArray(surface):  # pylint: disable=unused-argument
...
+
+
+def __getattr__(name):
```
Starbuck5, line 270: This whole lazy loading system is a weird bit of code, there should be a comment to explain what this is doing and why surfarray and sndarray are being singled out for this treatement.

Starbuck5:
There's a block in here called `def packager_imports():` for showing packagers imports that might be missed. I'm not sure it's necessary for this situation but adding explicit surfarray/sndarray imports there would make me feel more confident in this solution.

Also I think this could be worth a small versionchanged or note in the docs.

Starbuck5:
Also, for reviewer reference, the command to test this is `python -X importtime -c "import pygame"`

src_py/__init__.py
aatle: The `MissingModule` class is deleted at the end of the module, and `__getattr__` still needs it, so it must be stored somewhere (not as local variable).
A few options:
- Private alias of `MissingModule`
- Move `MissingModule` to another file
- Bind `MissingModule` as a default keyword argument in `__getattr__`

docs/reST/ref/sndarray.rst
```diff
@@ -23,6 +23,8 @@ Each sample is an 8-bit or 16-bit integer, depending on the data format. A
 stereo sound file has two values per sample, while a mono sound file only has
 one.
 
+.. versionchanged:: 2.5.3 sndarray module is lazily loaded
```
Starbuck5, line 26: As a passing user, I read this.

Why is it lazily loaded?

src_py/__init__.py
Starbuck5: Here's an idea:

```py
>>> a = 36
>>>
>>> def func():
...     print(func.b)
...
>>>
>>> func.b = a
>>> del a
>>> func()
36
```

aatle: Well I suggested that before but ultimately opted for a private alias because function attributes are kind of weird and dynamic. From my reasoning, it would've look strange and I would rather avoid these rare python features.
If it is necessary to avoid exposing `_MissingModule` in the pygame namespace, then sure, but I think the current approach is okay.

aatle:
@damusss uncovered a problem
Normal attribute accesses to the `pygame` module are affected by the presence of `__getattr__` too, leading to slower attribute accesses.
This most likely means a different implementation will have to be used.

MyreMylar:
Looks like this PR is superseded by the other one.

PR #3233, opened 2024-11-21
Fix SDL3 CI apt issues

ankith26:
Our usual fix for the weird apt update issues that happen on github actions was missed from this particular CI, leading to unrelated fails as seen in the recent PR #3232 

This PR adds that fix


oddbookworm (approved):
LGTM!

Starbuck5 (approved):

PR #3234, opened 2024-11-22
Deprecate `Window` `foreign` kwarg and `Window.focus` `input_only` kwarg

ankith26:
I was looking at compiling `Window` under SDL3, and seems like `SDL_WINDOW_FOREIGN` is now gone, with no mention or a replacement in SDL3. The easiest thing to do now seems to be undocumenting, and the code now raises a deprecation warning if the `foreign` kwarg is received.

### DIFF
--- a/src_c/window.c
+++ b/src_c/window.c
@@ -289,6 +289,12 @@ window_focus(pgWindowObject *self, PyObject *args, PyObject *kwargs)
         return NULL;
     }
     if (input_only) {
+        if (PyErr_WarnEx(PyExc_DeprecationWarning,
+                         "The input_only kwarg has been deprecated and may be "
+                         "removed in a future version",
+                         1) == -1) {
+            return NULL;
+        }
         if (SDL_SetWindowInputFocus(self->_win)) {
             return RAISE(pgExc_SDLError, SDL_GetError());
         }
--- a/src_c/window.c
+++ b/src_c/window.c
@@ -923,6 +923,12 @@ window_init(pgWindowObject *self, PyObject *args, PyObject *kwargs)
                     }
                 }
                 else if (!strcmp(_key_str, "foreign")) {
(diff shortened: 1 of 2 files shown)


ankith26:
Also added another deprecation for the `input_only` kwarg which does not have an SDL3 equivalent.

Now the only think left to address in porting `Window` in my view, is `fullscreen_desktop`. I think SDL3 didn't technically remove it, but they seem to have changed it a lot and I may be stuck on how to handle this one

damusss (approved):
Makes sense, LGTM 👍

src_c/window.c
```diff
@@ -289,6 +289,12 @@ window_focus(pgWindowObject *self, PyObject *args, PyObject *kwargs)
         return NULL;
     }
     if (input_only) {
+        if (PyErr_WarnEx(PyExc_DeprecationWarning,
+                         "The input_only kwarg has been deprecated and may be "
+                         "removed in a future version",
+                         1) == -1) {
+            return NULL;
+        }
         if (SDL_SetWindowInputFocus(self->_win)) {
```
damusss, line 298: When approving I assumed you will version check this call with your SDL3 port of window.c

yunline (commented):
Accroding to [this commit](https://github.com/libsdl-org/SDL/commit/e0c45c6c98bc74b49a01ed7320eed284ca0ec085) the `SDL_WINDOW_FOREIGN` is not removed. It's renamed to `SDL_WINDOW_EXTERNAL`.

However I approve to deprecate this kwarg. 

In `Window.__init__` we create a SDL window, However `SDL_WINDOW_EXTERNAL` is a flag for "window not created by SDL". Setting this kwarg doesn't make any sence.

yunline (approved):
LGTM

MyreMylar (approved):
LGTM I guess, external would be a clearer keyword anyway if we need it for some reason (integration with Tkinter maybe?)

PR #3237, opened 2024-11-24
Use `ruff` to also sort imports

Matiiss:
Just a capability that was not being utilized (it was done before when we were using `black` + `isort`)

From `ruff` docs:
> [Sorting imports](https://docs.astral.sh/ruff/formatter/#sorting-imports)
>
> Currently, the Ruff formatter does not sort imports. In order to both sort imports and format, call the Ruff linter and then the formatter:
> ```
> ruff check --select I --fix
> ruff format
> ```
> A unified command for both linting and formatting is [planned](https://github.com/astral-sh/ruff/issues/8232).


.pre-commit-config.yaml
```diff
@@ -25,6 +25,13 @@ repos:
               | docs/LGPL.txt
           )$
 
+  - repo: https://github.com/astral-sh/ruff-pre-commit
+    rev: v0.4.2
+    hooks:  # See pyproject.toml for configuration options.
+      - id: ruff
+        types_or: [ python, pyi, jupyter ]
+        args: [--select, I, --fix]
```
ankith26, line 33: This causes ruff to be run twice, once with these args and once without.

I would prefer a one step solution if possible, also probably one that edits the ruff config already present in `pyproject.toml` for consistency

Matiiss: The pre-commit config does exactly what the docs suggest:
```
ruff check --select I --fix
ruff format
```

And there are two hooks for the pre-commit config: https://github.com/astral-sh/ruff-pre-commit/blob/main/.pre-commit-hooks.yaml

ankith26: Oh okay, I now see that the `id`s are different. But still,
```
  - repo: https://github.com/astral-sh/ruff-pre-commit
    rev: v0.4.2
    hooks:  # See pyproject.toml for configuration options.
      - id: ruff
        types_or: [ python, pyi, jupyter ]
        args: [--fix]
      - id: ruff-format  # Run the formatter
        types_or: [ python, pyi, jupyter ]
```
would reduce some duplication

And we can still do something like below in pyproject.toml, and probably also expand on it as needed
```toml
[tool.ruff.lint]
select = ["I"]
```

Matiiss: Ooh, didn't know you could hook them hooks like that.

Also, this import sorting caused that function in _debug.py to get 3 more statements and apparently that's a statement too many for pylint... How shall that be addressed?

Matiiss: On a second thought, since this is specifically to do with formatting, I'm afraid we'll have to explicitly do this in the hook instead of in pyproject.toml because we may want to use ruff for linting in the future and then if we're gonna have only I selected for linting, well, that won't really work out that great, however, if we select most of the rules, then they will run during this pre-commit hook run.

aatle: The pre-commit hook run will mostly ignore the `pyproject.toml` because of the `--select I` option, so we should have `select = ["I"]`. Also, don't we want the pre-commit hook to run all selected rules, so it shouldn't have a `--select`?

pyproject.toml
```diff
@@ -96,6 +96,9 @@ exclude = [
   "setup.py",
 ]
 
+[tool.ruff.lint]
+fixable = ["I"]
```
aatle, line 100: I don't think this is necessary for the pre-commit. It should probably be `select = ["I"]`

pyproject.toml
```diff
@@ -96,6 +96,9 @@ exclude = [
...
+
```
aatle, line 101: I would recommend adding:
```toml
[tool.ruff.lint.isort]
combine-as-imports = true
```
This will make it so that isort does not put every `from ... import ... as ...` on its own line.

pyproject.toml
Matiiss: This will be necessary later when further configuring ruff as the linter

.pre-commit-config.yaml
Matiiss: No, this hook here is purely for sorting imports as part of other formatting, so we must override the selection of rules because otherwise on top of sorting the imports it will also lint according to the other rules and if those include fixable rules, they will also be fixed, but the goal here is to just sort the imports.

pyproject.toml
aatle: Okay. Could you elaborate more?
- reason for using `fixable` and not `select`
- reason it is necessary to include `fixable = ["I"]`

Matiiss: As explained in my other comment, this is not where we should configure linting rules using `select` because we may want to add other rules on top and then it would lint those rules during the `pre-commit` run, which we don't want because it is meant purely for formatting the code. So, we override the entire `select` for formatting using the CLI so that it only cares about that rule and also only fixes that rule. The `fixable` config won't have any impact on the `pre-commit` runs in this configuration.
It's not strictly necessary to include the `fixable` config, but it might help if someone runs `ruff check --fix` because it won't create a massive diff that they did not want to cause with that command. It's a guard rail in a sense.

aatle: If `select` has no impact on this `pre-commit` run because of the `--select` option, then we should include `select`.

With your current changes, the ruff select list implicitly is this (the ruff default, as we do not set it):
```toml
select = ["E4", "E7", "E9", "F"]
```
When the user uses `ruff check` and `ruff check --fix`, we only want it to warn them about the rules that we use (in this case, only isort), so that they may fix it before the `pre-commit` fixes it (including for isort 'formatting').
Rather than restricting just what is *fixed*, we should restrict what is *checked*.
What is fixed is always a subset of what is checked, so it is erroneous to include isort in `fixable` without including it in `select`. With the current changes, `ruff check [--fix]` will not detect or fix import sorting because the rule is not selected.
```suggestion
select = ["I"]
```
Ruff considers isort to be linting, so we must configure linting rules even if we consider it formatting.

Matiiss: Fair enough, I get what you're saying, if a contributor runs `ruff check` themselves, this would check those default rules and if they did run `ruff check --fix`, it would not fix anything at all (which is certainly not the expected behavior, since sorting the imports is part of what is expected).
I'm not going to remove the `fixable = ["I"]` configuration though, I don't see the need to do that and again, we're likely to move to ruff as the linter, so, I would like to avoid the case that it gets released without this fixable rule restriction and fixes rules that contributors may not expect it to fix.

aatle: I see your reasoning with `fixable`, but I slightly disagree with it, drawing from my experience using `ruff`. If we introduce more linting rules in the future, then it means they are enforced within the pygame-ce source, and contributors should be able to fix it.
> fixes rules that contributors may not expect it to fix

A contributor would actually expect basically all auto-fixable rules to be fixed, the default. This generally doesn't break anything either, ruff is very careful with auto-fixing.

The intended purpose of `fixable` and `unfixable` is to allow the user to manually correct certain rules in case the auto-fix for it may break something. (Of course, many rules in ruff don't support auto-fix in the first place.)
Considering this, we would instead use the more common **`unfixable` list** instead of `fixable` list, so exceptions are for the excluded instead of the included. (The default for `fixable` is `"ALL"`, then we narrow it down.)

Matiiss: After thinking about this for a while, I have concluded that it should be fine to just remove fixable for now

MyreMylar (approved):
Seems reasonable to me. 

I'd rather have these automagically sorted in a consistent manner aligned with python coding standards than having to figure it out myself.

src_py/sprite.py
```diff
@@ -84,15 +84,13 @@
 # specific ones that aren't quite so general but fit into common
 # specialized cases.
 
-import types
```
aatle, line 87: This import is missing and causing test failure.

The import is required by `types.GenericAlias` in `AbstractGroup.__class_getitem__`, which was merged in from main.

Matiiss:
I add back the missing import and also rewrote the method to how I've seen it typically typed out elsewhere, such as the CPython repository: https://github.com/search?q=repo%3Apython%2Fcpython%20%22classmethod(types.GenericAlias)%22&type=code

aatle:
Does the import sorting not also apply to stub files?

Matiiss:
> Does the import sorting not also apply to stub files?

Nope, `pyproject.toml` contains this
```
[tool.ruff]
exclude = [
  "buildconfig/",
  "docs/reST/",
  "setup.py",
]
```

aatle (approved):
LGTM;

Matiiss:
There's a little TODO still left:
- [ ] configure `pyproject.toml` to include `buildconfig/stubs/pygame` and run formatting again

aatle:
I assumed that would be a separate PR, but I guess it could be put with this one for convenience, placing the large changes together.

aatle:
This configuration should properly configure the ruff configuration:
```toml
[tool.ruff]
extend-exclude = [
    "buildconfig/*.py",  # *.pyi stub files still included
    "docs/reST",
    "setup.py",
]
```
Alternatively, `include` can be set to specific directories.

Matiiss:
> This configuration should properly configure the ruff configuration:
> 
> ```toml
> [tool.ruff]
> extend-exclude = [
>     "buildconfig/*.py",  # *.pyi stub files still included
>     "docs/reST",
>     "setup.py",
> ]
> ```
> 
> Alternatively, `include` can be set to specific directories.

Both options have a flaw I dislike, with the exclude it's not being quite as specific, for example, if there were other `.py` files in some other directories, they'd still get formatted.
Whereas the `include` requires being overly specific.

That said, considering that the reason we don't want `buildconfig/*.py` files formatted is because they're legacy files, if there were new files to be added in some other subdirectory, they'd likely not be a legacy thing and could be easily formatted, so I think I'll go with that approach for now.

I did consult `ruff` developers, however, it seems there isn't (currently) really a way to include a directory but exclude a specific subdirectory inside of it. (When I say include I mean including in the `exclude` and similar with excluding from `exclude`). There was some other quirky behavior I found, but that's besides the point :grin:

buildconfig/stubs/pygame/__init__.pyi
```diff
@@ -3,63 +3,48 @@
 # IMPORTANT NOTE: Do not edit this file by hand!
```
aatle, line 3: This file is autogenerated and cannot be import sorted.
Should add `# ruff: noqa: I001` to first line of this file by `gen_stubs.py`.

Matiiss: What about using gen_stubs.py to sort the imports instead?

buildconfig/stubs/pygame/typing.pyi
```diff
@@ -27,7 +27,6 @@ FileLike = Union[_PathLike, IO[bytes], IO[str]]
 
 _T_co = TypeVar("_T_co", covariant=True)
 
-
```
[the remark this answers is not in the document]
Matiiss: Why?

Matiiss:
> Can you also add .ruff_cache dir into the gitignore image

To be fair, it would be ignored by `git` anyway because it contains an auto-generated `.gitignore` inside of it, but I added it to the top level `.gitignore` anyway. (At least it's now visually consistent with other ignored directories :grin:)

buildconfig/stubs/pygame/typing.pyi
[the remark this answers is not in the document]
aatle: Black / ruff formatter mandates that module-level classes be surrounded by two blank lines on each side. Deleting those lines causes tests to fail and also makes `typing.pyi` inconsistent with `typing.py`.

buildconfig/stubs/pygame/__init__.pyi
aatle: That's a possibility. Having `gen_stubs.py` need to run ruff might not be worth it though.
I also checked how the imports will be sorted and I don't think it is good, the alphabetical sorting splits many of the key constants and organizes the imports in a less useful way.

pyproject.toml
```diff
@@ -92,11 +92,17 @@ setup-args = [
 
 [tool.ruff]
 exclude = [
```
aatle, line 94:
```suggestion
extend-exclude = [
```
(to not overwrite default excluded paths)

Matiiss: I really don't think it's an issue for us, `ruff` respects (and prioritizes) `.gitignore` rules, so if there's something we're not committing due to those rules, it won't be touched by `ruff` either and if it is something we can commit, we can then decide if we want to make `ruff` ignore it or not.

buildconfig/stubs/pygame/typing.pyi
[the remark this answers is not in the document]
Matiiss: Well, who do you think deleted those lines? 😄 It definitely wasn't me. It was `ruff`'s doing. Also these are stub files, so it's alright to make them more compact anyway IMO.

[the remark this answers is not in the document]
aatle: Yes, I see the problem now. Ruff formats `.pyi` files differently (more compact) than `.py` files, which causes a conflict when `gen_stubs.py` attempts to copy over `typing.py` to `typing.pyi`, and leads to failing tests.

Therefore, either `typing.pyi` should be formatted with ruff from `gen_stubs.py`, or `fmt: off` should be added to the top of the file (similar decision as to auto-generated `__init__.pyi`).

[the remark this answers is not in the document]
Matiiss: Alright, so that makes it two things that would benefit from ruff being run in `gen_stubs.py`. Or we could just reorder the action items in `python dev.py all` such that formatting happens after building the stubs.

buildconfig/stubs/pygame/__init__.pyi
Matiiss: I don't really think it matters how they're sorted there, given that there is another file that requires some additional sorting stuffs, I don't think including this into that (the reordering of formatting and stubs such that formatting happens after the stubs) would really matter.

aatle (approved):
LGTM; thanks!

(Before merging, pull in from main and run pre-commit so that tests pass.)

src_py/sprite.py
```diff
@@ -372,8 +371,7 @@ class AbstractGroup:
 
     """
 
-    def __class_getitem__(cls, generic):
-        return types.GenericAlias(cls, generic)
+    __class_getitem__ = classmethod(types.GenericAlias)
```
ankith26, line 374: How is this a formatting change?
I mean, it looks equivalent but still.
Not a PR blocker though

ankith26 (approved):
LGTM, thanks for the PR :+1:

src_py/sprite.py
Matiiss: Well, it looks different, doesn't it? Hence formatting /partial s
I honestly forgot I made that change here... but anyway, this is how it's done in CPython, so thought we should adhere to that style (still not sure why it's in this PR, but yeah)

Issue #1328, opened 2023-02-12
pygame.mixer.Sound.copy (2577)


GalacticEmperor1:
#  **Issue &numero;2577 opened by *[fortwoone](https://github.com/fortwoone)* at 2021-05-04 15:38:39**

I noticed there were copy protocols for some important Pygame objects, such as Surfaces and Rects. I wondered if it would have some interest to add one to Sound objects too, because it would help in some situations, for example if you don't want to use imported sounds directly (also to avoid direct instance editing as we may encounter with Surfaces if we don't copy them before editing)...<hr>

#  Comments

# # #  **[fortwoone](https://github.com/fortwoone)* commented at 2021-05-13 19:00:45*

And so I found a technique to copy Sound objects without using a copy method : you only need to make an array from the source Sound and then make another Sound object from the array
<hr>

# # #  **[illume](https://github.com/illume)* commented at 2021-05-22 20:12:49*

Would be useful to me too.
<hr>

# # #  **[fortwoone](https://github.com/fortwoone)* commented at 2021-05-23 10:36:01*

> 
> 
> Would be useful to me too.

Do you want me to give you my piece of code ?

bilhox:
hi,
I implemented it 2 weeks ago in my side, but didn't have time to fix some bugs. At least it is coming soon.

thomasgrgi:
@bilhox 
We'd be interested in taking a look at the code and working on it with my project group if you don't have time to work on the issues, and hopefully we can find a solution.

thomasgrgi:
@bilhox
Many thanks! Of course, if we manage to fix the bugs, you'll be added as a co-author on the commit.

bilhox:
Here is my branch : https://github.com/bilhox/pygame-ce/tree/sound-copy

ankith26:
This has been implemented in #3556, closing

PR #3238, opened 2024-11-24
Sound copy - co-authored @bilhox @thomasgrgi

Borishkof:
Create a function for deep copying sound.
Create some tests (failing tests on .xm and .opus for un unknown reason).

issue : https://github.com/pygame-community/pygame-ce/issues/1328#issuecomment-2431599386

### DIFF
--- a/src_c/mixer.c
+++ b/src_c/mixer.c
@@ -831,6 +831,31 @@ snd_get_samples_address(PyObject *self, PyObject *closure)
 #endif
 }
 
+static PyObject *
+snd_copy(pgSoundObject *self, PyObject *_null)
+{
+    Mix_Chunk *chunk = pgSound_AsChunk(self);
+    pgSoundObject *new_sound;
+    Mix_Chunk *new_chunk;
+
+    CHECK_CHUNK_VALID(chunk, -1);
+
+    new_sound =
+        (pgSoundObject *)pgSound_Type.tp_new(Py_TYPE(self), NULL, NULL);
+
+    if (chunk->allocated) {
+        new_chunk = Mix_QuickLoad_RAW(chunk->abuf, chunk->alen);
+        if (!new_chunk)
+            RAISE(pgExc_BufferError,
+                  "there is a problem I couldn't describe ;-;");
+
+        new_chunk->volume = chunk->volume;
+        new_sound->chunk = new_chunk;
+    }
+
+    return (PyObject *)new_sound;
+}
+
 PyMethodDef sound_methods[] = {
     {"play", (PyCFunction)pgSound_Play, METH_VARARGS | METH_KEYWORDS,
      DOC_MIXER_SOUND_PLAY},
@@ -842,6 +867,7 @@ PyMethodDef sound_methods[] = {
     {"get_volume", snd_get_volume, METH_NOARGS, DOC_MIXER_SOUND_GETVOLUME},
     {"get_length", snd_get_length, METH_NOARGS, DOC_MIXER_SOUND_GETLENGTH},
     {"get_raw", snd_get_raw, METH_NOARGS, DOC_MIXER_SOUND_GETRAW},
+    {"copy", snd_copy, METH_NOARGS, ""},
     {NULL, NULL, 0, NULL}};
 
 static PyGetSetDef sound_getset[] = {
--- a/src_c/mixer.c
+++ b/src_c/mixer.c
@@ -838,17 +838,59 @@ snd_copy(pgSoundObject *self, PyObject *_null)
     pgSoundObject *new_sound;
     Mix_Chunk *new_chunk;
 
-    CHECK_CHUNK_VALID(chunk, -1);
+    // Validate the input chunk
+    CHECK_CHUNK_VALID(chunk, NULL);
 
+    // Create a new sound object
     new_sound =
         (pgSoundObject *)pgSound_Type.tp_new(Py_TYPE(self), NULL, NULL);
+    if (!new_sound) {
+        PyErr_SetString(PyExc_RuntimeError,
+                        "Failed to allocate memory for new sound object");
+        return NULL;
+    }
 
+    // Handle chunk allocation type
     if (chunk->allocated) {
-        new_chunk = Mix_QuickLoad_RAW(chunk->abuf, chunk->alen);
-        if (!new_chunk)
-            RAISE(pgExc_BufferError, "there is a problem I can't describe :)");
+        // Create a deep copy of the audio buffer for allocated chunks
+        Uint8 *buffer_copy = (Uint8 *)malloc(chunk->alen);
+        if (!buffer_copy) {
+            Py_DECREF(new_sound);
+            PyErr_SetString(PyExc_MemoryError,
+                            "Failed to allocate memory for sound buffer");
+            return NULL;
+        }
+        memcpy(buffer_copy, chunk->abuf, chunk->alen);
 
+        // Create a new Mix_Chunk
+        new_chunk = Mix_QuickLoad_RAW(buffer_copy, chunk->alen);
+        if (!new_chunk) {
+            free(buffer_copy);
+            Py_DECREF(new_sound);
+            PyErr_SetString(PyExc_RuntimeError,
+                            "Failed to create new sound chunk");
+            return NULL;
+        }
         new_chunk->volume = chunk->volume;
+        new_sound->chunk = new_chunk;
+    }
+    else {
+        // For non-allocated chunks (e.g., formats like .xm), create a full
+        // copy
+        new_chunk = (Mix_Chunk *)malloc(sizeof(Mix_Chunk));
+        if (!new_chunk) {
+            Py_DECREF(new_sound);
+            PyErr_SetString(PyExc_MemoryError,
+                            "Failed to allocate memory for sound chunk");
+            return NULL;
+        }
+        *new_chunk = *chunk;  // Copy the entire structure
+
+        // For safety, ensure the copied chunk doesn't share pointers
+        new_chunk->abuf =
+            NULL;  // Prevent double-free if original gets deallocated
+        new_chunk->allocated = 0;
+
         new_sound->chunk = new_chunk;
     }
 
--- a/src_c/mixer.c
+++ b/src_c/mixer.c
@@ -846,8 +846,7 @@ snd_copy(pgSoundObject *self, PyObject *_null)
     if (chunk->allocated) {
         new_chunk = Mix_QuickLoad_RAW(chunk->abuf, chunk->alen);
         if (!new_chunk)
-            RAISE(pgExc_BufferError,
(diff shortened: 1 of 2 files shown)


src_c/mixer.c
```diff
@@ -831,6 +831,72 @@ snd_get_samples_address(PyObject *self, PyObject *closure)
 #endif
 }
 
+static PyObject *
+snd_copy(pgSoundObject *self, PyObject *_null)
```
ankith26, line 835:
```suggestion
snd_copy(PyObject *self, PyObject *_null)
```
This fixes compiler warnings

You may also need to do other changes wherever `self` is being used, to manually cast it to the required types

ankith26 (commented):
Thanks for contributing to pygame-ce!

This change would need documentation and type stubs for complete approval

docs/reST/ref/mixer.rst
```diff
@@ -479,6 +479,17 @@ The following file formats are supported
 
       .. ## Sound.get_raw ##
 
+   .. method:: copy
+
+      | :sl:`return a new Sound object that is a deep copy of this one`
+      | :sg:`copy() -> Sound`
+
+      Return a new Sound object that is a deep copy of this one. The new Sound will
+      be playable just like the original. If the copy fails, the method returns ``NULL``.
```
damusss, line 488: The end of the description isn't correct. The C-API function does indeed return NULL, but the python function would raise an exception (various in this case). (NULL doesn't actually exists in python). You should probably list the possible exceptions, but that's not an hard rule.
```suggestion
      be playable just like the original. If the copy fails a ``MemoryError`` or a :meth:`pygame.error` exception will be raised.
```

src_c/mixer.c
```diff
@@ -831,6 +831,72 @@ snd_get_samples_address(PyObject *self, PyObject *closure)
...
+    Mix_Chunk *new_chunk;
+
+    // Validate the input chunk
+    CHECK_CHUNK_VALID(chunk, NULL);
+
+    // Create a new sound object
+    new_sound =
+        (pgSoundObject *)pgSound_Type.tp_new(Py_TYPE(self), NULL, NULL);
+    if (!new_sound) {
+        PyErr_SetString(PyExc_RuntimeError,
```
damusss, line 848: `PyExc_MemoryError` feels better for this case.
```suggestion
        PyErr_SetString(PyExc_MemoryError,
```

src_c/mixer.c
```diff
@@ -831,6 +831,72 @@ snd_get_samples_address(PyObject *self, PyObject *closure)
...
+            return NULL;
+        }
+        memcpy(buffer_copy, chunk->abuf, chunk->alen);
+
+        // Create a new Mix_Chunk
+        new_chunk = Mix_QuickLoad_RAW(buffer_copy, chunk->alen);
+        if (!new_chunk) {
+            free(buffer_copy);
+            Py_DECREF(new_sound);
+            PyErr_SetString(PyExc_RuntimeError,
```
damusss, line 870: Since the error is caused by SDL, pygame.error is preferred over runtime error. (called pgExc_SDLError in the capi)
```suggestion
            PyErr_SetString(pgExc_SDLError,
```

src_c/mixer.c
```diff
@@ -842,6 +908,7 @@ PyMethodDef sound_methods[] = {
     {"get_volume", snd_get_volume, METH_NOARGS, DOC_MIXER_SOUND_GETVOLUME},
     {"get_length", snd_get_length, METH_NOARGS, DOC_MIXER_SOUND_GETLENGTH},
     {"get_raw", snd_get_raw, METH_NOARGS, DOC_MIXER_SOUND_GETRAW},
+    {"copy", snd_copy, METH_NOARGS, DOC_MIXER_SOUND_COPY},
```
damusss, line 911: To be coherent with the rest of the API, the same `snd_copy` c function should be used for the `__copy__` method. Both methods should also be added to the stubs (both extended, `__copy__` doesn't need to be in the docs).
```suggestion
    {"copy", snd_copy, METH_NOARGS, DOC_MIXER_SOUND_COPY}, {"__copy__", snd_copy, METH_NOARGS, DOC_MIXER_SOUND_COPY},
```

damusss (changes requested):
Things I noticed beside what ankith said. Also some error messages could be improved grammar wise.

Borishkof:
Thanks for your help, I will update my PR.

test/mixer_test.py
```diff
@@ -711,6 +711,33 @@ def test_get_sdl_mixer_version__linked_equals_compiled(self):
 
         self.assertTupleEqual(linked_version, compiled_version)
 
+    def test_snd_copy(self):
+        mixer.init()
+
+        filenames = ["house_lo.ogg", "house_lo.wav", "house_lo.flac"]
```
bilhox, line 717: If possible, you should include other file formats so we can be sure it doesn't fail.

bilhox (changes requested):
Overall LGTM, I left some requested changes.

test/mixer_test.py
Borishkof: I added support for .xm, .opus, and .mid formats. They work too (it's just that some of them are not playable with channel = sound.play(), but the other tests pass successfully). Additionally, for .mid files, I get fluidsynth warnings, but the tests still pass successfully.

Should I add all three formats to the tests, or just .xm and .opus to avoid warnings for every tests on mixer_test.py (or other formats)?

bilhox: just ``xm`` and ``opus`` will be fine.

yunline:
> It seems like xm or opus is "Unrecognized audio format" (using mixer.Sound() ) for the distribution that does the automatic test I failed.
> But it works on my distrib and some examples in mixer_music_test use those formats (without specific version checking).
> What should I do ?

You are writing unittest for Sound.copy, not for other functionalities. 
The test case should be decoupled with other functionalities like "reading files" since those should be taken care of by other test cases.

src_c/mixer.c
```diff
@@ -831,6 +831,72 @@ snd_get_samples_address(PyObject *self, PyObject *closure)
...
+    if (!new_sound) {
+        PyErr_SetString(PyExc_MemoryError,
+                        "Failed to allocate memory for new sound object");
+        return NULL;
+    }
+
+    // Handle chunk allocation type
+    if (chunk->allocated) {
+        // Create a deep copy of the audio buffer for allocated chunks
+        Uint8 *buffer_copy = (Uint8 *)malloc(chunk->alen);
```
Starbuck5, line 856: I believe this is a memory leak. This is allocated but never freed, and it can't be freed here because it needs to live as long as the chunk does.

src_c/mixer.c
```diff
@@ -831,6 +831,72 @@ snd_get_samples_address(PyObject *self, PyObject *closure)
...
+        // For non-allocated chunks (e.g., formats like .xm), create a full
+        // copy
+        new_chunk = (Mix_Chunk *)malloc(sizeof(Mix_Chunk));
+        if (!new_chunk) {
+            Py_DECREF(new_sound);
+            PyErr_SetString(PyExc_MemoryError,
+                            "Failed to allocate memory for sound chunk");
+            return NULL;
+        }
+        *new_chunk = *chunk;  // Copy the entire structure
```
Starbuck5, line 887: Copying an entire struct like this may work in this scenario but in my opinion is not something that is safe in general. From SDL's perspective these are all internal fields and they can do whatever they want with it. They're barely documented.

Starbuck5:
Thank you for working on this @Borishkof! I can tell this is a high effort contribution.

However, I feel that there is no safe way to do this within the SDL_mixer API, and I don't really see why this is a useful feature. I feel bad I didn't see the earlier discussion and say so then.

Copying a surface? Important, because they are editable. Edit one copy, leave the other copy the same.
Sounds on the other hand? The only thing you can edit currently is the volume. Not exactly as fundamental.

Talking myself through it I guess you could attach sounds to objects in a game. So like different sprites have different volume explosion noises. So you would want to give them each a copy of the sound with a different preset volume. But that would be very wasteful on the memory front. The ideal solution I think would be a bunch of Mix_Chunks all pointed at the same audio buffer, with different volumes set. So we already have Quickload_RAW from SDL_mixer, and we already can get the raw bytes using the routine `get_raw` uses right?

So maybe this can be done without this allocated branching? Whether or not the mix_chunk controls the allocation, the allocation can be copied out. We'd need some way of managing the life cycle of this new buffer that only some Sound objects have. If we wanted to share it between Sound copies for efficiency's sake it could be another Python object so we can ref-count it for garbage collection purposes.

OK new idea on this train of thought, can't this be implemented safely and simply in Python? I haven't tested this at all but couldn't you take the output of `Sound.get_raw()` and pipe it back into the `Sound` constructor with the `array=` keyword. If it's written in Python you could do my crazy memory proxy Python object idea with very low complexity.

Borishkof:
Thank you for the review :)

As I am a beginner in Pygame, I can't argue much about the solutions you propose and will leave it to the community to decide which is the most efficient way to tackle the issue.
That being said, I believe you are right about the poor memory management (I forgot to check it).

If the community reaches a consensus on a reasonably simple solution, I would be glad to try and implement it. Otherwise, I am happy to let anyone with good ideas take a shot at addressing this issue. 👍

bilhox:
I find the arguments gave by starbucks understandable, so I'm for implementation in python side now.

Starbuck5:
On second thought maybe sharing the memory is a bit too far. Sound data is mutable through https://pyga.me/docs/ref/sndarray.html#pygame.sndarray.samples, so that would need to protected against. Also I'm not sure the public Sound API supports zero copying the data.

But anyways here is a 1 line Python implementation of sound.copy, in a demonstration program.

```py
import pygame
pygame.init()

window = pygame.Window(resizable=True)
screen = window.get_surface()

class MySound(pygame.Sound):
    def copy(self):
        return self.__class__(buffer=self.get_raw())

a = MySound(r"C:\Users\charl\Desktop\pygame-ce\examples\data\secosmic_lo.wav")
b = a.copy()
b.set_volume(0.1)

clock = pygame.Clock()

while True:
        if event.type == pygame.QUIT:
            pygame.quit()
            raise SystemExit
        if event.type == pygame.KEYDOWN and event.key == pygame.K_1:
            a.play()

        if event.type == pygame.KEYDOWN and event.key == pygame.K_2:
            b.play()

    window.flip()
    clock.tick(144)
```

Starbuck5:
I've set this to a draft because it is not ready to merge in its current state. I was going to close it fully and send it back to the issue for a new PR, but the docs/tests/stubs bit of this PR is probably good. I just think the actual implementation needs to be overhauled.

I showed a Python example of how this could be achieved in one line. This same logic could be applied in C (it would take more than one line), or it could be monkey patched into the type in `__init__.py`.

MyreMylar:
PR taken over here: https://github.com/pygame-community/pygame-ce/pull/3556

PR #3242, opened 2024-11-25
Remove unnecessary calls to PyNumber_Index, PyLong_Checks

Starbuck5:
This simplifies code and improves performance. It makes `rect[x]`  ~7% faster and `rect[x] = y` ~9% faster. Although these functions are really cheap, and therefore I would be surprised if there was any measurable macro impact from this change. You can run them 10 million times in less than a second. I didn't test performance impact in pixelarray.

And in case anyone is curious `rect.y` is ~8% faster than `rect[1]` (after this PR). So for those micro-optimizers out there, that's the strategy to use.

**PyNumber_AsSsize_t doesn't need the value to be converted to an "index int" first.**

If you look at the CPython source, in Objects/abstract.c:
```c
Py_ssize_t
PyNumber_AsSsize_t(PyObject *item, PyObject *err)
{
    Py_ssize_t result;
    PyObject *runerr;
    PyObject *value = _PyNumber_Index(item);
    if (value == NULL)
        return -1;
    ...
```

The first thing it does is use PyNumber_Index anyway!

Plus it is documented as "Returns o converted to a [Py_ssize_t](https://docs.python.org/3/c-api/intro.html#c.Py_ssize_t) value if o can be interpreted as an integer". Having an `__index__` method counts as being interpretable as an integer.

**`PyIndex_Check(op) || PyLong_Check(op)`, the PyLong_Check is unnecessary.**

Python ints already pass PyIndex_Check, as you can tell because Python ints work in the Rect functions.

### DIFF
--- a/src_c/rect_impl.h
+++ b/src_c/rect_impl.h
@@ -2161,6 +2161,9 @@ RectExport_subscript(RectObject *self, PyObject *op)
 
     if (PyIndex_Check(op)) {
         Py_ssize_t i = PyNumber_AsSsize_t(op, NULL);
+        if (i == -1 && PyErr_Occurred()) {
+            return NULL;
+        }
         return RectExport_item(self, i);
     }
     else if (op == Py_Ellipsis) {
@@ -2207,6 +2210,9 @@ RectExport_assSubscript(RectObject *self, PyObject *op, PyObject *value)
     }
     if (PyIndex_Check(op)) {
         Py_ssize_t i = PyNumber_AsSsize_t(op, NULL);
+        if (i == -1 && PyErr_Occurred()) {
+            return -1;
+        }
         return RectExport_assItem(self, i, value);
     }
     else if (op == Py_Ellipsis) {


src_c/rect_impl.h
```diff
@@ -2213,15 +2206,7 @@ RectExport_assSubscript(RectObject *self, PyObject *op, PyObject *value)
...
-        PyObject *index;
-        Py_ssize_t i;
-
-        index = PyNumber_Index(op);
-        if (index == NULL) {
-            return -1;
-        }
-        i = PyNumber_AsSsize_t(index, NULL);
-        Py_DECREF(index);
+        Py_ssize_t i = PyNumber_AsSsize_t(op, NULL);
```
yunline, line 2209: Should add an error check here.
According to [python docs](https://docs.python.org/3.12/c-api/number.html#c.PyNumber_AsSsize_t), `PyNumber_AsSsize_t` could raise an error when returning `-1`

Starbuck5: I thought that, and then I get fooled by the passing NULL into the exception slot and thought it meant no errors. But you're right, thank you for pointing out.

yunline (approved):
LGTM

Testing code:
```py
import pygame
pygame.init()

class IndexLike:
    def __index__(self):
        raise RuntimeError("Nope")

r = pygame.Rect(0,0,1,1)

r[1<<256] # IndexError: Invalid rect Index
r[IndexLike()] # RuntimeError: Nope
```

ankith26 (approved):
LGTM, thanks! :balloon:

PR #3246, opened 2024-11-29
updated 'in' operator for sys.platform linux check 

sonotopes:
As Pygame currently uses CPython 3.8>=, the 'in' operator in the sys.platform check of get_backends() could be considered unnecessary, as no major version of Linux will be returned when accessing sys.platform. According to [Python documentation](https://docs.python.org/3/library/sys.html#sys.platform ), as of 3.3, sys.platform will only return the string, 'linux' (lowercased). Therefore, the == operator could potentially be more appropriate. Just a thought.


bilhox (approved):
Fair changes, thanks for your contribution !

ankith26 (approved):
LGTM, thanks for contributing! :tada:

PR #3179, opened 2024-10-17
Added ```Line``` methods and ```length``` attribute

MrRedstone058:
Added

### Attributes
- Line.length

### Methods
- Line.move()
- Line.move_ip()
- Line.update()
- Line.scale()
- Line.scale_ip()
- Line.flip_ab()
- Line.flip_ab_ip()

## Credits

**Geometry Project:**
For code, docs and tests:
@novialriptide @Emc2356 @itzpr3d4t0r @ScriptLineStudios @avaxar @Matiiss @newpaxonian @maqa41 @blankRiot96 Also thanks to @Starbuck5 for kickstarting the idea and the occasional help!

**Functionality added in this PR**
- `line.c` https://github.com/pygame-community/pygame-geometry/blame/main/src_c/line.c Credits to @Emc2356 @itzpr3d4t0r @novialriptide @ScriptLineStudios @avaxar
- `geometry.pyi` https://github.com/pygame-community/pygame-geometry/blame/main/geometry.pyi Credits to @Emc2356 @itzpr3d4t0r @novialriptide @ScriptLineStudios @avaxar
- `geometry_test.py` https://github.com/pygame-community/pygame-geometry/blame/main/test/test_line.py Credits to @Emc2356 @itzpr3d4t0r @novialriptide @ScriptLineStudios @avaxar
- `geometry.rst` https://github.com/pygame-community/pygame-geometry/blame/main/docs/line.rst Credits to @itzpr3d4t0r @avaxar @ScriptLineStudios
- `geometry.h` https://github.com/pygame-community/pygame-geometry/blame/main/src_c/include/geometry.h Credits to @Emc2356 @itzpr3d4t0r @maqa41

PR #3248, opened 2024-12-01
Fix failing tests: change `Tuple[float, float]` to `Point` in `geometry.pyi`

aatle:
Missed in merged #3179 somehow.

Mypy tests are failing because `Tuple` is no longer available in `geometry.pyi`. It should be `Point` anyway.


aatle:
Failing test: https://github.com/pygame-community/pygame-ce/actions/runs/12108442750/job/33756440051
False negative passed test: https://github.com/pygame-community/pygame-ce/actions/runs/11458498255/job/31880993053

oddbookworm (approved):
LGTM! Thanks for fixing this. It's just a classic example of the PR being old enough that a change from another pull request broke something in this PR without causing a merge conflict, so CI never reran without a new pull from main or a new commit. We should probably have some kind of policy where we require a fresh pull from main on new merges? @pygame-community/pygame-contributing-admins thoughts?

damusss (approved):
LGTM 🔝

PR #3249, opened 2024-12-01
New approach to lazy loading of pygame submodules (`surfarray`, `sndarray`)

aatle:
A new approach to lazy loading pygame submodules after #3232 had a problem with slower attribute accesses (`__getattr__`).

**Implementation explanation:**
The implementation is twofold, where the second part has many alternatives.

First, pygame checks if `numpy` exists.
Then, each submodule is loaded if numpy exists and the submodule's other pygame module dependencies (e.g. `mixer`) load successfully. If numpy is missing or a module dependency fails, then the submodule becomes `MissingModule`. It is necessary to get an error to use its message in `MissingModule.reason`, so `try-except` is used.

If successful, the submodule is created and put into `sys.modules`, using the documented [`LazyLoader`](https://docs.python.org/3/library/importlib.html#implementing-lazy-imports) implementation, relying on import machinery.
This makes it so that the submodule exists, but its contents are not executed until the first attribute or method access on the submodule.
It does this by changing the module class temporarily to `_LazyModule`; after the module is actually loaded, the module class is set back to the normal module type.

**Potential differences from current behavior:**
The initial checks are not completely perfect. It is assumed that if `numpy` is importable, it will not raise an `ImportError` nor `OSError`.
Obviously, `numpy` not being loaded by pygame automatically is an intended difference. Instead, it is loaded upon the first attribute access of either submodule. I do not expect this to cause any lag spikes in-game: the user probably imports numpy themselves, and it will also load if directly importing one of the submodule functions, or using one during game loading.
After a submodule is loaded successfully, there should not be any noticeable difference from current behavior. The module class is the same, the original loader is put back, the module code was not changed.
https://docs.python.org/3/library/importlib.html#importlib.util.LazyLoader
>**Note:** For projects where startup time is critical, this class allows for potentially minimizing the cost of loading a module if it is never used. For projects where startup time is not essential then use of this class is **heavily** discouraged due to error messages created during loading being postponed and thus occurring out of context.

There are no expected errors during postponed loading of these submodules, as they should have all been handled at the start.

Lazy loading of numpy has great potential benefits, so it's worth the extra complexity.

Closes #3232

### DIFF
--- a/docs/reST/ref/sndarray.rst
+++ b/docs/reST/ref/sndarray.rst
@@ -23,6 +23,8 @@ Each sample is an 8-bit or 16-bit integer, depending on the data format. A
 stereo sound file has two values per sample, while a mono sound file only has
 one.
 
+.. versionchanged:: 2.5.3 sndarray module is lazily loaded to avoid loading NumPy needlessly
+
 .. function:: array
 
    | :sl:`copy Sound samples into an array`
--- a/docs/reST/ref/surfarray.rst
+++ b/docs/reST/ref/surfarray.rst
@@ -35,6 +35,8 @@ pixels from the surface and any changes performed to the array will make changes
 in the surface. As this last functions share memory with the surface, this one
 will be locked during the lifetime of the array.
 
+.. versionchanged:: 2.5.3 surfarray module is lazily loaded to avoid loading NumPy needlessly
+
 .. function:: array2d
 
    | :sl:`Copy pixels into a 2d array`
--- a/src_py/__init__.py
+++ b/src_py/__init__.py
@@ -322,22 +322,30 @@ def lazy_import(name):
     return module
 
 
-numpy_exists = find_spec("numpy") is not None
+# Check if numpy is available for surfarray and sndarray modules
+numpy_missing = find_spec("numpy") is None
 
-# Preserve MissingModule behavior when numpy is not installed
-# and when their pygame module dependencies are unavailable
-
-if numpy_exists and not isinstance(pixelcopy, MissingModule):
-    surfarray = lazy_import("surfarray")
-else:
+try:
+    if numpy_missing:
+        # Always fails here. Need the error message for MissingModule.reason
+        import numpy
+    # Check that module dependencies are not missing, or get error message
+    import pygame.pixelcopy
+except (ImportError, OSError):
     surfarray = MissingModule("surfarray", urgent=0)
-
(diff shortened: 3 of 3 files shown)


docs/reST/ref/sndarray.rst
```diff
@@ -23,6 +23,8 @@ Each sample is an 8-bit or 16-bit integer, depending on the data format. A
 stereo sound file has two values per sample, while a mono sound file only has
 one.
 
+.. versionchanged:: 2.5.3 sndarray module is lazily loaded to avoid loading NumPy needlessly
```
Starbuck5, line 26: This still doesn't explain why the action is actually being taken.

Why is it bad to load NumPy?

Starbuck5:
@damusss You cast some doubt about this PR working with PyInstaller but did not elaborate.

If there's a problem here can you explain what you did please.

damusss:
> If there's a problem here can you explain what you did please.

🫡 I tested both normal pygame-ce and this branch, and it looks like both imports are successful. when I got that error I must have been using an incomplete version of the pull request
image

MyreMylar:
Looks like this change has uncovered a pylint bug.

ankith26 (commented):
I would love to see some benchmarking/numbers on how much this PR is helping. Just so we can weigh in the added complexity/maintenance-burden against it. The benchmark should cover both:
- A simple `import pygame`
- getattr of affected submodules

aatle:
Both are with numpy installed. (Without numpy or after loading numpy, times are the same.)

Simple import `python -X importtime -c "import pygame"`
import

First attribute access after importing pygame `pygame.surfarray.array2d`
access

(note: if accessing just `pygame.surfarray`, it is the same, because for lazy it doesn't get loaded, and for main it is already loaded.)

ankith26:
These benchmarks makes sense, thanks for working on it. You mention "first attribute access" I wonder how the performance of subsequent attribute accesses are affected (if it is)

aatle:
After the first attribute access, I fairly certain both have the same access times. (There is no module `__getattr__`, both modules are of type `types.ModuleType`.) So the modules should be indistinguishable afterwards.

ankith26 (approved):

aatle:
@coderabbitai full review please

MyreMylar (approved):
OK, LGTM. I can verify the speedup in importing on my machine locally.

PR #3251, opened 2024-12-02
Port window.c to SDL3

ankith26:
Continuing the SDL3 porting work, `window.c` now compiles in SDL3.


src_c/window.c
```diff
@@ -288,6 +376,10 @@ window_focus(pgWindowObject *self, PyObject *args, PyObject *kwargs)
     if (!PyArg_ParseTupleAndKeywords(args, kwargs, "|p", kwids, &input_only)) {
         return NULL;
     }
+#if SDL_VERSION_ATLEAST(3, 0, 0)
+    /* input_only ignored on SDL3 */
+    SDL_RaiseWindow(self->_win);
```
yunline, line 381: In SDL2, `SDL_RaiseWindow` returns void.
In SDL3, `SDL_RaiseWindow` returns a bool to indicate if an error occured.
So an error check is needed here.

yunline (commented):
SDL3 changed the return values of some functions. In SDL2 they return void. In SDL3 they return bool. We need to add error check for them in SDL3.
Here is a list of this kind of function (may be incomplete):
- `SDL_RaiseWindow`
- `SDL_HideWindow`
- `SDL_ShowWindow`
- `SDL_RestoreWindow`
- `SDL_MaximizeWindow`
- `SDL_MinimizeWindow`
- `SDL_SetWindowIcon`
- `SDL_SetWindowMouseGrab`
- `SDL_SetWindowKeyboardGrab`
- `SDL_SetWindowTitle`
- `SDL_SetWindowResizable`
- `SDL_SetWindowBordered`
- `SDL_SetWindowAlwaysOnTop`
- `SDL_SetWindowSize`
- `SDL_GetWindowSize`
- `SDL_GetWindowMinimumSize`
- `SDL_SetWindowMinimumSize`
- `SDL_GetWindowMaximumSize`
- `SDL_SetWindowMaximumSize`
- `SDL_SetWindowPosition`
- `SDL_GetWindowPosition`

ankith26:
Definitely a good catch here, thanks!

But it has gotten me thinking - raising errors is technically a behaviour change and while I think it is definitely a step in the right direction, I wonder if it's gonna be very disruptive.

damusss:
> But it has gotten me thinking - raising errors is technically a behaviour change and while I think it is definitely a step in the right direction, I wonder if it's gonna be very disruptive.

I'm not sure if it's a good idea to throw out so many errors. I wouldn't exactly want my program to crash just because non-critical functionality X didn't work. (for example, not sure how could changing the title fail, but even if it failed, I wouldn't want my program crashing). Otherwise you'd need to try-except basically every window operation and with that you basically lose the flexibility of OOP. If we do want to do something I think warnings are the best strategy.

yunline:
> But it has gotten me thinking - raising errors is technically a behaviour change and while I think it is definitely a step in the right direction, I wonder if it's gonna be very disruptive.

The errors added in SDL3 are more like "sanity check". If users don't write code in a really weirld way in SDL2 (like setting an invalid surface as the window icon), they won't get an error in SDL3. So I think it's OK to raise those errors.

> I'm not sure if it's a good idea to throw out so many errors. I wouldn't exactly want my program to crash just because non-critical functionality X didn't work. (for example, not sure how could changing the title fail, but even if it failed, I wouldn't want my program crashing). Otherwise you'd need to try-except basically every window operation and with that you basically lose the flexibility of OOP. If we do want to do something I think warnings are the best strategy.

If SDL raises an error, it's usually because the user is giving a wrong input to the function. To eliminate the error, what user needs to do is not simply adding a "try" statement to ignore the error. They need to fix where the code is wrong, making the code more robust. So I think there is no need to replace errors with warnings to decrease the use of "try". Users can fix those errors without "try" easily.

ankith26:
Consider `SDL_SetWindowPosition` for instance. For now on SDL2 it will silently fail on platforms like wayland, for reasons completely out of the users control. Whereas SDL3 will actively raise an error in this case.

yunline:
One way to solve this is to add a option for users to turn on or turn off these errors.
Like:
```py
pygame.compat_ignore_sdl3_errors(True)
```

damusss:
> If SDL raises an error, it's usually because the user is giving a wrong input to the function. To eliminate the error, what user needs to do is not simply adding a "try" statement to ignore the error. They need to fix where the code is wrong, making the code more robust. So I think there is no need to replace errors with warnings to decrease the use of "try". Users can fix those errors without "try" easily.
> 

I get what you mean, but I'm not exactly talking about that. Things that fail for user input are already raising errors. I'm talking about nonsensical errors like ankith mentioned (wayland doesn't allow to change window position). or like I said if your title string has no errors the code should not error just because the os failed to update the title for some reason. a warning is enough. I don't want to try excpet `window.title = "test"`

ankith26:
Thoughts on converting such SDL errors to python warnings instead?

ankith26:
Anyways, I think that this discussion can be had in a future PR. The goal of this PR is to simply support SDL3 with the current API while keeping SDL2 stuff as it is.

damusss (approved):
You are right, error or warning is partially an api discussion which this PR isn't aiming to change. As long as it compiles with SDL3 it's good.
LGTM, Thanks for the PR :)

src_c/window.c
```diff
@@ -262,9 +297,57 @@ window_set_windowed(pgWindowObject *self, PyObject *_null)
...
+        if (!modes) {
+            goto end;
+        }
+        chosen_mode = modes[0];
+        if (!chosen_mode) {
+            SDL_SetError("Could not get fullscreen display mode");
+            goto end;
+        }
+    }
+    if (!SDL_SetWindowFullscreenMode(window, chosen_mode)) {
```
Starbuck5, line 324: I believe you call this or SDL_SetWindowFullscreen, not both.

src_c/window.c
```diff
@@ -262,9 +297,57 @@ window_set_windowed(pgWindowObject *self, PyObject *_null)
...
+end:
+    SDL_free(modes);
+    return ret;
+}
+#endif
+
 static PyObject *
 window_set_fullscreen(pgWindowObject *self, PyObject *args, PyObject *kwargs)
 {
+#if SDL_VERSION_ATLEAST(3, 0, 0)
```
Starbuck5, line 341: Both SDL2 and SDL3 have their own python argument parsing here, can that be reduced.

src_c/window.c
ankith26: I assumed `SDL_SetWindowFullscreenMode` just set some internal state without actually updating the winow, but just checked the source, and both functions both call the same internal window update function, so you must be right. I have fixed this in the latest push

src_c/window.c
ankith26: True, fixed it.

Starbuck5 (approved):
Alright, looks good.

PR #3263, opened 2024-12-11
this statement seems to be false, removing

JovialKnoll:
From a quick test `fblits` seems to work fine (not drawing anything) with an empty sequence. Unless I'm missing something I think this sentence can be removed.


JovialKnoll:
from looking at history: I believe that this sentence was correct when `fblits` was first added, just doesn't seem to be anymore.

ankith26 (approved):
LGTM. I went over the `fblits` code and found nothing that should cause problems if a 0 length sequence is passed. Thanks for the PR :tada:

ankith26:
Requesting a review from the author of the fblits PR as their approval would be best to merge this PR

itzpr3d4t0r (approved):
Funny to see that turning out as a false statement, and even funnier me non remembering why i put that there!
Anyway it looks good to me, tested it with some random sequences and all worked well!

PR #3264, opened 2024-12-15
Correct `pygame.display` stubs

aatle:
`pygame.display.get_surface()` can return `None` if no display is initialized yet.
This correction may cause type checking errors in existing codebases.

`pygame.display.update()` can take `float`s for pixel coordinates, truncating them as usual.


buildconfig/stubs/pygame/display.pyi
```diff
@@ -44,14 +44,14 @@ def set_mode(
     display: int = 0,
     vsync: int = 0,
 ) -> Surface: ...
-def get_surface() -> Surface: ...
+def get_surface() -> Optional[Surface]: ...
```
ankith26, line 47: The `:sg:` line in the docs of this signature could also be corrected in the same way for consistency.

ankith26 (approved):
Nice catch, thanks! :tada: 

Left a minor review comment, but that won't stop my approval

damusss:
I'm slightly unsure about this. When I work on apps or games I don't enable type checkers, so this won't really affect me, but when I work on mili I do for people's sake and every time there is X|None I have to guard it at every usage. The reason I'm not sure it's because whether the function returns None doesn't depend on input or calculations but rather on the state of the program that a type checker can never know, meaning it'll always complain, even if you are sure the code path is going to run after display initialization.

I guess you can just blame this on the limitations of type checkers, and that the stubs mustn't lie to a user just because of it, and I can't complain on that, so I'll not stop this. I don't use this function either (window api on TOP), I was just rambling in the perspective on someone else (but at the same time, if they are good enough to use type checkers, they are good enough to use the window api lol)

buildconfig/stubs/pygame/display.pyi
damusss: @ankith26 can you not do `Surface | None` in python 3.9?

ankith26: Nope, that is a 3.10 thing

damusss: crazy

aatle:
Added an overload for no-argument `display.update()` to better reflect its behavior (differs from `display.update(None)`).
Updated docs, including listing more overloads for `display.update()`.

aatle:
I agree that it may inconvenience users who rely on this function to grab the display. They will probably have to use `cast`.
However, `pygame.display.get_surface()` is one of the main ways to check if the display is initialized, and it is important to correctly document its behavior. (Incorrect type annotations can lead to crashes when the type checker thought it was safe.)

Starbuck5 (approved):
Thanks!

PR #3268, opened 2024-12-16
Add more `Line` attributes

itzpr3d4t0r:
Adds the following `Line` attributes:
- center
- centerx/y
- angle
- slope

## Credits

**Geometry Project:**
For code, docs and tests:
Also thanks to @Starbuck5 for kickstarting the idea and the occasional help!

**Functionality added in this PR**
- `geometry.pyi`
https://github.com/pygame-community/pygame-geometry/blame/main/geometry.pyi
Credits to @Emc2356 @itzpr3d4t0r @novialriptide @ScriptLineStudios @avaxar

- `geometry.rst`
https://github.com/pygame-community/pygame-geometry/blame/main/docs/line.rst
Credits to @Emc2356 @itzpr3d4t0r @ScriptLineStudios @avaxar

- `line.c`
https://github.com/pygame-community/pygame-geometry/blame/main/src_c/line.c
Credits to @Emc2356 @itzpr3d4t0r @novialriptide @ScriptLineStudios @avaxar

- `geometry.h`
https://github.com/pygame-community/pygame-geometry/blame/main/src_c/include/geometry.h
Credits to @Emc2356 @itzpr3d4t0r @maqa41

- `geometry_test.py`
https://github.com/pygame-community/pygame-geometry/blame/main/test/test_line.py
Credits to @Emc2356 @itzpr3d4t0r @novialriptide @ScriptLineStudios @avaxar


docs/reST/ref/geometry.rst
```diff
@@ -610,6 +610,76 @@
 
          .. ## Line.length ##
 
+   .. attribute:: center
+
+         | :sl:`the coordinate of the middle point of the line`
+         | :sg:`center -> (float, float)`
+
+         The center of the line. Calculated using the `((xa + xb) / 2, (ya + yb) / 2)` formula.
```
damusss, line 618: This is for consistency, new docs should use `ax` and `ay`

docs/reST/ref/geometry.rst
```diff
@@ -610,6 +610,76 @@
...
+         .. ## Line.centery ##
+
+   .. attribute:: angle
+
+         | :sl:`the angle of the line`
+         | :sg:`angle -> float`
+
+         The angle of the line representing its orientation. Calculated using
+         the `atan2(yb - ya, xb - xa)` formula. This attribute is read-only, it cannot
+         be reassigned. To change the line's angle use the `rotate` method or change
```
damusss, line 663: Unless I'm blind, the rotate method doesn't exist yet, does it have to be ported or was it descarted?

damusss (changes requested):

src_c/doc/geometry_doc.h
```diff
@@ -37,6 +37,11 @@
 #define DOC_LINE_A "a -> (float, float)\nthe first point of the line"
 #define DOC_LINE_B "b -> (float, float)\nthe second point of the line"
 #define DOC_LINE_LENGTH "length -> float\nthe length of the line"
+#define DOC_LINE_CENTER "center -> (float, float)\nthe coordinate of the middle point of the line"
```
MyreMylar, line 40:
```suggestion
#define DOC_LINE_CENTER "center -> (float, float)\nthe coordinates of the middle point of the line"
```

docs/reST/ref/geometry.rst
```diff
@@ -610,6 +610,67 @@
 
          .. ## Line.length ##
 
+   .. attribute:: center
+
+         | :sl:`the coordinate of the middle point of the line`
```
MyreMylar, line 615:
```suggestion
         | :sl:`the coordinates of the middle point of the line`
```

MyreMylar (commented):
The angle attribute seems like it is a bit off from what I would expect in pygame. Not sure if this is because of the reversed y axis, or the difference in chosen calculations

I was expecting results something like this I guess:

image

If I apply rotations to a Vector2 of `(0,-10)` then it lines up with the image above. 

However a line of `((0,0),(0,-10))` currently has an angle of 90° and a line of `((0,0),(10,0))` has a reported angle of -0.0°.

How does everyone else feel about this difference?

MyreMylar:
> How does everyone else feel about this difference?

I guess I should also ask - is there a `line.rotate(angle, origin)` function on the way? And what are the results of it with an origin of (0,0) compared to doing the same rotation on a Vector2?

The way we rotate stuff geometrically doesn't particularly matter - as long as we are internally consistent.

aatle:
@MyreMylar you are comparing rotating a line by a certain angle, with the inherent "angle" of a line.
The angle of a line corresponds with the proposed `Vector2` `angle` and `angle_rad` properties in #3222, and should have similar behavior for special cases.

The angle of a line is similar to `atan(line.slope)` (quadrant incorrect), and is better described as the angle of a vector representing the line, where `a` is the origin.

The `angle` should be in degrees, for consistency. There should probably be both a degrees and a radians implementation for every angle method, also for circle too. This would be consistent with `Vector2`.

aatle:
Perhaps the vector corresponding to a line may be a useful attribute. `(bx - ax, by - ay)`.

src_c/line.c
```diff
@@ -347,6 +347,115 @@ pg_line_getlength(pgLineObject *self, void *closure)
...
+
+    return PyFloat_FromDouble(-RAD_TO_DEG(atan(gradient)));
+}
+
+static PyObject *
+pg_line_getslope(pgLineObject *self, void *closure)
+{
+    double dem = self->line.bx - self->line.ax;
+    if (dem == 0) {
+        return PyFloat_FromDouble(0);
```
aatle, line 451: I think this is inaccurate. The result here should either be an exception or `+-inf`.
If `inf`, then I would think that for `by - ay == 0` case it would be `nan` instead.

In any case, the vertical line behavior needs to be documented.

src_c/line.c
```diff
@@ -347,6 +347,115 @@ pg_line_getlength(pgLineObject *self, void *closure)
...
+    double dx = self->line.bx - self->line.ax;
+
+    if (dx == 0.0)
+        return (self->line.by > self->line.ay) ? PyFloat_FromDouble(-90.0)
+                                               : PyFloat_FromDouble(90.0);
+
+    double dy = self->line.by - self->line.ay;
+    double gradient = dy / dx;
+
+    return PyFloat_FromDouble(-RAD_TO_DEG(atan(gradient)));
```
aatle, line 443: I don't think the angle should be negated.
Also it should use some form of `atan2` instead of `atan`, to reach the full 360 deg range, as had been specified in the docs.
There's also special cases like `a == b`, `inf`, etc., where it maybe should emulate python `math.atan2(line.by - line.ay, line.bx - line.ax)` (also see #3222).

PR #3271, opened 2024-12-18
Bump SDL2_image to 2.8.3

### LINKED ISSUE #3036: Certain PNGs load incorrectly.
Reported by @geometrian on discord

1_img_okay.png:
1_img_okay

2_img_broken.png
2_img_broken

Inside pygame-ce however, image 1 load correctly and image 2 does not.

image

Basic display script:
```py
import pygame

pygame.init()

im1 = pygame.image.load("1_img_okay.png")
print("load 1 finished")
im2 = pygame.image.load("2_img_broken.png")

print(im1.get_at((20,20)), im2.get_at((20,20)))
print(im1, im1.get_shifts())
print(im2, im2.get_shifts())

running = True

print(screen, screen.get_shifts())

while running:
        if event.type == pygame.QUIT:
            running = False

    screen.fill("grey")
    screen.blit(im1, (0,0))
    screen.blit(im2, (im1.get_width(),0))

    pygame.display.flip()

pygame.quit()
```

ankith26:
fixes #3036


Starbuck5 (approved):
Confirmed again it fixes the linked issue as expected.

Thanks!

P.S. I'm quite pleased you were able to use the SDL_image fork to generate a new Windows prebuilt.

Starbuck5:
I'm glad. It's just something I thought I had expected I would need to do myself.

Matiiss (approved):
LGTM :rocket: 
Does appear to fix the linked issue too

P.S. it was very painful getting pygame-ce to compile with these changed dependencies, maybe we should streamline this? Because the wiki mentions installing development dependencies from package repositories, however, in this case, they have to be got directly from the SDL repo. I'd propose adjusting, for example, `buildconfig/manylinux-build/docker_base/sdl_libs/build-sdl2-libs.sh` so that it can be used locally.

ankith26:
The best way to test this PR would be to simply download the wheels generated by CI, that is what I also did.

I think generalizing the docker scripts is more effort than it is useful, due to the fact that we pick some settings specifically for the purpose of bundling the dependencies in wheels, you would not want those settings used in your local SDL install.

PR #3272, opened 2024-12-19
Docs updates (mainly to mixer)

Starbuck5:
> The mixer module is optional and depends on SDL_mixer. Your program should test that :mod:`pygame.mixer` is available and initialized before using it.

It's part of our API. Any remotely compliant build of pygame-ce has this, it's not something to be afraid of. I don't expect people's programs to test for pygame.mixer.

> Pygame will default to reasonable values, but pygame cannot perform Sound resampling, so the mixer should be initialized to match the values of your audio resources.

Technically, true, pygame-ce does not perform resampling. But SDL does, and they will resample into the format of your audio resources, this is highly misleading.

### DIFF
--- a/docs/reST/ref/image.rst
+++ b/docs/reST/ref/image.rst
@@ -245,7 +245,7 @@ following formats.
 
    .. note:: The use of this function is recommended over :func:`tostring` as of pygame 2.1.3.
              This function was introduced so it matches nicely with other
-             libraries (PIL, numpy, etc), and with people's expectations.
+             libraries (PIL, NumPy, etc), and with people's expectations.
 
    .. versionadded:: 2.1.3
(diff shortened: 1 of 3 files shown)


yunline (approved):
LGTM

ankith26 (commented):
I wonder if imageext, font and freetype should also be getting the same treatment? They are also in the same boat of being "technically optional but used fairly widely so not actually"

Starbuck5:
> I wonder if imageext, font and freetype should also be getting the same treatment? They are also in the same boat of being "technically optional but used fairly widely so not actually"

I don't see any such blatantly outdated text in their documentations.

ankith26 (approved):
LGTM, thanks :tada:

PR #3274, opened 2024-12-30
Fix *most of* the Deprecation Warnings in surfarray_test

MyreMylar:
Noticed that there were several fixable Deprecation Warnings coming out of our surfarray tests.

I feel in general that we should try and minimise the outputting of 'expected' warnings from our tests so that we are more likely to find the unexpected ones when running a full test run.

I spotted the warnings by running:

```
pytest test/surfarray_test.py -s
```

There are two main fixes applied:

1. Using..
```
with self.assertWarns(DeprecationWarning):
```
as a context manager when we are deliberately calling a deprecated function at the moment - we can remove tests like this  if/when the deprecated function itself is finally removed.

2. Applying the recommendation from Numpy's deprecation warning where we have deprecated behaviour:
```
DeprecationWarning: NumPy will stop allowing conversion of out-of-bound Python integers to integer arrays.  The conversion of -10185016 to uint32 will fail in the future.
  For the old behavior, usually:
      np.array(value).astype(dtype)
  will give the desired result (the cast overflows).
    arr[:5, 6:] = surf.map_rgb(palette[3])
```

### DIFF
--- a/test/surfarray_test.py
+++ b/test/surfarray_test.py
@@ -13,7 +13,7 @@ from numpy import (
     rint,
     arange,
     __version__ as np_version,
-    array
+    array,
 )
 
 import pygame
--- a/test/surfarray_test.py
+++ b/test/surfarray_test.py
@@ -13,6 +13,7 @@ from numpy import (
     rint,
     arange,
     __version__ as np_version,
+    array
 )
 
 import pygame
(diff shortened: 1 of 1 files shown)


damusss (approved):
Nice fix, thanks :3

test/surfarray_test.py
```diff
@@ -114,10 +115,10 @@ def _make_array3d(self, dtype):
...
     def _fill_array2d(self, arr, surf):
         palette = self.test_palette
-        arr[:5, :6] = surf.map_rgb(palette[1])
-        arr[5:, :6] = surf.map_rgb(palette[2])
-        arr[:5, 6:] = surf.map_rgb(palette[3])
-        arr[5:, 6:] = surf.map_rgb(palette[4])
+        arr[:5, :6] = array(surf.map_rgb(palette[1])).astype(int)
+        arr[5:, :6] = array(surf.map_rgb(palette[2])).astype(int)
+        arr[:5, 6:] = array(surf.map_rgb(palette[3])).astype(int)
+        arr[5:, 6:] = array(surf.map_rgb(palette[4])).astype(int)
```
ankith26, line 121: I believe this is one of the things that is already biting us in numpy v2. And from what I can remember it isn't trivial to change `surfarray`/`map_rgb` such that this works and nothing else breaks. But it should be possible to find a solution that does minimal breaks.

So basically, I think we need a solution that can fix this usage without this `array.astype` hack.

MyreMylar: OK, if you think there is a code solution then I'm happy to revert this change and leave it as one of the warnings we should fix in a more in-depth way.

But in a way this is a good example of the issue with us having so many warnings being spat out from the tests - the actual warnings we might want to investigate more fully are getting lost.

ankith26 (approved):
LGTM, thanks! :tada:

PR #3276, opened 2024-12-30
Rename TestTags so it doesn't get confused for a test

MyreMylar:
Noticed this warning in the console:

```
test\test_utils\test_machinery.py:47
  C:\Users\dan\Programming\pygame-ce\test\test_utils\test_machinery.py:47: PytestCollectionWarning: cannot collect test class 'TestTags' because it has a __init__ constructor (from: test/test_utils/test_machinery.py)
    class TestTags:
```

The class name starting with `Test` is confusing the testing machinery that looks for tests in the `tests` folder. Seems a simple enough fix to rename the class slightly so it doesn't begin with `Test`.


damusss (approved):
Makes sense, LGTM :)
(the failing test should get fixed by your latest pr)

ankith26 (approved):
LGTM, thanks! :tada:

PR #3277, opened 2024-12-30
Fix mypy to 1.13.0 to avoid __all__ duplication

MyreMylar:
The relevant mypy issue: https://github.com/python/mypy/issues/13300
and PR: https://github.com/python/mypy/pull/18005

Basically MyPy updated and added a new feature of erroring on `__all__` not being included in the stubs. I'm not sure we want to do this as constants `__all__` is very large. Either way it is some annoying work to solve in the future that is causing the unit tests to fail right now.

This PR just locks us to the previous version before this change was added.

### DIFF
--- a/.github/workflows/build-ubuntu-sdist.yml
+++ b/.github/workflows/build-ubuntu-sdist.yml
@@ -84,7 +84,7 @@ jobs:
 
     - name: Test typestubs
       run: |
-        pip3 install mypy
+        pip3 install mypy==1.13.0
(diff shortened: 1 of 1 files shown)


damusss (approved):
I agree. it's an unnecessary update, if you don't specify all in a stub it's clear you want everything non-private to be included

ankith26 (approved):
Yeah I agree this is a good solution for now, and that we could probably change it in the future.

PR #3280, opened 2025-01-05
Remove unnecessary lock in gfxdraw

Starbuck5:
All the surrounding code is written to assume the surface has already been locked (which it has). Possibly this was a copy paste error many years ago, it's been in pygame/pygame-ce for at least 9 years.

If you look at the source of `pixelColor` vs `pixelColorNolock`, they do the exact same thing, except the former locks and unlocks the surface, which is unnecessary.

I noticed this when looking into porting gfxdraw to SDL3.


damusss (approved):
Damn, how did you spot this

ankith26 (approved):
LGTM, nice catch! :tada:

PR #3282, opened 2025-01-05
Add `Font.set_linesize()`  (TTF 2.24.0 feature)

itzpr3d4t0r:
We've long had the ability to get the line size with `get_linesize()`, but we've never had the ability to set the spacing between lines when rendering multiline text. Thanks to SDL_TTF 2.24.0 this PR adds the functionality.


damusss (approved):
As much as I don't like set_/get_ methods, get_linesize was already there, so not adding it would be rather weird (we can have just properties with pygame 3, I guess!). And `Font.linesize` can be added by another PR, too, so this LGTM, thanks for adding it :)

docs/reST/ref/font.rst
```diff
@@ -511,6 +511,16 @@ solves no longer exists, it will likely be removed in the future.
...
 
+   .. method:: set_linesize
+
+      | :sl:`set the line space of the font text`
+      | :sg:`set_linesize(linesize) -> int`
+
+        Set the height in pixels for a line of text with the font. When rendering
+        multiple lines of text this refers to the amount of space between lines.
+
+      .. ## Font.set_linesize ##
```
Starbuck5, line 522: Needs a "new in version" entry

src_c/font.c
```diff
@@ -226,6 +226,32 @@ font_get_linesize(PyObject *self, PyObject *_null)
...
+    }
+
+#if SDL_TTF_VERSION_ATLEAST(2, 24, 0)
+    TTF_Font *font = PyFont_AsFont(self);
+    int linesize = PyLong_AsLong(arg);
+    if (PyErr_Occurred())
+        return NULL;
+
+    if (linesize < 0)
+        return RAISE(PyExc_ValueError, "linesize must be >= 0");
```
Starbuck5, line 243: What does SDL_ttf do if this isn't true? Also could be good to specify in the docs that this needs to be a positive and nonzero number.

src_c/font.c
```diff
@@ -226,6 +226,32 @@ font_get_linesize(PyObject *self, PyObject *_null)
...
+static PyObject *
+font_set_linesize(PyObject *self, PyObject *arg)
+{
+    if (!PgFont_GenerationCheck(self)) {
+        return RAISE_FONT_QUIT_ERROR();
+    }
+
+#if SDL_TTF_VERSION_ATLEAST(2, 24, 0)
+    TTF_Font *font = PyFont_AsFont(self);
+    int linesize = PyLong_AsLong(arg);
```
Starbuck5, line 238: I tried passing a float in and got a deprecation error:
```
C:\Users\charl\Desktop\test.py:31: DeprecationWarning: an integer is required (got type float).  Implicit conversion to integers using __int__ is deprecated, and may be removed in a future version of Python.
  font.set_linesize(0.11)
```

Starbuck5:
This will be a great feature to get in! Left a few small notes / a question.

Wrote a sample program to test around with:

```py
import pygame

pygame.init()
screen = pygame.display.set_mode((680, 420))
clock = pygame.time.Clock()
running = True

font = pygame.font.SysFont("Arial", 32)

def generate_text():
    return font.render(
        "Hello world\nThis is an example of\nmultiline text\nusing different spacings.",
        True,
        "black",
    )

text = generate_text()

while running:
        if event.type == pygame.QUIT:
            running = False

        if event.type == pygame.KEYDOWN and event.key == pygame.K_UP:
            font.set_linesize(font.get_linesize() + 5)
            print(font.get_linesize())
            text = generate_text()

        if event.type == pygame.KEYDOWN and event.key == pygame.K_DOWN:
            font.set_linesize(font.get_linesize() - 5)
            print(font.get_linesize())
            text = generate_text()

    screen.fill("purple")

    for i in range(4):
        pygame.draw.rect(
            screen,
            "orangered" if i % 2 else "lightblue",
            [0, i * font.get_linesize(), 300, font.get_linesize()],
        )

    screen.blit(text)
    pygame.display.flip()
    clock.tick(60)

pygame.quit()
```

Starbuck5:
Last time I talked to itzpr he was busy IRL and okay with people taking over his PRs. 

@aatle I was wondering if you were interested in taking this over, since you’re a member now you can push to itzprs branch. This could be a good first C PR for you, as it just needs some polishing up to get over the finish line.

src_c/font.c
aatle: When I tried allowing negative sizes, `font.render()` raised `pygame.error` for a large negative size (possibly when greater than height).
Documented constraint.

Starbuck5 (approved):
Thanks @itzpr3d4t0r and @aatle, looks good to me.

ankith26 (approved):

PR #3283, opened 2025-01-05
event.peek return bool, minor doc/stub fixes

### LINKED ISSUE #1196: pygame.event.peek docs missing a detail (2329)

ankith26:
My take on #3122

fixes #1196

Credits: The commit in this PR has been based on work done by Zoldalma and MyreMylar in the linked PR


MyreMylar (approved):
OK, LGTM 👍

PR #3285, opened 2025-01-08
Implement `display.gl_get_proc`

### LINKED ISSUE #3284: Expose SDL_GL_GetProcAddress
Pygame is a binding to SDL, which has functionality to  query opengl procedure addresses (https://wiki.libsdl.org/SDL2/SDL_GL_GetProcAddress) however this does not appear to be visible in pygame.

why it’s important: currently pygame works well with PyOpenGL, however the PyOpenGL binding does not include all available extensions (eg. Bindless textures) in glfw, these extensions can be fetched directly (eg. What to do when PyOpenGL doesn't support an OpenGL function (load_proc_addr)
https://youtu.be/GGfuJj2ubzI). However I’d rather use Pygame over glfw as it’s more widely supported.

yunline:
Trys to fix  #3284

Example:
```py
import pygame
import ctypes

win = pygame.Window(opengl=True)

GL_VENDOR = 0x1F00
GL_RENDERER = 0x1F01
GL_VERSION = 0x1F02
GL_EXTENSIONS = 0x1F03

glGetString = pygame.display.gl_get_proc("glGetString")
glGetString.restype = ctypes.c_char_p
glGetString.argtypes = (ctypes.c_int,)
print(glGetString(GL_VENDOR))

# prints b'ATI Technologies Inc.' on my computer
```

**WARNING**
This PR introduces a new dependency: `ctypes`

This PR is experimental. If you have better idea, feel free to open another PR.

### DIFF
--- a/docs/reST/ref/display.rst
+++ b/docs/reST/ref/display.rst
@@ -540,6 +540,13 @@ required).
 
    .. ## pygame.display.gl_set_attribute ##
 
+.. function:: gl_get_proc
+
+   | :sl:`Get an OpenGL function by name`
+   | :sg:`gl_get_proc(proc_name) -> ctypes._FuncPointer`
+
+   .. # pygame.display.gl_get_proc ##
+
 .. function:: get_active
 
    | :sl:`Returns True when the display is active on the screen`
--- a/src_c/display.c
+++ b/src_c/display.c
@@ -3078,7 +3078,8 @@ static PyMethodDef _pg_display_methods[] = {
      METH_VARARGS | METH_KEYWORDS, DOC_DISPLAY_SETALLOWSCREENSAVER},
     {"message_box", (PyCFunction)pg_message_box, METH_VARARGS | METH_KEYWORDS,
      DOC_DISPLAY_MESSAGEBOX},
-    {"gl_get_proc", (PyCFunction)pg_gl_get_proc, METH_O, "doc"},
+    {"gl_get_proc", (PyCFunction)pg_gl_get_proc, METH_O,
+     DOC_DISPLAY_GLGETPROC},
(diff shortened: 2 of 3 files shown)


Starbuck5:
Thoughts:

Maybe it would be simpler to import ctypes instead of a custom module into display.c, and then use PyObject_call stuff and ifdef WIN32 to do the same logic as the custom ffi.py module

Maybe we should give the pointer only, and let the user worry about ctypes, since they're going to need to worry about ctypes anyway to extract any value (as I understand it). Plus it makes it simpler code for this seemingly-niche use case.

docs/reST/ref/display.rst
```diff
@@ -540,6 +540,13 @@ required).
 
    .. ## pygame.display.gl_set_attribute ##
 
+.. function:: gl_get_proc
+
+   | :sl:`Get an OpenGL function by name`
+   | :sg:`gl_get_proc(proc_name) -> ctypes._FuncPointer`
+
```
damusss, line 547: Since this is an uncommon function it could be explained with a bit more detail, for example mentioning that it will try to import ctypes if it is not imported, the fact that it can fail and raise errors, and if it's relevant roughly how to use the return value with an embedded example.

damusss:
[assuming the "changes" are addressed/discussed and that the code actually works]
It's true that it's uncommon but sdl still exports it and it makes sense to reside in display, given other GL functions are already there. plus it makes sense if one needs to do something simple with opengl and doesn't want to import something big like PyOpenGL, or if it is using something like moderngl and only has a missing function or two, it can use this function instead.

I think it's ok for it to return a ctypes object since it is imported only on first function call. if we didn't use ctypes the return value wouldn't be much of an abstraction, so it would be less useful, I think.

Issue #3239, opened 2024-11-25
Deprecate and remove the legacy types like `SoundType`


yunline:
- `CameraType`
- `EventType`
- `FontType`
- `JoystickType`
- `MaskType` 
- `SoundType`
- `ChannelType`
- `RectType`
- `FRectType` (what?)
- `SurfaceType`

These `*Type`s has a long history. You can find them since the first commit of pygame on github.
They are aliases of the original type object and are not recommended to use.

### What are `*Type` used for?
Before python 2.3, there is no `tp_new` slot in `PyTypeObject` struct. People use a function to "new" a class defined by c extension.
```c
// Copied from https://docs.python.org/2.2/ext/dnt-basics.html
...

static PyMethodDef noddy_methods[] = {
    {"new_noddy", noddy_new_noddy, METH_VARARGS,
     "Create a new Noddy object."},
    {NULL, NULL, 0, NULL}
};

DL_EXPORT(void)
initnoddy(void) 
{
    noddy_NoddyType.ob_type = &PyType_Type;

    Py_InitModule("noddy", noddy_methods);
}
```
Take `Event` as an example, at that time, `EventType` is the type of event, people need to call `Event()` function, to create an `EventType` object.

After python2.3, API changed, people no longger need a explicit function to "new" a class
```c
// Copied from https://docs.python.org/2.3/ext/dnt-basics.html
...

static PyMethodDef noddy_methods[] = {
    {NULL}  /* Sentinel */
};

#ifndef PyMODINIT_FUNC	/* declarations for DLL import/export */
#define PyMODINIT_FUNC void
#endif
PyMODINIT_FUNC
initnoddy(void) 
{
    PyObject* m;

    noddy_NoddyType.tp_new = PyType_GenericNew;
    if (PyType_Ready(&noddy_NoddyType) < 0)
        return;

    m = Py_InitModule3("noddy", noddy_methods,
                       "Example module that creates an extension type.");

    Py_INCREF(&noddy_NoddyType);
    PyModule_AddObject(m, "Noddy", (PyObject *)&noddy_NoddyType);
}
```
Again, take `Event` as example, now people only need to call `EventType()` to create an `EventType` object

To keep the compatibility of the code written before python 2.2, the `Event` become the type name and `EventType` become an alias.
People can still call `Event()` to get an `Event` object.

These could explain why there are so many `*Type` in pygame.

### Deprecate and remove
[...]

PR #3288, opened 2025-01-10
Mark old xType aliases as deprecated

Starbuck5:
Starts to handle https://github.com/pygame-community/pygame-ce/issues/3239

It would be a huge pain to put in deprecation warnings about this, but the deprecation decorator is easy and we're now using it in other places so lets use it here.

Doesn't touch JoystickType because the C code still works the old way I think, doesn't touch CameraType because CameraType is not in the stubs. In fact CameraType might be removable as an internal implementation detail.


buildconfig/stubs/pygame/rect.pyi
```diff
@@ -276,5 +277,8 @@ class Rect(_GenericRect[int]):
 class FRect(_GenericRect[float]):
     ...
 
-RectType = Rect
-FRectType = FRect
+@deprecated("Use `Rect` instead (this is an old alias)")
+class RectType(Rect): ...
+
+@deprecated("Use `FRect` instead (this is an old alias)")
+class FRectType(FRect): ...
```
ankith26, line 284: Not really a review but an observation, it's funny that `FRectType` exists in the first place because I'm fairly sure `FRect` was added after we decided that `*Type` was not needed anymore and stopped doing it.

buildconfig/stubs/pygame/event.pyi
```diff
@@ -1,9 +1,37 @@
...
+    # before any uses of the dict[] typehinting because of the same naming
+    @property
+    def dict(self) -> dict[str, Any]: ...
+
+@final
+@deprecated("Use `Event` instead (this is an old alias)")
+class EventType:
+    # Duplicated Event fields here because it is marked as final and can't be
+    # subclassed like the other xType aliases are to allow @deprecated.
+
```
ankith26, line 34: I like how all other instances are elegantly marked as deprecated without duplication, and there's there's this that sticks out.
I get why it's done, but can we find a workaround that doesn't need the duplication? Something like

`EventType = deprecated("Use `Event` instead (this is an old alias)")(Event)` should be a one-liner that could work? If it does, this style could be used everywhere

Starbuck5: That is a clever idea, but I get a stubtest error.

```
error: pygame.event.EventType variable differs from runtime type builtins.type
Stub: in file C:\Users\charl\Desktop\pygame-ce\buildconfig\stubs\pygame\event.pyi:29
def (type: builtins.int, dict: builtins.dict[builtins.str, Any] =, **kwargs: Any) -> pygame.event.Event
Runtime:
<class 'pygame.event.Event'>
```

Starbuck5: I did get it almost perfect with

```
class _EventStuff:
    # Event fields here duplicated below into EventType because Event is
    # marked as final and can't be subclassed like the other xType aliases
    # are to allow @deprecated.

    @property
    def type(self) -> int: ...
    __dict__: dict[str, Any]
    __hash__: None  # type: ignore
    def __init__(
        self, type: int, dict: dict[str, Any] = ..., **kwargs: Any
    ) -> None: ...
    def __getattribute__(self, name: str) -> Any: ...
    def __setattr__(self, name: str, value: Any) -> None: ...
    def __delattr__(self, name: str) -> None: ...
    def __bool__(self) -> bool: ...

    # this is at the bottom because mypy complains if this declaration comes
    # before any uses of the dict[] typehinting because of the same naming
    @property
    def dict(self) -> dict[str, Any]: ...

@final
class Event(_EventStuff):
    pass

@final
@deprecated("Use `Event` instead (this is an old alias)")
class EventType(_EventStuff):
    pass
```

The problem is that _EventStuff is also exported, shows up in the list of suggestions when typing pygame.event. I tried `del _EventStuff` but that didn't change it.

ankith26: I think the `_EventStuff` approach is fine, given we already have stuff like `_GenericVector` and `_GenericRect` in the stubs already

MyreMylar (approved):
LGTM 👍

buildconfig/stubs/pygame/event.pyi
Starbuck5: Alright, implemented it as _GenericEvent (because that sounds better than EventStuff).

buildconfig/stubs/pygame/event.pyi
```diff
@@ -21,6 +24,15 @@ class Event:
     @property
     def dict(self) -> dict[str, Any]: ...
 
+@final
+class Event(_GenericEvent):
+    pass
+
+@final
+@deprecated("Use `Event` instead (this is an old alias)")
```
aatle, line 32: It might be a confusing to use "this" there, it looks like it would refer to `Event`.
```suggestion
@deprecated("Use `Event` instead of old alias")
```

Starbuck5: Pushed a change to refer to "EventType" instead of "this" so it's clear what the message is talking about.

damusss: @Starbuck5  Shouldn't all XType aliases that you deprecated use XType instead of 'this' aswell then?

Starbuck5: Yep yep, fixed.

damusss (approved):
Looking good, thanks!
_long live deprecation!_

PR #3292, opened 2025-01-16
Vector caching.

gresm:
As in the title. Benchmarks would be nice if anyone has time for doing them.

### DIFF
--- a/src_c/math.c
+++ b/src_c/math.c
@@ -87,7 +87,11 @@ static PyTypeObject pgVectorIter_Type;
 
 #define pgVector2_Check(x) (PyType_IsSubtype(Py_TYPE(x), &pgVector2_Type))
 #define pgVector3_Check(x) (PyType_IsSubtype(Py_TYPE(x), &pgVector3_Type))
+#define pgVector2_CheckExact(x) (Py_TYPE(x) == &pgVector2_Type)
+#define pgVector3_CheckExact(x) (Py_TYPE(x) == &pgVector3_Type)
 #define pgVector_Check(x) (pgVector2_Check(x) || pgVector3_Check(x))
+#define pgVector_CheckExact(x) \
+    (pgVector2_CheckExact(x) || pgVector3_CheckExact(x))
 #define vector_elementwiseproxy_Check(x) \
     (Py_TYPE(x) == &pgVectorElementwiseProxy_Type)
 #define _vector_subtype_new(x) \
@@ -110,6 +114,31 @@ typedef struct {
 typedef struct {
     PyObject_HEAD pgVector *vec;
 } vector_elementwiseproxy;
+
+#define VECTOR_CACHE_SIZE 100
+
+static pgVector *vec2_cache[VECTOR_CACHE_SIZE] = {};
+static int vec2_cache_num = 0;
+static pgVector *vec3_cache[VECTOR_CACHE_SIZE] = {};
+static int vec3_cache_num = 0;
+
+#define _VEC_CACHE_ADD(x, v)                          \
+    {                                                 \
+        if (vec##x##_cache_num < VECTOR_CACHE_SIZE) { \
+            Py_INCREF(v);                             \
+            vec##x##_cache[vec##x##_cache_num] = v;   \
+            vec##x##_cache_num++;                     \
+        }                                             \
+    }
+
+#define _VEC_CACHE_POP(x, v)                           \
+    {                                                  \
+        if (vec##x##_cache_num > 0) {                  \
+            vec##x##_cache_num--;                      \
+            (v) = vec##x##_cache[vec##x##_cache_num];  \
+            vec##x##_cache[vec##x##_cache_num] = NULL; \
+        }                                              \
+    }
 
 /* further forward declarations */
 /* math functions */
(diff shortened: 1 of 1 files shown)


src_c/math.c
```diff
@@ -628,9 +657,33 @@ pgVector_NEW(Py_ssize_t dim)
 static void
 vector_dealloc(pgVector *self)
 {
+    if (pgVector_CheckExact(self) &&
+        PyObject_CallFinalizerFromDealloc((PyObject *)self) < 0) {
```
ankith26, line 661: I see no mention of this function in the C API docs. What does it do exactly?

Also, I think there is a possibility of introducing a memory leak here. What happens when the cache is full? I would expect vectors to be dealloc'd normally in that case, but it looks like those vectors will be leaked in this implementation

src_c/math.c
```diff
@@ -111,6 +115,31 @@ typedef struct {
     PyObject_HEAD pgVector *vec;
 } vector_elementwiseproxy;
 
+#define VECTOR_CACHE_SIZE 100
```
ankith26, line 118: I wonder why this is 100, I mean, I would expect it to be bigger.
To be fair, these numbers are hard to get right though. I would expect numbers like this to be backed up by empirical test data. A simple test app that uses vectors extensively would be a good start.

src_c/math.c
```diff
@@ -628,9 +657,33 @@ pgVector_NEW(Py_ssize_t dim)
...
 {
+    if (pgVector_CheckExact(self) &&
+        PyObject_CallFinalizerFromDealloc((PyObject *)self) < 0) {
+        return;
+    }
     Py_TYPE(self)->tp_free((PyObject *)self);
 }
 
+static void
+vector_finalize(PyObject *self)
```
ankith26, line 668: Why is this logic kept in a separate finalize function, is there a particular reason it should not be a part of `dealloc`?

src_c/math.c
gresm: > I see no mention of this function in the C API docs. What does it do exactly?

https://github.com/python/cpython/issues/78090

> Also, I think there is a possibility of introducing a memory leak here. What happens when the cache is full?

``PyObject_CallFinalizerFromDealloc`` returns non-zero value only when object is resurrected [[1]](https://github.com/python/cpython/blob/5ed5572cac7ef204767ddf8e8888e15672ba558e/Objects/object.c#L575) (the ref-count is increased while finalizing, which only happens when cache is not full - ``_VEC_CACHE_ADD``).

src_c/math.c
gresm: Yes, that's true. I don't have experience in that.

src_c/math.c
ankith26: Oh okay, yeah that makes sense

src_c/math.c
ankith26: It would be nice if you could write a simple benchmark that demonstrates the exact performance enhancements this approach can give.

src_c/math.c
ankith26: I'm a bit skeptical on using undocumented Python C API, but for now I think it's okay

Starbuck5:
> Benchmarks would be nice if anyone has time for doing them.

I think it is the PR author's job to justify why the change should be added. Writing the code is often not the hard part, it's doing the research, doing experimentation, making sure others understand. From what I've seen that still needs to happen, so I'm marking this PR as a draft.

gresm:
@Starbuck5
---
Using this timing script:
```py
import time

def test(rounds):
    import pygame
    start = time.perf_counter()
    for _ in range(rounds):
        vec = pygame.Vector3(1, 2, 3)
        vec2 = vec.copy()
    end = time.perf_counter()
    print(f"Total time: {end - start}, on average: {(end - start) / rounds}")
```

## For ``test(10**7)`` and ``VECTOR_CACHE_SIZE==100000``
Before:
```
Total time: 1.37842830199952, on average: 1.37842830199952e-07
```
After:
```
Total time: 1.2690909360026126, on average: 1.2690909360026126e-07
```
## For ``test(10**7)`` and ``VECTOR_CACHE_SIZE==10000``
Before:
```
Total time: 1.3938012459984748, on average: 1.3938012459984747e-07
```
After:
```
Total time: 1.2679740370003856, on average: 1.2679740370003856e-07
```

## For ``test(10**7)`` and ``VECTOR_CACHE_SIZE==1000``
Before:
```
Total time: 1.392446205001761, on average: 1.392446205001761e-07
```
After:
```
Total time: 1.2817346929987252, on average: 1.2817346929987253e-07
```

## For ``test(10**7)`` and ``VECTOR_CACHE_SIZE==100``
Before:
```
Total time: 1.3991181530000176, on average: 1.3991181530000176e-07
```
After:
```
Total time: 1.2715259210017393, on average: 1.2715259210017393e-07
```

## For ``test(10**7)`` and ``VECTOR_CACHE_SIZE==10``
Before:
```
Total time: 1.3997546829996281, on average: 1.399754682999628e-07
```
After:
```
Total time: 1.2715557740011718, on average: 1.271555774001172e-07
```

## For ``test(10**7)`` and ``VECTOR_CACHE_SIZE==1``
Before:
```
Total time: 1.4705402879990288, on average: 1.4705402879990288e-07
```
After:
```
Total time: 1.2530362709985639, on average: 1.253036270998564e-07
```
## Conclusion
There's a difference between them, but I probably messed up benchmarks a bit. Maybe instancing multiple vectors at the same time would make more sense.

src_c/math.c
gresm: @ankith26  Done

PR #3293, opened 2025-01-17
Drop circleci, use gh actions for arm manylinux

ankith26:
https://github.blog/changelog/2025-01-16-linux-arm64-hosted-runners-now-available-for-free-in-public-repositories-public-preview/

By removing circleci we are reducing our maintenance burden, and now have one unified config to maintain for manylinux instead of 2.


Starbuck5 (approved):
Checked the size and checked the same wheels generated, looks good to me.

The arm runners are a public preview though so they could go away in the future.

ankith26:
We can always switch back to circleci easily if the need arises in the future

MyreMylar:
Looks like CircleCI still tried to run on this PR.

ankith26:
I have disabled CircleCI from the web UI now, so it shouldn't try to build and fail after this PR is merged

PR #3294, opened 2025-01-17
Introduce PG_SURF_FORMATENUM macro (SDL3 compat)

Starbuck5:
- Brings SDL3 just a bit closer. Rotozoom.c for example now compiles, previously did not.
- Simplify 2 pixel format checks (in mask and transform)
- Remove unneeded include from transform.c (SDL_cpuinfo.h comes in fine from SDL.h, and fails to include on SDL3)

This macro is the same as PG_SurfaceFormatEnum from gresm's surface PR #3215, but named as a macro and applied throughout the project. Except for window.c, because I didn't want to cause a merge conflict with Ankith's window SDL3 PR.

### DIFF
--- a/src_c/_camera.c
+++ b/src_c/_camera.c
@@ -109,7 +109,7 @@ surf_colorspace(PyObject *self, PyObject *arg)
     surf = pgSurface_AsSurface(surfobj);
 
     if (!surfobj2) {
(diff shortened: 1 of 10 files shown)


ankith26 (approved):

PS: you can apply the macro on window as well, I don't mind dealing with the conflict. I will revive that PR after this one merges.

MyreMylar (approved):
LGTM 👍

PR #3297, opened 2025-01-19
Optimize correct colordict entries

damusss:
as @Starbuck5 noticed string methods are called even if the color name existed raw in the color dictionary. This PR adds a fast path when that is the case. I tested with timeit before and after this PR and correct names got a major performance increase (from 5.5 to 1.8) while other cases are basically left unchanged. I believe a fast path should also be considered for hex colors, but it's beyond the scope of this PR. MRE:
```py
import pygame
import timeit

# change the string for different cases
print(timeit.timeit("pygame.Color('white')", globals=globals(), number=10_000_000))

# OLD
# correct: 5.5-5.8
# hex: 8-8.2
# case: 5.6-5.8
# space: 5.9+

# NEW
# correct: 1.8-1.9
# hex: 8-8.2
# case: 5.7-5.9
# space: 6+
```

### DIFF
--- a/src_c/color.c
+++ b/src_c/color.c
@@ -598,29 +598,42 @@ _parse_color_from_text(PyObject *str_obj, Uint8 *rgba)
     /* We assume the caller handled this check for us. */
     assert(PyUnicode_Check(str_obj));
 
-    name1 = PyObject_CallMethod(str_obj, "replace", "(ss)", " ", "");
-    if (!name1) {
-        return -1;
-    }
-    name2 = PyObject_CallMethod(name1, "lower", NULL);
-    Py_DECREF(name1);
-    if (!name2) {
-        return -1;
-    }
-    color = PyDict_GetItem(_COLORDICT, name2);
-    Py_DECREF(name2);
+    color = PyDict_GetItem(_COLORDICT,
+                           str_obj);  // optimize for correct color names
     if (!color) {
-        switch (_hexcolor(str_obj, rgba)) {
-            case TRISTATE_FAIL:
-                PyErr_SetString(PyExc_ValueError, "invalid color name");
-                return -1;
-            case TRISTATE_ERROR:
-                return -1;
-            default:
-                break;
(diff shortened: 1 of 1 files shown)


src_c/color.c
```diff
@@ -598,36 +598,50 @@ _parse_color_from_text(PyObject *str_obj, Uint8 *rgba)
...
+        else {
+            goto found_color;
+        }
+    }
+    else {
+        goto found_color;
     }
-    else if (!pg_RGBAFromObjEx(color, rgba, PG_COLOR_HANDLE_RESTRICT_SEQ)) {
+
+found_color:
```
ankith26, line 635: All gotos to this point are redundant if the codepath is going to flow into this label naturally. I suggest simplifying this slightly by removing this label altogether.

ankith26: This is just for code quality purposes. With modern compilers I'm mostly sure this sort of thing has no performance impacts.

damusss: yeah, you are right. only thing I need to keep is the success label otherwise an hex color will segfault

ankith26: You can replace that goto too with a simple `return 0;` if you'd like

ankith26 (approved):

Starbuck5 (approved):
Nice work, you turned me complaining about something into something productive! I did not have this in mind at all.

I tested with set_at to get an example of it doing something real (but also cheap).

```py
import time
import pygame

pygame.init()

screen = pygame.Surface((300,300))

start = time.time()
for _ in range(1000000):
    screen.set_at((4,35), "red")
print(time.time() - start)
```

This went from 0.67 seconds to 0.29 seconds, a more than 2x boost.

Saving 0.38 seconds over a million reps means this saves 380 nanoseconds per call. If we assume the mapped tuple is absolutely zero color mapping cost there are only 120 nanoseconds left to save.

For the curious, using a mapped color was 0.17 seconds, using a 3 element tuple took 0.25 seconds, using a 4 element tuple took 0.26 seconds, and using a hex string took 1.1 seconds.

PR #3301, opened 2025-01-20
Freetype typing fixes, enhancements to dict/tuple/ellipsis typing

ankith26:
A bunch of minor general typing fixes and enhancements

- Ellipsis is allowed in rect subscripted. Fixed the typing here, and also updated the usage in pixelarray to the more modern way.
- Enhanced dict typing with `TypedDict` where possible. This will convey more typing info to users.
- Fixed bare `tuple` usage with type info
- In freetype, most methods that take `text` also accept `None` values, so fixed that. Also `set_default_resolution` can be called without arguments (in which case it takes the value 0)


damusss (approved):
Actually great typehint enhancements, Nice :)

MyreMylar (approved):
Great! 👍

PR #3302, opened 2025-01-23
Optimize for str hex color values

ankith26:
Inspired by https://github.com/pygame-community/pygame-ce/pull/3297

That PR optimized str colordict access but left hexcolor optimization for the future, this PR tackles that now. After this PR, a similar 2-3x performance enhacement is seen for the hex color case as well.


Starbuck5 (approved):
Looks good to me.

I like the PyUnicode_AsUTF8AndSize simplification as well.

damusss (approved):
Tested it an the speedups are there as promised :) Thanks for "completing" my work

PR #3303, opened 2025-01-23
Move get_rect helper to c_api

MightyJosip:
This PR adds helper used to create the return value of  `Surface.get_(f)rect` to public base api. So far that function was only used inside of Surface class, but since that same functionality is present in `pygame._sdl2.Texture` and `pygame._sdl2.Image` classes I moved it to public api to make it easier for implementation in the future (when we rewrite Texture/Image in C). That will also make it easier to implement extra stuff for those classes (like FRect support for both, or keyword support for Image.get_rect)


Starbuck5 (approved):
Haven't tested, but this looks fine.

I got the same test error on one of my PRs, which makes even less sense for it to error than this one, so it's not this PRs fault.

damusss (approved):
Makes sense, LGTM!

damusss:
Joystick segfaulting doesn't really seem to have anything to do with this PR so I think it's safe to merge

PR #3307, opened 2025-01-26
Mask SDL3

Starbuck5:

This PR is those macros + the mask module.

### DIFF
--- a/src_c/_pygame.h
+++ b/src_c/_pygame.h
@@ -108,6 +108,33 @@ PG_UnlockMutex(SDL_mutex *mutex)
 #define PG_FORMAT_BytesPerPixel(format) format->bytes_per_pixel
 #define PG_SURF_FORMATENUM(surf) surf->format
 
+#define PG_FORMAT_R_LOSS(format) (8 - format->Rbits)
+#define PG_FORMAT_G_LOSS(format) (8 - format->Gbits)
+#define PG_FORMAT_B_LOSS(format) (8 - format->Bbits)
+#define PG_FORMAT_A_LOSS(format) (8 - format->Abits)
+
+#define PG_PixelFormat const SDL_PixelFormatDetails
+
+static inline bool
+PG_GetSurfaceDetails(SDL_Surface *surf, PG_PixelFormat **format_p,
+                     SDL_Palette **palette_p)
+{
+    *palette_p = SDL_GetSurfacePalette(surf);
+    *format_p = SDL_GetPixelFormatDetails(surf->format);
+    return *format_p != NULL;
+}
+
+static inline PG_PixelFormat *
+PG_GetSurfaceFormat(SDL_Surface *surf)
+{
+    return SDL_GetPixelFormatDetails(surf->format);
+}
+
+#define PG_GetRGBA SDL_GetRGBA
+#define PG_GetRGB SDL_GetRGB
+#define PG_MapRGBA SDL_MapRGBA
+#define PG_MapRGB SDL_MapRGB
+
 /* Mask to test if surface flags are in a fullscreen window. */
 #define PG_WINDOW_FULLSCREEN_INCLUSIVE SDL_WINDOW_FULLSCREEN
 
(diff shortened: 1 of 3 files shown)


src_c/_pygame.h
```diff
@@ -108,6 +108,33 @@ PG_UnlockMutex(SDL_mutex *mutex)
 #define PG_FORMAT_BytesPerPixel(format) format->bytes_per_pixel
 #define PG_SURF_FORMATENUM(surf) surf->format
 
+#define PG_FORMAT_R_LOSS(format) format->Rbits
+#define PG_FORMAT_G_LOSS(format) format->Gbits
+#define PG_FORMAT_B_LOSS(format) format->Bbits
+#define PG_FORMAT_A_LOSS(format) format->Abits
```
ankith26, line 114:
```suggestion
#define PG_FORMAT_R_LOSS(format) (8 - format->Rbits)
#define PG_FORMAT_G_LOSS(format) (8 - format->Gbits)
#define PG_FORMAT_B_LOSS(format) (8 - format->Bbits)
#define PG_FORMAT_A_LOSS(format) (8 - format->Abits)
```
bits and loss are not the same thing, but they are complementary. Ref: https://github.com/libsdl-org/SDL/pull/10187

src_c/_pygame.h
```diff
@@ -176,6 +203,56 @@ PG_UnlockMutex(SDL_mutex *mutex)
...
+{
+    return SDL_MapRGBA(format, r, g, b, a);
+}
+
+static inline Uint32
+PG_MapRGB(PG_PixelFormat *format, const SDL_Palette *palette, Uint8 r, Uint8 g,
+          Uint8 b)
+{
+    return SDL_MapRGB(format, r, g, b);
+}
```
ankith26, line 254: These compat functions are not equivalent in behaviour. The SDL2 versions are ignoring the palette but the SDL3 path seemingly will handle it?

Starbuck5: In the SDL2 codepath the palette is part of the PixelFormat struct, so it does handle it.

I chose to have all callers extract the palette on SDL2 and SDL3 and then send both into the compat functions, so that the the 2 codepaths have the same variables and signatures.

Starbuck5: For extra assurance (for both of us), here is the source of SDL_MapRGB on SDL2.

```c
/* Find the pixel value corresponding to an RGBA quadruple */
Uint32 SDL_MapRGBA(const SDL_PixelFormat *format, Uint8 r, Uint8 g, Uint8 b, Uint8 a)
{
    if (!format) {
        SDL_InvalidParamError("format");
        return 0;
    }
    if (!format->palette) {
        return (r >> format->Rloss) << format->Rshift | (g >> format->Gloss) << format->Gshift | (b >> format->Bloss) << format->Bshift | ((Uint32)(a >> format->Aloss) << format->Ashift & format->Amask);
    } else {
        return SDL_FindColor(format->palette, r, g, b, a);
    }
}
```

src_c/_pygame.h
Starbuck5: Good spot.

src_c/_pygame.h
ankith26: Oh yeah, that makes sense. I didn't realise the fact that the palette is part of the format in SDL2

src_c/mask.c
```diff
@@ -947,47 +956,57 @@ palette_colors - this only affects surfaces with a palette
...
     Uint32 the_color, the_color2, rmask, gmask, bmask, rmask2, gmask2, bmask2;
     Uint8 *pix;
     Uint8 r, g, b, a;
     Uint8 tr, tg, tb, ta;
     int bpp1, bpp2;
+    PG_PixelFormat *format, *format2;
+    SDL_Palette *palette, *palette2;
+
+    if (!PG_GetSurfaceDetails(surf, &format, &palette)) {
+        PyErr_SetString(pgExc_SDLError, SDL_GetError());
```
ankith26, line 975: This function is called with GIL released, cannot set python exception here like this.

src_c/mask.c
```diff
@@ -947,47 +956,57 @@ palette_colors - this only affects surfaces with a palette
...
-    bpp1 = PG_SURF_BytesPerPixel(surf);
+    rloss = PG_FORMAT_R_LOSS(format);
+    gloss = PG_FORMAT_G_LOSS(format);
+    bloss = PG_FORMAT_B_LOSS(format);
+    bpp1 = PG_FORMAT_BytesPerPixel(format);
 
     if (surf2) {
-        format2 = surf2->format;
+        if (!PG_GetSurfaceDetails(surf2, &format2, &palette2)) {
+            PyErr_SetString(pgExc_SDLError, SDL_GetError());
```
ankith26, line 992: Same as above. We can just call this same line in the caller though after the function exits and the GIL is reheld

Starbuck5: Damn, missed that. Sort of a bummer, I would rather have more of the internal funcs call this rather than overloading all the function calls with a bunch of extra parameters. Anyways, reworked and pushed new commit. I would like to squash the two mask.c commits together before merge, but leaving it on its own now so you can see what I changed.

ankith26: The new changes LGTM, you can go ahead with the squashing

src_c/mask.c
```diff
@@ -1110,11 +1117,14 @@ mask_from_threshold(PyObject *self, PyObject *args, PyObject *kwargs)
...
     pgMaskObject *maskobj = NULL;
     SDL_Surface *surf = NULL, *surf2 = NULL;
+    PG_PixelFormat *surf_format, *surf2_format = NULL;
+    SDL_Palette *surf_palette;
     PyObject *rgba_obj_color, *rgba_obj_threshold = NULL;
     Uint8 rgba_threshold[4] = {0, 0, 0, 255};
     Uint32 color;
     Uint32 color_threshold;
     int palette_colors = 1;
+    int threshold_ret = 0;
```
ankith26, line 1127: This is an unneeded variable now.

src_c/mask.c
```diff
@@ -1158,15 +1179,20 @@ mask_from_threshold(PyObject *self, PyObject *args, PyObject *kwargs)
...
 
     pgSurface_Unlock(surfobj);
     if (surfobj2) {
         pgSurface_Unlock(surfobj2);
     }
 
+    // Python exception already set.
+    if (threshold_ret < 0) {
+        return NULL;
+    }
```
ankith26, line 1194: This is dead code as well

ankith26 (approved):
LGTM, thanks for the PR! Left a few more minor reviews, but this PR should be good to go in after those reviews are addressed.

src_c/mask.c
Starbuck5: Oops, missed those in my restructure. Thanks for finding.

MyreMylar (approved):
OK, LGTM. Changes make sense - especially now we've seen these sorts of pixel format changes a few times now. It all compiles and runs the tests fine locally as well 👍

PR #3309, opened 2025-01-26
Pixelarray SDL3

Starbuck5:

This PR is those macros + the pixelarray module.

### DIFF
--- a/src_c/pixelarray.c
+++ b/src_c/pixelarray.c
@@ -23,6 +23,7 @@
 #include "pygame.h"
 
 #include "pgcompat.h"
+#include <stddef.h>
 
 #include "doc/pixelarray_doc.h"
 
@@ -971,6 +972,13 @@ _array_assign_array(pgPixelArrayObject *array, Py_ssize_t low, Py_ssize_t high,
         return -1;
     }
 
+    PG_PixelFormat *surf_format = PG_GetSurfaceFormat(surf);
+    PG_PixelFormat *val_surf_format = PG_GetSurfaceFormat(val_surf);
+    if (surf_format == NULL || val_surf_format == NULL) {
+        PyErr_SetString(pgExc_SDLError, SDL_GetError());
+        return -1;
+    }
+
     /* If we reassign the same array, we need to copy the pixels
      * first. */
     if (SURFACE_EQUALS(array, val)) {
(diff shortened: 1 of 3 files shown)


src_c/pixelarray_methods.c
```diff
@@ -28,6 +28,18 @@
...
+    if (!ppa) {                                                      \
+        _sA = 255;                                                   \
+    }
+
+#define PXM_GET_PIXELVALS_1(sr, sg, sb, sa, _src, _palette) \
+    sr = _palette->colors[*((Uint8 *)(_src))].r;            \
+    sg = _palette->colors[*((Uint8 *)(_src))].g;            \
+    sb = _palette->colors[*((Uint8 *)(_src))].b;            \
+    sa = 255;
+
```
Starbuck5, line 42: Once surface.h is ported over (where the non PXM_ variants of these macros live), these should be removed and everyone should use the one consistent set.

src_c/pixelarray_methods.c
```diff
@@ -828,13 +859,12 @@ _compare(pgPixelArrayObject *array, PyObject *args, PyObject *kwds)
         return 0;
     }
 
-    new_format = surf->format;
```
Starbuck5, line 831: This function goes from 3 pixel formats to 2. `new_format == format` as the code was written previously.

Should it actually be a format derived from new_surf instead of surf? Was the previous implementation a typo?

I'm not fully sure, but this translates it as was written.

src_c/pixelarray.c
```diff
@@ -1021,20 +1029,23 @@ _array_assign_array(pgPixelArrayObject *array, Py_ssize_t low, Py_ssize_t high,
             }
             break;
         case 3: {
+// Note:
+// Why is the 24 bit case pixelformat aware but none of the rest are?
+// - Starbuck, jan. 2025
```
ankith26, line 1034: How are the rest of the paths not pixelformat aware? I think they are assuming that source and destination pixelformats are same and therefore directly copying over the data. But I didn't find this check enforced in the code anywhere on a quick search?

Starbuck5:
Rebased so it uses latest SDL2/SDL3 updates. (has it on the branch, so you don't need to swap dep versions between main and this).

src_c/pixelarray.c
Starbuck5: I just mean they don't take into account the pixelformat. As you say, it looks like they just assume the source and destination are the same. Going through the code we have a lot of pixel stuff in pygame-ce and lots of it has little suspicious items like that.

src_c/pixelarray_methods.c
ankith26: Can this be made into a code comment?

Starbuck5: I missed this comment or I would've updated sooner. Good idea. Done.

damusss (approved):
Pretty understandable, and LGTM from what I can understand. Thanks for making it compile :) 👍

src_c/pixelarray.c
ankith26: oh well, guess that can be addressed in future PRs, it shouldn't block this one

ankith26 (approved):
LGTM, thanks for the PR. :tada: 

IG this PR touches on some confusing code, but overall everything make sense so I am approving.

PR #3310, opened 2025-01-26
Move Surface towards compiling w/ SDL3

Starbuck5:

This PR is those macros + the low hanging fruit in surface.c.

This is my take on a #3215 PR.

### DIFF
--- a/src_c/surface.c
+++ b/src_c/surface.c
@@ -712,7 +712,6 @@ static PyObject *
 surf_get_at(PyObject *self, PyObject *position)
 {
     SDL_Surface *surf = pgSurface_AsSurface(self);
-    SDL_PixelFormat *format = NULL;
     Uint8 *pixels = NULL;
     int x, y;
     Uint32 color;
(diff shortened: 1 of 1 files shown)


MyreMylar (approved):
OK, LGTM. On theme with the other SDL3 pixel format conversion PRs. The code makes sense to me and the tests pass locally.

src_c/surface.c
```diff
@@ -775,7 +780,6 @@ surf_set_at(PyObject *self, PyObject *const *args, Py_ssize_t nargs)
 {
```
Starbuck5, line 780: If anyone is worried about performance impact I used set_at here as a test of overhead.

Used this script
```py
import time
import pygame

pygame.init()

screen = pygame.Surface((300,300))

start = time.time()
for _ in range(1000000):
    screen.set_at((4,35), 0x12345678)
print(time.time() - start)
```

I saw no appreciable difference in runtime between main and this branch.

src_c/surface.c
```diff
@@ -823,6 +836,8 @@ surf_set_at(PyObject *self, PyObject *const *args, Py_ssize_t nargs)
             break;
         case 3:
             byte_buf = (Uint8 *)(pixels + y * surf->pitch) + x * 3;
+            // Shouldn't this be able to happen without awareness of shifts?
+            // mapped color -> pixel and all.
```
ankith26, line 840: I think this cannot be done because we don't have a `Uint24` type, as most common architectures likely does not support 3-byte mov as a single instruction. So we have to break down the mapped color object back to 1-byte components for the assignment.

Starbuck5:
Rebased so it uses latest SDL2/SDL3 updates.

ankith26 (approved):

damusss (approved):
That's a lot of work, thanks :)

PR #3314, opened 2025-01-27
Move transform towards compiling w/ SDL3

Starbuck5:
#3310 but for transform.

### DIFF
--- a/src_c/scale2x.c
+++ b/src_c/scale2x.c
@@ -73,7 +73,7 @@ scale2x(SDL_Surface *src, SDL_Surface *dst)
     const int height = src->h;
 
 #if SDL_VERSION_ATLEAST(3, 0, 0)
-    const Uint8 Bpp = src->format->bytes_per_pixel;
+    const Uint8 Bpp = SDL_BYTESPERPIXEL(src->format);
 #else
     const Uint8 Bpp = src->format->BytesPerPixel;
 #endif
--- a/src_c/transform.c
+++ b/src_c/transform.c
@@ -131,9 +131,26 @@ newsurf_fromsurf(SDL_Surface *surf, int width, int height)
 
     /* Copy palette, colorkey, etc info */
     if (SDL_ISPIXELFORMAT_INDEXED(PG_SURF_FORMATENUM(surf))) {
-        if (SDL_SetPaletteColors(newsurf->format->palette,
-                                 surf->format->palette->colors, 0,
-                                 surf->format->palette->ncolors) != 0) {
+        SDL_Palette *newsurf_palette = PG_GetSurfacePalette(newsurf);
+        SDL_Palette *surf_palette = PG_GetSurfacePalette(newsurf);
+
+        if (newsurf_palette == NULL) {
+            PyErr_SetString(
+                pgExc_SDLError,
+                "Palette expected (newsurf) but no palette found.");
+            SDL_FreeSurface(newsurf);
+            return NULL;
+        }
+
+        if (surf_palette == NULL) {
+            PyErr_SetString(pgExc_SDLError,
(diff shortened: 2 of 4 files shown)


src_c/transform.c
```diff
@@ -2358,16 +2409,18 @@ solid_overlay(pgSurfaceObject *srcobj, Uint32 color, pgSurfaceObject *dstobj,
...
         for (y = 0; y < src->h; y++) {
             for (x = 0; x < src->w; x++) {
-                SDL_GetRGBA((Uint32)*src_row, fmt, &r, &g, &b, &a);
+                PG_GetRGBA((Uint32)*src_row, fmt, src_palette, &r, &g, &b, &a);
 
                 if (a) {
                     if (keep_alpha)
-                        *dst_row = (Uint16)SDL_MapRGBA(fmt, Cr, Cg, Cb, a);
+                        *dst_row = (Uint16)PG_MapRGBA(
+                            newsurf_format, newsurf_palette, Cr, Cg, Cb, a);
```
Starbuck5, line 2423: So previously the `color16=` and `*dst_row = (Uint16)SDL_MapRGBA` lines here used the `fmt` variable, which represents the format of the surface "src".

However, these values are being mapped to "newsurf", not "src". So why is it written like this?

Well, the format enums are checked earlier, so the author of this code probably thought it made no difference.

However I can think of a case where passing in a destination surface with a custom palette into this would be different.

So I've changed the behavior here very slightly to be more correct, by mapping colors using the data of the surface being mapped to (newsurf). I don't think anyone will notice, I feel like this a very weird edge case, and this is a very new function as well. Regardless, if this change complicates this PR too much I can pull it out into a different PR or drop it.

src_c/transform.c
```diff
@@ -2605,7 +2660,7 @@ modify_hsl(SDL_Surface *surf, SDL_Surface *dst, float h, float s, float l)
                 }
 
                 HSL_to_RGB(s_h, s_s, s_l, &r, &g, &b);
-                pixel = SDL_MapRGBA(fmt, r, g, b, a);
+                pixel = PG_MapRGBA(dst_format, dst_palette, r, g, b, a);
```
Starbuck5, line 2663: There is a similar situation here. Previously, it was using fmt, which represents the other surface, not the surface being written to.

Starbuck5:
Rebased so it uses latest SDL2/SDL3 updates. (has it on the branch, so you don't need to swap dep versions between main and this).

src_c/scale2x.c
```diff
@@ -73,7 +73,7 @@ scale2x(SDL_Surface *src, SDL_Surface *dst)
     const int height = src->h;
 
 #if SDL_VERSION_ATLEAST(3, 0, 0)
-    const Uint8 Bpp = src->format->bytes_per_pixel;
+    const Uint8 Bpp = SDL_BYTESPERPIXEL(src->format);
```
ankith26, line 76: minor nit: We can use `PG_SURF_BytesPerPixel` here to cover for this whole block and reduce the loc.

src_c/transform.c
ankith26: I think this fix is good, it can be a part of this PR

src_c/scale2x.c
Starbuck5: Good spot, but this file doesn't include _pygame.h. So rather than messing with the includes I chose to to a manual fix here.

ankith26: Okay, fair enough then

ankith26 (approved):

damusss (approved):
Looks like some good progress, thanks! 💯

oddbookworm:
@coderabbitai review this pull request

PR #3317, opened 2025-01-28
Add interface for _sdl2.video classes

MightyJosip:
I imagined this as the first PR towards the goal of rewriting _sdl2.video stuff in C code. Since those classes are used in combination with each other, I wanted to add them all, so then for example when I will write the implementation for Renderer, I can do it without any compilation errors where I am missing the declaration for Texture. At this moment, this code itself doesn't do anything, you can write
```py
import pygame
import pygame._render

window = pygame.Window()
renderer = pygame._render.Renderer(window)
texture = pygame._render.Texture(renderer)
image = pygame._render.Image(texture)

print(renderer)
print(texture)
print(image)

```
for example, and you get 
```
<pygame._render.Renderer object at 0x0000026FE34668E0>
<pygame._render.Texture object at 0x0000026FE3467FC0>
<pygame._render.Image object at 0x0000026FE34966F0>
```
as the output. But that is everything that has been implemented. Also, I didn't include `_render` in pygame.__init__, so you need to manually import it, otherwise it won't work, so at the moment, this module is basically hidden, just so it doesn't break anything


Starbuck5:
Hey, good to see this moving forward, this is important for us to do!

Some organizational things:
I'm surprised to see the 3 classes all getting their own module. I would think they would all be part of one module, as they are currently. (You could still have separate files for implementation and have them in one module).

Also, I would think these new files would live in the _sdl2 folder. I believe when Window was ported it went there first.

damusss:
I agree that they should live in one module. a good name would be `pygame.renderer` (privatized at the moment) in my opinion. Just like pygame.gpu should hold all gpu features in the future. it makes so things that are directly related to each other and that depend on each other under the same name (you can't use a texture without a renderer, but you can use a window with regular surfaces, you get the idea)

MightyJosip:
Yea you are correct about the multiple modules, I will add it in a single module. Is there an example of that in the repo at this moment, I can probably do that with the multiple header files, but if we already have more elegant solution I would like to use it, probably call it pygame._renderer like damusss said.

About _sdl2, I am pretty sure Window was added outside of it in the first commit https://github.com/pygame-community/pygame-ce/pull/2114/files#diff-3279c6977adb71e8d262e5974889267e57e33c5ef0d5c76a0146999fba0b40c6

Starbuck5:
The code looks fine reading through, but I disagree with damus, this should be called pygame._render.

"Render" is what SDL2 and SDL3 both call it. https://wiki.libsdl.org/SDL2/CategoryRender, https://wiki.libsdl.org/SDL3/CategoryRender

Starbuck5 (approved):
OK! looks good.

damusss (approved):
I thought this was super big, instead just the interface (who could have guessed, reading the title?). LGTM, Thanks!

PR #3321, opened 2025-02-01
Add `special_flags` to `Group.draw()`

### LINKED ISSUE #3306: Allow passing `special_flags` to `Group.draw`
**Description**
Add a `special_flags` argument to `Group.draw` (and all the other group classes).
This would be useful as with this feature, I could write code like this:
```python
group.draw(surface, special_flags=pygame.BLEND_RGBA_MAX)
```
instead of
```python
for s in group.sprites():
    surface.blit(s.image, s.rect, special_flags=pygame.BLEND_RGBA_MAX)
```
Also, for more complicated group classes (eg. `RenderUpdates`), you have to copy the entire code of the `.draw` functions and change just the line where the `.blit` method is called which is quite annoying.

See pygame/pygame#3719 for the issue I opened over there (before I found pygame-ce).

MarcellPerger1:
Add `special_flags` to `Group.draw()`, fixes #3306.
I implemented this in the same way as in pygame/pygame#3722 for compatibility with pygame/pygame:
- I added `bgsurf` as a (sometimes unused) 2nd argument to to all the `Group.draw()` methods so that `special_flags` is consistently the 3rd argument across all subclasses.
- For `LayeredUpdates.draw()`, `special_flags` (if provided as non-`None`) overrides the `blendmode` of sprites.

In the docs, I haven't added the `versionchanged` because I don't know if the next version (or version where this will be merged) will be `2.5.3` or `2.6.0`.

I haven't used `fblits` in this PR as I thought it might be better to do that in separate PR (let me know if I should do it in this PR)


Starbuck5:
> In the docs, I haven't added the versionchanged because I don't know if the next version (or version where this will be merged) will be 2.5.3 or 2.6.0.

2.5.4, as of today :)

docs/reST/ref/sprite.rst
```diff
@@ -551,10 +555,11 @@ Sprites are not thread safe. So lock them yourself if using threads.
    .. method:: draw
 
       | :sl:`draw all sprites in the right order onto the passed surface.`
-      | :sg:`draw(surface, bgd=None) -> Rect_list`
+      | :sg:`draw(surface, bgsurf=None, special_flags=0) -> Rect_list`
```
aatle, line 558: `special_flags` should default to `None`.

aatle:
The feature is nice, but the implementation looks strange (it can't be changed though, for `pygame` compatibility).
Rather than adding a confusing unused `bgsurf` argument for all sprite groups to keep consistency with a single, rarely used `LayeredDirty` subclass, it would have been better to make `special_flags` keyword-only. Every code example given has used it as a keyword argument.
I also don't see why has `bgd` been renamed to `bgsurf`. This is technically a breaking change for `LayeredDirty` (though it probably won't have much impact).

MarcellPerger1:
> I also don't see why has `bgd` been renamed to `bgsurf`. This is technically a breaking change for `LayeredDirty` (though it probably won't have much impact).

The only reason I changed it here is for compatibility with `pygame`, although it might be better just to keep it as `bgd` for compatibility with previous versions of `pygame-ce` (personally, I think compatibility with previous versions of `pygame-ce` is more important than compatibility with `pygame`).
I can change it back to `bgd` if you think that would be better.

Starbuck5:
> (personally, I think compatibility with previous versions of pygame-ce is more important than compatibility with pygame).

I agree.

MarcellPerger1:
I've changed `bgsurf` back to `bgd`. Do I need to do anything else to help get this merged?

ankith26 (approved):
I'm not super familiar with sprite stuff, but this PR LGTM at a glance.

Thanks for contributing to pygame-ce! :partying_face:

Starbuck5:
@aatle if this looks okay to you I would be okay with merging.

src_py/sprite.py
```diff
@@ -1211,8 +1229,11 @@ def draw(self, surface, bgd=None):  # noqa pylint: disable=arguments-differ; una
         return local_ret
 
     @staticmethod
-    def _draw_dirty_internal(_old_rect, _rect, _sprites, _surf_blit, _update):
+    def _draw_dirty_internal(
+        _old_rect, _rect, _sprites, _surf_blit, _update, _special_flags
+    ):
         for spr in _sprites:
+            flags = spr.blendmode if _special_flags is None else _special_flags
```
aatle, line 1236: It is not possible for `_special_flags` to be `None` because it is already checked for in `draw()`.

MarcellPerger1: I think it is possible for it to be `None` here because when the code checks for `None` in `draw()`, the result is assigned to a different variable (`flags`) and isn't assigned back to the `special_flags` variable.
I implemented it in this way because if `special_flags` isn't specified, it defaults back to `0` when drawing the background but defaults to `spr.blendmode` when drawing each sprite.

aatle: I see.
It is a bit confusing with all of the flags. Perhaps rename the background flags from `flags` to `bgd_flags`?

aatle (approved):
It looks good to me.
Thanks!

Starbuck5 (approved):

PR #3322, opened 2025-02-01
General maintenance: Minor dev.py fixes, SDL(2+3) and deps updates

ankith26:
Some general maintenance updates:
- Fixes minor `dev.py` issues.
- Build deps updates (meson and uv)
- mac/manylinux deps updates (and also cleanups - I have removed the scripts for modules we don't use anymore)


Starbuck5 (approved):
Reviewing this from my phone, so it would be pretty hard to check- how does this change the output size of the wheels?

Anyways all the code changes look good, thanks Ankith.

ankith26:
In the previous commit, I see no significant increase in wheel size (checked the combined size of all linux wheels reported on github UI)
I took the opportunity to also update SDL 2 and 3 to the latest patch releases in this PR in the latest commit

Starbuck5 (approved):

PR #3325, opened 2025-02-02
Add `pygame.system.get_theme()` (w/ sdl3)

damusss:
I implemented fully what the title says, "fixing" half of https://github.com/pygame-community/pygame-ce/issues/2760#issuecomment-2496022434

The codebase doesn't compile with sdl3 yet, so I'm not sure what kind of versionadded to put.

Depending on which PR gets in before compared to https://github.com/pygame-community/pygame-ce/pull/3296 the other will have to be updated I guess.


Starbuck5:
I don't think we're ready to have PRs like this yet. We don't have any branches, any releases targeting SDL3 right now.

We need to get stuff compiling on SDL3 first.

Let's say this merges, now everyone who reads the system docs sees a function that they have no way of using. And the description doesn't even say that this function is impossible to use, it only says it requires SDL3, which is not a statement most of our users will understand the implications of.

damusss:
@Starbuck5 don't worry I'm aware that it will take a while for it to get in, i just following roughly the indications ankith gave me so not to have too many stagnant branches in my fork. btw what % are we in sdl3 porting?

Starbuck5:
> btw what % are we in sdl3 porting?

That's a difficult question. Maybe 50% ?

damusss:
I was lazy to fix the conflicts so I created a new branch instead. After receiving a design direction I'll open a new pull request.

PR #3326, opened 2025-02-03
Port font to SDL3(_ttf)

ankith26:
SDL3_ttf preview release: https://github.com/libsdl-org/SDL_ttf/releases/tag/preview-3.1.0

With this PR `pygame.font` compiles on SDL3(_ttf)


damusss (changes requested):
A lot if not all TTF functions in sdl3 (I assume?) have 0 on success and -1 on error, like `TTF_SizeText` or `TTF_Init`, so checking `if (TTF_X) {error}` is actually wrong and should be `if (TTF_X < 0) {error}`. I know this PR is supposed to compile it, but without this an extra PR can't be avoided to fix this so you might aswell change it now.

damusss:
> Correction: SDL3 uses more of the bool type and in many places it returns `true` (1) to indicate success and `false` (0) for failure. In SDL2 it is (0) for success (no error) and (-1) for failure, so `if (SDL_Func() < 0) {err}` and `if (SDL_Func()) {err}` works for SDL2 and `if (!SDL_Func()) {err}` for SDL3, so there's no version that would work for both SDL2 and SDL3, different code paths are needed.

I said my 0 and -1 returns while looking at my IDE docstrings, but my IDE was actually looking at SDL2 functions, so no < 0 is incorrect and will never raise (that's why those bug errors disappeared) so yeah a not operator is needed like you said.

src_c/font.c
```diff
@@ -683,15 +726,24 @@ font_size(PyObject *self, PyObject *text)
             return NULL;
         }
         string = PyBytes_AS_STRING(bytes);
+#if SDL_TTF_VERSION_ATLEAST(3, 0, 0)
+        ecode = TTF_GetStringSize(font, string, 0, &w, &h);
```
gresm, line 730:
```suggestion
        ecode = !TTF_GetStringSize(font, string, 0, &w, &h);
```

gresm, line 730:
```suggestion
        ecode = !TTF_GetStringSize(font, string, 0, &w, &h);
```

src_c/font.c
```diff
@@ -683,15 +726,24 @@ font_size(PyObject *self, PyObject *text)
...
         Py_DECREF(bytes);
         if (ecode) {
             return RAISE(pgExc_SDLError, TTF_GetError());
         }
     }
     else if (PyBytes_Check(text)) {
         string = PyBytes_AS_STRING(text);
-        if (TTF_SizeText(font, string, &w, &h)) {
+#if SDL_TTF_VERSION_ATLEAST(3, 0, 0)
+        if (TTF_GetStringSize(font, string, 0, &w, &h))
```
gresm, line 742:
```suggestion
        if (!TTF_GetStringSize(font, string, 0, &w, &h))
```

src_c/font.c
```diff
@@ -878,13 +950,19 @@ font_metrics(PyObject *self, PyObject *textobj)
...
-         * TTF_GlyphMetrics() seems to return a value for any character,
-         * using the default invalid character, if the char is not found.
-         */
+/* TODO:
+ * TTF_GlyphMetrics() seems to return a value for any character,
+ * using the default invalid character, if the char is not found.
+ */
+#if SDL_TTF_VERSION_ATLEAST(3, 0, 0)
+        if (!surrogate && /* conditional and */
+            !TTF_GetGlyphMetrics(font, (Uint16)ch, &minx, &maxx, &miny, &maxy,
```
gresm, line 959:
```suggestion
            TTF_GetGlyphMetrics(font, ch, &minx, &maxx, &miny, &maxy,
```
Is the SDL error set if the call fails. If yes, should it be cleared? The cast to (Uint16) seem to be pointless. Also in SDL3, Uint32 is used, in SDL2 there's ``TTF_GlyphMetrics32``. Is there any reason to use (Uint16) instead of (Uint32)?

gresm (changes requested):
Some things that I've caught, definitely not checked exhaustively.

src_c/font.c
ankith26: Good catch, thanks for pointing it out!

src_c/font.c
ankith26: Well, `TTF_GlyphMetrics32` was added in SDL 2.0.18, while our minimum supported version at the moment is SDL 2.0.15. The fix for this is a bit more involved and probably out of scope of this PR anyways, but I will look into it in another PR

src_c/font.c
```diff
@@ -935,7 +1021,13 @@ font_set_script(PyObject *self, PyObject *arg)
                      "script code must be exactly 4 characters");
     }
 
-    if (TTF_SetFontScriptName(font, script_code) < 0) {
+#if SDL_TTF_VERSION_ATLEAST(3, 0, 0)
+    if (!TTF_SetFontScript(font, HB_TAG(script_code[0], script_code[1],
+                                        script_code[2], script_code[3])))
+#else
+    if (TTF_SetFontScriptName(font, script_code) < 0)
```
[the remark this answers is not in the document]
Starbuck5: Opened https://github.com/libsdl-org/SDL_ttf/issues/505

src_c/font.c
```diff
@@ -878,13 +955,19 @@ font_metrics(PyObject *self, PyObject *textobj)
...
         ch = buffer[i];
         surrogate = Py_UNICODE_IS_SURROGATE(ch);
-        /* TODO:
-         * TTF_GlyphMetrics() seems to return a value for any character,
-         * using the default invalid character, if the char is not found.
-         */
+/* TODO:
+ * TTF_GlyphMetrics() seems to return a value for any character,
+ * using the default invalid character, if the char is not found.
+ */
```
Starbuck5, line 961: very minor thing but if you left this comment indented how it was before this wouldn't be part of the diff.

src_c/font.c
```diff
@@ -912,6 +995,14 @@ font_metrics(PyObject *self, PyObject *textobj)
     return list;
 }
 
+/* This is taken from the harfbuzz header file. It coverts script name in the
```
Starbuck5, line 998: typo: converts

ankith26: Good catch, fixed it

ankith26:
Just force pushed with fixes to your 2 review points, without any other changes.

damusss (approved):
Alright, looking good now, thanks :3

PR #3327, opened 2025-02-03
Port Renderer to C code

MightyJosip:
This PR converts pygame._sdl2.video.Renderer to C code. It has the implementation of all the Renderer methods and attributes. It doesn't have docs/type hints/tests or stuff like that.


Starbuck5:
Pulled the relevant bit of logs for the WASM build failure:

```
src_c/render.c:412:19: error: call to undeclared function 'pgSurface_New'; ISO C99 and later do not support implicit function declarations [-Wimplicit-function-declaration]
        surface = pgSurface_New(surf);
                  ^
src_c/render.c:412:19: note: did you mean 'pgSurface_New2'?
src_c/surface.c:343:1: note: 'pgSurface_New2' declared here
pgSurface_New2(SDL_Surface *s, int owner)
^
In file included from src_c/static.c:372:
src_c/render.c:412:17: error: incompatible integer to pointer conversion assigning to 'pgSurfaceObject *' from 'int' [-Wint-conversion]
        surface = pgSurface_New(surf);
                ^ ~~~~~~~~~~~~~~~~~~~
```

src_c/render.c
```diff
@@ -12,9 +12,673 @@ static PyTypeObject pgTexture_Type;
...
+static void
+image_renderer_draw(pgImageObject *self, PyObject *area, PyObject *dest);
+
+/* Renderer implementation */
+static PyObject *
+from_window(PyTypeObject *cls, PyObject *args, PyObject *kwargs)
+{
+    PyObject *window;
+    static char *keywords[] = {"window", NULL};
+    if (!PyArg_ParseTupleAndKeywords(args, kwargs, "O", keywords, &window)) {
```
Starbuck5, line 45: Would it work to use "O!" here to only accept Window objects, and have the type error handled already?

src_c/render.c
```diff
@@ -12,9 +12,673 @@ static PyTypeObject pgTexture_Type;
...
+
+static void
+texture_renderer_draw(pgTextureObject *self, PyObject *area, PyObject *dest);
+
+static void
+image_renderer_draw(pgImageObject *self, PyObject *area, PyObject *dest);
+
+/* Renderer implementation */
+static PyObject *
+from_window(PyTypeObject *cls, PyObject *args, PyObject *kwargs)
```
Starbuck5, line 41: I noticed the static methods `from_window` and `compose_custom_blend_mode` aren't prefaced with renderer, unlike the others. I'd think it's still worth starting with renderer, because they are renderer methods still, even if they're not instance methods.

src_c/render.c
```diff
@@ -12,9 +12,673 @@ static PyTypeObject pgTexture_Type;
...
+    pgWindowObject *window;
+    int index = -1;
+    int accelerated = -1;
+    int vsync = 0;
+    int target_texture = 0;
+    Uint32 flags = 0;
+
+    char *keywords[] = {"window", "index",          "accelerated",
+                        "vsync",  "target_texture", NULL};
+    if (!PyArg_ParseTupleAndKeywords(args, kwargs, "O|iipp", keywords, &window,
```
Starbuck5, line 580: As far as I can tell you're assuming the argument that comes in here is a pgWindowObject. This could use "O!" as well?

src_c/render.c
```diff
@@ -12,9 +12,673 @@ static PyTypeObject pgTexture_Type;
...
+}
+
+static PyObject *
+compose_custom_blend_mode(PyObject *self, PyObject *args, PyObject *kwargs)
+{
+    PyObject *color_mode, *alpha_mode;
+    int mode[6];
+    SDL_BlendMode blend_mode;
+    static char *keywords[] = {"color_mode", "alpha_mode", NULL};
+    if (!PyArg_ParseTupleAndKeywords(args, kwargs, "OO", keywords, &color_mode,
```
Starbuck5, line 330: Instead of "OO" could this be "(iii)(iii)" ?

src_c/render.c
```diff
@@ -12,9 +12,673 @@ static PyTypeObject pgTexture_Type;
...
+        if (draw_method && PyCallable_Check(draw_method)) {
+            PyObject_CallMethodObjArgs(sourceobj, PyUnicode_FromString("draw"),
+                                       areaobj, destobj, NULL);
+            Py_DECREF(draw_method);
+        }
+        else {
+            return RAISE(PyExc_AttributeError, "source.draw is not callable");
+        }
+    }
+    else {
```
Starbuck5, line 443: I feel like this code could be simpler.
It checks a property, grabs the function, checks if its a callable, then calls it by looking it up again. -- Wouldn't PyUnicode_FromString("draw") leak?

If you're going to check that it's callable first you should use https://docs.python.org/3/c-api/call.html#c.PyObject_CallFunctionObjArgs because you've already looked it up you don't need to look it up again.

But I'm sorta thinking instead of fetching it, checking it, and then calling it you could just try to call it? Let Python do those checks for you.

src_c/render.c
```diff
@@ -12,9 +12,673 @@ static PyTypeObject pgTexture_Type;
...
+}
+
+static PyObject *
+renderer_get_target(pgRendererObject *self, void *closure)
+{
+    if (self->target == NULL) {
+        Py_RETURN_NONE;
+    }
+    Py_INCREF(self->target);
+    return (PyObject *)self->target;
```
Starbuck5, line 543: If you stored the None object instead of NULL you could have a unified codepath here. Not sure if that makes it harder other places though.

src_c/render.c
```diff
@@ -12,9 +12,673 @@ static PyTypeObject pgTexture_Type;
...
+    else {
+        SDL_RenderGetViewport(self->renderer, rect);
+        areaparam = NULL;
+    }
+    if (!Py_IsNone(surfobj)) {
+        if (!(pgSurface_Check(surfobj))) {
+            return RAISE(PyExc_TypeError, "surface must be None or a Surface");
+        }
+        surface = (pgSurfaceObject *)surfobj;
+        Py_INCREF(surface);
```
Starbuck5, line 389: I'm suspicious of this incref. We're not taking ownership of the surface, we just are using it randomly in this function.

MightyJosip: It is needed because in line 410 we return that surface. In the else branch (without passing surface as the argument to that function) it works ok, because I guess pgSurface_New() increases the ref, so I put it only in that if branch. Without it, the following code

```py
surface = pygame.Surface((100, 100))
window = pygame.Window(size=(100, 100))
renderer = _render.Renderer(window)
renderer.to_surface(surface)
print(surface)
```
segfaults on `print(surface)` line (or basically any call of the surface)

src_c/render.c
MightyJosip: Great catch, thanks, now the function looks much more beautiful (all this argument parsing below was really ugly)

Fixed

src_py/typing.py
```diff
@@ -63,6 +63,13 @@ def rect(self) -> Union["RectLike", Callable[[], "RectLike"]]: ...
 ]
 
 
+class DrawableClass(Protocol):
```
damusss, line 66: Hi. @MightyJosip . I can see that this typehint is only used in the `blit` method of `Renderer`, and nowhere else. `pygame.typing` is a public module, meaning this would become part of the public API and could hardly be removed in the future. Since it's only used once in the stubs, it could either be left here, but prefixed with an underscore or it could be copy and pasted inside `render.pyi` (with an underscore) directly, similar to what happens with surface formats for `pygame.image`. In the future, there is always the possibility of making it public, but perhaps it could stay out of scope for this PR and keep New API limited to the port.

src_c/render.c
MightyJosip: Ok I changed it, and when testing it looks kinda ok. Can you check if that is what you meant.

src_c/render.c
MightyJosip: Yea I see what you meant, and here indeed it would be better. Though the `target` in `pgRendererObject` is defined as `pgTextureObject`, and I think it is better than `PyObject` that could then accept Py_None. I would go with the current version, though if you want I can change it to `PyObject`

src_c/render.c
MightyJosip: Thanks, I realised I actually don't have any type check in this function so the program segfaults if I call it like `renderer = _render.Renderer("a")`, probably in the line 595.

Fixed

src_c/render.c
```diff
@@ -12,9 +12,673 @@ static PyTypeObject pgTexture_Type;
...
+#if SDL_VERSION_ATLEAST(2, 0, 18)
+    PyObject *p1, *p2, *p3, *p4;
+    SDL_Vertex vertices[6];
+    SDL_Color color;
+    static char *keywords[] = {"p1", "p2", "p3", "p4", NULL};
+    if (!PyArg_ParseTupleAndKeywords(args, kwargs, "OOOO", keywords, &p1, &p2,
+                                     &p3, &p4)) {
+        return NULL;
+    }
+    if (!pg_TwoFloatsFromObj(p1, &vertices[0].position.x,
```
damusss, line 244: I hope this doesn't sound nitpicky.  In this draw and fill methods depending on the amount of points there is a great repetition of the pg_TwoFloatsFromObj code with an increasing vertex index and later the repetition of vertices[index].color = color. I think both of them could be put into C loops for all of the methods. If you don't want to make a temporary array with p1, p2, p3... you can create a temporary macro before this methods and undefine it after, since that code pattern is used a lot around here. You can also ignore all of this, it's only about my biased view on code quality.

src_c/render.c
```diff
@@ -12,9 +12,640 @@ static PyTypeObject pgTexture_Type;
...
+    return pgRect_New(&rect);
+}
+
+static PyObject *
+renderer_set_viewport(pgRendererObject *self, PyObject *args, PyObject *kwargs)
+{
+    PyObject *rectobj = Py_None;
+    SDL_Rect *rect = NULL, temp;
+    static char *keywords[] = {"area", NULL};
+    if (!PyArg_ParseTupleAndKeywords(args, kwargs, "|O", keywords, &rectobj)) {
```
damusss, line 304: First of all, I think `Renderer.viewport` (property) would be much better, but if this was like this before the porting there's not much you can do.

I noticed that this method was ported with a slight modification. Before you could do either of this:
```
renderer.set_viewport(rect)
renderer.set_viewport(None)
```
But using the pipe format in `PyArg_ParseTupleAndKeywords` you are allowing this:
```
renderer.set_viewport(rect)
renderer.set_viewport(None)
renderer.set_viewport()
```
You are aware of this, because you ported this behaviour to the stubs, but I think it makes no sense to use a setter method without passing in anything. Someone could forget the argument and wonder why the viewport is gone. If someone wants to remove the viewport, they have to conciously specify it as None, the same way as in `Surface.set_clip`

damusss (dismissed):
🔽

src_c/render.c
MightyJosip: Ofc, it is not nitpicky, I am always happy for good review. Yea I agree with you, I changed it a bit, can you check if it is something you also thought or you would like some extra changes

src_c/render.c
MightyJosip: One of the changes when we will do pygame-ce2 -> pygame-ce3 could definitely be converting that to the property, I completely agree with you.

About no argument, fixed it in implementation/typing, now it requires the argument

src_c/render.c
```diff
@@ -12,9 +12,598 @@ static PyTypeObject pgTexture_Type;
...
+    }
+
+    if (pgTexture_Check(sourceobj)) {
+        texture_renderer_draw((pgTextureObject *)sourceobj, areaobj, destobj);
+    }
+    else if (pgImage_Check(sourceobj)) {
+        image_renderer_draw((pgImageObject *)sourceobj, areaobj, destobj);
+    }
+    else {
+        PyObject_CallFunctionObjArgs(PyObject_GetAttrString(sourceobj, "draw"),
```
damusss, line 368: If this function returns NULL it means there was an error. Shouldn't you check for it and return early if it's the case?

MightyJosip: Ok, changed that

damusss:
Alright, thanks. As soon as I test and confirm the ported features I'll approve it (if everything goes right)

damusss (approved):
Tested the ported functions, everything works right. Thanks!

PR #3330, opened 2025-02-07
Port Texture to C code

MightyJosip:
This PR converts pygame._sdl2.video.Texture to C code. It has the implementation of all the Texture methods and attributes.


src_c/render.c
```diff
@@ -629,12 +1234,14 @@ static PyTypeObject pgRenderer_Type = {
     .tp_getset = renderer_getset};
 
 static PyTypeObject pgTexture_Type = {
-    PyVarObject_HEAD_INIT(NULL, 0).tp_name = "pygame._render.Texture",
+    PyVarObject_HEAD_INIT(NULL, 0).tp_name = "pygame._renderer.Texture",
```
Starbuck5: It's the render module, not the renderer module

src_c/render.c
```diff
@@ -546,7 +563,568 @@ renderer_dealloc(pgRendererObject *self, PyObject *_null)
...
+            RAISERETURN(PyExc_ValueError,
+                        "only one of static, streaming, or target can be true",
+                        -1)
+        }
+        access = SDL_TEXTUREACCESS_TARGET;
+    }
+    else {
+        access = SDL_TEXTUREACCESS_STATIC;
+    }
+    self->renderer = renderer;
```
Starbuck5, line 1081: Wouldn't this need to be incref-ed?

src_c/render.c
```diff
@@ -546,7 +563,568 @@ renderer_dealloc(pgRendererObject *self, PyObject *_null)
...
+    new_texture =
+        (pgTextureObject *)(&(pgTexture_Type))->tp_alloc(&pgTexture_Type, 0);
+    surf = pgSurface_AsSurface(surfobj);
+    SURF_INIT_CHECK(surf)
+    new_texture->texture =
+        SDL_CreateTextureFromSurface(renderer->renderer, surf);
+    if (!new_texture->texture) {
+        return RAISE(pgExc_SDLError, SDL_GetError());
+    }
+    new_texture->renderer = renderer;
```
Starbuck5, line 626: Same concern here about refcounting.

src_c/render.c
```diff
@@ -546,7 +563,568 @@ renderer_dealloc(pgRendererObject *self, PyObject *_null)
...
+    SDL_BlendMode blend_mode;
+    RENDERER_ERROR_CHECK(SDL_GetTextureBlendMode(self->texture, &blend_mode));
+    return PyLong_FromLong((long)blend_mode);
+}
+
+static int
+texture_set_blend_mode(pgTextureObject *self, PyObject *arg, void *closure)
+{
+    if (PyLong_Check(arg)) {
+        long longval = PyLong_AsLong(arg);
```
Starbuck5: Technically this function could fail and return -1 with an error set, which you're not checking.

Rather than doing PyLong_Check + PyLong_AsLong you could just do PyLong_AsLong and check it succeeds?

src_c/render.c
```diff
@@ -546,7 +563,568 @@ renderer_dealloc(pgRendererObject *self, PyObject *_null)
...
+            }
+            dstrect.w = (float)self->width;
+            dstrect.h = (float)self->height;
+        }
+    }
+    if (!Py_IsNone(originobj)) {
+        if (!pg_TwoFloatsFromObj(originobj, &origin.x, &origin.y)) {
+            return RAISE(PyExc_ValueError, "origin must be a point or None");
+        }
+        originptr = &origin;
```
Starbuck5, line 679: You could pass &origin into `SDL_RenderCopyExF` directly rather than needing this extra variable. (I see the Cython has it, maybe it's a weird Cython thing or the person who wrote it wasn't as experienced with pointers).

JoKing-1999: It is used because &origin cannot be NULL, but in some cases SDL_RenderCopyExF expects the center argument to be NULL

src_c/render.c
```diff
@@ -477,12 +494,14 @@ renderer_set_target(pgRendererObject *self, PyObject *arg, void *closure)
 {
     if (Py_IsNone(arg)) {
         self->target = NULL;
+        Py_XDECREF(self->target);
```
zoldalma999: This line seems to do nothing because you set target to NULL in the line above. But I think `self->target` should be xdecref'd at the beginning of the function no matter what, since we don't use it anymore.

src_c/render.c
```diff
@@ -546,7 +565,570 @@ renderer_dealloc(pgRendererObject *self, PyObject *_null)
...
+        return NULL; /* Exception already set. */
+    }
+    new_texture =
+        (pgTextureObject *)(&(pgTexture_Type))->tp_alloc(&pgTexture_Type, 0);
+    surf = pgSurface_AsSurface(surfobj);
+    SURF_INIT_CHECK(surf)
+    new_texture->texture =
+        SDL_CreateTextureFromSurface(renderer->renderer, surf);
+    if (!new_texture->texture) {
+        return RAISE(pgExc_SDLError, SDL_GetError());
```
zoldalma999, line 621: I think new_texture should be freed before the RAISE?

src_c/render.c
```diff
@@ -546,7 +565,570 @@ renderer_dealloc(pgRendererObject *self, PyObject *_null)
...
+    vertices[0].color.b = (int)b_mod * p1_mod[2];
+    vertices[0].color.a = (int)a_mod * p1_mod[3];
+    vertices[1].color.r = (int)r_mod * p2_mod[0];
+    vertices[1].color.g = (int)g_mod * p2_mod[1];
+    vertices[1].color.b = (int)b_mod * p2_mod[2];
+    vertices[1].color.a = (int)a_mod * p2_mod[3];
+    vertices[2].color.r = (int)r_mod * p3_mod[0];
+    vertices[2].color.g = (int)g_mod * p3_mod[1];
+    vertices[2].color.b = (int)b_mod * p3_mod[2];
+    vertices[2].color.a = (int)a_mod * p3_mod[3];
```
zoldalma999: Casting (to int) happens before the multiplication, so `(int)r_mod` will be 0 unless that channel has 255 as a color mod.

buildconfig/stubs/pygame/__init__.pyi
```diff
@@ -204,6 +204,9 @@ from .constants import (
     FLASH_BRIEFLY as FLASH_BRIEFLY,
     FLASH_CANCEL as FLASH_CANCEL,
     FLASH_UNTIL_FOCUSED as FLASH_UNTIL_FOCUSED,
+    FLIP_HORIZONTAL as FLIP_HORIZONTAL,
+    FLIP_NONE as FLIP_NONE,
+    FLIP_VERTICAL as FLIP_VERTICAL,
```
zoldalma999, line 209: Do we need to expose these constants? As far as I can tell, these are always handled by keyword arguments like `flip_x` and `flip_y`

src_c/render.c
```diff
@@ -546,7 +565,570 @@ renderer_dealloc(pgRendererObject *self, PyObject *_null)
...
+    vertices[3].color = vertices[2].color;
+    vertices[4].color.r = (int)r_mod * p4_mod[0];
+    vertices[4].color.g = (int)g_mod * p4_mod[1];
+    vertices[4].color.b = (int)b_mod * p4_mod[2];
+    vertices[4].color.a = (int)a_mod * p4_mod[3];
+    vertices[5].color = vertices[0].color;
```
zoldalma999: Same issue here with the casting

src_c/render.c
```diff
@@ -35,6 +35,23 @@ static PyTypeObject pgImage_Type;
...
+        a = 255;                                                         \
+    }                                                                    \
+    else {                                                               \
+        if (!pg_IntFromObjIndex(obj, 0, &r) ||                           \
+            !pg_IntFromObjIndex(obj, 1, &g) ||                           \
+            !pg_IntFromObjIndex(obj, 2, &b)) {                           \
+            return RAISE(PyExc_TypeError, "invalid " #name " argument"); \
+        }                                                                \
+        if (PySequence_Size(obj) == 4)                                   \
+            pg_IntFromObjIndex(obj, 3, &a);                              \
```
zoldalma999: This block should probably use `pg_RGBAFromObjEx`.

src_c/render.c
```diff
@@ -546,7 +565,570 @@ renderer_dealloc(pgRendererObject *self, PyObject *_null)
...
+    }
+    else {
+        access = SDL_TEXTUREACCESS_STATIC;
+    }
+    self->renderer = renderer;
+    Py_XINCREF(self->renderer);
+    self->texture =
+        SDL_CreateTexture(renderer->renderer, format, access, width, height);
+    if (!self->texture) {
+        RAISERETURN(PyExc_ValueError, "Debug?", -1)
```
zoldalma999: Left in debug message?

src_c/render.c
```diff
@@ -546,7 +565,570 @@ renderer_dealloc(pgRendererObject *self, PyObject *_null)
...
+        if (!(srcrectptr = pgRect_FromObject(area, &srcrect))) {
+            PyErr_SetString(PyExc_ValueError,
+                            "srcrect must be a Rect or None");
+        }
+    }
+    if (!Py_IsNone(dest)) {
+        if (!(dstrectptr = pgFRect_FromObject(dest, &dstrect))) {
+            if (!pg_TwoFloatsFromObj(dest, &dstrect.x, &dstrect.y)) {
+                PyErr_SetString(PyExc_ValueError,
+                                "dstrect must be a point, Rect, or None");
```
zoldalma999, line 589: After setting this error, this function should probably indicate that somehow to the caller. Otherwise caller will (probably) return a value with an exception also set here, and python will error out.

src_c/render.c
```diff
@@ -546,7 +565,570 @@ renderer_dealloc(pgRendererObject *self, PyObject *_null)
...
+            }
+            dstrect.w = (float)self->width;
+            dstrect.h = (float)self->height;
+        }
+    }
+    if (SDL_RenderCopyExF(self->renderer->renderer, self->texture, srcrectptr,
+                          dstrectptr, 0, NULL, SDL_FLIP_NONE) < 0) {
+        PyErr_SetString(pgExc_SDLError, SDL_GetError());
```
zoldalma999, line 598: Same here

buildconfig/stubs/pygame/__init__.pyi
JoKing-1999: I can remove them, used just as a possible improvement for the future if we change flip x/y arguments with the single flip that would accept something like FLIP_HORIZONTAL | FLIP_VERTICAL

src_c/render.c
JoKing-1999: Changed

src_c/render.c
JoKing-1999: You are right, fixed

src_c/render.c
JoKing-1999: Makes sense, changed

src_c/render.c
JoKing-1999: Good catch, rewrote the function a bit so free isn't needed in current version, but otherwise you are correct, can you check again

src_c/render.c
JoKing-1999: Fixed a bit and moved repetitive lines to macro (at the start of the file), can you check if the current version is correct

src_c/render.c
JoKing-1999: Fixed

oddbookworm (approved):
LGTM! My functional testing produced results in line with `_sdl2`

PR #3334, opened 2025-02-10
Switch host image to arm on debian multiarch

ankith26:
An attempt to fix the consistent fails we are dealing with on our multiarch builds


Starbuck5 (approved):
Looks like it worked?

gresm (approved):
Thanks  :+1:

Starbuck5:
@gresm Not sure if you've known before, but when you merge you should add the PR to the current milestone so it can be tracked 🙂

PR #3335, opened 2025-02-12
General improvements and fixes to stubs

aatle:
Collection of general fixes to stubs.
Especially, making some attributes that seemed to be read-only when I tested, read-only.
Plus minor corrections, internal renaming.

Notes:
It seems that stubtest didn't catch the incorrect return type of `sndarray.use_arraytype()` even though `sndarray` is a python module.
I found that `midi.Input.device_id` was undocumented. Also `midi.__theclasses__`, an unused variable.
There are 2 other larger stub issues that I will address in later PRs.


buildconfig/stubs/pygame/event.pyi
```diff
@@ -10,11 +10,11 @@ class _GenericEvent:
     @property
     def type(self) -> int: ...
     __dict__: dict[str, Any]
-    __hash__: None  # type: ignore
+    __hash__: ClassVar[None]  # type: ignore[assignment]
     def __init__(
         self, type: int, dict: dict[str, Any] = ..., **kwargs: Any
     ) -> None: ...
-    def __getattribute__(self, name: str) -> Any: ...
+    def __getattr__(self, name: str) -> Any: ...
```
ankith26, line 17: I don't get why this change is being made. I mean, the older one was correct right?

aatle: The previous one does seem to have the same behavior in mypy, they're both correct I think.
If you think about it, `__getattr__` is more fitting though.
Also see https://typing.readthedocs.io/en/latest/guides/writing_stubs.html#attribute-access, which promotes `__getattr__` over `__getattribute__` a few times.
> Note that only `__getattr__`, not `__getattribute__`, is guaranteed to be supported in stubs.

On a side note, stubcheck prevents me from changing all `__getattribute__` to `__getattr__` in the stubs.

ankith26 (approved):

buildconfig/stubs/pygame/image.pyi
```diff
@@ -6,11 +6,11 @@ from pygame.surface import Surface
 
 from pygame.typing import FileLike, IntPoint, Point
 
-_BufferStyle = Union[BufferProxy, bytes, bytearray, memoryview]
+_BufferLike = Union[BufferProxy, bytes, bytearray, memoryview]
```
MightyJosip, line 9: Should this be moved to pygame.typing as BufferLike?

aatle: No, it is only used once and there are already built-in types for this now.

There are improvements to be made for the stubs about buffers (`__buffer__` and `collections.abc.Buffer`), and I will address this in a different PR.

damusss (approved):
Fixes looks good to me, thank you!

PR #3337, opened 2025-02-19
Prep surface_fill.c for SDL3

Starbuck5:
Putting this one out as a simple PR, hopefully straightforward to review.

It brings some of the SDL3 compat macros that have been developed in previous PRs into the surface_fill.c's sphere of influence. To get surface_fill.c to fully compile I additionally had to mess with the GET_PIXELVALS macros, which I have not included in this PR because I am keeping it simple and because I am waiting to reunify them with the macros I split out in https://github.com/pygame-community/pygame-ce/pull/3309.

PR #3338, opened 2025-02-19
Prep alphablit.c for SDL3

Starbuck5:
Same idea as https://github.com/pygame-community/pygame-ce/pull/3337, but for alphablit.c

These are low hanging fruit changes. They need to happen, and even though I've gone further in other branches there is no need to make this PR super complex.


damusss (approved):
Makes sense, thanks a lot :)

ankith26 (approved):

PR #3340, opened 2025-02-20
Port scap.c to SDL3

ankith26:
This used to be a part of #3208 but as that PR is not moving due to joystick, thought it's better to separate out this into its own PR because it's pretty trivial.


Starbuck5 (approved):
Thanks!

damusss (approved):
LGTM, Thanks! 👍

PR #3341, opened 2025-02-20
Port camera to SDL3

ankith26:
camera now fully compiles on SDL3


damusss (approved):
Thanks for porting, looks good :)

PR #3343, opened 2025-02-22
GET_PIXELVALS SDL3 compat

Starbuck5:
Changes
- Ports the macros GET_PIXELVALS, GET_PIXELVALS_1, and SET_PIXELVAL to SDL3, consolidating back in the pixelarray methods version which I introduced recently to port the pixelarray module only
- Ports the non-SIMD premultiply to SDL3 (since it was using the macro)

These changes lead to surface_fill.c and alphablit.c fully compiling on SDL3.

### DIFF
--- a/src_c/alphablit.c
+++ b/src_c/alphablit.c
@@ -601,7 +601,9 @@ blit_blend_rgba_add(SDL_BlitInfo *info)
     int dstpxskip = info->d_pxskip;
     int dstskip = info->d_skip;
     PG_PixelFormat *srcfmt = info->src;
+    SDL_Palette *srcpal = info->src_palette;
     PG_PixelFormat *dstfmt = info->dst;
+    SDL_Palette *dstpal = info->dst_palette;
     int srcbpp = PG_FORMAT_BytesPerPixel(srcfmt);
     int dstbpp = PG_FORMAT_BytesPerPixel(dstfmt);
     Uint8 dR, dG, dB, dA, sR, sG, sB, sA;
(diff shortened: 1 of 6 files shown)


src_c/alphablit.c
```diff
@@ -1696,9 +1746,10 @@ blit_blend_sub(SDL_BlitInfo *info)
             while (height--) {
                 LOOP_UNROLLED4(
                     {
-                        GET_PIXELVALS_1(sR, sG, sB, sA, src, srcfmt);
+                        GET_PIXELVALS_1(sR, sG, sB, sA, src, srcpal);
                         GET_PIXEL(pixel, dstbpp, dst);
-                        GET_PIXELVALS(dR, dG, dB, dA, pixel, dstfmt, dstfmt);
+                        GET_PIXELVALS(dR, dG, dB, dA, pixel, dstfmt, dstpal,
+                                      dstppa);
```
Starbuck5, line 1752: Unlike all other instances, this GET_PIXELVALS passes `dstfmt` into the macro twice. Once as the format, once as the "ppa" (per pixel alpha).

I believe this was a copy paste bug and have removed this behavior.

It now properly uses the per pixel alpha of the surface.

I'm classifying this as a super minor bugfix no one will notice. To notice you would have to be blit blend subtracting an 8 bit surface onto a 24 bit surface on a platform we don't do SIMD  for (otherwise it would use a different implementation). I have not attempted to reproduce the issue.

damusss: Sure, and nice catch too.
(we should reaaaalllyyy constrain surfaces to 32bits in pygame 3, imagine how much less branching and focus towards what really matters)

damusss (approved):
Damn, good job!
Also those files are basically made entirely of macros, just how big is the processed code 😭

ankith26 (approved):
LGTM, thanks for the PR :balloon:

src_c/alphablit.c
Starbuck5: I understand where you're coming from but forcing everything 32 bit will be difficult and removes paths that people can currently use.

I would like to do a review of all pixel handling code for pygame-ce 3, I think there are inconsistencies between places that could be resolved with higher level macros perhaps.

One other thing is that SDL3 introduces more exotic types of surfaces than ever before, and we need to make sure our APIs properly refuse to handle them or can actually handle them. Because it will be quite challenging to make sure they never leak into pygame-ce from SDL. Like imagine using the camera API and it gives you a YUV surface, or you load a weird png and it's a 64 bit surface.

PR #3344, opened 2025-02-22
A few surface.c SDL3 fixes

Starbuck5:
This PR is more low hanging fruit towards porting surface.c to SDL3.

The remaining complex things in this file are
- surface.init
- surface.convert
- RLEACCEL changes

I have patches staged that get those items to compile as well, but they are more complex or provisional (in the case of RLEACCEL). So I will open them in future PRs.

### DIFF
--- a/src_c/surface.c
+++ b/src_c/surface.c
@@ -1054,7 +1054,7 @@ surf_get_palette(PyObject *self, PyObject *_null)
 
     SURF_INIT_CHECK(surf)
 
-    pal = surf->format->palette;
+    pal = PG_GetSurfacePalette(surf);
 
     if (!pal)
         return RAISE(pgExc_SDLError, "Surface has no palette to get\n");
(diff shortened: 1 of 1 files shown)


src_c/surface.c
```diff
@@ -2517,11 +2517,15 @@ surf_scroll(PyObject *self, PyObject *args, PyObject *keywds)
...
 
-    clip_rect = &surf->clip_rect;
+    SDL_Rect clip_rect;
+    if (!PG_GetSurfaceClipRect(surf, &clip_rect)) {
+        return RAISE(pgExc_SDLError, SDL_GetError());
+    }
+
     SDL_Rect surf_rect = {0, 0, surf->w, surf->h};
 
     // In SDL3, SDL_IntersectRect is renamed to SDL_GetRectIntersection
```
damusss, line 2527: Following this comment, should branching be made to use the correct name? should a macro be made? There are roughly 6 usages of this function in the codebase, in pygame._sdl2 aswell.
I guess you didn't change it here because it is out of scope from surface-only related fixes, do you plan on fixing it in another PR?

damusss (approved):
Looking good, and my comment is just for info. Thanks :)

src_c/surface.c
ankith26: It's probably covered by `SDL_ENABLE_OLD_NAMES` already.

ankith26 (approved):

PR #3345, opened 2025-02-22
Fix build with debug python

oddbookworm:
Our build currently sucks when we use python with debug symbols. This at least gets us in a working spot. First commit is to baseline the CI and demo the problem. When that's done, I'll push the fix commit

### DIFF
--- a/.github/workflows/build-ubuntu-debug-python.yml
+++ b/.github/workflows/build-ubuntu-debug-python.yml
@@ -1,7 +1,7 @@
-# this workflow runs a build with a python version that has debug symbols
-# Note that for intrinsics, it only runs what gets compiled
-# and would naturally run. It also is limited to what can run in
-# a CI environment
+# This workflow runs a build with a python version that has debug symbols
+# Update this workflow when our min/max python minor versions update
+# This workflow is necessary to ensure that we can build and run with
+# a debug python build without too much worrying about SIGABRT being thrown
 # IMPORTANT: binaries are not to be uploaded from this workflow!
 
 name: Ubuntu debug python
@@ -9,8 +9,8 @@ defaults:
   run:
     shell: bash -leo pipefail {0}
 
-# Run CI only when a release is created, on changes to main branch, or any PR
-# to main. Do not run CI on any other branch. Also, skip any non-source changes
+# Run CI only on changes to main branch, or any PR to main.
+# Do not run CI on any other branch. Also, skip any non-source changes
 # from running on CI
 on:
   push:
--- a/.github/workflows/build-ubuntu-debug-python.yml
+++ b/.github/workflows/build-ubuntu-debug-python.yml
@@ -87,7 +87,7 @@ jobs:
       uses: actions/cache@v4.2.1
       with:
         key: ${{ matrix.python }}
-        path: $PYENV_ROOT/versions/${{ matrix.python }}-debug/**
+        path: ~/.pyenv/versions/${{ matrix.python }}-debug/**
 
     - name: Build debug python
       id: build
--- a/.github/workflows/build-ubuntu-debug-python.yml
+++ b/.github/workflows/build-ubuntu-debug-python.yml
@@ -85,7 +85,7 @@ jobs:
     - name: Build debug python and pygame-ce
       id: build
       run: |
-        pyenv install ${{ matrix.python }} --debug
+        pyenv install ${{ matrix.python }} --debug -v
         pyenv global ${{ matrix.python }}-debug
         python dev.py build --lax
 
--- a/.github/workflows/build-ubuntu-debug-python.yml
+++ b/.github/workflows/build-ubuntu-debug-python.yml
@@ -5,6 +5,9 @@
 # IMPORTANT: binaries are not to be uploaded from this workflow!
 
 name: Ubuntu debug python
+defaults:
+  run:
+    shell: bash -leo pipefail {0}
 
 # Run CI only when a release is created, on changes to main branch, or any PR
 # to main. Do not run CI on any other branch. Also, skip any non-source changes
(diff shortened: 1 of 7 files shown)


oddbookworm:
[debug_fix.zip](https://github.com/user-attachments/files/18924184/debug_fix.zip)

For anyone too impatient to wait for the CI to be finished, here's the patch (in a zip because github doesn't like the raw patch file lol)

.github/workflows/build-ubuntu-debug-python.yml
```diff
@@ -0,0 +1,119 @@
...
+      - 'docs/**'
+      - 'examples/**'
+      - '.gitignore'
+      - '*.rst'
+      - '*.md'
+      - '.github/workflows/*.yml'
+      # gcov/lcov only gets C coverage
+      - 'src_py/**'
+      # re-include current file to not be excluded
+      - '!.github/workflows/build-ubuntu-coverage.yml'
```
ankith26, line 25: this line would need to be updated

.github/workflows/build-ubuntu-debug-python.yml
```diff
@@ -0,0 +1,119 @@
...
+    branches: main
+    paths-ignore:
+      - 'docs/**'
+      - 'examples/**'
+      - '.gitignore'
+      - '*.rst'
+      - '*.md'
+      - '.github/workflows/*.yml'
+      # gcov/lcov only gets C coverage
+      - 'src_py/**'
```
ankith26, line 23: this should not be here right?

oddbookworm: Correct, I thought I deleted that lol

.github/workflows/build-ubuntu-debug-python.yml
oddbookworm: Apparently I updated this file in my imagination and just forgot to actually do it

src_c/surflock.c
```diff
@@ -114,6 +114,14 @@ pgSurface_LockBy(pgSurfaceObject *surfobj, PyObject *lockobj)
 static int
 pgSurface_UnlockBy(pgSurfaceObject *surfobj, PyObject *lockobj)
 {
+#if PY_MINOR_VERSION >= 12
```
ankith26, line 117: It is probably better to compare against PY_VERSION_HEX here so that the major version is factored into the calculation as well.

.github/workflows/build-ubuntu-debug-python.yml
```diff
@@ -0,0 +1,98 @@
...
+        echo -e 'export PYENV_ROOT="$HOME/.pyenv"' >> ~/.bash_profile
+        echo -e 'export PATH="$PYENV_ROOT/bin:$PATH"' >> ~/.bash_profile
+        echo -e 'eval "$(pyenv init --path)"' >> ~/.bash_profile
+        echo -e 'eval "$(pyenv init -)"' >> ~/.bash_profile
+
+    - name: Build debug python and pygame-ce
+      id: build
+      run: |
+        pyenv install ${{ matrix.python }} --debug -v
+        pyenv global ${{ matrix.python }}-debug
```
ankith26, line 89: I wonder if we can cache pyenv builds so that every subsequent run does not waste CI time.

src_c/surflock.c
oddbookworm: Fair, Python 4 when?

.github/workflows/build-ubuntu-debug-python.yml
oddbookworm: Next commit *should* fix this

oddbookworm: https://github.com/pygame-community/pygame-ce/actions/runs/13513571348/job/37758191618?pr=3345#step:5:19

It works

src_c/surflock.c
```diff
@@ -153,12 +164,17 @@ pgSurface_UnlockBy(pgSurfaceObject *surfobj, PyObject *lockobj)
                 noerror = 0;
             }
             else if (weakref_getref_result == 0) {
+                // Need to cache any currently set exceptions before calling
+                // PySequence_DelItem
+                RESTORE_EXCEPTION
```
ankith26, line 169: should this not be `CACHE_EXCEPTION`?

src_c/include/_pygame.h
```diff
@@ -738,3 +738,33 @@ pg_PointList_FromArrayDouble(double const *array, int arr_length)
 
     return sequence;
 }
+
+#if PY_VERSION_HEX >= 0x030C0000
+
+#define CREATE_CACHE_VARIABLES static PyObject *__cached_exception = NULL;
```
ankith26, line 744: Can probably name this better lol, "cache variables" sounds misleading. Maybe `DECLARE_EXC`, `GET_EXC` and `SET_EXC`?

src_c/surflock.c
oddbookworm: Oops

src_c/include/_pygame.h
oddbookworm: Maybe `DECLARE_EXCEPTION_CACHE`, `CACHE_EXCEPTION`, `RESTORE_EXCEPTION`? I'm not as big a fan of `GET_EXC` and `SET_EXC` because it's not as clear IMO what they're actually doing, and `DECLARE_EXC` makes it sound like you're declaring an exception itself, not a cache to be used

ankith26: names do be hard :skull: 

IG my confusion is with the word "cache" because it's not really a cache, is it? I mean, IG we can have `DECLARE_EXCEPTION_STORE`, `STORE_EXCEPTION` and `RESTORE_EXCEPTION` but I'm not too fond of these names either.

oddbookworm: Renamed to something @ankith26 approved of on discrod

ankith26 (approved):

gresm (approved):
LGTM :+1:

PR #3347, opened 2025-02-24
Added tolerance for Color.lerp

oddbookworm:
Small little bug that plagued me while working on my free-threading stuff

### DIFF
--- a/src_c/color.c
+++ b/src_c/color.c
@@ -808,7 +808,19 @@ _color_lerp(pgColorObject *self, PyObject *args, PyObject *kw)
         return NULL;
     }
 
-    if (amt < 0 || amt > 1) {
+    // TOLERANCE to account for double precison floating point inaccuracy at
+    // the very limits, like if you're LERP'ing by 0.01. When you hit the end,
+    // something stupid like this might happen
+    /*  >>> value = 0
+        >>> offset = 0.01
+        >>> while value < 1:
+        ...     value += offset
+        ...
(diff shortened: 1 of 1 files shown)


damusss (approved):
I was about to say to add something in the docs, but I guess the change is so small and not api-breaking so actually it's not necessary, LGTM!

gresm (approved):
LGTM, but alternatively we could just clamp the amount to the range.

oddbookworm:
> LGTM, but alternatively we could just clamp the amount to the range.

That would be a behavior change that would be harder to justify. I don't see a compelling reason to clamp it beyond this tiny little bit, and rounding will take care of it anyway

src_c/color.c
```diff
@@ -808,7 +808,19 @@ _color_lerp(pgColorObject *self, PyObject *args, PyObject *kw)
...
+    /*  >>> value = 0
+        >>> offset = 0.01
+        >>> while value < 1:
+        ...     value += offset
+        ...
+        >>> print(value)
+        1.0000000000000007
+    */
+    static const double TOLERANCE = 1e-6;
+    if ((amt < -TOLERANCE) || (amt > (1.0 + TOLERANCE))) {
```
ankith26, line 823: I think I'd rather add clamping instead of this for the edge cases.
So -`tol < amt < 0` => `amt = 0` and `1 < amt < 1 + tol` => `amt = 1`

PR #3352, opened 2025-02-26
Refactor and consolidate all SIMD handlers

ankith26:
- We currently have "SIMD handler" code across various header files. To reduce duplication and make things more robust, I wanted to see all SIMD related common functionalities in one file. We already have `simd_shared.h`, but stuff here was pointlessly duplicated across various headers, and this header was not being used everywhere either. This PR fixes that.
- The strategy of "define an avx2 function to error at runtime when called if compile time support is missing" was a bad idea IMO (and I am probably partly to blame here for putting some code like this). This is a situation that can quite easily prevent right at compile time, by guarding the function declaration/implementation.
- The "new strategy" I want to introduce via this PR is having the runtime check *also serve as the* compile time check. I implemented this via a macro that can be tested at compile time with an `ifdef`, and then at runtime is just the underlying SDL SIMD feature check function. Why? This ensures that we don't call the runtime check when compile time support is missing, even accidentaly. Doing so would now result in a compile time error after this PR (instead of runtime error with the current code). This solves the issue I mentioned in the above point.
- Added a new macro `PG_DISABLE_SIMD` that can be set (in one place only) to disable all SIMD at compile time.

How to test this PR? This is probably the hard part of reviewing/approving this PR, but:
check that there's no performance regressions in anything SIMD related.
- go through CI logs and check that there's no "your system supports X but X was not compiled" type warnings that surface import can raise.


ankith26:
Making this PR a draft because it needs more thought and testing from my side, I may have overlooked a thing or two

ankith26:
Closing this PR because the implementation here isn't feasible.

Issue #3353, opened 2025-02-28
Cannot open "tutorials" in the offline documentation for Pygame CE 2.5.3


quitelight:
**Environment:**

```
pygame-ce 2.5.3
Platform:		MacOS 15.3.1
System:			MacOS
Processor:		Apple M1 Pro
Python:			    CPython 3.13.2

```

**Current behavior:**

When launching the offline pygame docs via:

`python3 -m pygame.docs   `

The docs loads successfuly however. trying to view any of the tutorials results in the following appearing:

<img width="386" alt="Image" src="https://github.com/user-attachments/assets/517ebad5-d6a9-4edb-ae5b-57a0547d93e5" />

**Expected behavior:**

The tutorial docs loads successfully like other parts of the documentation.

**Screenshots**

As above.

**Steps to reproduce:**

As above.

Starbuck5:
This is semi intentional (in that it was noticed when we changed our buildconfig but not fixed).

The argument for it is that it's a size reduction. The tutorial directory is 1.16 mb, duplicated across all of our ~60 wheels every single release adds up. Especially since we have a global cap on the size of all our uploads to pypi.

makix222:
On first download / install, I would appreciate a notification that documentation can be downloaded for offline work, and then providing a command which would download the files directly into the project. I often work offline, and each time I realized that I have forgotten to download the manuals.

Starbuck5:
@makix222 I'm not aware of any mechanism to notify users of arbitrary information on a pip install of a package, nor that this would be the most important message to put in said message.

You do not need to download documentation, it comes with the project already. `python -m pygame.docs`.

quitelight:
> [@makix222](https://github.com/makix222) I'm not aware of any mechanism to notify users of arbitrary information on a pip install of a package, nor that this would be the most important message to put in said message.
> 
> You do not need to download documentation, it comes with the project already. `python -m pygame.docs`.

Is there a way to download the entire pygame docs and included tutorials? I understand its already available via the command but it doesn't contain the tutorial pages.

Starbuck5:
We do not provide any sort of download page, but it's straightforward to build the documentation yourself if you want a stand alone copy.

I didn't ask before, but why are the tutorials something you want downloaded in the first place?

quitelight:
Oh I'm just learning Pygame atm and wanted an offline source as I'm travelling in an area with limited internet atm.

Starbuck5:
[...]

PR #3358, opened 2025-03-01
Use webp to minify logo images

Starbuck5:
This brings the wheel size down (for my python version / system) from 12181 kb to 11139 kb (-1042kb).

It accomplishes this by using webp to losslessly store logo images that are currently PNGs.

This could also be done for the tutorial assets, but I'm keeping this PR to just these ones.

I was inspired by https://github.com/pygame-community/pygame-ce/issues/3353 to take a look around for size reductions.


damusss (approved):
Very good idea! Thanks

ankith26 (approved):
LGTM, nice file size savings! :sunglasses:

PR #3371, opened 2025-03-03
Restructure and port surface init and convert to SDL3

Starbuck5:
Surface init--
```
# Before, each codepath would output bpp and masks, which was then used to construct a format enum.
depth + masks -> bpp and masks
depth -> bpp + default masks
surface -> format -> bpp and masks
default window surface -> format -> bpp and masks

# Now, each codepath outputs a format enum itself.
depth + masks -> format enum
depth -> default format enum
surface -> format enum
default window surface -> format enum
```
Why change this now? It's not just that it's nicer, one of the old code paths used an SDL_PixelFormat as a writeable thing, and since that no longer exists the code could not be directly ported.

Surface convert--
Surface.convert had a similar thing but even more dubious. The code `memcpy(&format, surf->format, sizeof(format));` does not seem particularly safe! These objects contain internal refcounts, internal references.

It used a bunch of the same default masks, so just like surface init I translated them all into pixel format enums. This eliminates some logic around building formats up, although I did still have to build a format struct in the SDL2 code path to convert to a surface with a palette. In SDL3 I had access to a function that just took a pixelformat enum and a palette, like all the SDL3 things, so I used that instead.

### DIFF
--- a/src_c/surface.c
+++ b/src_c/surface.c
@@ -483,7 +483,13 @@ surface_init(pgSurfaceObject *self, PyObject *args, PyObject *kwds)
     int bpp;
     Uint32 Rmask, Gmask, Bmask, Amask;
     SDL_Surface *surface;
+
+#if SDL_VERSION_ATLEAST(3, 0, 0)
     PG_PixelFormatEnum format = SDL_PIXELFORMAT_UNKNOWN;
+#else
+    SDL_PixelFormat default_format;
+    default_format.palette = NULL;
+#endif
 
     char *kwids[] = {"size", "flags", "depth", "masks", NULL};
     if (!PyArg_ParseTupleAndKeywords(args, kwds, "O|iOO", kwids, &size, &flags,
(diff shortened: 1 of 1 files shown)


damusss (approved):
Alright, the code looks much cleaner, the changes makes sense, and the sdl3 differences are clear. I guess that shocking code should be dealt with sooner or later. Thanks, LGTM! 👍

src_c/surface.c
```diff
@@ -1544,104 +1519,109 @@ surf_convert(pgSurfaceObject *self, PyObject *args)
...
-#endif
-            if (PG_FORMAT_BitsPerPixel((&format)) > 8) {
-                /* Allow a 8 bit source surface with an empty palette to be
-                 * converted to a format without a palette (pygame-ce issue
-                 * #146). If the target format has a non-NULL palette pointer
-                 * then SDL_ConvertSurface checks that the palette is not
-                 * empty-- that at least one entry is not black.
-                 */
-                format.palette = NULL;
-            }
```
ankith26, line 1626: This whole block of changes is kinda confusing. I'm not clear what is exactly happening to this block in particular, how is the new code equivalent to this?

Starbuck5:
@ankith26 @damusss As discussed on discord I've now made this PR less risky and easier to review by putting the SDL3 codepath fully separate from the SDL2 codepath, rather than modifying the SDL2 codepath to work in SDL3. I'd like to get this in, it significantly reduces the amount of compilations errors in surface.

ankith26 (approved):
LGTM, much easier to approve now! thanks :tada:

PR #3372, opened 2025-03-03
Add `image.load_animation`

### LINKED ISSUE #1054: pygame.image.load("filename.gif") should be able to return one surface for each frame in the gif file (1968)

ankith26:
My take on the PR #2218

fixes #1054

While that PR did some nice work and set the base up for this work, I believe the general consensus was that `image.load_animation` should just be pretty barebones in terms of the API, just returning a python list of the frames, closely wrapping the underlying SDL function.
If we have a builtin `AnimatedSurface` API in pygame-ce, it could now be implemented in pure python and use this function internally, and that should be decoupled from image entirely.

Here is a simple example of how a GIF player would look like

```py
import itertools
import pygame

pygame.init()

frames = pygame.image.load_animation("animated_sample.gif")
print(frames)

screen = pygame.display.set_mode(frames[0][0].size)

frames_cycle = itertools.cycle(frames)
while True:
        if event.type == pygame.QUIT:
            raise SystemExit

    frame, frame_duration = next(frames_cycle)
    screen.blit(frame, (0, 0))
    pygame.display.flip()
    pygame.time.wait(frame_duration)
```

Here is the example file being used:

animated_sample


robertpfeiffer (approved):
:+1:

src_c/image.c
```diff
@@ -1885,6 +1886,17 @@ image_load_sized_svg(PyObject *self, PyObject *args, PyObject *kwargs)
...
 
+static PyObject *
+image_load_animation(PyObject *self, PyObject *args, PyObject *kwargs)
+{
+    if (ext_load_animation) {
+        return PyObject_Call(ext_load_animation, args, kwargs);
+    }
+
+    return RAISE(PyExc_NotImplementedError,
+                 "Support for sized svg image loading was not compiled in.");
```
MyreMylar, line 1897:
```suggestion
                 "Support for animation loading was not compiled in.");
```

src_c/doc/image_doc.h
```diff
@@ -2,6 +2,7 @@
 #define DOC_IMAGE "pygame module for image transfer"
 #define DOC_IMAGE_LOAD "load(file) -> Surface\nload(file, namehint="") -> Surface\nload new image from a file (or file-like object)"
 #define DOC_IMAGE_LOADSIZEDSVG "load_sized_svg(file, size) -> Surface\nload an SVG image from a file (or file-like object) with the given size"
+#define DOC_IMAGE_LOADANIMATION "load_sized_svg(file, namehint="") -> list[tuple[Surface, int]]\nload an animation (GIF/WEBP) from a file (or file-like object)"
```
MyreMylar, line 5:
```suggestion
#define DOC_IMAGE_LOADANIMATION "load_animation(file, namehint="") -> list[tuple[Surface, int]]\nload an animation (GIF/WEBP) from a file (or file-like object)"
```

docs/reST/ref/image.rst
```diff
@@ -135,6 +135,28 @@ following formats.
 
    .. ## pygame.image.load_sized_svg ##
 
+.. function:: load_animation
+
+   | :sl:`load an animation (GIF/WEBP) from a file (or file-like object)`
+   | :sg:`load_sized_svg(file, namehint="") -> list[tuple[Surface, int]]`
```
MyreMylar, line 141:
```suggestion
   | :sg:`load_animation(file, namehint="") -> list[tuple[Surface, int]]`
```

docs/reST/ref/image.rst
```diff
@@ -135,6 +135,28 @@ following formats.
...
+   This returns a list of tuples (corresponding to every frame of the animation),
+   where each tuple is a (surface, delay) pair for that frame.
+
+   This function requires SDL_image 2.6.0 or above. If pygame was compiled with
+   an older version, ``pygame.error`` will be raised when this function is
+   called.
+
+   .. versionadded:: 2.5.4
+
+   .. ## pygame.image.load_sized_svg ##
```
MyreMylar, line 158:
```suggestion
   .. ## pygame.image.load_animation ##
```

MyreMylar (approved):
OK, LGTM 👍 I've tested the animation locally and it all looks good to me. I like the simple API - not much to argue about but easy for people to expand on further (as they do already when loading animations frame by frame).

For some reason github is having a moment with downloading pygame and breaking the MSYS build but other than that it all looks good to merge. I'll leave it for a bit longer to give everyone a chance to have a look at the API as it's new.

damusss (approved):
Tested locally, checked the code, looks immaculate and well explained, awesome New API sauce, thanks!
Not going to press the shiny green merge button either at the moment

Starbuck5:
I would like to weigh in before this is merged, thank you.

test/image_test.py
```diff
@@ -1360,6 +1360,37 @@ def test_load_sized_svg_erroring(self):
                     value_error_size,
                 )
 
+    @unittest.skipIf(
+        pygame.image.get_sdl_image_version() < (2, 6, 0),
+        "load_sized_svg requires SDL_image 2.6.0+",
```
Starbuck5, line 1365: Copy paste error here.

src_c/imageext.c
```diff
@@ -206,6 +206,84 @@ imageext_load_sized_svg(PyObject *self, PyObject *arg, PyObject *kwargs)
...
+        return NULL;
+    }
+
+    rw = pgRWops_FromObject(obj, &ext);
+    if (rw == NULL) { /* stop on NULL, error already set */
+        return NULL;
+    }
+
+    if (name) { /* override extension with namehint if given */
+        type = (name[0] != '\0') ? iext_find_extension(name) : NULL;
```
Starbuck5, line 229: This code confused me for a bit but I think I get it now. You're saying if the namehint is an empty string we shouldn't consider it but we also should ignore the file extension. That is different behavior than other namehint locations?

I think it would be better to use strlen rather than directly checking if the first char is null.

docs/reST/ref/image.rst
```diff
@@ -135,6 +135,28 @@ following formats.
 
    .. ## pygame.image.load_sized_svg ##
 
+.. function:: load_animation
+
+   | :sl:`load an animation (GIF/WEBP) from a file (or file-like object)`
+   | :sg:`load_animation(file, namehint="") -> list[tuple[Surface, int]]`
+
+   Load an animation (GIF/WEBP) from a file source. You can pass either a
```
Starbuck5, line 143: What happens if the file source does not contain an animation?

Starbuck5:
> While that PR did some nice work and set the base up for this work, I believe the general consensus was that image.load_animation should just be pretty barebones in terms of the API, just returning a python list of the frames, closely wrapping the underlying SDL function.

I've been consistently pushing for load_animation -> AnimatedSurface. But I'm okay with this approach instead, given that you and others support it. It doesn't let perfect be the enemy of the good.

test/image_test.py
ankith26: good catch, will fix it soon

src_c/imageext.c
ankith26: Yes this is slightly different compared to the other locations, but I think it makes the best sense given the documentation/typehinting of the function. There `""` is the default value of the argument, so it implies that not passing the same argument is the same as passing `""`. therefore our code should ideally handle both cases the same (and now, both will handle it with `type = NULL`).

ankith26:
I think if/when we are adding `AnimatedSurface`, we can just support
- `AnimatedSurface.__init__(frames: list[tuple[Surface, int]])` (so `AnimatedSurface(image.load_animation(...))` would be supported)
- For GIF/WEBP loading, we can have a classmethod shorthand `AnimatedSurface.load(...)` that is equivalent to the above.

This is more pythonic IMO, even if we wanted to discard this PR and go with the class only approach.

Starbuck5:
In a similar vein, what do you think about bring image things into the Surface class overall?

`image.load` -> `Surface.load`
`image.tobytes` -> `Surface.tobytes`

ankith26:
Love that idea as well. The same would go for `Surface.save` as well IMO.

src_c/imageext.c
ankith26: I have implemented your `strlen` suggestion in the latest commit.

Starbuck5:
Tested with the iconic animation from the last April fools' event, there is some issue with webp. Presumably this would be fixed by an update to https://github.com/libsdl-org/SDL_image/releases/tag/release-2.8.5

image


[pygame-real-engine.zip](https://github.com/user-attachments/files/19177391/pygame-real-engine.zip)

Starbuck5 (approved):
I think this is good to go in now, although SDL image should be handled before next release.

LeWolfYT:
Will there ever be support for saving animations?

gresm:
@LeWolfYT Very unlikely, as SDL doesn't support it.

damusss:
> Will there ever be support for saving animations?

@LeWolfYT you could technically request this in the SDL library, which would make the chances of us adding it a lot higher

PR #3373, opened 2025-03-05
Improve SysFont documentation and add font verification warning. Closes #2643

### LINKED ISSUE #2643: Unnecessary warning gets issued when constructing a SysFont with the default font explicitly
When running this line of code
```python
pygame.font.SysFont(pygame.font.get_default_font(), font_size)
```
A warning gets issued to the console
```
/home/oddbookworm/Projects/pygame-ce/venv/lib/python3.11/site-packages/pygame/sysfont.py:492: UserWarning: The system font 'freesansbold.ttf' couldn't be found. Did you mean: 'freesansbold'? Using the default font instead.
  warnings.warn(
```

I don't think a warning should be emitted here

mv805:
Hello. This is my propsal to solve the usability issue noted here #2643 where I give my analysis into the issue.
I decided just to update the docs with a minor clarification and improve the warning to help prompt future users to double check their name inputs. This should prevent the issue from occuring as long as the user reads the docs and warnings 😃 

The following short code can be ran to confirm the new warning:

```
import pygame
pygame.init()

font_size = 20  # some default font they might want

font = pygame.font.SysFont(pygame.font.get_default_font(), font_size)  # gives warning and defaults to freesansbold. Now you should see the updated warning prompting to double check name input.

print("Font loaded:", font.name)  # confirm that freesansbold loaded by default

pygame.quit()
```


src_py/sysfont.py
```diff
@@ -469,12 +470,14 @@ def SysFont(name, size, bold=False, italic=False, constructor=None):
                 break
 
         else:
+            font_verification_message = "Verify your font name input."
+            " Using the default font instead."
```
ankith26, line 474: These two lines are not being concatenated into the same string, the second line is actually a no-op here. You could do the concatenation explicitly by using a `+`, or alternatively put both strings within a single parenthesis, so that they are concatenated implicitly.

gresm: A backslash should also work (with indentation).

mv805: This is corrected in the new commit

ankith26 (approved):

gresm (approved):
LGTM 👍

PR #3379, opened 2025-03-11
Add `Color.hex`

damusss:
Followed the New API routine for this property. FAQ below.
### Why, if it can be done in one line?
That's not a reason not to add a shortcut. Other color shortcuts like Color.normalized can be achieved in one line aswell, plus, how to get the Color hex with one line is not very straight forward.
### Can't you use the built in `hex` function?
That function wasn't made for correct representation of hex colors, so `Color(hex(Color))` can easily error, for example in the following cases: `hex(Color(1, 2, 3)) -> "0x10203ff"` (no trailing zero; ValueError) or `hex(Color(0, 0, 0, 255)) -> "'0xff"` (not long enough; ValueError).


src_c/color.c
```diff
@@ -1543,6 +1549,33 @@ _color_set_normalized(pgColorObject *color, PyObject *value, void *closure)
...
+static PyObject *
+_color_get_hex(pgColorObject *color, void *closure)
+{
+    return PyUnicode_FromFormat("#%02x%02x%02x%02x", color->data[0],
+                                color->data[1], color->data[2],
+                                color->data[3]);
+}
+
+static int
+_color_set_hex(pgColorObject *color, PyObject *value, void *closure)
```
ankith26, line 1561: Needs `DEL_ATTR_NOT_SUPPORTED_CHECK`

ankith26 (approved):
LGTM, thanks! :100: 

As it is new API, requesting atleast 2 more approvals in addition to mine, having more eyes on this PR will be better.

docs/reST/ref/color.rst
```diff
@@ -210,6 +210,23 @@
 
       .. ## Color.normalized ##
 
+   .. attribute:: hex
+
+      | :sl:`Gets or sets the stringified hexadecimal representation of the Color.`
+      | :sg:`hex -> str`
+
+      The strigified hexadecimal representation of the Color. The hexadecimal string
```
Matiiss, line 218: strigified -> stringified

docs/reST/ref/color.rst
```diff
@@ -210,6 +210,23 @@
...
+      is formatted as ``"#rrggbbaa"`` where rr, gg, bb, and aa are 2 digits hex numbers
```
Matiiss, line 219: digits -> digit
optional: 2 -> two

test/color_test.py
```diff
@@ -950,6 +950,33 @@ def test_normalize__all_elements_within_limits(self):
...
+        with self.assertRaises(AttributeError):
+            del color.hex
+
+        for value in ["FFFFFFFF", "#FFzzFF00", "0x FFFFFF", "#FF"]:
+            with self.assertRaises(ValueError):
+                color.hex = value
+
+        for value in ["#FFFFFFFF", "#FFFFFF", "0xFFFFFFFF", "0xFFFFFF"]:
+            color.hex = value
+            self.assertEqual((color.r, color.g, color.b, color.a), (255, 255, 255, 255))
```
Matiiss, line 978: Optional: iterate over the list again and check again with `.lower()`

Matiiss (approved):
LGTM, left a couple comments, but nothing that would necessarily have to block this PR

damusss:
all suggestions are good I'll fix them today

ankith26:
The latest commit broke formatting of a test file.

Starbuck5:
I don't really see this why should be added.

The FAQ in the PR description seems to assume this is a thing people want to do, but I can't think of why I would want to generate a hex string from a color object.

aatle:
The primary reason is for displaying and inputting for user, not really for internal code. The hex rgb representation is the standard way to display, port, input colors; *every* color picker supports this format. Therefore it makes sense to add a shortcut.

If a pygame program wanted to allow the user to pick any color for something, this would be useful for both labelling and inputting the color.

aatle (changes requested):
You need to add a corresponding classmethod `from_hex()`, in accordance with every other settable color format. So that it's possible to use the setter in an expression.

Matiiss:
> You need to add a corresponding classmethod `from_hex()`, in accordance with every other settable color format. So that it's possible to use the setter in an expression.

Not necessary, it's already possible like so
```py
pygame.Color("#hexstr")
```

aatle:
I still think it could be beneficial to add it. It is explicit and could differentiate hex strings from color name strings (useful for input), plus more consistency.
Otherwise I think a note should be added, it's easy to miss the hex overload in the mass of constructor overloads.

Starbuck5 (approved):
I'm okay with this, I suppose. Aatle's stated use case makes sense.

I don't see a need to add from_hex in this PR.

Starbuck5:
Also IDK what's up with the tests, but maybe if you rebase @damusss it would rerun and be green and pretty?

PR #3380, opened 2025-03-11
Include tutorials docs in release build (fixes #3353)

### LINKED ISSUE #3353: Cannot open "tutorials" in the offline documentation for Pygame CE 2.5.3

damusss:
Did as the title says as sturbuck requested in https://github.com/pygame-community/pygame-ce/issues/3353#issuecomment-2712954833, fixing https://github.com/pygame-community/pygame-ce/issues/3353.

### DIFF
--- a/docs/meson.build
+++ b/docs/meson.build
@@ -4,7 +4,6 @@ if fs.is_dir('generated')
         'generated',
         exclude_directories: [
             'doctrees',
-            'tutorials',
             '_sources/tutorials',
             '_sources/c_api',
         ],


ankith26:
Just curious, what is the wheel size change after this PR?

damusss:
> Just curious, what is the wheel size change after this PR?

Probably around 1.16 MB. I'm not a fan of this, but it's true that we technically did save enough space to "afford" it. You should chat with Starbuck about wether it's worth adding it or not.

Starbuck5:
In my testing it is not 1.16 MB, it is 227 KB. That is today's main vs this branch (wheel files).

Maybe you were not accounting for the fact that wheels are zip files, so text is compressed inside them. Where as with images it's more 1-to-1 because they're already compressed.

Starbuck5 (approved):
Thanks!

ankith26 (approved):
I think the increase in size is fine, especially given the reductions we have gotten in recently.

LGTM, thanks for the PR! :sunglasses:

Issue #3385, opened 2025-03-14
pygame.display.set_mode printing RuntimeWarning when size set to (0, 0)


KennedyRichard:
Hello, dear dev team of the pygame-ce project,

Here's info on my system:

pygame-ce 2.5.3 (SDL 2.30.12, Python 3.10.12) on Xubuntu (GNU/Linux Ubuntu distro, 22.04.5 LTS)

Setting the screen size when calling `pygame.display.set_mode` with the size set to `(0, 0)` is printing the following warning:

```
RuntimeWarning: Requested window size was smaller than minimum supported window size on platform. Using (1920, 1031) instead.
```

I don't know when this new behaviour started, but as a developer, this looks unfitting, since `(0, 0)` is an acceptable value with expected consequences as explained in the  [documentation for pygame.display.set_mode](https://pyga.me/docs/ref/display.html#pygame.display.set_mode). In other words, if I'm not mistaken, there shouldn't be any warning at all.

For users that see the warning printed on their console this must be even more confusing and potentially cause them to think there's some sort of error in the app they are using if the app sets the size to `(0, 0)`.

In addition to all that, the documentation for `pygame.display.set_mode` doesn't even mention such warning. I don't know when this was introduced, but I recently updated my pygame-ce from 2.5.0 to 2.5.3 and in 2.5.0 there wasn't any warning.

Please, let me know whether this is normal behaviour or a bug (in which case I can suppress the warning to avoid confusing the end user).

damusss:
I believe this to be a bug. The code that raises the warning uses a variable w_actual (and h_actual) but that variable is not updated in the special case where the size is (0, 0), so the code thinks that the OS rejected the size, even if it's not the case. I should be able to fix it. The warning is really useful if you request something like (1, 0) on windows (non-borderless window) for example, because your window cannot be less wide than the controls buttons, so Windows will resize it to the smallest available (or at least I'm pretty sure that's what happening? you get the idea)

oddbookworm:
This is something I introduced because of linux behavior forcing a resize when the window was too small. It's when the OS sends a resize event to us automatically before `set_mode` is finished. I didn't consider the case of `(0, 0)` though when writing that patch, so oops

oddbookworm:
If anyone wants to beat me to a patch (I'm not at my pc right now), it's these lines: https://github.com/pygame-community/pygame-ce/blob/main/src_c/display.c#L1434-L1449

oddbookworm:
[...]

PR #3386, opened 2025-03-14
Remove warning with zero size & RESIZABLE in display.set_mode

damusss:
Probably fixes https://github.com/pygame-community/pygame-ce/issues/3385 .
NEW approach: don't warn if (0, 0) size is used with `pygame.RESIZABLE`. Explanation below
I humbly ask @oddbookworm (original implementation) and Kennedy Richard (issue opener) to test this PR if they can to see if it is fixed.

### DIFF
--- a/src_c/display.c
+++ b/src_c/display.c
@@ -1440,10 +1440,10 @@ pg_set_mode(PyObject *self, PyObject *arg, PyObject *kwds)
             ((surface->surf->flags & SDL_WINDOW_FULLSCREEN_DESKTOP) != 0)) {
             char buffer[150];
             char *format_string =
-                "Requested window size was smaller than minimum supported "
-                "window size on platform. Using (%d, %d) instead.";
-            snprintf(buffer, sizeof(buffer), format_string, surface->surf->w,
-                     surface->surf->h);
+                "Requested window was forcibly resized by the OS.\n\t"
+                "Requested window size: (%d, %d)\n\tNew window size: (%d, %d)";
+            snprintf(buffer, sizeof(buffer), format_string, w_actual, h_actual,
+                     surface->surf->w, surface->surf->h);
             if (PyErr_WarnEx(PyExc_RuntimeWarning, buffer, 1) != 0) {
                 return NULL;
             }
--- a/src_c/display.c
+++ b/src_c/display.c
@@ -988,6 +988,8 @@ pg_set_mode(PyObject *self, PyObject *arg, PyObject *kwds)
            desktop resolution without breaking compatibility. */
             w = display_mode.w;
             h = display_mode.h;
+            w_actual = w; /* avoid warning false positives */
+            h_actual = h;
         }
 
         if (flags & PGS_FULLSCREEN) {
--- a/src_c/display.c
+++ b/src_c/display.c
@@ -871,6 +871,7 @@ pg_set_mode(PyObject *self, PyObject *arg, PyObject *kwds)
     SDL_Surface *newownedsurf = NULL;
     int depth = 0;
     int flags = 0;
+    int zero_size = 0;
     int w, h, w_actual, h_actual;
     PyObject *size = NULL;
     int vsync = SDL_FALSE;
@@ -988,8 +989,7 @@ pg_set_mode(PyObject *self, PyObject *arg, PyObject *kwds)
            desktop resolution without breaking compatibility. */
             w = display_mode.w;
             h = display_mode.h;
-            w_actual = w; /* avoid warning false positives */
-            h_actual = h;
+            zero_size = 1;
         }
 
         if (flags & PGS_FULLSCREEN) {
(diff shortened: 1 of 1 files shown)


KennedyRichard:
> I humbly ask @\oddbookworm (original implementation) and Kennedy Richard (issue opener) to test this PR if they can to see if it is fixed.

I'd love too, but I'm not tech savvy when it comes to using pygame-ce directly from the repo (I usually only install it directly from pip). Is there any compilation steps required? Whichever the case, if there's any sort of guide I can follow available, I'd love to follow it and report my results.

oddbookworm:
> I'd love too, but I'm not tech savvy when it comes to using pygame-ce directly from the repo (I usually only install it directly from pip). Is there any compilation steps required? Whichever the case, if there's any sort of guide I can follow available, I'd love to follow it and report my results.

You just need to install the correct wheel from the CI. Since you're on linux and python 3.10, I'll grab the appropriate wheel for you and post that here for you in a sec. Then you can just pip install it

oddbookworm:
Unzip this and then pip install it (I'm assuming you're on x86_64 lol)
[pygame_ce-2.5.4.dev1-cp310-cp310-manylinux_2_17_x86_64.manylinux2014_x86_64.zip](https://github.com/user-attachments/files/19256209/pygame_ce-2.5.4.dev1-cp310-cp310-manylinux_2_17_x86_64.manylinux2014_x86_64.zip)

KennedyRichard:
Sorry for the wait, I had to test under different circumstances to check a few assumptions.

It didn't solve the problem (at least in my system). However, I noticed that the warning appears when I set the flag to RESIZABLE. If I don't set a flag, the warning doesn't appear, even though I set the size to `(0, 0)`. I didn't test with other flags though.

I tested the wheel both with Nodezator and with this small fixture I put together:

```python

from random import randint

import pygame as pg



pg.init()

sc = pg.display.set_mode((0, 0), pg.RESIZABLE)

scr = sc.get_rect()
bg = pg.Surface(sc.get_size()).convert()
bg.fill('white')

circle = pg.font.Font(None, 200).render('\N{bullet}', True, 'black').convert_alpha()
circle_rect = circle.get_rect()
circle_rect.center = scr.center


running = True

maintain_fps = pg.time.Clock().tick

while True:

    maintain_fps(30)

    ###

    for event in pg.event.get():

        if event.type == pg.QUIT:

            pg.quit()
            quit()

        elif event.type == pg.KEYDOWN:

            if event.key == pg.K_ESCAPE:

                pg.quit()
                quit()

    ###

    sc.blit(bg, (0, 0))

    circle_rect.move_ip(randint(0, 20), randint(0, 20))

    if not scr.colliderect(circle_rect):
        circle_rect.center = scr.center

    sc.blit(circle, circle_rect)

    ###

    pg.display.update()
```

KennedyRichard:
Here where I live it is starting to get late, so I have to retire for the night. I also have a commitment for the entire morning, so if you still require my assistance I'll be glad to help when I get back to my pc tomorrow after lunch.

oddbookworm:
Unfortunately, I can't currently reproduce this on my end either. My current linux install is being too nice, so you might be the only one currently available to test this...

oddbookworm:
Ok, I couldn't reproduce the issue you reported, but the issue present with RESIZABLE windows is reproducible. Here's a snip of some outputs I produced by slightly modifying the output message
image

oddbookworm:
So RESIZBLE windows are actually still subject to being force resized by the OS, and my warning is properly working. We're properly scaling it to the monitor's size, but then the OS is slighly scaling it down (probably to account for things like the task bar). We can probably add in an exception to just suppress the warning in this very specific case where we scale up a (0, 0) window with the RESIZABLE flag and the OS resizes it back down

damusss:
Alright @oddbookworm I'll do that. It makes sense, because we were already deciding the size of the window ourselves, so it doesn't make sense to warn if the size we chose wasn't liked by the OS. The size was going to be different from (0,0) anyways

KennedyRichard:
> Ok, I couldn't reproduce the issue you reported, but the issue present with RESIZABLE windows is reproducible.

Don't worry, the RESIZABLE flag was also causing the problem. It just happens that when I posted the issue I hadn't realized this yet, but as I said after, the RESIZABLE flag was indeed part of the problem:

> I noticed that the warning appears when I set the flag to RESIZABLE. If I don't set a flag, the warning doesn't appear, even though I set the size to (0, 0)

Since your new solution takes this into account I believe it should work now for everyone. Even so, I'll test it again here in my system, I think I figured out where to grab a wheel (on the Actions tab of the repo, from the item that has the same name as this PR).

KennedyRichard:
Okay, confirmed: the new changes solve the issue here on my end as well.

Thank you all again for the awesome work. :+1:

KennedyRichard:
Just one more question, if I may: would it be safe to assume these changes will make into 2.5.4?

Just wanted to know because I put an if-else clause in Nodezator that will [suppress the warning](https://docs.python.org/3/library/warnings.html#temporarily-suppressing-warnings) in case the size is `(0, 0)`, the flag is RESIZABLE and pygame-ce versions are 2.5.2 (when the bug was introduced) or 2.5.3 (the current version where the bug persists).

damusss:
@KennedyRichard yes, this bugfix will be released with version 2.5.4 since the previous ones are already released. You may suppress this warning for 2.5.2 and 2.5.3.

ankith26 (approved):
LGTM, thanks! :star:

oddbookworm (approved):
LGTM! Thanks

PR #3389, opened 2025-03-19
Move Surface, draw, transform and image docs to stubs

zoldalma999:
More progress on #2757


Starbuck5:
This PR has a few merge conflicts.

ankith26:
I did a side by side comparison of the old docs and the new docs

## Surface
- `get_frect` does not have a versionadded (in both versions) despite it being a relatively new method. Not the fault of this PR, just something I noticed.
- `_pixels_address` is documented before this PR, but this PR removes documentation for it. I think this is something to be fixed here.

## Tranform

- `pygame.transform.threshold` has an example section before this, which is not in this PR. If this change was intentional, I think it is fine.

zoldalma999:
> get_frect does not have a versionadded (in both versions) despite it being a relatively new method. Not the fault of this PR, just something I noticed.

Fixed

> _pixels_address is documented before this PR, but this PR removes documentation for it. I think this is something to be fixed here.

Good catch, forgot that private members need to be added to the rst explicitly. Fixed.

> pygame.transform.threshold has an example section before this, which is not in this PR. If this change was intentional, I think it is fine.

Yep, that is intentional. It was a test that was included anyway, I don't think that is a good example. But of course can be added back if others disagree.

ankith26 (approved):

Starbuck5:
I also combed through it, found a few things.

Surface.get_locks() return type looks strange
Surface.get_at() changes what previously seemed to be a note into a changed in 1.9 notice, drops a created in 1.9 notice.
image.load_basic has a line break between the 3rd and 4th sentence not in current docs

Starbuck5 (approved):
I trust you to resolve those notes as practical, if worth it. But I'm also going to approve, thanks for the PR!

ankith26:
> Surface.get_at() changes what previously seemed to be a note into a changed in 1.9 notice, drops a created in 1.9 notice.

I noticed that too, and assumed it's a fix made in this PR

PR #3395, opened 2025-04-04
Fix CI by avoiding bugged version of `pyproject-metadata` package

aatle:
`pyproject-metadata` version `0.9.1` has an incorrect dependency minimum version requirement for `packaging`, sometimes causing crashes during building, when `pyproject-metadata==0.9.1` and `packaging<23.2`. See https://github.com/pypa/pyproject-metadata/pull/239, issue got fixed but version `0.9.1` was released with the bug.

This affected #3379. Then it also failed tests on my fork's main branch, at https://github.com/aatle/pygame-ce/actions/runs/14256274454, so the fact that tests aren't failing on main right now is by chance.

Fix: Require that `pyproject-metadata` is not `0.9.1`.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -59,6 +59,7 @@ requires = [
   "cython<=3.0.11",
   "sphinx<=8.1.3",
   "sphinx-autoapi<=3.3.2",
+  "pyproject-metadata!=0.9.1",
 ]
 build-backend = 'mesonpy'
 


oddbookworm (approved):
LGTM! But I’d prefer to see this restriction lifted when the most recent release works

oddbookworm:
> I'm not sure I understand. Won't we still need the restriction when there is a new release of `pyproject-metadata`?

Isn't this due to a borked dependency in a specific release version?

Starbuck5 (approved):
Thanks for looking into it!

PR #3399, opened 2025-04-09
Update dependencies

Starbuck5:
SDL2 2.32.2 -> 2.32.4
SDL3 3.2.6 -> 3.2.10
SDL3_image 3.2.0 -> 3.2.4
SDL3_tff 3.2.0 -> 3.2.2


Starbuck5:
I was hoping to do that in a separate PR. But I agree, we should get their Webp animation fix in before we release.

ankith26 (approved):
I don't mind that being done in a future PR if it's not done in this one.

Thanks! :partying_face:

PR #3400, opened 2025-04-12
Run `pre-commit autoupdate`

ankith26:
This PR fixes the latest `dev.py format` fails on main.

While I was at it I also took the liberty of updating our formatting a little bit and getting the latest versions of stuff


Starbuck5 (approved):
Thanks!

Issue #3019, opened 2024-07-25
Method Proposal: pygame.geometry.Line.closest_point(point, do_clamp=True)


damusss:
The `Line` object has not been ported yet, but for the time it will, I propose to add the method in the title.

The method returns the closest point on the line relative to another point. The math for it isn't super straight forward, and unless you studied that somehow you probably need to search it online, that's why I think it should be added.

The `do_clamp` parameter controls weather the endpoints are ignored. The "default" math returns a point that can exceed A and B, but this could be what the user wants, so it could be optionally clamped.

bilhox:
I didn't understand what the feature does, it might be more elaboration and a schema.

MrRedstone058:
> The `Line` object has not been ported yet, but for the time it will, I propose to add the method in the title.
> 
> The method returns the closest point on the line relative to another point. The math for it isn't super straight forward, and unless you studied that somehow you probably need to search it online, that's why I think it should be added.
> 
> The `do_clamp` parameter controls weather the endpoints are ignored. The "default" math returns a point that can exceed A and B, but this could be what the user wants, so it could be optionally clamped.

to understand what youre saying, is is basically finding the point in a line that is directly perpendicular to the point? (in the case of do_clamp = false, the closest point follows the gradient of the original line)?

damusss:
@bilhox @MrRedstone058 I drawed what this method is supposed to do:
![image](https://github.com/user-attachments/assets/bf235846-1f26-4314-8b4f-af45c3df9979)
Now, with do_clamp = False, we have:
![image](https://github.com/user-attachments/assets/76b2b3ce-71b1-4f94-a683-f5ad72f019bb)
and with do_clamp = True, we have:
![image](https://github.com/user-attachments/assets/9ba9e85b-0064-4307-888d-53c450775f95)

MrRedstone058:
yup, just as i thought. i do not know the best way to do the math for this, but i can try to work it out as a base

bilhox:
After a discussion, we agreed to rename this function to ``Line.project``.

PR #3402, opened 2025-04-14
Implemented the Line.project

XFajk:
This PR should add the feature requested in #3019 
I added the `project()` method to the `pygame.geometry.Line` object I added the docs and test for the method 

## Summary by CodeRabbit

* **New Features**
  * Added Line.project(point, clamp=False), which returns the closest point on a line to a given point. When clamp is True, the result is restricted to the line segment. Handles zero-length lines safely.

* **Documentation**
  * Added API documentation for Line.project with usage details, figures, version note, and a warning about zero-length lines.

* **Tests**
  * Added unit tests covering projections, clamped behavior, and degenerate-line scenarios.


src_c/line.c
```diff
@@ -219,6 +283,54 @@ pg_line_scale_ip(pgLineObject *self, PyObject *const *args, Py_ssize_t nargs)
...
+    int do_clamp = 0;
+
+    static char *kwlist[] = {"point", "do_clamp", NULL};
+
+    if (!PyArg_ParseTupleAndKeywords(args, kwnames, "O|p:project", kwlist,
+                                     &point_obj, &do_clamp)) {
+        return RAISE(
+            PyExc_TypeError,
+            "project requires a sequence(point) and an optional clamp flag");
+    }
```
itzpr3d4t0r, line 299: All other geometry methods are METH_FASTCALL, so it's best to stick with that imo, especially because of the performance benefit and the small number of parameters.

src_c/line.c
```diff
@@ -179,6 +179,70 @@ _line_scale_helper(pgLineBase *line, double factor, double origin)
...
+            projection[0] = 0;
+            projection[1] = 0;
+        }
+    }
+
+    double projected_point[2] = {line->ax + projection[0],
+                                 line->ay + projection[1]};
+
+    PyObject *projected_tuple =
+        Py_BuildValue("(dd)", projected_point[0], projected_point[1]);
```
itzpr3d4t0r, line 241: We have a faster alternative for this in pygame-ce.

src_c/line.c
```diff
@@ -219,6 +283,54 @@ pg_line_scale_ip(pgLineObject *self, PyObject *const *args, Py_ssize_t nargs)
...
+    double point[2];
+
+    if (!PySequence_Check(point_obj) || PySequence_Size(point_obj) != 2) {
+        return RAISE(
+            PyExc_ValueError,
+            "project requires the point to be a sequence of 2 elements");
+    }
+
+    item = PySequence_GetItem(point_obj, 0);
+    point[0] = PyFloat_AsDouble(item);
```
itzpr3d4t0r, line 311: We use custom api to get doubles from pyobjects. Look at similar occurences inside the geometry module.

itzpr3d4t0r (changes requested):

src_c/line.c
XFajk: can I ask where and how is it called

src_c/line.c
XFajk: I noticed this my self that all the other methods on the `Line` are `METH_FASTCALL`. The reason why I dindnt do this for this method  because it takes kwargs and I am not sure if I can implement it correctly using `METH_FASTCALL`. 

if I need to also want to get kwargs with a `METH_FASTCALL` method I can do `METH_FASTCALL | METH_KEYWORDS` which does give me a dictionary object containing the kwargs in the function so the function could look  like this 
```C
static PyObject *
pg_line_project(pgLineObject *self, PyObject *const *args, Py_ssize_t nargs, PyObject *kwnames)
```
The problem comes now. You indicated in the other problem with my code that I should use the `pg_TwoDoublesFromFastcallArgs()` function but if I use that it kinda creates a buggy API in my opinion.
because now lets say I do this 
```C
    double point[2] = {0.f, 0.f};
    int do_clamp = 0;

    if (!pg_TwoDoublesFromFastcallArgs(args, nargs, &point[0], &point[1])) {
        return RAISE(PyExc_TypeError,
                     "project requires a sequence of two numbers");
    }
    // here could be the code that grabs the do_clamp out of the kwargs dictionary I just didn't write it yet
```
now if I do write the function in these ways
```python
l.project((50, 50))
l.project((60, 60), do_clamp=True)
```
but if I do
```python
l.project((50, 50), True)
```
it would fail because of how `pg_TwoDoublesFromFastcallArgs()` works (witch I dont fully understand) but it is a problem because this should work there is no reason why calling the method like this should cause an error from the perspective of the user. Another way to ditch the `pg_TwoDoublesFromFastcallArgs()` for this method wich I could do but it would really work in my opinion similarly to how it works now that I would have to get the object from args[0] get the numbers from it and I would be doing really the same thing it would probably be faseter still so should I do that? or can should I just not care that `l.project((50, 50), True)` dosent work.

src_c/line.c
```diff
@@ -179,6 +179,70 @@ _line_scale_helper(pgLineBase *line, double factor, double origin)
...
+
+    double dot_product =
+        vector_from_line_start_to_point[0] * normalized_line_vector[0] +
+        vector_from_line_start_to_point[1] * normalized_line_vector[1];
+
+    double projection[2] = {dot_product * normalized_line_vector[0],
+                            dot_product * normalized_line_vector[1]};
+
+    if (do_clamp) {
+        if (dot_product > line_length) {
```
MightyJosip, line 227: Since line_length uses sqrt, which is relatively slow function, can you square both sides. So you will end up with `if (dot_product * dot_product > line_vector[0] * line_vector[0] + line_vector[1] * line_vector[1])`. This would help with the performance of this function.

Atm you also have sqrt that in the normalization part. Can you check https://stackoverflow.com/a/54010814, At first it looks like it is the algorithm without the square root to me, maybe it can be used here as well

src_c/line.c
MightyJosip: `pg_tuple_couple_from_values_double`.

So you want to replace lines 240-243 with `return pg_tuple_couple_from_values_double(projected_point[0], projected_point[1])`

buildconfig/stubs/pygame/geometry.pyi
```diff
@@ -197,3 +197,4 @@ class Line:
     def scale_ip(self, factor_and_origin: Point, /) -> None: ...
     def flip_ab(self) -> Line: ...
     def flip_ab_ip(self) -> None: ...
+    def project(self, point: Point, do_clamp: bool = False) -> Point: ...
```
aatle, line 200: Use `tuple[float, float]` as the return type instead of `Point`. Also use that in the docs.

docs/reST/ref/geometry.rst
```diff
@@ -740,3 +740,31 @@
          .. versionadded:: 2.5.3
 
       .. ## Line.flip_ab_ip ##
+
+   .. method:: project
+
+         | :sl:`projects the line onto the given line`
+         | :sg:`project(point, do_clamp=False) -> point`
+
+         Returns a new point(tuple[float, float]) that is projected onto the line like so
```
aatle, line 749: I think the docs are lacking for this new method. The examples are good, but they cannot be the only description of the behavior.
- Add a technical definition involving orthagonality or something.
- Add an intuitive definition: The result is the *closest* point on the extended line (or line segment if `do_clamp` is true) to the given point.
- In the signature, replace the return type `point` with `tuple[float, float]`, and then remove `(tuple[float, float])` from the description.

docs/reST/ref/geometry.rst
```diff
@@ -740,3 +740,31 @@
...
+
+         .. figure:: code_examples/project1.png
+            :alt: project method image
+
+            Example of what project method does if you input the green point you get back the yellow one.
```
aatle, line 754: Confusing grammar.
```suggestion
            Example of what project method does.
            Green point: input argument. Yellow point: return value.
```
Additionally, maybe the yellow point should be purple instead for consistency with the other examples.

aatle:
(discussion)
API suggestion: the `do_clamp` argument should be renamed to `clamp`, because the `do_` is redundant and unpythonic, especially for new API.

Personal API opinion: the positional/keyword clamp argument should be made keyword-only because positional boolean arguments are bad - see https://docs.astral.sh/ruff/rules/boolean-type-hint-positional-argument/ for reasoning.

src_c/line.c
itzpr3d4t0r: I think nargs refers to the number of positionals so you'll have to check the dict size to know if there are any kwargs.

XFajk:
I tried to satisfy every request I got here but I made one choice that might should be evaluated.
I decided to not check for the number or name of the kwargs passed it just checks if some kwargs where passed in there and if yes it check if the first one is truthy aka what that means even `l.project((10, 10), fdafads=True)` would work. When I was writing this I was like I am doing this in the name of performance but after I pushed I realized that we might want to support doing this `l.project(point=(10, 10))` and this would not work now so I can still change it or I can change the stub to look like this 
```python
def project(self, point: tuple[float, float], / clamp: bool = False) -> tuple[float, float]: ...
```
aka saying that the first arg is positional but I want to get this approved I guess

buildconfig/stubs/pygame/geometry.pyi
```diff
@@ -197,3 +197,6 @@ class Line:
     def scale_ip(self, factor_and_origin: Point, /) -> None: ...
     def flip_ab(self) -> Line: ...
     def flip_ab_ip(self) -> None: ...
+    def project(
+        self, point: tuple[float, float], clamp: bool = False
```
aatle, line 201: Sorry I wasn't clear. Only the return type should be changed to `tuple[float, float]`; the argument should stay `Point`.

The reason is that while the input argument should be as broad and abstract of a type as possible, the return type needs to be more concrete and narrow to allow easy usage.

docs/reST/ref/geometry.rst
```diff
@@ -740,3 +740,27 @@
...
 
       .. ## Line.flip_ab_ip ##
+
+   .. method:: project
+
+         | :sl:`projects the line onto the given line`
+         | :sg:`project(point: tuple[float, float], clamp=False) -> tuple[float, float]`
+
+         This method takes in a point and one boolean keyword argument clamp. It outputs an orthogonally projected point onto the line.
+         If clamp is True it makes sure that the outputted point will be on the line segment.
```
aatle, line 750:
```suggestion
         If clamp is `True` it makes sure that the outputted point will be on the line segment (which might not be orthogonal), and if it is `False` (the default) then any point on the infinitely extended line may be outputted.
         This method can be used to find the closest point on a line to the given point. The output is the unique point on the line or line segment that is the smallest distance away from the given point.
```
- Clarify `clamp` argument more
- Add intuitive definition (biggest use case)

docs/reST/ref/geometry.rst
```diff
@@ -740,3 +740,27 @@
...
+         .. figure:: code_examples/project.png
+            :alt: project method image
+
+            Example of how it projects the point on to the line purplish point is the point we want to project and the blue point is what you would get as a result
+
+
+         .. figure:: code_examples/project_clamp.png
+            :alt: project do_clamp=True image
+
+            Example of what the clamp argument changes if it is true it stays on the line segment
```
aatle, line 762:
```suggestion
            :alt: project clamp argument image

            Example of what the clamp argument changes. If it is `True`, the point is bounded between the line segment ends.
```

buildconfig/stubs/pygame/geometry.pyi
MightyJosip: Now when talking about this, not related to this PR, but maybe we should rename the typing.(Int)Point to typing.(Int)PointLike, just to be the same as the rest of typing module, and a little bit more clear to avoid stuff like that in the future. On the other hand, it is hard to do that now since the typing module is exposed in pygame API, so maybe in pg3

test/geometry_test.py
```diff
@@ -2201,6 +2201,24 @@ def test_meth_update(self):
         with self.assertRaises(TypeError):
             line.update(1, 2, 3)
 
+    def test_meth_project(self):
+        line = Line(0, 0, 100, 100)
+        test_point1 = (25, 75)
+        test_clamp_point1 = (100, 300)
+        test_clamp_point2 = (-50, -150)
+
```
aatle, line 2209: For testing, make sure that all code paths are covered. It seems that the `clamp=True`, `dot_product < 0` case is not covered.
Also rename `do_clamp` to `clamp`.

src_c/line.c
```diff
@@ -179,6 +179,66 @@ _line_scale_helper(pgLineBase *line, double factor, double origin)
...
+    double dot_product =
+        (vector_from_line_start_to_point[0] * line_vector[0] +
+         vector_from_line_start_to_point[1] * line_vector[1]) /
+        (line_vector[0] * line_vector[0] + line_vector[1] * line_vector[1]);
+
+    double projection[2] = {dot_product * line_vector[0],
+                            dot_product * line_vector[1]};
+
+    if (clamp) {
+        if (projection[0] * projection[0] + projection[1] * projection[1] >
```
aatle, line 223: There might be a potential bug with the condition.
When `dot_product < 0`, the line start point should be used instead of the endpoint. However, after this condition was changed, it may now occur when the projected point is beyond and behind the line start point.
This is because when both sides were squared to remove the expensive square root, the sign of the left value was lost (it is now always positive) and an extraneous solution was introduced from the squaring.

Test case: `line = Line(0, 0, 100, 100)`, and the input is `(-200, -200)`, `clamp=True`, expected output should be `(0, 0)`.

I think this can be easily fixed by moving the `dot_product < 0` case to be the first one.

buildconfig/stubs/pygame/geometry.pyi
aatle: I would normally agree with increasing naming consistency.
However, `Point` is not necessarily like the other types in that it is not a union of similar, acceptable types. It is also not a raw protocol like `SequenceLike`.
The main reason of course is verbosity; originally, `CoordinateLike` would be too long which might be the original reason, but `PointLike` is still long too.

src_c/line.c
XFajk: I tried the test case it passed so I dont think this is a problem because we are checking if the `dot_product` is bellow zero. and the only change from how I originally implemented the `dot_product` is that I made the dot product with the `line_vector` wich does not affect the sighn. And then I devided by the squared lenght of the `line_vector`. Wich will be always positive so that dosen't effect the signh either so it works.

Starbuck5:
I may have gone a little overboard, but here's my test script.

```py
"""
https://github.com/pygame-community/pygame-ce/pull/3402 testing
"""

import math

import pygame
import pygame.geometry

pygame.init()

screen_width, screen_height = 720, 500

screen = pygame.display.set_mode((screen_width, screen_height))
pygame.display.set_caption("Line.project example")

clock = pygame.Clock()
font = pygame.font.SysFont("Arial", 24)

line = pygame.geometry.Line((screen_width/2-20,250), (screen_width/2+20,250))

instructions = """
Use WASD/arrows to adjust the two points, respectively.
The position of the mouse is projected onto the blue line.
Toggle clamp by pressing C.
"""

instructions_surf = font.render(instructions.strip(), True, "black")

speed = 250
clamp = False

while True:
    dt = clock.tick(144) / 1000

        if event.type == pygame.QUIT or (event.type == pygame.KEYDOWN and event.key == pygame.K_ESCAPE):
            pygame.quit()
            raise SystemExit
        
        if event.type == pygame.KEYDOWN and event.key == pygame.K_c:
            clamp = not clamp

    if keys[pygame.K_w]:
        line.ay -= speed * dt
    if keys[pygame.K_a]:
        line.ax -= speed * dt
    if keys[pygame.K_s]:
        line.ay += speed * dt
    if keys[pygame.K_d]:
        line.ax += speed * dt
    if keys[pygame.K_UP]:
        line.by -= speed * dt
    if keys[pygame.K_LEFT]:
        line.bx -= speed * dt
    if keys[pygame.K_DOWN]:
        line.by += speed * dt
    if keys[pygame.K_RIGHT]:
        line.bx += speed * dt

    screen.fill("purple")

    screen.blit(instructions_surf, (10, 10))

    pygame.draw.line(screen, "blue", line.a, line.b, width=3)
    pygame.draw.circle(screen, "blue", line.a, 5)
    pygame.draw.circle(screen, "blue", line.b, 5)

    mouse = pygame.mouse.get_pos()

    proj_point = line.project(mouse, clamp=clamp)
    if not math.isnan(proj_point[0]) and not math.isnan(proj_point[1]):
        pygame.draw.line(screen, "red", mouse, proj_point)

    pygame.display.flip()
```

Starbuck5:
I ran into problems with this function returning NaNs (not a number) in some scenarios, pygame.draw.line freezes indefinitely when it gets one of them, opened an issue for that: https://github.com/pygame-community/pygame-ce/issues/3412. Is there a way this could not return NaN? Like if it's going to do that, return the original coordinate instead? IDK about implementation or whether that makes sense. Another option would be to add a warning in the docs that this can return NaN, because I think people will want to do stuff with this value and having it be NaN will mess those calculations up.
- Is there prior art from other game engines/libs we could steal? How have others dealt with this?

On another subject, the current implementation of fastcall and keywords has several issues. Yes, the code that should work works, but also code that shouldn't work works.

For example, this line completely works: `proj_point = line.project(mouse, damusssss=clamp, arugula="yadafasdf")`

I checked, there are 0 instances of a normal function usage of fastcall and keywords in the pygame-ce codebase, so this is not a standard this PR should be held to. Manual kwarg implementations are error prone and will have inconsistent error messaging relative to PyArg_ParseTupleAndKeywords.

I think this PR should either be positional only or should do keywords the normal way.

itzpr3d4t0r:
> I checked, there are 0 instances of a normal function usage of fastcall and keywords in the pygame-ce codebase, so this is not a standard this PR should be held to.

Surface.get_rect and get_frect are one example.
<https://github.com/pygame-community/pygame-ce/blob/be3d05a8f3ac9bcf52261593cded6c17c2112dfb/src_c/surface.c#L292>

Starbuck5:
Yeah, those aren't normal functions. Normal functions use the kwargs directly, surface.get_rect and rect.move_to (the other example) both just forward the keywords to another function. AKA they don't process them themselves.

XFajk:
So I can again rewrite it again to not be a fastcall I agree that implementing kwargs parsing for fastcall would be super error prone. I will also look into why it returns NaN and fix it.

> * Is there prior art from other game engines/libs we could steal? How have others dealt with this?

I can look but I dont think so. What problems do you have with the art? I can try to improve it

Starbuck5:
> I can look but I dont think so. What problems do you have with the art? I can try to improve it

Just to be clear, I wasn't referring to the graphics you put in the docs, I was referring to the NaN problem.

XFajk:
Okay, so I want to discuss how we are going to handle the NaN problem. The issue arises when the line has a length of zero, meaning both points A and B are at the same position. Mathematically, there is no way to project the point. If clamping is on, I can technically say the point should be equal to A or B, but if clamping is off, I don't know what it would return. The only solution I see as a possibility is to throw a ZeroDivisionError and mention it in the docs

XFajk:
I need help I dont understand why the github test aren't passing

MightyJosip:
> D:\a\pygame-ce\pygame-ce\src_c\line.c(5): fatal error C1083: Cannot open include file: 'pytypedefs.h': No such file or directory

Do you need those extra imports in lines 4/5. Because I can't see them anywhere else

XFajk:
> > D:\a\pygame-ce\pygame-ce\src_c\line.c(5): fatal error C1083: Cannot open include file: 'pytypedefs.h': No such file or directory
> 
> Do you need those extra imports in lines 4/5. Because I can't see them anywhere else

no that my stupid editor adding them in I will delete them

XFajk:
> > D:\a\pygame-ce\pygame-ce\src_c\line.c(5): fatal error C1083: Cannot open include file: 'pytypedefs.h': No such file or directory
> 
> Do you need those extra imports in lines 4/5. Because I can't see them anywhere else

no that my stupid editor adding them in I will delete them

XFajk:
> > D:\a\pygame-ce\pygame-ce\src_c\line.c(5): fatal error C1083: Cannot open include file: 'pytypedefs.h': No such file or directory
> 
> Do you need those extra imports in lines 4/5. Because I can't see them anywhere else
yeah I will delete them my editor added them I didnt need them

XFajk:
> > D:\a\pygame-ce\pygame-ce\src_c\line.c(5): fatal error C1083: Cannot open include file: 'pytypedefs.h': No such file or directory
> 
> Do you need those extra imports in lines 4/5. Because I can't see them anywhere else


yeah I will delete them my editor added them I didnt need them thanks I didnt know that can cause an issue

XFajk:
all new requested changes anything else?

.clangd
```diff
@@ -0,0 +1,2 @@
+CompileFlags:
+  CompilationDatabase: ./.mesonpy-build/
```
ankith26: what is this file doing, can you remove it?

XFajk: definitely my editor just needed that and I forgot to remove it before pushing

buildconfig/stubs/pygame/geometry.pyi
```diff
@@ -197,3 +197,4 @@ class Line:
     def scale_ip(self, factor_and_origin: Point, /) -> None: ...
     def flip_ab(self) -> Line: ...
     def flip_ab_ip(self) -> None: ...
+    def project(self, point: Point, /, clamp: bool = False) -> tuple[float, float]: ...
```
ankith26:
```suggestion
    def project(self, point: Point, clamp: bool = False) -> tuple[float, float]: ...
```
The slash signifies that everything before it is positional only, but in this case `point` could be passed as a keyword argument in the implementation and therefore isn't positional only.

docs/reST/ref/geometry.rst
```diff
@@ -740,3 +740,30 @@
...
+
+
+         .. figure:: code_examples/project_clamp.png
+            :alt: project clamp argument image
+
+            Example of what the clamp argument changes. If it is `True`, the point is bounded between the line segment ends.
+
+            WARNING: This method has to have some length or the clamp parameter must be true for it to work and not throw a `ValueError`
+
+         .. versionadded:: 2.5.4
```
ankith26: this would need to be updated to 2.5.6 after we have consensus on this PR

src_c/line.c
```diff
@@ -219,6 +219,86 @@ pg_line_scale_ip(pgLineObject *self, PyObject *const *args, Py_ssize_t nargs)
...
+
+    // this is a vector that goes from the start of the line to the point we
+    // are projecting onto the line
+    double vector_from_line_start_to_point[2] = {point[0] - line->ax,
+                                                 point[1] - line->ay};
+
+    double dot_product =
+        (vector_from_line_start_to_point[0] * line_vector[0] +
+         vector_from_line_start_to_point[1] * line_vector[1]) /
+        (line_vector[0] * line_vector[0] + line_vector[1] * line_vector[1]);
```
ankith26:
```suggestion
        squared_line_length;
```
we can make use of the precomputed value here

src_c/line.c
```diff
@@ -219,6 +219,86 @@ pg_line_scale_ip(pgLineObject *self, PyObject *const *args, Py_ssize_t nargs)
...
+
+    if (clamp) {
+        if (dot_product < 0) {
+            projection[0] = 0;
+            projection[1] = 0;
+        }
+        else if (projection[0] * projection[0] +
+                     projection[1] * projection[1] >
+                 line_vector[0] * line_vector[0] +
+                     line_vector[1] * line_vector[1]) {
```
ankith26:
```suggestion
                 squared_line_length) {
```
again, can make use of precomputed value here

src_c/line.c
```diff
@@ -219,6 +219,86 @@ pg_line_scale_ip(pgLineObject *self, PyObject *const *args, Py_ssize_t nargs)
     Py_RETURN_NONE;
 }
 
+static PyObject *
+_line_project_helper(pgLineBase *line, double *point, int clamp)
+{
+    // this is a vector that goes from one point of the line to another
+    double line_vector[2] = {line->bx - line->ax, line->by - line->ay};
+    double squred_line_length =
```
ankith26:
```suggestion
    double squared_line_length =
```
minor typo here

ankith26 (commented):
Left a few more code quality related reviews, but assuming those are solved this PR is looking good to me!

src_c/line.c
```diff
@@ -219,6 +219,85 @@ pg_line_scale_ip(pgLineObject *self, PyObject *const *args, Py_ssize_t nargs)
...
+        projected_point[0] = line->ax;
+        projected_point[1] = line->ay;
+        return pg_tuple_couple_from_values_double(projected_point[0],
+                                                  projected_point[1]);
+    }
+    else if (squared_line_length == 0.0) {
+        return RAISE(PyExc_ValueError,
+                     "The Line has to have some length or this method has to "
+                     "be clamped to work");
+    }
```
ankith26: API discussion point: is the current implementation better, or would it be better if the clamp and no clamp behaviour was identical when `squared_line_length == 0.0`? It seems like that essentially becomes "project a point onto a point" where there should be no notion of clamping because there are no two distinct points.

Me personally I don't have much of an opinion either way but I'd like to get other reviewers say on this front, for now I am not asking you to change anything here.

ankith26 (approved):
Left a point up for API discussion but apart from that this PR LGTM. I played around with the script from starbuck and got no funny nan-behaviour.

Thanks for contributing to pygame-ce and putting up with all our reviews! :tada: :smile:

src_c/line.c
damusss: I believe it should not error when the line has no length in any case. actually, it makes less sense to error when it is not clamped. there is one good available spot to project the point, which is the only one. The clamped-0-length behavior is the one that actually has an "exception" since the method doesn't know in what direction to project, resulting in the only valid spot.

damusss: No one would want their code to error because while the mouse was dragging or while the objects overlapped, the line got a length of zero. If you really do care about it working differently in that case, checking it before hand is straight forward.

aatle: I agree that these edge case errors should be avoided when possible.

For the 0-length non-clamped projection, the method could possibly either return the line point or the argument point. (Clamped returns the line point.) I think returning the line point is accurate though.

MyreMylar: OK, lets switch it to return the line point in all cases.

docs/reST/ref/geometry.rst
```diff
@@ -740,3 +740,30 @@
...
+
+            Example of how it projects the point onto the line. The red point is the point we want to project and the blue point is what you would get as a result.
+
+
+         .. figure:: code_examples/project_clamp.png
+            :alt: project clamp argument image
+
+            Example of what the clamp argument changes. If it is `True`, the point is bounded between the line segment ends.
+
+            WARNING: This method has to have some length or the clamp parameter must be true for it to work and not throw a `ValueError`
```
MyreMylar:
```suggestion
            WARNING:  If the line has no length (i.e. the start and end points are the same) then the returned point of this function will be the same point as both ends of the line.
```

src_c/line.c
MyreMylar:
```suggestion
    if (squared_line_length == 0.0) {
        double projected_point[2];
        projected_point[0] = line->ax;
        projected_point[1] = line->ay;
        return pg_tuple_couple_from_values_double(projected_point[0],
                                                  projected_point[1]);
    }
```

test/geometry_test.py
```diff
@@ -2201,6 +2201,40 @@ def test_meth_update(self):
...
+        self.assertEqual(math.ceil(projected_point[0]), 0)
+        self.assertEqual(math.ceil(projected_point[1]), 0)
+
+        projected_point = bad_line.project(test_bad_line_point, clamp=True)
+        self.assertEqual(math.ceil(projected_point[0]), 0)
+        self.assertEqual(math.ceil(projected_point[1]), 0)
+
+        # testing if the method fails when it should
+        with self.assertRaises(ValueError):
+            bad_line.project(test_bad_line_point)
```
MyreMylar:
```suggestion
        projected_point = bad_line.project(test_bad_line_point)
        self.assertEqual(math.ceil(projected_point[0]), 0)
```

PR #3406, opened 2025-04-19
Replace AddObject() with AddObjectRef()

MightyJosip:
Since Python 3.13 PyModule_AddObject() is [soft deprecated](https://docs.python.org/3.14/c-api/module.html#c.PyModule_AddObject). This PR replaces it with newer function, PyModule_AddObjectRef(). It was changed following the way adviced in [the official documentation](https://docs.python.org/3.14/extending/newtypes_tutorial.html#the-basics) between versions 3.9 <-> 3.14
image

### DIFF
--- a/src_c/bufferproxy.c
+++ b/src_c/bufferproxy.c
@@ -617,7 +617,8 @@ MODINIT_DEFINE(bufferproxy)
     c_api[2] = pgBufferProxy_GetParent;
     c_api[3] = pgBufferProxy_Trip;
     apiobj = encapsulate_api(c_api, "bufferproxy");
-    if (PyModule_AddObjectRef(module, PYGAMEAPI_LOCAL_ENTRY, apiobj)) {
+    if (PyModule_AddObject(module, PYGAMEAPI_LOCAL_ENTRY, apiobj)) {
+        Py_XDECREF(apiobj);
         Py_DECREF(module);
         return NULL;
     }
--- a/src_c/bufferproxy.c
+++ b/src_c/bufferproxy.c
@@ -604,10 +604,8 @@ MODINIT_DEFINE(bufferproxy)
         return NULL;
     }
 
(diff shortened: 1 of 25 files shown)


src_c/bufferproxy.c
```diff
@@ -604,10 +604,8 @@ MODINIT_DEFINE(bufferproxy)
         return NULL;
     }
 
-    Py_INCREF(&pgBufferProxy_Type);
-    if (PyModule_AddObject(module, "BufferProxy",
-                           (PyObject *)&pgBufferProxy_Type)) {
-        Py_DECREF(&pgBufferProxy_Type);
+    if (PyModule_AddObjectRef(module, "BufferProxy",
```
Starbuck5, line 607: If we're going through these anyways, might as well use PyModule_AddType so you don't need to do the PyType_Ready's either.

Starbuck5:
It's also confusing to see replacements for things that didn't previously have an incref.

MightyJosip:
> It's also confusing to see replacements for things that didn't previously have an incref.

Yea you are right, that comment is mostly for encapsulate_api, and in the current master if I print Py_REFCNT of apiobj, it is 1. However, after this PR it is indeed 2 like you said. I will remove it from those places, and use it only when we actually increase the ref

src_c/bufferproxy.c
MightyJosip: Actually, the problem with that is the following: at the moment, we still allow XType classes (like SurfaceType, RectType, EventType... are still declared. And AddType would use the name from .tp_name, so it would be regular "Surface".

Maybe do this for pygame3, once we remove those XType classes?

Starbuck5 (approved):
Changes make sense to me, thanks!

zoldalma999 (approved):
LGTM!

PR #3408, opened 2025-04-20
Fix `pygame.Vector{2,3}` initialization from numpy arrays

Matiiss:
Originally reported by @damusss on [the Pygame Community discord server](https://discord.com/channels/772505616680878080/772940667231928360/1363535610995806368)

numpy arrays (evidently) implement the numeric protocol, thus they go through the first check that checks if the object is a number and subsequently fail on conversion to a double because the array is >1 dimensional

MRE:
```py
import numpy as np
import pygame

arr = np.array([1, 0])
pygame.Vector2(arr)
```
Traceback:
```bash
pygame-ce 2.5.3 (SDL 2.30.12, Python 3.12.8)
TypeError: only length-1 arrays can be converted to Python scalars

The above exception was the direct cause of the following exception:

  File "/home/matiiss/projects/Something/for_so.py", line 5, in <module>
    pygame.Vector2(arr)
SystemError: <class 'pygame.math.Vector2'> returned a result with an exception set
```

### DIFF
--- a/src_c/math.c
+++ b/src_c/math.c
@@ -2280,19 +2280,23 @@ static int
 _vector2_set(pgVector *self, PyObject *xOrSequence, PyObject *y)
 {
     if (xOrSequence) {
-        if (RealNumber_Check(xOrSequence)) {
-            self->coords[0] = PyFloat_AsDouble(xOrSequence);
-            /* scalar constructor. */
-            if (y == NULL) {
-                self->coords[1] = self->coords[0];
-                return 0;
-            }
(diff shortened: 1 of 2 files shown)


damusss:
checking from phone but I think you copy pasted tests to the vector 3 without changing them to vector 3?

ankith26 (approved):
LGTM, thanks for the PR :+1:

damusss (approved):
Thank you for fixing this! :) looks good to me

PR #3410, opened 2025-04-21
Fix `__class_getitem__` to be positional-only

ankith26:
As noted by @aatle in https://github.com/pygame-community/pygame-ce/pull/3398, `__class_getitem__` is positional only, and stubtest on python 3.13 complains about this issue. This PR fixes that.


gresm:
I've looked around and failed to find information about key parameter being required a positional-only:
[PEP-560: original spec](https://peps.python.org/pep-0560/#class-getitem)
[Python documentation](https://docs.python.org/3/reference/datamodel.html#object.__class_getitem__)
[Possible alternative that it's ``Generic`` stuff instead - also negative](https://github.com/python/typeshed/blob/10b4c620d4f27b429576e02698bd0aa733061358/stdlib/typing.pyi#L201)

aatle (approved):
If you would like, you can rename `generic` to `item` to be consistent with typeshed.

LGTM;

ankith26:
I updated the code to use `item` instead of `generic` as per the review.

I also want to keep code changes out of this stubs PR so I am not changing the parent classes in the implementation, though someone could take it up in the future if interested.

Starbuck5:
I played around a bit with list's implementation, it being positional only seems to be standard.

```py
>>> list.__class_getitem__(key=int)
  File "<python-input-9>", line 1, in <module>
    list.__class_getitem__(key=int)
    ~~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^
TypeError: list.__class_getitem__() takes no keyword arguments
```

Also ofc class getitem is meant to be used by type hints, and trying to use a keyword there is a syntax error.

```py
>>> list[key=int]
  File "<python-input-5>", line 1
    list[key=int]
         ^^^^^^^
SyntaxError: invalid syntax. Maybe you meant '==' or ':=' instead of '='?
```

Starbuck5 (approved):

damusss (approved):
LGTM, thanks!

PR #3417, opened 2025-04-28
Remove surflock as the module

MightyJosip:
Atm, surflock is empty module, without any function, also without documentation. However, it is needed for Surface. This PR removes the module, and moves the functionality to the surface.

Python users shouldn't notice any changes (nor they were ever supposed to use that module)

However, if someone has been using pygame C API, old functions have been moved from surflock to surface module, so they would need to update their code


Starbuck5:
Why not leave it as a separate file but use the functions in surface.c?

I think the original intention of this file was to split up some related surface functionality. Not have files so long.

>  I find it absolutely impossible for this change to break anyone's code
> obsolete pygame submodules

Yes this could break someone's code, no it's not an obsolete submodule just because it doesn't export functions directly to the user. pygame.base doesn't export functions to the user either. Surflock has a C API and if anyone's using it, this will absolutely break their code. I know we don't typically care about C API backwards compat but the C API is publicly documented.

Starbuck5:
Last year I spent quite a while trying to merge pygame-ce submodules, rwobject was my early target. I was hopeful it could increase the compiler ability to optimize, make our codebase more typical for a C codebase, and I got promising results that a smaller number of binaries would be meaningfully more space efficient. I couldn't get a solution good enough to incrementally port some modules using the classic C API and some modules not using it.

Anyways I support this effort in general but our files are already so long, and the surflock stuff is so nicely separated into a different unit of abstraction, I would like to keep it a support file even if it's not technically a Python module.

MightyJosip:
Valid points. My main goal for this was to remove this part https://github.com/pygame-community/pygame-ce/pull/3417/files#diff-41261be2b7f66f350124346cb97d5e6af46016b07c7d82d4fa9120b359def0bfL208-L243, since it has no use for the python side. Anyways, I guess I will wait a bit and reconsider this, coz I have really big refactor changes in progress

PR #3418, opened 2025-04-28
Remove controller.c from gitignore

MightyJosip:
controller.c is not cython generated file (?) so it shouldn't be ignored by git


ankith26 (approved):
Good catch, just checked the source and yes `controller.c` does exist in there. Thanks for the PR :tada:

damusss (approved):
Thanks!

PR #3421, opened 2025-04-29
Move modules to their own directories

MightyJosip:
Big pygame refactor


Starbuck5:
Hey MightyJosip, this is a very interesting branch to navigate around to navigate the concept. It's a very cool concept, I'm still in support.

Things I notice navigating around:
I noticed that while colordict is technically a private dep of the color module and goes in the color folder, pypm/surflock/constants/locals are put as separate modules from their public facing brethren.

I feel like more grouping is a better, like what you have as _debug, base, constants, locals, macosx, pkgdata, pygame, and rwobject could all be one grouping. Pypm should definitely go part of midi.

I'm not sure a good place for static and resources, you've put them here like they're modules, which is a little strange but I don't have a better suggestion right now.

----------------------------

I don't believe we can or should do this as one huge PR, rather an incremental effort with smaller PRs.

MightyJosip:
Hi, thanks for comment. Yea this is more like proof of concept, we can still discuss it in more details, and split the commit.

Just one question, what is your opinion about X_test.py files, should they be moved together with the code as well, or it should be in the specific folder (so we would have src/ and test/ split up, just like now). In this commit it is together, but I can split it again

Starbuck5:
I'm okay with tests in the respective folders, but they can also be migrated over in separate PRs, in the sake of making the PRs simpler.

PR #3428, opened 2025-05-12
Port `display.c` to SDL3

ankith26:
WIP to port `display.c` to SDL3

## Summary by CodeRabbit

- New Features
  - SDL3 compatibility for display and window management.
  - Updated display enumeration, mode queries, and refresh rate reporting.
  - Unified integer scaling behavior across SDL2 and SDL3.
  - Improved window/WM info retrieval and fullscreen handling.

- Bug Fixes
  - More reliable fullscreen transitions and state detection across platforms.
  - Safer pixel format handling.

- Refactor
  - Broad internal updates to support dual SDL2/SDL3 code paths without changing public APIs.

- Chores
  - Display module is now always built.

### DIFF
--- a/src_c/display.c
+++ b/src_c/display.c
@@ -417,7 +417,11 @@ pgVidInfo_New(const pg_VideoInfo *i)
 static pg_VideoInfo *
 pg_GetVideoInfo(pg_VideoInfo *info)
 {
+#if SDL_VERSION_ATLEAST(3, 0, 0)
+    const SDL_DisplayMode *mode_ptr;
+#else
     SDL_DisplayMode mode;
+#endif
     PG_PixelFormat *tempformat;
     Uint32 formatenum;
     pgSurfaceObject *winsurfobj;
@@ -443,11 +447,19 @@ pg_GetVideoInfo(pg_VideoInfo *info)
         info->vfmt = &info->vfmt_data;
     }
     else {
+#if SDL_VERSION_ATLEAST(3, 0, 0)
+        if ((mode_ptr = SDL_GetCurrentDisplayMode(0))) {
+            info->current_w = mode_ptr->w;
+            info->current_h = mode_ptr->h;
+            formatenum = mode_ptr->format;
+        }
+#else
         if (SDL_GetCurrentDisplayMode(0, &mode) == 0) {
             info->current_w = mode.w;
             info->current_h = mode.h;
             formatenum = mode.format;
         }
+#endif
         else {
             info->current_w = -1;
             info->current_h = -1;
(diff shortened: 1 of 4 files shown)


ankith26:
This PR should be done from my side.

Given that this is a kinda big PR to review, I have already organized everything into commits that take up stuff one at a time and focus on it fully. So, if it helps reviewing I can split this up into multiple PRs each with subsets of commits from here.

src_c/window.c
```diff
@@ -331,6 +331,10 @@ pg_window_set_fullscreen(SDL_Window *window, int desktop)
             goto end;
         }
     }
+
```
damusss, line 334: What is this doing?

ankith26: Basically while working on fullscreen changes in display I saw that `SDL_SetWindowFullscreenMode` in SDL3 has effect only if the window is already in fullscreen mode according to the docs. So I added a call here as well that does that first.

src_c/display.c
```diff
@@ -2787,31 +3476,27 @@ pg_display_resize_event(PyObject *self, PyObject *event)
...
         return PyLong_FromLong(-1);
     }
 
-    // could also take the size of the old display surface
-    SDL_GetWindowSize(win, &window_w, &window_h);
-    window_display = SDL_GetWindowDisplayIndex(win);
-    if (SDL_GetDesktopDisplayMode(window_display, &display_mode) != 0) {
-        return RAISE(pgExc_SDLError, SDL_GetError());
-    }
-
```
Starbuck5, line 2814: This was just unused?

Starbuck5 (approved):
Thanks Ankith! I looked through the changes and it seems good.

This did take a significant amount of time to read through, maybe if some of the smaller commits were split into their own PRs they could be reviewed more quickly by others? Not sure the best move.

When I combine with my surface branch I'm unable to open a window, but this PR is for compiling and not necessarily for getting everything to run.

```
>>> screen = pygame.display.set_mode((500, 500))
pygame.error: Invalid display
```

src_c/display.c
ankith26: Yeah it was just unused in this function. Just a relic of copy pasting stuff IG

ankith26:
I rebased this PR on main to get the latest surface compilation changes (hence the force push). While I was at it I also fixed a super minor return code related issue in an SDL3 specific part of this PR, and as a result `set_mode` now works on SDL3! * 

*: `set_mode` on SDL3 only works when called after `pygame.init()` as of now. This can be fixed but needs a bit more code restructuring that would also affect SDL2 paths, hence I am keeping that out of this PR.

With this PR we have some examples in a somewhat usable/runnable state on SDL3

ankith26:
@coderabbitai full review

src_c/display.c
```diff
@@ -427,15 +450,28 @@ pg_GetVideoInfo(pg_VideoInfo *info)
...
         info->current_w = winsurf->w;
         info->current_h = winsurf->h;
-        info->vfmt_data = *(winsurf->format);
+        PG_PixelFormat *fmt = PG_GetSurfaceFormat(winsurf);
+        if (!fmt) {
+            PyErr_SetString(pgExc_SDLError, SDL_GetError());
+            return (pg_VideoInfo *)NULL;
+        }
+        info->vfmt_data = *fmt;
         info->vfmt = &info->vfmt_data;
```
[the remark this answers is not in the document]
ankith26: no

PR #3429, opened 2025-05-13
Document `pygame._render`

damusss:
Ported the docs from sdl2_video to `pygame._render`, using the autoapi docs system (I added one line to make it work). I added the module to the list of experimentals, warning about it being currently incomplete. The constructor of Renderer and the blit methods are the only one currently referencing things that have not been ported. Tho it won't be like this anymore since the PRs to port them are already up. I can change if needed, but I don't think so as this is an experimental, privated module with a big warning.

If you don't want the people to know it exists, I can remove the link inside index.rst. But the _render.rst and the docs in the stubs should still be present so it doesn't have to be added all at once. I want to propose new features aswell, and I'd like for the basic docs to already exist to be able to do that.


ankith26:
#3519 and #3520 may have implications on this PR

MyreMylar (approved):
OK, LGTM. Keeping these up to date is a moving target as render adds more features, but this is a good start!

PR #3432, opened 2025-05-17
Fix SRCALPHA TGA image saving

ankith26:
This PR is a fix for an issue @fladd raised on the og pygame repo.

Essentially, images with per pixel alpha were not being saved to TGA correctly because there was blending happening in intermediate steps that should not happen. This was being prevented for blanket alpha and color key alpha cases already in the code, but not for per pixel alpha cases. This PR implements a more generic blendmode based approach so that all cases are handled correctly.

Reproducer code for this issue (by @fladd, from the original issue)

```python3
import pygame

pygame.init()
screen = pygame.display.set_mode((800, 600))

circle_color = (255, 0, 0, 255)
circle_surface = pygame.Surface((800, 600), pygame.SRCALPHA)
pygame.draw.circle(circle_surface, circle_color, (400, 300), 100)

# Save the circle surface as an image
pygame.image.save(circle_surface, 'image.tga')
```

### DIFF
--- a/src_c/image.c
+++ b/src_c/image.c
@@ -1690,7 +1690,7 @@ SaveTGA_RW(SDL_Surface *surface, SDL_RWops *out, int rle)
     int alpha = 0;
     struct TGAheader h;
     int srcbpp;
-    Uint8 surf_alpha;
+    SDL_BlendMode surf_blendmode;
     int have_surf_colorkey = 0;
     Uint32 surf_colorkey;
     SDL_Rect r;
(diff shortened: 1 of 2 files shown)


ankith26:
Note that with this PR my aim is to not change existing behaviour for other types of images, even though I have noticed inconsistencies with the code.
- blanket alpha is completely ignored in all cases... which is fine but it is also easily fixable, but may also break someones usecase
- colorkey is handled in the 8 bit path but not in the 24/32 bit path

Starbuck5:
What do you think about dropping TGA in 3.0? It's not exactly a popular format.

ankith26:
Given the inconsistencies and the fact that we maintain this code on our side, it makes sense to drop it in a breaking release IMO

ankith26:
@coderabbitai full review

MyreMylar (approved):
Can confirm this fixes the 'smear' bug identified in the original issue and that saving to tga now works fine with alpha and non-alpha per-pixel pixels preserved.

👍

PR #3434, opened 2025-05-22
Port over some SDL functions to SDL3

Starbuck5:
There are a lot of functions that used to return 0 on success, negative error code on failure that in SDL3 return true on success and false on failure. This means our old code compiles but fails at runtime. This PR ports a few of them.

This PR introduces little PG_ compat macros for SDL_SetPaletteColors, SDL_SetSurfaceBlendMode, SDL_GetSurfaceBlendMode, SDL_GetSurfaceAlphaMod, SDL_SetSurfaceAlphaMod.

I only replaced instances of these functions that are actually error checked, I didn't replace them globally. I also fixed a typo in an error message in surf_set_palette.


ankith26 (approved):
LGTM thanks for the PR :tada:

MightyJosip (approved):
LGTM, you defined all the modified functions, and change conditions the right way with the SDL3 (from == 0 to != 0, and other alternatives)

👍

PR #3435, opened 2025-05-24
Surface module compiling in SDL3

Starbuck5:
This PR consists of 4 commits, and it gets the surface module to compile with SDL3. It still fails lots of tests when run in SDL3 mode, but there is lots of SDL-function porting to be done with int->bool stuff.

#### Commit 1 focuses on Surface.c

This commit does not change the SDL2 codepath at all, except for a cliprect thing similar to what I've done in other PRs.

The RLEaccel stuff is confusing, even digging through the SDL2 source it's hard to understand. Our code uses `surf->flags & SDL_RLEACCEL` to check, interchangeably with `SDL_SurfaceHasRLE(surf)`, but if you check the source SDL_SurfaceHasRLE does not check the same flags. However, in testing it seems to be equivalent, somehow.

I also just declined to port a whole section that covers a blitting edge case, we'll get to it later, for now I think it will be valuable to have people able to compile at all.

EDIT: also added a commit 5 for surface.c.

#### Commit 2 and 3 focus on filling and blitting respectively

There are a lot of lines changed but it's just passing formats around, nothing complex

#### Commit 4 makes intrinsics compile properly

I took Ankith's prior work on this and moved it up to _surface.h, as well as making sure it defines __SSE2__ and __AVX2__ itself if necessary. On my (Windows) system it doesn't compile at least the SSE2 stuff without this, so in SDL2 we must be getting that define from SDL itself.

### DIFF
--- a/src_c/surface.c
+++ b/src_c/surface.c
@@ -1496,8 +1496,6 @@ surf_set_alpha(pgSurfaceObject *self, PyObject *args)
     PyObject *alpha_obj = NULL, *intobj = NULL;
     Uint8 alpha;
     int alphaval = 255;
-    SDL_Rect sdlrect;
-    SDL_Surface *surface;
 
     if (!PyArg_ParseTuple(args, "|Oi", &alpha_obj, &flags)) {
         return NULL;
@@ -1546,24 +1544,24 @@ surf_set_alpha(pgSurfaceObject *self, PyObject *args)
     bool success =
         PG_SetSurfaceRLE(surf, (flags & PGS_RLEACCEL) ? SDL_TRUE : SDL_FALSE);
     /* HACK HACK HACK */
-#if SDL_VERSION_ATLEAST(3, 0, 0)
-    if (SDL_SurfaceHasRLE(surf) && (!(flags & PGS_RLEACCEL)))
-#else
-    if ((surf->flags & SDL_RLEACCEL) && (!(flags & PGS_RLEACCEL)))
-#endif
-    {
+    // TODO: SDL3: figure out how to port this or if it's relevant to SDL3.
+#if !SDL_VERSION_ATLEAST(3, 0, 0)
+    if ((surf->flags & SDL_RLEACCEL) && (!(flags & PGS_RLEACCEL))) {
         /* hack to strip SDL_RLEACCEL flag off surface immediately when
            it is not requested */
+        SDL_Rect sdlrect;
         sdlrect.x = 0;
         sdlrect.y = 0;
         sdlrect.h = 0;
         sdlrect.w = 0;
 
-        surface = PG_CreateSurface(1, 1, PG_SURF_FORMATENUM(surf));
+        SDL_Surface *surface =
+            PG_CreateSurface(1, 1, PG_SURF_FORMATENUM(surf));
 
         SDL_LowerBlit(surf, &sdlrect, surface, &sdlrect);
         SDL_FreeSurface(surface);
     }
+#endif
     /* HACK HACK HACK */
     if (success) {
         success = PG_SetSurfaceAlphaMod(surf, alpha);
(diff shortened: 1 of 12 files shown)


src_c/surface.c
```diff
@@ -1546,8 +1546,14 @@ surf_set_alpha(pgSurfaceObject *self, PyObject *args)
     pgSurface_Prep(self);
     result =
         SDL_SetSurfaceRLE(surf, (flags & PGS_RLEACCEL) ? SDL_TRUE : SDL_FALSE);
+
     /* HACK HACK HACK */
-    if ((surf->flags & SDL_RLEACCEL) && (!(flags & PGS_RLEACCEL))) {
+#if SDL_VERSION_ATLEAST(3, 0, 0)
+    if (SDL_SurfaceHasRLE(surf) && (!(flags & PGS_RLEACCEL)))
+#else
+    if ((surf->flags & SDL_RLEACCEL) && (!(flags & PGS_RLEACCEL)))
```
ankith26: `SDL_SurfaceHasRLE` basically checks if RLE was requested on a surface (and in this context should always be equivalent to `(flags & PGS_RLEACCEL)` because of the lines above this) which would basically make this whole block a `if (false)` thing that never runs.
`surf->flags & SDL_RLEACCEL` OTOH checks if the surface actually got RLE.

The problem here is that SDL doesn't expose the equivalent of `(surf->flags & SDL_RLEACCEL)` anymore. This is something that bites us with sdl2-compat as well (ref: https://github.com/libsdl-org/sdl2-compat/issues/476) and as we have unit tests that fail because of this, we can tell that they don't have the same behavior.

As for this specific block, I think a more appropriate thing that reduces confusion would be to simply ifdef out the whole block in SDL3 and add a comment explaining that there's no way to do the check anymore (and its probably not even required in SDL3)

Starbuck5: I'm unaware of any changes to RLE internals in SDL3, so if this block isn't necessary in SDL3 is it necessary in SDL2?

ankith26: Added a print statement inside this block and ran the tests, it got into that path once. Found out that we have one test case that depends on this behaviour: `test_solarwolf_rle_usage_2`. So seems like this logic is necessary actually.

ankith26: ~~I think that using `SDL_MUSTLOCK` (which is equivalent to `(((S)->flags & SDL_RLEACCEL) != 0)` on SDL2) does the same job on SDL3 as well, as an implementation detail~~
NVM `SDL_MUSTLOCK` is also now equivalent to `SDL_SurfaceHasRLE` apparently in SDL3

Starbuck5: @ankith26 Just getting back to this, I pushed a new commit that comments out the whole block in SDL3, as you initially suggested. Also added a "TODO" comment. If it's something we need to worry about with SDL3 we can worry about the fine details later, right now I'd just like to get surface compiling.

I also rebased.

src_c/surface.c
```diff
@@ -3017,9 +3021,15 @@ surf_get_flags(PyObject *self, PyObject *_null)
     if (PG_SurfaceHasRLE(surf)) {
         flags |= PGS_RLEACCELOK;
     }
+#if SDL_VERSION_ATLEAST(3, 0, 0)
+    if (SDL_SurfaceHasRLE(surf)) {
```
ankith26, line 3025: It would be good to also have a TODO comment here

src_c/surface.c
```diff
@@ -4381,7 +4391,6 @@ surface_do_overlap(SDL_Surface *src, SDL_Rect *srcrect, SDL_Surface *dst,
     int x, y;
     int w = srcrect->w, h = srcrect->h;
     int maxw, maxh;
-    SDL_Rect *clip = &dst->clip_rect;
```
ankith26, line 4383: Here the clip is `dstclip` but you are replacing it with `srcclip`? Am I missing something, do they happen to be equivalent?

Starbuck5: Ugh, sorry about that. Fixed it. Thanks for spotting.

ankith26 (approved):

PR #3441, opened 2025-05-24
Move mouse and key docs to stubs

zoldalma999:
#2757


ankith26 (approved):
LGTM thanks for the PR :tada:

aatle:
I think a `type: ignore[misc]` comment can be added after the header of each implementation definition.

buildconfig/stubs/pygame/mouse.pyi
```diff
@@ -1,4 +1,74 @@
...
+                  pygame.quit()
+                  return
+               elif event.type == MOUSEWHEEL:
+                  print(event)
+                  print(event.x, event.y)
+                  print(event.flipped)
+                  print(event.which)
+                  # can access properties with
+                  # proper notation(ex: event.y)
+         clock.tick(60)
```
aatle: Small improvement while we are here.
```suggestion
   def main():
       while True:
           for event in pygame.event.get():
               if event.type == QUIT:
                   pygame.quit()
                   return
               elif event.type == MOUSEWHEEL:
                   print(event)
                   print(event.x, event.y)
                   print(event.flipped)
                   print(event.which)
                   # can access properties with
                   # proper notation(ex: event.y)
           clock.tick(60)
```

PR #3442, opened 2025-05-24
Use `namedtuple` for `pygame.image.load_animation`

damusss:
As requested in the pygame discord server, I made load_animation return a list of namedtuples so frame.surface and frame.delay are possible. The typehinting ensures proper types for both attribute access and item access.

### DIFF
--- a/src_c/imageext.c
+++ b/src_c/imageext.c
@@ -69,6 +69,7 @@
 static SDL_mutex *_pg_img_mutex = 0;
 #endif
 */
+static PyTypeObject *animation_frame_type = NULL;
 
 static char *
 iext_find_extension(char *fullname)
@@ -216,6 +217,8 @@ imageext_load_animation(PyObject *self, PyObject *arg, PyObject *kwargs)
     SDL_RWops *rw = NULL;
     static char *kwds[] = {"file", "namehint", NULL};
 
+    assert(animation_frame_type);
+
     if (!PyArg_ParseTupleAndKeywords(arg, kwargs, "O|s", kwds, &obj, &name)) {
         return NULL;
     }
@@ -263,12 +266,22 @@ imageext_load_animation(PyObject *self, PyObject *arg, PyObject *kwargs)
         /* The python surface object now "owns" the sdl surface, so set it
          * to null in the animation to prevent double free */
         surfs->frames[i] = NULL;
-
-        PyObject *listentry = Py_BuildValue("(Oi)", frame, surfs->delays[i]);
-        Py_DECREF(frame);
+        PyObject *listentry = PyStructSequence_New(animation_frame_type);
         if (!listentry) {
+            Py_DECREF(frame);
             goto error;
         }
+
+        PyStructSequence_SetItem(listentry, 0, frame);
+        PyObject *delay_obj = PyLong_FromLong(surfs->delays[i]);
+        if (!delay_obj) {
+            Py_DECREF(listentry);
+            Py_DECREF(frame);
+            goto error;
+        }
+        PyStructSequence_SetItem(listentry, 1, delay_obj);
+        Py_DECREF(delay_obj);
+
         PyList_SET_ITEM(ret, i, listentry);
     }
     IMG_FreeAnimation(surfs);
@@ -461,7 +474,6 @@ MODINIT_DEFINE(imageext)
         return NULL;
     }
     import_pygame_rwobject();
-
     if (PyErr_Occurred()) {
         return NULL;
     }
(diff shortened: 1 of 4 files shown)


src_c/imageext.c
```diff
@@ -476,6 +488,22 @@ MODINIT_DEFINE(imageext)
...
+        .doc = NULL,
+        .fields = _namedtuple_fields,
+        .n_in_sequence = 2,
+    };
+
+    animation_frame_type = PyStructSequence_NewType(&_namedtuple_descr);
+    if (animation_frame_type == NULL) {
+        return NULL;  // exception set
+    }
+    Py_INCREF(animation_frame_type);
```
ankith26, line 505: Isn't `animation_frame_type` already a strong reference at this point, this feels extra

src_c/imageext.c
```diff
@@ -263,12 +266,22 @@ imageext_load_animation(PyObject *self, PyObject *arg, PyObject *kwargs)
...
+
+        PyStructSequence_SetItem(listentry, 0, frame);
+        PyObject *delay_obj = PyLong_FromLong(surfs->delays[i]);
+        if (!delay_obj) {
+            Py_DECREF(listentry);
+            Py_DECREF(frame);
             goto error;
         }
+        PyStructSequence_SetItem(listentry, 1, delay_obj);
+        Py_DECREF(delay_obj);
```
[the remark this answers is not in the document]
damusss: It does steal it!

src_c/imageext.c
```diff
@@ -263,12 +268,14 @@ imageext_load_animation(PyObject *self, PyObject *arg, PyObject *kwargs)
         /* The python surface object now "owns" the sdl surface, so set it
          * to null in the animation to prevent double free */
         surfs->frames[i] = NULL;
-
-        PyObject *listentry = Py_BuildValue("(Oi)", frame, surfs->delays[i]);
-        Py_DECREF(frame);
+        PyObject *listentry = PyStructSequence_New(animation_frame_type);
         if (!listentry) {
+            Py_DECREF(frame);
             goto error;
```
ankith26, line 274: needs a `delay_obj` decref here.

src_c/imageext.c
```diff
@@ -254,6 +255,10 @@ imageext_load_animation(PyObject *self, PyObject *arg, PyObject *kwargs)
     }
 
     for (int i = 0; i < surfs->count; i++) {
+        PyObject *delay_obj = PyLong_FromLong(surfs->delays[i]);
+        if (!delay_obj) {
+            goto error;
+        }
         PyObject *frame = (PyObject *)pgSurface_New(surfs->frames[i]);
         if (!frame) {
```
ankith26, line 263: needs a `delay_obj` decref here.

src_c/imageext.c
damusss: right, since I moved it

src_c/imageext.c
```diff
@@ -254,21 +255,29 @@ imageext_load_animation(PyObject *self, PyObject *arg, PyObject *kwargs)
     }
 
     for (int i = 0; i < surfs->count; i++) {
+        PyObject *delay_obj = PyLong_FromLong(surfs->delays[i]);
```
ankith26, line 258: slightly off PR scope but now that I noticed this, should we make this a float to be future-proof? Yes, this is still an int in `SDL3_image` but who knows what will happen in `SDL{N > 3}_image`? People who need to work with ints can always round it on their side.

damusss:
This PR should wait for Starbuck's opinion before merging

src_c/imageext.c
```diff
@@ -477,6 +484,21 @@ MODINIT_DEFINE(imageext)
...
+        .name = "_AnimationFrame",
+        .doc = NULL,
+        .fields = _namedtuple_fields,
+        .n_in_sequence = 2,
+    };
+
+    animation_frame_type = PyStructSequence_NewType(&_namedtuple_descr);
+    if (animation_frame_type == NULL) {
+        return NULL;  // exception set
+    }
```
MightyJosip, line 500: Should this part be moved to 448 (before MODINIT)

damusss: Maybe the _namedtuple_fields and _namedtuple_descr, NewType probably needs to happen there

Starbuck5:
Ankith and I discussed this a little bit, and we decided it would be best to leave it as is for now-- given that it's less code for us, standard types can be easier to understand, and we might just replace the whole thing with an animation class in the future.

Thank you for working up the PR.

PR #3443, opened 2025-05-24
Bump build dependencies

ankith26:
Bumps our build dependencies.

The sphinx/sphinx-autoapi bump fixes an issue where docs fail to build due to the older versions not being constrained on astroid releases properly.


Starbuck5:
Did you check that our autoapi generated pages still generate properly?

ankith26:
It fixes a warning/error I was having locally, but when I tried opening the docs they were also not being generated properly I'm assuming due to some issues in my setup. I'm yet to investigate this in depth.

ankith26:
I just checked out the wheels generated by CI from this PR and the docs on there looked completely fine, going over a few modules with a side by side check against pyga.me docs

PR #3444, opened 2025-05-25
Use float for `load_animation` delay, improve docs

damusss:
Since we are still in time, convert to using floats as the delay for animation frames. SDL has started using floats after int multiple times when changing version so this future proofs us. Since the docs specify milliseconds (and, if the named tuple is added, also the attribute) there is no ambiguity.

### DIFF
--- a/src_c/imageext.c
+++ b/src_c/imageext.c
@@ -264,7 +264,8 @@ imageext_load_animation(PyObject *self, PyObject *arg, PyObject *kwargs)
          * to null in the animation to prevent double free */
         surfs->frames[i] = NULL;
 
-        PyObject *listentry = Py_BuildValue("(Oi)", frame, surfs->delays[i]);
+        PyObject *listentry =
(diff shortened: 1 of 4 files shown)


src_c/imageext.c
```diff
@@ -264,7 +264,8 @@ imageext_load_animation(PyObject *self, PyObject *arg, PyObject *kwargs)
          * to null in the animation to prevent double free */
         surfs->frames[i] = NULL;
 
-        PyObject *listentry = Py_BuildValue("(Oi)", frame, surfs->delays[i]);
+        PyObject *listentry =
+            Py_BuildValue("(Of)", frame, (float)surfs->delays[i]);
```
ankith26, line 268:
```suggestion
            Py_BuildValue("(Od)", frame, (double)surfs->delays[i]);
```
nit: lets just pick double so that we are a little more precise.

Issue #3423, opened 2025-05-05
Add `pygame.math.Vector2.to_tuple()`, `pygame.math.Vector2.to_list()`, `pygame.math.Vector3.to_tuple()`and `pygame.math.Vector3.to_list()`


LittleBread69:
Although I believe the title is self-explanatiory, I will try to clarify what I mean by `.to_tuple()`, `.to_list()`.

Here's an example code:
```
import pygame

my_vector = pygame.math.Vector2(300, 200)
my_big_vector = pygame.math.Vector3(100, 300, 200)
```
Let's say that there is a situation when I want to pass `my_vector` into a function like this:
```
def some_function(argument: tuple[ int | float, int | float ]:
    #The code does something
    ...
```
If I try to pass `my_vector` in the `argument slot`, I will get the type checker marking it as an error, and in many cases, it will throw an error.
The way to bypass this would be to make a new `tuple` with `my_vector.x` and `my_vector.y`: `tupled_vector2 = (my_vector.x, my_vector.y)`. This is counterproductive.
Thus, I propose adding a `.to_tuple()` to both `Vector2` and `Vector3`.

In strict python code, it would look something alike to:
```
from typing import Literal, Optional

class Vector2:
    #All the code already present in the class would be still here

    def to_tuple(self,\
                order: Optional[Literal['xy', 'xx', 'yx', 'yy']]\
                ) -> tuple[float, float]:
        if order is None:
            return (self.x, self.y)
        match order:
            case 'xy':
                return (self.x, self.y)
            case 'xx':
                return (self.x, self.x)
            case 'yx':
                return (self.y, self.x)
             case 'yy':
                return (self.y, self.y)
            case _:
                raise ValueError(f"Invalid order: {order}. Must be 'xy', 'xx', 'yx', 'yy' or None.")
```
_The reason there's no else, is because the return statement prevents the execution of the following code if order is None. I made the code on purpose [PEP 484](https://peps.python.org/pep-0484/) compliant._

With that said, the only difference with `Vector2.to_list()` would be that it returns a list instead of a tuple, and `Vector3.to_tuple()` and `Vector3.to_list()` would do the same, although with a length of 3. _Also, for Vector3, there would be considerably more possibilities, one of the reasons I used `match-case` instead of `if-elif`._

[...]

PR #3446, opened 2025-05-27
add doc for vec conversion

ToniDevStuff:
https://github.com/pygame-community/pygame-ce/issues/3423

### DIFF
--- a/docs/reST/ref/math.rst
+++ b/docs/reST/ref/math.rst
@@ -35,7 +35,7 @@ subscripts
    v.y == v[1]
    v.z == v[2]
 
-Multiple coordinates can be set using swizzling or slices
+Multiple coordinates can be set using slices or swizzling
 
 ::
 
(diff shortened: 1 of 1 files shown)


oddbookworm:
Here's what the current version looks like:
image

Although, the original issue had a bit more in the example given, specifically swizzling, which isn't in your example. Could you add that part of the example?

Also, keep in mind that swizzling isn't just for two-length
```
>>> v = pygame.Vector2(1, 2)
>>> 
>>> list(v) == [1.0, 2.0]
True
>>> tuple(v) == (1.0, 2.0)
True
>>> set(v) == {1.0, 2.0}
True
>>> set(v)
{1.0, 2.0}
>>> set(v)
{1.0, 2.0}
>>> set(v) == {2.0, 1.0}
True
>>> tuple(v.xxy)
(1.0, 1.0, 2.0)
```

oddbookworm:
But overall, great job! Thanks for contributing! 🥳

docs/reST/ref/math.rst
```diff
@@ -35,14 +35,38 @@ subscripts
...
+
+    v = pygame.Vector3(1, 2, 3)
+
+    list(v.xz) == [1.0, 3.0]
+    tuple(v.yyy) == (2.0, 2.0, 2.0)
+    set(v.xxyyzz) == {1.0, 2.0, 3.0} # sets remove duplicates
+
+    list(v[:]) == [1.0, 2.0, 3.0]
+    tuple(v[::-1]) == (3.0, 2.0, 1.0)
+    set(v[1:3]) == {2.0, 3.0}
```
oddbookworm, line 68: I would like to see order changed up some. Everything here is either in strictly increasing or strictly decreasing order (or x, y, z order). Maybe throw in some stuff like xyxzxzzyyz just to get the point across

oddbookworm (changes requested):
Just got a couple small comments and it should be good to go soon

docs/reST/ref/math.rst
```diff
@@ -35,14 +35,38 @@ subscripts
    v.y == v[1]
    v.z == v[2]
 
-Multiple coordinates can be set using slices or swizzling
+Multiple coordinates can be set using swizzling or slices
```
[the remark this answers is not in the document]
ToniDevStuff: To be honest, I didn’t realize myself that I had changed this line.
It kinda makes more sense now in my head since the following example uses swizzling first and then slices, but it’s kinda pointless unless you have OCD or something, lol.

docs/reST/ref/math.rst
ToniDevStuff: Good point, I have implemented your change.

oddbookworm: Thanks

oddbookworm (approved):
LGTM! Thanks! 🎉

PR #3447, opened 2025-05-28
Elaborate on Vector conversion to tuples/lists

DickerDackel:
Attempt at solving issue https://github.com/pygame-community/pygame-ce/issues/3423 - "Update vector doc to have a brief mention of how to convert to lists or tuples"

### DIFF
--- a/docs/reST/ref/math.rst
+++ b/docs/reST/ref/math.rst
@@ -35,13 +35,32 @@ subscripts
    v.y == v[1]
    v.z == v[2]
 
-Multiple coordinates can be set using slices or swizzling
+Multiple coordinates can be set and retrieved using slices or swizzling.
 
 ::
 
    v = pygame.Vector2()
    v.xy = 1, 2
    v[:] = 1, 2
+   print(v)  # Vector2(1, 2)
+   print(v.x)  # 1.0
+   print(v.y)  # 2.0
+   print(v.xy)  # Vector2(1, 2)
+   print(v.yx)  # Vector2(2, 1)
+   print(v.xyyx)  # (1.0, 2.0, 2.0, 1.0)
+
+Note above, that swizzlling with 2 components will return a Vector2 instance
+again, while more than 2 components will result in a tuple. But since vectors
+support the iterator protocol, they can be by unpacked, or converted to lists
+or tuples.
+
+::
+
+   v = Vector2(1, 2)
+   print(*v)  # 1.0 2.0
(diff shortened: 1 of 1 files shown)


docs/reST/ref/math.rst
```diff
@@ -35,13 +35,32 @@ subscripts
...
    v.xy = 1, 2
    v[:] = 1, 2
+   print(v)  # Vector2(1, 2)
+   print(v.x)  # 1.0
+   print(v.y)  # 2.0
+   print(v.xy)  # Vector2(1, 2)
+   print(v.yx)  # Vector2(2, 1)
+   print(v.xyyx)  # (1.0, 2.0, 2.0, 1.0)
+
+Note above, that swizzlling with 2 components will return a Vector2 instance
```
oddbookworm, line 52: Typo in "swizzling"

You don't have to actually commit this suggestion. Feel free to commit like normal through git CLI, I'm just demonstrating that a reviewer can suggest a change and you have the choice to then commit the suggestion directly from github if you want to. If you were to choose to commit this directly from github, it would add me as a co-author of the commit since I suggested the commit.
```suggestion
Note above, that swizzling with 2 components will return a Vector2 instance
```

oddbookworm (commented):
Thanks for the contribution! 🥳
I just have one minor nitpick about it. As I mentioned in [the contributing channel in the PGC discord server](https://discord.com/channels/772505616680878080/772940667231928360/1377458026532831284), I don't think this pull is mutually exclusive of #3446, I think it's quite possible for both pulls to merge (although there is a merge conflict in line 38, in which I think this pull's version should win out because it does actually add more detail as opposed to just swapping the order of the words "slices" and "swizzling", which I think are actually in the correct grammatical order here, but I don't really care about that lol)

docs/reST/ref/math.rst
```diff
@@ -35,13 +35,32 @@ subscripts
...
+again, while more than 2 components will result in a tuple. But since vectors
+support the iterator protocol, they can be by unpacked, or converted to lists
```
oddbookworm, line 54: "they can be by unpacked"
I didn't catch it the first time, but the grammar here is weird

DickerDackel: Right, fragment from sentence rework.

Fixed and pushed both of your remarks.  Thanks!

oddbookworm (approved):
LGTM! Thanks! 🎉

PR #3449, opened 2025-05-30
Deprecate all Surface format related methods

damusss:
The docs were already static "not needed for pygame usage" - then, why keep? SDL3 went in a direction of simplifying, it makes sense for us to do the same. As @ankith26 pointed out, while that information exists, you don't quite have the chance to use it. To remove this methods in pygame 3, it makes sense to deprecate them now, unless you have a reason not to do that.
[edit:]
Maybe some of the data, like the masks, could be useful. But I don't think they are all useful, perhaps we can choose what we deprecate and what not. If we do care about pixel information I originally thought about a PixelFormat class, but I don't know anymore.

I also want to bring the attention towards locking - Surfaces have 5 separate methods related to locking - but when do users actually lock them? Is this much method cluttering useful? I guess if the amounts of sequential operations are great for an hardware surface only locking once does indeed make sense - but at least `mustlock`, `get_locks` and `get_locked` could go away, as I still don't find a use for them (calling lock/unlock multiple times does no harm!). If you want I can "reduce" them to two, deprecating the others (to be removed in pygame 3). Thanks for the attention.

### DIFF
--- a/buildconfig/stubs/pygame/surface.pyi
+++ b/buildconfig/stubs/pygame/surface.pyi
@@ -776,18 +776,24 @@ class Surface:
         ..versionadded:: 2.3.0
         """
 
+    @deprecated("since 2.5.5. Not needed for pygame usage")
     def get_bitsize(self) -> int:
         """Get the bit depth of the Surface pixel format.
 
         Returns the number of bits used to represent each pixel. This value may
         not exactly fill the number of bytes used per pixel. For example a 15 bit
         Surface still requires a full 2 bytes.
+
+        .. deprecated: 2.5.5 This value is not needed for normal pygame usage.
         """
 
+    @deprecated("since 2.5.5. Not needed for pygame usage")
     def get_bytesize(self) -> int:
         """Get the bytes used per Surface pixel.
 
         Return the number of bytes used per pixel.
+
+        .. deprecated: 2.5.5 This value is not needed for normal pygame usage.
         """
 
     def get_flags(self) -> int:
@@ -821,6 +827,7 @@ class Surface:
             PREALLOC       0x01000000    # Surface uses preallocated memory
         """
 
+    @deprecated("since 2.5.5. Not needed for pygame usage")
     def get_pitch(self) -> int:
         """Get the number of bytes used per Surface row.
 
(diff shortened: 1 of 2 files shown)


damusss:
yesterday I was very bipolar! (I changed ideas about everything in a short amount of time, not in a medical sense)

I originally thought of implementing pygame.surface.PixelFormat because I remember us talking about it, and now that sdl3 was out it could be a thing (compatible with sdl2 of course). But then I was pointed out about how a lot of the info wasn't necessary, so I sort of panicked and went "oh wait maybe it's right, why did I want to port sdl stuff that pygame users wouldn't care about?" so I went on the completely extreme opposite side "then let's get rid of it all (add an evil laugh)"

I guess my mind mainly cared about cleaning up Surface's methods, regardless of whether an alternative was provided. Re-reading some GitHub conversations with you I also noticed how format in sdl3 is slightly incomplete compared to sdl2 because the palette is left out, but at the same time I'm not sure if that's "so bad", we could just think of the palette as something belonging to the surface and not to its format.

Since I had nothing better to do, I actually started implementing that class for "fun". if you (and reading what you said, I have to refresh my mind and re-connect that you are right) think that - masks -bitsize -bytssize - formatname then maybe, just maybe that class still makes sense, unless the useful information is too "low" so it can be left in Surface and not worry about it forever. (Also pitch might be useful for making raw surface data from other libraries.) I'm not sure why I care to know about the name of a surface format even tho I don't think I'll ever need that information, maybe just curiosity? lol

@Starbuck5 sorry for the confusion my mind went through and tell me what you think about the PixelFormat "tantrum".

Starbuck5:
Treating a palette as part of the surface is already what we do, which is nice. E.g. get_palette.

In 3.0 I would like a format class, so we can more closely expose SDL functionality / easier way to provide that power to our users. But designing something to work in SDL3 and then back porting it to SDL2 based pygame-ce is not trivial at all. I also would like to see a design for how this would be used in other places in our API e.g. in convert or surface init.

Starbuck5:
I'd rather see a plan for the API than a branch with ideas, at this point in time. Do some research, look at what's available with SDL pixelformats, both functions, and macros. Also I think a lot of this can and should be written in Python.

PR #3451, opened 2025-05-31
Port over more SDL functions to SDL3

Starbuck5:
This PR handles SDL_LockSurface and SDL_FillRect.

This is a prerequisite to https://github.com/pygame-community/pygame-ce/pull/3435, since that one is giving me warnings (that I don't get locally) related to these functions.

This is similar to https://github.com/pygame-community/pygame-ce/pull/3434


damusss (approved):
I like the inline strategy to avoid having conditional preprocessives all over the place convoluting the code. LGTM!

MightyJosip (approved):
LGTM, you defined all the modified functions, and change conditions the right way with the SDL3 (from == 0 to != 0, and other alternatives)

👍

Starbuck5:
(Just force pushed a rebase)

> I like the inline strategy to avoid having conditional preprocessives all over the place convoluting the code. LGTM!

And they're all in the SDL3 style, so in the future when we only support SDL3 we can just switch PG_ to SDL_.

ankith26 (approved):
LGTM thanks for the PR :tada:

PR #3453, opened 2025-06-01
Simplify math module init

Starbuck5:
This removes the unused math C API, and switches to the new PyModule_AddType instead of PyType_Ready + PyModule_AddObjectRef.


MightyJosip (approved):
Good change

PR #3455, opened 2025-06-01
Improve, fix and clarify Surface docs

damusss:
Some parts of the docs had smaller errors. Some parts were basically lying. Big methods and the constructor where very convoluted throwing information randomly. I didn't change everything, but rephrased things. Also, in the constructor and in convert, I thought that the docs weren't even trying to explain how the arguments worked, so I looked at the actual code and refined them. I'm open for suggestions of course.


buildconfig/stubs/pygame/surface.pyi
```diff
@@ -46,77 +46,69 @@ class Surface:
...
-    matching the colorkey will not be drawn. The surface alpha value is a single
-    value that changes the transparency for the entire image. A surface alpha of
-    255 is opaque, and a value of 0 is completely transparent.
-
-    Per pixel alphas are different because they store a transparency value for
-    every pixel. This allows for the most precise transparency effects, but it
-    also the slowest. Per pixel alphas cannot be mixed with surface alpha and
-    colorkeys.
+    When creating a new Surface it will be cleared to all black or all transparent
+    depending on the format. The only required arguments is the size, a sequence
```
oddbookworm: singular "argument"

buildconfig/stubs/pygame/surface.pyi
```diff
@@ -46,77 +46,69 @@ class Surface:
...
+    containing the width and height of the Surface. With no additional arguments
+    the format will best match the display Surface.
```
oddbookworm: What if there is no display surface?

buildconfig/stubs/pygame/surface.pyi
```diff
@@ -46,77 +46,69 @@ class Surface:
...
+         allowed producing the corresponding formats: ``8``:``indexed``, ``12``:``XRGB4444``,
+         ``15``:``XRGB1555``, ``16``:``RGB565``, ``24``:``RGB/BGR24``, ``32``: ``XRGB8888``.
+       * Providing a bit depth and also a sequence of four integers describing the
+         bitmasks that will result in a precise format. Normal Surfaces don't require
+         masks as they are intended for advanced pixel Surface control.
+
+    An indexed format requires a palette for its colors. While pygame provides a default
+    palette, it can also be controlled with the appropriate methods.
+
+    Surfaces additionally have alpha, colorkey and rectangle clipping. These charateristics
```
oddbookworm: Misspelling of "characteristics"

buildconfig/stubs/pygame/surface.pyi
```diff
@@ -46,77 +46,69 @@ class Surface:
...
+    mainly effect how the Surface is blitted to other Surfaces. The blit routines will
```
oddbookworm: "affect"

buildconfig/stubs/pygame/surface.pyi
```diff
@@ -46,77 +46,69 @@ class Surface:
...
+    attempt to use hardware acceleration when possible, otherwise they will use highly
+    optimized software blitting methods. The default clipping area is the whole
+    Surface. Otherwise, only that area will be able to be modified.
+
+    There are three types of transparency supported for a pygame Surface:
+       * Per-pixel alpha. It requires a format with an alpha channel and cannot
+         be combined with other transparency options. Every pixel has its
+         own alpha value and is the slowest but most flexible method.
```
oddbookworm: I'm pretty sure we can mix per-pixel and global alpha

buildconfig/stubs/pygame/surface.pyi
```diff
@@ -46,77 +46,69 @@ class Surface:
...
+         :meth:`Surface.premul_alpha` for more.
+       * Global alpha: A single alpha values controlling all the pixels.
+       * Colorkey: A color flagged to be considered transparent in operations.
+
+    Gloabl alpha and colorkeys can be mixed. An alpha of 255 is opaque while an
+    alpha of 0 is fully transparent.
+
+    Surfaces can also reference sections of other Surfaces. These are created
+    with the :meth:`subsurface()` method. Any change to either Surface will
+    effect the other.
```
oddbookworm: "affect"

buildconfig/stubs/pygame/surface.pyi
```diff
@@ -46,77 +46,69 @@ class Surface:
...
+       * Premultiplied alpha: An advanced usage of per-pixel alpha. See
+         :meth:`Surface.premul_alpha` for more.
+       * Global alpha: A single alpha values controlling all the pixels.
+       * Colorkey: A color flagged to be considered transparent in operations.
+
+    Gloabl alpha and colorkeys can be mixed. An alpha of 255 is opaque while an
+    alpha of 0 is fully transparent.
+
+    Surfaces can also reference sections of other Surfaces. These are created
+    with the :meth:`subsurface()` method. Any change to either Surface will
```
oddbookworm: I'm not a fan of this phrasing. IMO it's not strong enough.
```suggestion
    Surfaces can be created that reference the pixel data of other Surfaces.
    These are called suburfaces and are created with the :meth:`subsurface()`
    method. Changing the pixels referenced by either the original Surface or
    the subsurface will have an effect on both.
```

buildconfig/stubs/pygame/surface.pyi
```diff
@@ -46,77 +46,69 @@ class Surface:
...
-    with the :meth:`subsurface()` method. Any change to either Surface will
-    effect the other.
-
-    Each Surface contains a clipping area. By default the clip area covers the
-    entire Surface. If it is changed, all drawing operations will only effect
-    the smaller area.
+    Any functions that directly access a Surface's pixel data will need that
+    Surface to be lock()'ed in some cases. Built in functions that manipulate
+    Surfaces will lock and unlock the Surface automatically. Only use manual
+    locking for a lot of sequential pixel modifications.
```
oddbookworm: Expand upon this that this is not a blocking lock, and should not be considered a mutex

buildconfig/stubs/pygame/surface.pyi
```diff
@@ -291,56 +283,60 @@ class Surface:
...
+        """Change the pixel format of a Surface.
+
+        Creates a new copy of the Surface with the pixel format changed. The
+        format can be controlled in different ways:
+           * If no arguments are passed the new Surface will have the same pixel
+             format as the display Surface which is always the fastest format
+             for blitting. It is a good idea to convert all Surfaces before
+             blitting them. In this case the converted Surface won't have any
+             alpha. See :meth:`convert_alpha()` for preserving or creating per-pixel alpha.
+             An exception is thrown if the display Surface was not initialized.
```
oddbookworm: What kind of exception?

buildconfig/stubs/pygame/surface.pyi
```diff
@@ -291,56 +283,60 @@ class Surface:
...
-        be optimized for fast alpha blitting to it.
+        Creates a new copy of the Surface with a different format that contains
+        an alpha channel. Unlike the :meth:`convert()` method, the pixel format
+        for the new Surface will not be exactly the same as the display Surface,
+        but it will be suited and optimized for fast alpha blitting to it.
 
-        As with :meth:`convert()` the returned surface has the same class as
-        the converted surface.
+        As with :meth:`convert()`, the returned Surface has the same class as this
+        Surface. An exception will be raised if the display Surface has not
```
oddbookworm: Same comment, what kind of exception?

buildconfig/stubs/pygame/surface.pyi
```diff
@@ -438,21 +434,22 @@ class Surface:
...
 
     def get_alpha(self) -> Optional[int]:
         """Get the current Surface transparency value.
 
-        Return the current alpha value for the Surface.
+        Return the current alpha value for the Surface. For a Surface
+        with per pixel alpha, blanket alpha is ignored and ``None`` is returned.
         """
 
     def lock(self) -> None:
```
oddbookworm: Maybe add in notes that these are not mutex lock/unlocks

buildconfig/stubs/pygame/surface.pyi
```diff
@@ -801,13 +797,11 @@ class Surface:
         ``SDL_video.h``
 
         ::
-
+            SRCALPHA       0x00010000    # Blit uses source alpha blending
```
oddbookworm: I'm not sure this moving here is the right call

oddbookworm (changes requested):
Just a few comments lol, thanks @damusss 🎉

buildconfig/stubs/pygame/surface.pyi
damusss: Why would SRCALPHA be private if you use it often when constructing a surface?

damusss:
I've addressed most of @oddbookworm 's comments and I also fixed get_alpha again. I took the liberty of moving locking related methods at the bottom, and I would like the liberty of moving the width, height and size properties at the top.

buildconfig/stubs/pygame/surface.pyi
```diff
@@ -46,77 +46,70 @@ class Surface:
...
-    entire Surface. If it is changed, all drawing operations will only effect
-    the smaller area.
+    When creating a new Surface it will be cleared to all black or all transparent
+    depending on the format. The only required argument is the size, a sequence
+    containing the width and height of the Surface. With no additional arguments
+    the format will best match the display Surface if one was initialized, otherwise
+    the most suitable format is chosen for the current platform.
+
+    One way to control the pixel format/charateristics is providing a bitmask of flags:
+       * ``SRCALPHA``: If it is possible the Surface will include an alpha channel.
```
ankith26: Is there a specific reason you are adding "if it is possible" here. I think this can be a source of confusion. What sort of cases would it not be possible? specifying incompatible depths?

buildconfig/stubs/pygame/surface.pyi
```diff
@@ -46,77 +46,70 @@ class Surface:
...
+       * ``HWSURFACE``: Creates the image in video memory but is obsolete in pygame 2.
+
+    Otherwise The pixel format can be controlled in different ways:
+       * Providing another ``Surface`` after the flags argument. In this case that Surface's
+         format will be used.
+       * Providing a bit depth (describing the data used by one pixel). If alpha is
+         requested, only bit depths of ``16`` and ``24`` are allowed, resulting in ``ARGB4444``
```
ankith26: You mean only 16 and 32? Either way, I would not want to mention the 16 case in the docs. We (most likely) don't support it in our SIMD and people should just not need to worry about it and use 32 anyways.

buildconfig/stubs/pygame/surface.pyi
damusss: the possibility of it not being possible was present in the old docs, I left it there but it's probably just referring to wrong arguments in the format, so it can be safely removed.

buildconfig/stubs/pygame/surface.pyi
damusss: Latest commit should address this I believe

buildconfig/stubs/pygame/surface.pyi
```diff
@@ -46,77 +46,66 @@ class Surface:
...
-    Each Surface contains a clipping area. By default the clip area covers the
-    entire Surface. If it is changed, all drawing operations will only effect
-    the smaller area.
+    When creating a new Surface it will be cleared to all black or all transparent
+    depending on the format. The only required argument is the size, a sequence
+    containing the width and height of the Surface. With no additional arguments
+    the format will best match the display Surface if one was initialized, otherwise
+    the most suitable format is chosen for the current platform.
+
+    One way to control the pixel format/charateristics is providing a bitmask of flags:
```
aatle: `characteristics` misspelling

buildconfig/stubs/pygame/surface.pyi
```diff
@@ -46,77 +46,66 @@ class Surface:
...
+       * ``SRCALPHA``: The Surface will include an alpha channel.
```
aatle: I think this should mention "per-pixel alpha", mention transparency in some way.
It is not really stated that `SRCALPHA` is how to initialize a surface with transparency; I often see people using `.convert_alpha()` immediately after Surface creation instead.

buildconfig/stubs/pygame/surface.pyi
```diff
@@ -46,77 +46,66 @@ class Surface:
...
+       * ``HWSURFACE``: Creates the image in video memory but is obsolete in pygame 2.
```
aatle: IMO the wording can be improved or rearranged about "obsolete": the original description with `(obsolete in pygame 2) creates the image in video memory` is more clear in my opinion.

buildconfig/stubs/pygame/surface.pyi
```diff
@@ -46,77 +46,66 @@ class Surface:
...
+
+    Otherwise The pixel format can be controlled in different ways:
```
aatle: The "Otherwise" is not appropriate because this is not mutually exclusive with the flags. Actually, the two options listed here are mutually exclusive, which is not really mentioned.
(also, there is a random capitalization of "The")
Perhaps this section could be merged with the flags section, as all describe how to initialize a Surface with characteristics and formats.

buildconfig/stubs/pygame/surface.pyi
```diff
@@ -46,77 +46,66 @@ class Surface:
...
+       * Providing another ``Surface`` after the flags argument. In this case that Surface's
```
aatle: Wording nitpick: I think the "In this case" is unnecessary.
Additionally, I noticed that the per-pixel alpha (SRCALPHA) flag does not seem to be used by the new surface. If that is accurate, it could be mentioned.

buildconfig/stubs/pygame/surface.pyi
```diff
@@ -46,77 +46,66 @@ class Surface:
...
+         format will be used.
+       * Advanced users can provide a bit depth and masks to precisely control how the
+         pixel's channels are organized in memory. If no masks are provided they will
+         be selected optimally from the bit depth. Normal usage of Surfaces should
+         only use the default format.
```
aatle: Few things:
- Change "the pixel's channels" to "the pixels' channels" or "a pixel's channels" (grammar)
- "Normal usage of Surfaces should only use the default format." What is it meant by default format, does that mean `SRCALPHA` should not be used? Suggestion: "Normal usage of Surfaces is to leave bit depth and masks as default"
- I once got confused by the fact that if `depth` was not provided, `masks` seemed to be ignored. Not sure if something should be mentioned, but just a note.
- Doesn't have to be changed here, but I found the information on what masks is to be lacking (in the original docs), although it is rarely used. I have to use masks to standardize pixel format when working with `zengl`

buildconfig/stubs/pygame/surface.pyi
```diff
@@ -46,77 +46,66 @@ class Surface:
...
+
+    An indexed format requires a palette for its colors. While pygame provides a default
```
aatle: I'm not too familiar with palettes, but it is not explained what an "indexed format" is. IMO the statement is not useful to beginners nor experts. Maybe more explanation on palettes would help?

buildconfig/stubs/pygame/surface.pyi
```diff
@@ -46,77 +46,66 @@ class Surface:
...
+    palette, it can also be controlled with the appropriate methods.
+
+    Surfaces additionally have alpha, colorkey and rectangle clipping. These characteristics
+    mainly affect how the Surface is blitted to other Surfaces. The blit routines will
+    attempt to use hardware acceleration when possible, otherwise they will use highly
+    optimized software blitting methods. The default clipping area is the whole
```
aatle: For the clipping area, the "Otherwise" does not make sense IMO; the clipping area is always the modifiable area, but the default covers the entire surface. Also, the definition of the term should be written before telling what the default value is.
Additionally, it does not seem fitting to place this information at the end of this paragraph. I think it should be its own independent paragraph as it was in the original.

buildconfig/stubs/pygame/surface.pyi
```diff
@@ -46,77 +46,66 @@ class Surface:
...
+    Surface. Otherwise, only that area will be able to be modified.
+
+    There are three types of transparency supported for a pygame Surface:
```
aatle: Four (not three) are listed

buildconfig/stubs/pygame/surface.pyi
```diff
@@ -46,77 +46,66 @@ class Surface:
...
+       * Per-pixel alpha. It requires a format with an alpha channel where
+         very pixel has its own alpha value, making it very flexible.
```
aatle: "very" -> "every" (typo)
Also, the phrasing sounds off to me: "It requires a ..." ending with ", making it very flexible". Change first part maybe?

Nitpick: "premultiplied" usually has a hyphen, so "pre-multiplied"

buildconfig/stubs/pygame/surface.pyi
```diff
@@ -46,77 +46,66 @@ class Surface:
...
+       * Premultiplied alpha: An advanced usage of per-pixel alpha. See
+         :meth:`Surface.premul_alpha` for more.
+       * Global alpha: A single alpha values controlling all the pixels.
```
aatle: "values" plural typo
Phrasing suggestion:
```suggestion
       * Global alpha: A single alpha value that applies to the whole surface.
```

buildconfig/stubs/pygame/surface.pyi
```diff
@@ -46,77 +46,66 @@ class Surface:
...
+       * Colorkey: A color flagged to be considered transparent in operations.
+
+    Every type of alpha can be mixed. An alpha of 255 is opaque while an
+    alpha of 0 is fully transparent.
```
aatle: Types of transparency, not alpha. Also, pre-multiplied currently does not consider global alpha nor colorkey.
```suggestion
    All types of transparency can be used together and correctly mix,
    except pre-multiplied alpha does not consider colorkey nor global alpha.
    Also, an alpha value of 255 is opaque while an alpha of 0 is fully transparent.
```

buildconfig/stubs/pygame/surface.pyi
```diff
@@ -46,77 +46,66 @@ class Surface:
...
+
+    Surfaces can be created that reference the pixel data of other Surfaces.
+    These are called suburfaces and are created with the :meth:`subsurface()`
```
aatle: `subsurfaces` typo

buildconfig/stubs/pygame/surface.pyi
```diff
@@ -46,77 +46,66 @@ class Surface:
...
+    method. Changing the pixels referenced by either the original Surface or
+    the subsurface will have an effect on both.
+
+    There is support for pixel access for Surfaces but it is slow on hardware
```
aatle: A note, it was hardly explained what "hardware acceleration" and "hardware surface" are in the original docs, and many have thought that hardware acceleration meant GPU.

buildconfig/stubs/pygame/surface.pyi
```diff
@@ -46,77 +46,66 @@ class Surface:
...
+    Surfaces. You can use the :meth:`get_at()` and :meth:`set_at()` functions
+    but while they are fine for simple access will be slow when doing pixel work
```
aatle: missing a subject between "access" and "will" (grammar)

buildconfig/stubs/pygame/surface.pyi
```diff
@@ -291,56 +280,61 @@ class Surface:
...
+             alpha. See :meth:`convert_alpha()` for preserving or creating per-pixel alpha.
+             A :class:`pygame.error` exception is thrown if the display Surface
+             was not initialized.
+           * If another Surface is provided, the format of that Surface will
+             be used.
+           * Passing the bit depth of a pixel that will result in the same formats
+             explained in the :class:`Surface` constructor. Flags can also be
+             provided, for example ``SRCALPHA`` will request an alpha channel.
+           * Passing the bitmasks of the pixels and flags. Note that the bit depth
+             could be calculated wrong, therefore it is not adviced to use this
```
aatle: `adviced` -> `advised` typo

buildconfig/stubs/pygame/surface.pyi
```diff
@@ -797,28 +725,22 @@ class Surface:
...
-            SWSURFACE      0x00000000    # Surface is in system memory
-            HWSURFACE      0x00000001    # (obsolete in pygame 2) Surface is in video memory
-            ASYNCBLIT      0x00000004    # (obsolete in pygame 2) Use asynchronous blits if possible
+        Here is a more complete list of flags:
+           * ``SRCALPHA``: Blit uses source alpha blending
+           * ``SWSURFACE``: Surface is in system memory
+           * ``HWACCEL``: Blit uses hardware acceleration
+           * ``SRCCOLORKEY``: Blit uses a source color key
+           * ``RLEACCELOK``: Private flag
+           * ``RLEACCEL``: Surface is RLE encoded
```
aatle: "RLE encoded" does not make sense ("run-length encoding encoded")
```suggestion
           * ``RLEACCEL``: Surface has run-length encoding
```

buildconfig/stubs/pygame/surface.pyi
```diff
@@ -46,77 +46,66 @@ class Surface:
...
+    Surface. Otherwise, only that area will be able to be modified.
+
+    There are three types of transparency supported for a pygame Surface:
+       * Per-pixel alpha. It requires a format with an alpha channel where
+         very pixel has its own alpha value, making it very flexible.
+       * Premultiplied alpha: An advanced usage of per-pixel alpha. See
+         :meth:`Surface.premul_alpha` for more.
+       * Global alpha: A single alpha values controlling all the pixels.
+       * Colorkey: A color flagged to be considered transparent in operations.
+
```
aatle, line 85: I think it would be useful to have information on which is generally fastest on what platforms.

aatle (commented):
nice work 👍, it was about time these docs got updated

buildconfig/stubs/pygame/surface.pyi
```diff
@@ -291,56 +280,61 @@ class Surface:
...
-
-        Creates a new copy of the surface with the desired pixel format. The new
-        surface will be in a format suited for quick blitting to the display surface
-        with per pixel alpha.
+        """Change the pixel format of a Surface including per pixel alphas.
 
-        Unlike the :meth:`convert()` method, the pixel format for the new
-        surface will not be exactly the same as the display surface, but it will
-        be optimized for fast alpha blitting to it.
+        Creates a new copy of the Surface with a different format that contains
```
Starbuck5: The original wording was "a new copy of the surface with the desired pixel format", which is more accurate, because the Surface could already be in the most ideal format.

Starbuck5:
Great idea for an improvement, Damus. It may be wise to split this out into smaller PRs to get it through, there's already been a ton of conversation and probably will be a bunch more, this is that kind of PR. But parts of this likely could have gotten through already, like you doing surface -> Surface all over the place.

damusss:
@Starbuck5  Made this stall because of the end of high school. I'll push a few smaller PRs so it can get in

buildconfig/stubs/pygame/surface.pyi
damusss: I'm not sure I know what hardware accelerations means in this case. But the HWSURFACE flag is obsolete, so it might be worth removing the mention alltogether

buildconfig/stubs/pygame/surface.pyi
damusss: Ah, didn't even know this is what it meant, thank you

damusss:
Thank you @aatle for noticing all those typos and inconsistencies. I tried to address them.

damusss:
The pull request was succesfully split in three parts (one still open, needing updates on my side), meaning there's no need for the parent pull request anymore.

PR #3456, opened 2025-06-02
add Color.from_hex

ToniDevStuff:


buildconfig/stubs/pygame/color.pyi
```diff
@@ -78,6 +78,9 @@ class Color(Collection[int]):
     @overload
     @classmethod
     def from_normalized(cls, r: float, g: float, b: float, a: float, /) -> Color: ...
+    @overload
```
oddbookworm: This isn't an overload, there's only one signature

buildconfig/stubs/pygame/color.pyi
```diff
@@ -78,6 +78,9 @@ class Color(Collection[int]):
...
+    @classmethod
+    def from_hex(cls, hex:str, /) -> Color: ...
```
oddbookworm: space between `hex:` and `str`
```suggestion
    def from_hex(cls, hex: str, /) -> Color: ...
```

docs/reST/ref/color.rst
```diff
@@ -292,6 +292,18 @@
...
 
+   .. classmethod:: from_hex
+
+      | :sl:`Returns a Color object from a Hexadecimal representation`
+      | :sg:`from_hex(hex, /) -> Color`
+
+      Creates a Color object from the given Hexadecimal components. Refer to :attr:`Color.hex`
+      for more information.
+
+      .. versionadded:: 2.5.0
```
oddbookworm: 2.5.5

AntonisPylos:
It fails with dev.py because _color_set_hex doesn't support multiple arguments. 
( In the test I had only one, that’s why it passed all the others.)

```shell
File "/usr/lib/python3.11/unittest/case.py", line 237, in handle
    callable_obj(*args, **kwargs)
File "/home/personal/Desktop/git/pygame-ce/test/color_test.py", line 858, in <lambda>
    self.assertRaises(ValueError, lambda: pygame.Color.from_hex("not-a-color", 32))
  ```
  
  So to add Color.from_hex the set_hex need re-write?

**Review by oddbookworm — CHANGES_REQUESTED** [review]
Run `python dev.py format` and commit the changes to fix the failure here. The one CI action is complaining about the formatting of some stuff

File: buildconfig/stubs/pygame/color.pyi (hunk 1 above)
Comment (AntonisPylos) [reply buildconfig/stubs/pygame/color.pyi]: changed it at 45801dd4051de6bf1e2a704d6fb5aa0bec285361

File: docs/reST/ref/color.rst (hunk 3 above)
Comment (AntonisPylos) [reply docs/reST/ref/color.rst]: Done

**Comment by AntonisPylos** [comment]
> Run `python dev.py format` and commit the changes to fix the failure here. The one CI action is complaining about the formatting of some stuff

I committed the format changed but it still fails, the problem is with:
```py
ValueError, lambda: pygame.Color.from_hex("not-a-color", "lel")
               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
TypeError: hex color must be a string
```
Even if I remove this assertRaises it's still going to fail because of  *args,**kwargs in  callable_obj when the tests are running.

Unless I'm misunderstanding something.

buildconfig/stubs/pygame/color.pyi
AntonisPylos: changed it at 45801dd4051de6bf1e2a704d6fb5aa0bec285361

test/color_test.py
```diff
@@ -848,6 +848,18 @@ def test_from_normalized(self):
...
 
+    def test_from_hex(self):
+        hex = pygame.Color.from_hex("#FFFFFF")
+
+        expected_hex = (255, 255, 255, 255)
+
+        self.assertEqual(expected_hex, hex)
+
+        self.assertRaises(ValueError, lambda: pygame.Color.from_hex("not-a-color"))
+        self.assertRaises(
```
oddbookworm, line 870: This should assert that it raises a `TypeError`, not a `ValueError`

damusss: also in our tests we do 
```py
with self.assertRaises(exception):
	code
```

AntonisPylos: > This should assert that it raises a `TypeError`, not a `ValueError`

Update: It appears everything works now. Thanks for the help

test/color_test.py
```diff
@@ -848,6 +848,18 @@ def test_from_normalized(self):
             ValueError, lambda: pygame.Color.from_normalized(1, 1, 1, 1, "lel")
         )
 
+    def test_from_hex(self):
+        hex = pygame.Color.from_hex("#FFFFFF")
+
+        expected_hex = (255, 255, 255, 255)
```
oddbookworm: I would like to see a variety of random colors tested, not just opaque white
Keep this one (I do like that it feeds in an RGB hex and expects an RGBA tuple (although the names `hex` and `expected_hex` might be a bit misleading because they're not in hex)

oddbookworm: Also, what happens if you feed it something that looks like hex, but isn't?
"#FFFFFFF"

AntonisPylos: I included all your suggestions. It should be good now.

docs/reST/ref/color.rst
```diff
@@ -292,6 +292,18 @@
...
 
+   .. classmethod:: from_hex
+
+      | :sl:`Returns a Color object from a Hexadecimal representation`
+      | :sg:`from_hex(hex, /) -> Color`
+
+      Creates a Color object from the given Hexadecimal components. Refer to :attr:`Color.hex`
+      for more information.
+
+      .. versionadded:: 2.5.5
```
ankith26: this needs to be 2.5.6 now

AntonisPylos: indeed

oddbookworm (approved):
LGTM! @damusss did mention the context manager for `self.assertRaises`, but I'm not going to enforce that here because the rest of the file doesn't do that consistently either. One other minor thing that I might recommend as a nice bonus if you want to (but I'm not asking for) is to add something into the test to show what happens if you call the classmethod on an instance
```py
pygame.Color(255, 0, 0, 255).from_hex("#00FF00")
```
What does this do?

ankith26 (approved):
LGTM, thanks for contributing! :sunglasses: 

Andrew mentioned updating the tests for checking classmethods on instances, but none of the other `from_*` methods do that either. So it need not necessarily be a part of this PR, but could be a future PR that implements the test for all `from_*` methods.

PR #3461, opened 2025-06-02
Remove colordict, rework parsing color from string

### LINKED ISSUE #1664: Move colordict.py into C (3367)
#  **Issue &numero;3367 opened by *[Starbuck5](https://github.com/Starbuck5)* at 2022-08-01 07:56:42**

This is a bit of a strange idea, but I thought I would put it out there.

colordict.py is 26kb
colordict.colors is 18.5kb of memory at runtime (measured using sys.getsizeof)
colordict.colors puts a python dict in between parsing a string color name and the translated color (performance?)

So, this might be worth implementing our own tiny dictionary inside color.c to handle this more efficiently. We can control the hash function and bucket size to maximize the benefit for our use case.

We also might be compress the data a bit more, filesize wise.

Here's a small example of what the frozen data could look like.
```c
struct ColorDataStruct {
    char* name1;
    char* name2;
    int hex;
}

static ColorDataStruct colordata[] = {
    {"grey0", "gray0", 0x000000FF},
    {"tomato", NULL, 0xFF6347FF},
    {NULL, NULL, NULL}
}
```

There are two name slots because between the 665 color names there are only 507 unique color outputs.

Additionally, a C compiler might do string interning on all the similar names "grey0", "grey1", ... "grey99" for instance, reducing binary size.<hr>

#  Comments

# # #  **[ankith26](https://github.com/ankith26)* commented at 2022-08-03 05:21:36*

I'm not sure a C compiler would do string interning on strings that are similar yet not the same (atleast I did not find anything online from a quick search)

I'm also not sure if this can give any significant change in performance (if benchmarks show otherwise I'd be happy to stand corrected), infact a naive C implementation (that uses iteration) could actually end up being slower than the python dict implementation (which as you know uses hashes so dictionary lookup is `O(1)`)

So I personally don't think chasing this issue is worth the efforts needed (gains, even if present, would likely be minor), but if anyone wants to do it I'd be happy to review a PR nonetheless.

MightyJosip:
This commit removes colordict module, moves parsing color from string to C code, didn't check for the speed, but this should improve memory usage. Also need to find the right way to test this change

Fix #1664


damusss:
Hi, while on the surface it doesn't seem bad, I don't think this change is available in the way you are proposing it fully. You are effectively removing a documented and type-hinted constant, breaking compatibility. A github search shows that a very big chunk of people's code use that constant, check for yourself:
https://github.com/search?q=THECOLORS+language%3APython+&type=code
Not sure if I'm wrong, but I doubt so.

ankith26:
This PR cannot be merged as it is for reasons damus stated. Though it would be fine if you also implement some "dict proxy" class that can compat the existing dict.

In my opinion this is too much work though, it would be instead better to wait for a breaking release and do this PR then

Starbuck5:
Hello, thanks for pursuing this. Due to the aforementioned discussion we can't do this until a breaking release (3.0 hype), so let's close for now to keep the PR list tidier.

PR #3462, opened 2025-06-02
Move surface to its own directory

JoKing-1999:
More discussion in #3421, this PR moves all surface stuff into its own directory. Commit also contain some other changes important for moving modules into their own directories, like the changes to make_docs. Decided to go with surface at first, because it is arguably the most important module, so it deserves its special place. This commit compared to the 3421 doesn't move tests, decided it would be the best if it is done separately.


src_c/pixelcopy.c
```diff
@@ -22,7 +22,7 @@
 
 #include <stddef.h>
 
-#include "palette.h"
+#include "include/palette.h"
```
ankith26: I don't get why this change is required here

src_c/surface/alphablit.c
```diff
@@ -24,7 +24,7 @@
 
 #define NO_PYGAME_C_API
 #include "_surface.h"
-#include "simd_shared.h"
+#include "../include/simd_shared.h"
```
ankith26: Is this still required if you are updating include paths in the buildconfig?

src_c/surface/meson.build
```diff
@@ -0,0 +1,55 @@
+# TODO: support SDL3
+if sdl_api != 3
```
ankith26: Minor nitpick: I would prefer this sdl_api check to be in `src_c/meson.build` just wrapping `subdir("surface")`. So we can remove one layer of indentation here in this file and also cause less merge conflicts with PRs already working on SDL3 support for surface.

src_c/transform.c
```diff
@@ -33,7 +33,7 @@
 #include <math.h>
 #include <string.h>
 
-#include "simd_shared.h"
+#include "include/simd_shared.h"
```
ankith26: same question as the other include

src_c/surface/meson.build
JoKing-1999: Ok yea you are completely right, I moved it to root

src_c/pixelcopy.c
JoKing-1999: Reworked includes, can you check this commit again

src_c/surface/meson.build
ankith26: You moved it there but forgot to remove it at this file :smile:

JoKing-1999: 🤣 lmao yea, now it should be ok

Starbuck5:
This PR has several merge conflicts now.

JoKing-1999:
Fixed

PR #2354, opened 2023-07-27
pygame.transform.pixelate

oddbookworm:
Addressing #2302 

TODO:
- [x] tests

## Summary by CodeRabbit

- New Features
  - Added transform.pixelate(surface, pixel_size, dest_surface=None) to create a pixelated version of a surface while preserving its size.
- Documentation
  - Documented the new pixelate API with usage details and version tag.
  - Updated type stubs to include the new function.
- Tests
  - Added tests for pixelate, including correctness at various pixel sizes and error handling.
- Chores
  - Updated build requirements to include a new build-time dependency.

PR #3464, opened 2025-06-03
Fix palettes in `newsurf_fromsurf`

### LINKED ISSUE #3370: Investigate for any regressions in scale_to
> I think this PR broke scale_to https://github.com/pygame-community/pygame-ce/commit/717d6d00cf36727badd3fa4db46b35985435f757#diff-02f1b57c8aaf1c6efca9e58df6187db5ff5c4d6b5e1e6968f132419d2a0c79d0, or at least changed how it behaves enought that pixelate isn't behaving anymore 
- @oddbookworm on discord

I'm the author of PR so I've re-read through it a couple more times and I still don't see what I did that could've changed it's behavior. But I don't want this note to be lost, so I'm opening an issue to be sure we look into it.

### LINKED ISSUE #3463: Image glitch when loading and immediately rotating a surface.
WIth pygame-ce 2.5.4 loading an image, immediately rotating it, and then convert_alpha, the surface returned is completely white.

```python
basic_enemy_img = pygame.transform.rotate(pygame.image.load(
    os.path.join("Assets/Enemy_assets", "basic_enemy.png")), 180).convert_alpha()
```

The code above worked just fine in 2.5.3 but as I said above, in pygame-ce 2.5.4 it returns a completely white surface. 
I was able to fix the issue with the code below, (converting the surface before rotating) but I'm pretty sure this behaviour is not intended.

```python
basic_enemy_img = pygame.transform.rotate(pygame.image.load(
    os.path.join("Assets/Enemy_assets", "basic_enemy.png")).convert_alpha(), 180)
```

oddbookworm:
#3314 had an uncaught issue that means `newsurf_fromsurf` instantiates palettes incorrectly, causing .... white. I reported this before when I was trying to revive #2354, but I could only point to the specific PR that broke it. Now that #3463 has been opened, I investigated further and finally found the source of the problem. I haven't tested it on my `pixelate` problem, but I suspect this is the cause there as well because it's the same symptom

Closes #3463 and should also close #3370, but I'll verify that one manually at a later date

### DIFF
--- a/src_c/transform.c
+++ b/src_c/transform.c
@@ -136,7 +136,7 @@ newsurf_fromsurf(SDL_Surface *surf, int width, int height)
     /* Copy palette, colorkey, etc info */
     if (SDL_ISPIXELFORMAT_INDEXED(PG_SURF_FORMATENUM(surf))) {
         SDL_Palette *newsurf_palette = PG_GetSurfacePalette(newsurf);
-        SDL_Palette *surf_palette = PG_GetSurfacePalette(newsurf);
+        SDL_Palette *surf_palette = PG_GetSurfacePalette(surf);
 
         if (newsurf_palette == NULL) {
             PyErr_SetString(
--- a/test/transform_test.py
+++ b/test/transform_test.py
@@ -1449,6 +1449,39 @@ class TransformModuleTest(unittest.TestCase):
         for pt, color in gradient:
             self.assertTrue(s.get_at(pt) == color)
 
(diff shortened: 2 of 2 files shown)


oddbookworm:
Also fixes https://github.com/pygame-community/pygame-ce/pull/2354
That PR currently has a failing test, but it works locally if I apply the fix from this PR

Starbuck5 (approved):
Thank you

PR #3465, opened 2025-06-03
Fix dst palette in modify_hsl

ankith26:
After seeing #3464 I reviewed #3314 more closely (which I should have done in the first place while reviewing that), and I found another minor issue. Fortunately this one shouldn't be the cause any regression, because the old code also had the same issue as the current code.


MightyJosip (approved):
https://github.com/pygame-community/pygame-ce/pull/3465/files#diff-02f1b57c8aaf1c6efca9e58df6187db5ff5c4d6b5e1e6968f132419d2a0c79d0R2795

Variable is used only here, and it definitely makes sense to actually be the palette of the destination, so LGTM, ty 👍

Starbuck5:
I did flag this in https://github.com/pygame-community/pygame-ce/pull/3314#discussion_r1930092597, but I dealt with it weirdly by using the format from one and the palette from the other. I should've used both from one or both from the other.

PR #3467, opened 2025-06-03
Fix draw halting when using NaN argument

### LINKED ISSUE #3412: NaNs completely deadlock draw.line
```py
import math
import pygame

pygame.init()

surf = pygame.Surface((500,500))

print("Starting")

pygame.draw.line(surf, "black", (0, 0), (math.nan, 100))

print("Done!")
```

This test script makes pygame-ce completely unresponsive, it can't even be CTRL-Ced.

I haven't looked into other draw functions.

MightyJosip:
Fix #3412

When using NaN as argument, skip the drawing (of that part of function

### DIFF
--- a/src_c/draw.c
+++ b/src_c/draw.c
@@ -113,7 +113,7 @@ draw_round_rect(SDL_Surface *surf, SDL_Rect surf_clip_rect, int x1, int y1,
         return NULL;                                     \
     }
 
-#define CHECK_NAN(value) value == 0x80000000
+#define CHECK_NAN(value) (value == 0x80000000)
 
 /* Definition of functions that get called in Python */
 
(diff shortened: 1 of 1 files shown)


src_c/draw.c
```diff
@@ -113,6 +113,8 @@ draw_round_rect(SDL_Surface *surf, SDL_Rect surf_clip_rect, int x1, int y1,
         return NULL;                                     \
     }
 
+#define CHECK_NAN(value) ((unsigned int)value == 0x80000000)
```
ankith26, line 116: I'd rather have something like `isnan` from math.h instead of this

ankith26 (commented):
It would be nice if you added testcases for this

Starbuck5:
Hello, marking this as a draft because there are unresolved comments.

Issue #3287, opened 2025-01-10
Use pygame-ce to test SDL2-compat


Starbuck5:
One of the SDL folks mentioned they could use some additional testing on SDL2-compat as SDL3 gets closer to release, I think one really comprehensive way to do that would be to try to get pygame-ce running on SDL2-compat (SDL3).

Opening an issue to see if anyone is interested in tackling this.

Starbuck5:
To do this (Windows) I cloned down https://github.com/libsdl-org/sdl2-compat, built as instructed + `-DCMAKE_PREFIX_PATH=C:\Users\charl\Desktop\pygame-ce-stuff\sdl2-compat\SDL3-devel-3.1.10-VC\SDL3-3.1.10`

Put the SDL3 prerelease into the files.

Then I got an SDL2 DLL out of it, which I manually placed into the site-packages/pygame folder after building pygame-ce.

If others want to test here are the two executables to put into the folder:  [SDL3_compat_lib_bundle.zip](https://github.com/user-attachments/files/18468392/SDL3_compat_lib_bundle.zip)

Starbuck5:
pygame-ce is probably like worst case scenario for this library, but that's why it's a good test.

I segfaulted several times between running the tests and running selected examples, haven't done any debugging on why.

transform.scale doesn't work because SDL_SoftStretch (RIP) worked while the surface was locked and SDL_BlitScaled doesn't.

I managed to trigger assertions about surface width/height, as well as about event watchers. I feel like I get the warnings once and then I don't get them again, so it's hard to say which examples triggered them?

In `python -m pygame.examples.video` neither of the two windows can be exited by clicking the x button.

ankith26:
Just an update on this issue, with pygame-ce and sdl2-compat built from their respective `main`s, all tests should be passing in theory.

Starbuck5:
With Ankith's work on this on getting us and them more mutually compatible, I don't think there's any reason to keep this issue anymore. Nice!

PR #3470, opened 2025-06-04
Fix windowevent translation in sdl2-compat

ankith26:
WINDOWEVENT translation was broken when compiled with sdl2-compat.

The issue was in our `_pg_translate_windowevent` function. The current code was directly modifying the event type there but it seems like it is simply an implementation detail that we were relying on, and one isn't supposed to actually modify events in `SDL_FilterEvents` based filtering. Plus, this is easier to fix on our side than fix on sdl2-compat side.

I also took the opportunity to improve the comments while I was giving the implementation a good think.

This fixes a failing unit test so I don't see the need to add more tests here.


ankith26:
from @Starbuck5 in #3287
> In `python -m pygame.examples.video` neither of the two windows can be exited by clicking the x button. (Probably related to whatever it didn't like about the event watch?)

This PR should fix the issue.

ankith26:
```py
import pygame
import sys

pygame.init()
screen = pygame.display.set_mode((400, 200), pygame.RESIZABLE)
pygame.display.set_caption("Key Repeat & Block Test")
clock = pygame.time.Clock()
font = pygame.font.Font(None, 36)

counter = 0
key_repeat_on = True
key_blocked_on = False

pygame.key.set_repeat(300, 100)

running = True
while running:
    screen.fill((30, 30, 30))

    text = font.render(f"Counter: {counter}", True, (255, 255, 255))
    info1 = font.render(f"Repeat: {'ON' if key_repeat_on else 'OFF'}", True, (180, 180, 180))
    info2 = font.render(f"Blocked: {'YES' if key_blocked_on else 'NO'}", True, (180, 180, 180))

    screen.blit(text, (50, 40))
    screen.blit(info1, (50, 90))
    screen.blit(info2, (50, 130))
    
    pygame.display.flip()

        if event.type == pygame.QUIT:
            running = False

            print(f"KEYDOWN: {pygame.key.name(event.key)}")

                counter += 1

            elif event.key == pygame.K_r:
                key_repeat_on = not key_repeat_on
                if key_repeat_on:
                    pygame.key.set_repeat(300, 100)
                else:
                    pygame.key.set_repeat()

        elif event.type == pygame.MOUSEBUTTONDOWN:
            key_blocked_on = not key_blocked_on
            if key_blocked_on:
                pygame.event.set_blocked(pygame.KEYDOWN)
            else:
                pygame.event.set_allowed(pygame.KEYDOWN)

        elif event.type >= pygame.WINDOWSHOWN and event.type <= pygame.WINDOWDISPLAYCHANGED:
            print(event)

    clock.tick(60)

pygame.quit()
sys.exit()
```
test script for this PR

Starbuck5:
I tested with your script and SDL2 compat and key repeat doesn't seem to work.

Starbuck5:
> from @Starbuck5 in #3287
> 
> > In `python -m pygame.examples.video` neither of the two windows can be exited by clicking the x button. (Probably related to whatever it didn't like about the event watch?)
> 
> This PR should fix the issue.

It doesn't in my testing. I'm using the latest SDL 2.32.56 release from them.

ankith26:
hmm, it works for me. Are you sure you are on this branch? Also, the `r` key toggles key repeat in the code, just in case you missed it.

Starbuck5:
image

`pygame-ce 2.5.6.dev1 (SDL 2.32.56, Python 3.9.5)`

PR #3471, opened 2025-06-04
Add skip to set_gamma tests

ankith26:


test/display_test.py
```diff
@@ -515,6 +515,10 @@ def test_quit__multiple(self):
         os.environ.get("SDL_VIDEODRIVER") == pygame.NULL_VIDEODRIVER,
         "Needs a non-null videodriver",
     )
+    @unittest.skipIf(
+        pygame.version.SDL >= (2, 32, 50),
```
JoKing-1999, line 519: Maybe `pygame.version.SDL >= (3, 0, 0)` would be more understandable as not SDL3

ankith26: This is done to handle sdl2-compat releases in that case we have versions like 2.32.56 and such.

MightyJosip (approved):
LGTM

damusss (approved):
Sure!

PR #3472, opened 2025-06-04
Updated outdated docs for pygame.freetype module and added note at top

Eshan-Shah:
To the docs for the pygame.freetype module, I added a note at the top warning and suggesting users to use the better maintained pygame.font module. I also removed any false statements from the longer description of the pygame.freetype module.


damusss (changes requested):
Hi, thank you for contributing :) I have some small requests
- freetype isn't older than pygame.font, it was an attempt to replace it but eventually failed. you can just say that it's an outdated alternative to pygame.font (plus the other things you said)
- the note needs to be on the same indentation level as the module, with the proper spaces. run docs --full again after your change.
- the grammar in the note should be improved

src_c/doc/freetype_doc.h
oddbookworm, line 1: This file should not be modified by hand, as the header says it's autogenerated. It's actually an important file and pygame-ce will not compile without it (it's the built-in docs for the `pygame.freetype` module)

docs/reST/ref/freetype.rst
```diff
@@ -5,12 +5,15 @@
 
 .. module:: pygame.freetype
    :synopsis: Enhanced pygame module for loading and rendering computer fonts
+   .. note:: The ``pygame.freetype`` module is no longer being maintained,
+      and instead use of :mod:`pygame.font` is strongly recommended.
+
 
 | :sl:`Enhanced pygame module for loading and rendering computer fonts`
```
oddbookworm, line 12: This slug line should be updated too

docs/reST/ref/freetype.rst
```diff
@@ -5,12 +5,15 @@
...
 
-The ``pygame.freetype`` module is a replacement for :mod:`pygame.font`.
-It has all of the functionality of the original, plus many new features.
-Yet is has absolutely no dependencies on the SDL_ttf library.
+The ``pygame.freetype`` module is an old pygame module for font functionality
+and has since been replaced by the newer :mod:`pygame.font` module.
+It has absolutely no dependencies on the SDL_ttf library.
 It is implemented directly on the FreeType 2 library.
 The ``pygame.freetype`` module is not itself backward compatible with
 :mod:`pygame.font`.
```
oddbookworm, line 19: Below here, there should be some changes too because `ftfont` is a `pygame.font`-like API that wraps `pygame.freetype`, and it's still written in a way that intends to replace `font`

oddbookworm (changes requested):
Good start! But I see a few places for improvement.

Thanks for the contribution 🎉

src_c/doc/freetype_doc.h
oddbookworm: This header file gets regenerated when you run `python dev.py docs --full`

Starbuck5:
Since the PR has been adrift for a long time, I sent in and got merged a PR to resolve the issue.

Issue #3218, opened 2024-11-09
`test_from_display_module` fail in `Window.opengl` attribute


ankith26:
**Environment:**

```
Platform:		Linux-6.11.0-9-generic-x86_64-with-glibc2.40
System:			Linux
System Version:		#9-Ubuntu SMP PREEMPT_DYNAMIC Mon Oct 14 13:19:59 UTC 2024
Processor:		x86_64	SSE2: Yes	AVX2: Yes	NEON: No
Architecture:		Bits: 64bit	Linkage: ELF

Python:			CPython 3.12.7 (main, Oct  3 2024, 15:15:22) [GCC 14.2.0]
pygame version:		2.5.2
SDL versions:		Linked: 2.30.8	Compiled: 2.30.8
SDL Mixer versions:	Linked: 2.8.0	Compiled: 2.8.0
SDL Font versions:	Linked: 2.22.0	Compiled: 2.22.0
SDL Image versions:	Linked: 2.8.2	Compiled: 2.8.2
Freetype versions:	Linked: 2.13.3	Compiled: 2.13.3

Display Driver:		x11 ( xwayland == True )
Mixer Driver:		pulseaudio
```

On a standard Ubuntu 24.10 install, have the following failing test.

```
======================================================================
FAIL: test_from_display_module (pygame.tests.window_test.WindowTypeTest.test_from_display_module)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/home/ankith/base_env/lib/python3.12/site-packages/pygame/tests/window_test.py", line 327, in test_from_display_module
    self.assertFalse(win1.opengl)
AssertionError: True is not false
```

It's been like this for a while and I used to brush aside this fail because the `Window` module was not stable API, but now that it is, we should look into fixing it. I think cause of this issue is basically same as the issue that the PR https://github.com/pygame-community/pygame-ce/pull/3059 was addressing, SDL defaulting to framebuffer acceleration.

PR #3474, opened 2025-06-05
Fix and improve `pygame.Window` tests

ankith26:
fixes https://github.com/pygame-community/pygame-ce/issues/3218


MyreMylar (approved):
LGTM 👍

PR #3478, opened 2025-06-06
Re exclude korean tutorials in release wheels and run important CI everytime

### LINKED ISSUE #3468: Notarization fails for pygame-ce 2.5.4, but succeeds for 2.5.3
**Environment:**

```
Platform:		macOS-15.5-arm64-arm-64bit
System:			Darwin
System Version:		Darwin Kernel Version 24.5.0: Tue Apr 22 19:54:25 PDT 2025; root:xnu-11417.121.6~2/RELEASE_ARM64_T6020
Processor:		arm	SSE2: No	AVX2: No	NEON: Yes
Architecture:		Bits: 64bit	Linkage:

Python:			CPython 3.10.13 | packaged by conda-forge | (main, Dec 23 2023, 15:35:25) [Clang 16.0.6 ]
GIL Enabled:		True
pygame version:		2.5.4
SDL versions:		Linked: 2.32.6	Compiled: 2.32.6
SDL Mixer versions:	Linked: 2.8.1	Compiled: 2.8.1
SDL Font versions:	Linked: 2.24.0	Compiled: 2.24.0
SDL Image versions:	Linked: 2.8.8	Compiled: 2.8.8
Freetype versions:	Linked: 2.13.3	Compiled: 2.13.3

Display Driver:		Display Not Initialized
Mixer Driver:		Mixer Not Initialized
```

**Current behavior:**

I use Beeware's Briefcase to build an .app bundle, sign it, and submit it for notarization. I used the default minimal pygame template, so there is hardly any code. It just imports pygame and includes it in the binary. The notarization step fails and Apple's system gives the following unhelpful error message:
```
The signature of the binary is invalid.
```
The problem does _not_ occur if I replace version 2.5.4 with 2.5.3.

**Expected behavior:**

The notarization step completes successfully when using pygame-ce 2.5.3.

**Steps to reproduce:**

Please explain the steps required to duplicate the issue, especially if you are able to provide a sample application.
if the bug is caused by a specific file (image, font, sound, level, please upload it as an attachment

1. Install briefcase, make sure you have a code signing identity from Apple.
2. `briefcase new`, choose a pygame template
2. Replace `pygame` with `pygame-ce==2.5.4` or `pygame-ce==2.5.3` in `pyproject.toml`
3. `briefcase create`
4. `briefcase build`
5. `briefcase package`
6. Failure for 2.5.4, success for 2.5.3.

ankith26:
PR to hopefully fix #3468

### DIFF
--- a/.github/workflows/build-macos.yml
+++ b/.github/workflows/build-macos.yml
@@ -1,32 +1,13 @@
 name: macOS
 
 # Run CI only when a release is created, on changes to main branch, or any PR
-# to main. Do not run CI on any other branch. Also, skip any non-source changes
-# from running on CI
+# to main.
 on:
   push:
     branches: main
-    paths-ignore:
-      - 'docs/**'
-      - 'examples/**'
-      - '.gitignore'
-      - '*.rst'
-      - '*.md'
-      - '.github/workflows/*.yml'
-      # re-include current file to not be excluded
-      - '!.github/workflows/build-macos.yml'
 
   pull_request:
     branches: main
-    paths-ignore:
-      - 'docs/**'
-      - 'examples/**'
-      - '.gitignore'
-      - '*.rst'
-      - '*.md'
-      - '.github/workflows/*.yml'
-      # re-include current file to not be excluded
-      - '!.github/workflows/build-macos.yml'
 
   # the github release drafter can call this workflow
   workflow_call:
--- a/.github/workflows/build-manylinux.yml
+++ b/.github/workflows/build-manylinux.yml
@@ -1,32 +1,13 @@
 name: ManyLinux
 
 # Run CI only when a release is created, on changes to main branch, or any PR
-# to main. Do not run CI on any other branch. Also, skip any non-source changes
-# from running on CI
+# to main.
 on:
   push:
     branches: main
-    paths-ignore:
-      - 'docs/**'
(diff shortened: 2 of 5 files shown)


oddbookworm:
https://github.com/pygame-community/pygame-ce/issues/3353
We'll regress this if we merge this as-is. Can we try to fix the encoding issue instead potentially?

ankith26:
There are a lot of files and we would have to trial and error manually. The first suspect is all korean files. I don't have a mac so someone else would need to do this.

oddbookworm:
It has to be someone with a developer license from apple I think, I tried with my mac, but it worked for me (but the sign was only valid on my own mac)

ankith26:
I think for now we have a good compromise - only the korean tuts are excluded. Anyone who wants the english/spanish tuts can still access it from the wheel. The korean tuts should continue to be available on pyga.me so I think excluding it in the wheels doesn't really cause problems to anyone.

@davidfokkema has confirmed that the current state of this PR fixes the issue (thanks again for all the testing!)

oddbookworm (approved):
LGTM! We'll have to keep in mind to exclude any non-ASCII docs names in the future

Starbuck5 (approved):
Thanks Ankith!

PR #3481, opened 2025-06-07
Update to SDL 2.32.8

Starbuck5:
Also updates libdrm because the old setup was failing to build.


Starbuck5:
I'm unsure how to deal with drm failure @ankith26.

We could move 123->124 and update the checksum?
We could complain to them that their release changed? Do we think it's a security issue?

ankith26:
I don't think they changed the release file, its more likely that it was just some network issue or something else that lead to this. We can try doing a rerun in a few hours/days and seeing if this issue persists

Starbuck5:
I checked locally several attempts, the file hash does not match.

ankith26:
You are right... the hash did actually change. I extracted the archive and yes, the name of the top level folder definitely changed.

> We could move 123->124 and update the checksum?

I think this makes sense.

oddbookworm (approved):
LGTM!

PR #3482, opened 2025-06-07
Remove redundant tp_getattro sets

Starbuck5:
tp_getattro = PyObject_GenericGetAttr is the default, so we don't need to set it.

See https://docs.python.org/3/c-api/typeobj.html#c.PyTypeObject.tp_getattro

I verified this by printf-ing the value of tp_getattro before it gets set, it's already been set to the address of PyObject_GenericGetAttr by PyType_Ready.

For the curious, this the pixelarray one was originally committed in https://github.com/pygame-community/pygame-ce/commit/284267bb114c3ee0bc74006f4973c99203941727

### DIFF
--- a/src_c/math.c
+++ b/src_c/math.c
@@ -3732,7 +3732,6 @@ static PyTypeObject pgVectorIter_Type = {
     PyVarObject_HEAD_INIT(NULL, 0).tp_name = "pygame.math.VectorIterator",
     .tp_basicsize = sizeof(vectoriter),
     .tp_dealloc = (destructor)vectoriter_dealloc,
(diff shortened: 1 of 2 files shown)


JoKing-1999:
When you are doing this, can you also remove some other generic method pointers, like for example PyObject_GenericGetAttr from Event/Vector

Starbuck5:
> When you are doing this, can you also remove some other generic method pointers, like for example PyObject_GenericGetAttr from Event/Vector

I removed from the vector iterator as requested, I don't want to remove from Event because while it makes the code shorter, I think it makes it harder to understand (given pypy has a different codepath).

MightyJosip (approved):
👍

PR #3483, opened 2025-06-07
Combine all "checks" CI into a single workflow

ankith26:
To reduce the number of concurrent CI I propose using just one workflow for all "checks" related things where everything can go on sequentially. In this PR
- coverage and debug python builds are merged into one build to save CI time (and merging them also makes sense)
- the cppcheck build is now going to execute sequentially after the above build... for no particular reason apart from reducing concurrent CI usage.
- Any future "checks" builds can be appended to this file in a similar fashion


ankith26:
This PR is a precursor to Andrews https://github.com/pygame-community/pygame-ce/pull/3477 and uses some code/ideas from there

.github/workflows/run-ubuntu-checks.yml
```diff
@@ -107,3 +110,49 @@ jobs:
...
+        lcov --capture --directory . --output-file ./coverage.info
+        genhtml ./coverage.info --output-directory ./out
+
+    # We upload the generated files under github actions assets
+    - name: Upload coverage html
+      # want to continue only if the coverage generation was successful
+      if: ${{ steps.gen-coverage.conclusion == 'success' && !cancelled() }}
+      uses: actions/upload-artifact@v4
+      with:
+        name: pygame-wheels-coverage
```
oddbookworm:
```suggestion
        name: pygame-wheels-coverage-{ matrix.os }-{ matrix.python }
```

oddbookworm: Should fix the fail with duplicate file names I think?

ankith26: True, I have fixed that issue

oddbookworm (approved):
LGTM! Thanks for doing this! 🎉

PR #3484, opened 2025-06-07
Add versionchanged note to Mask.to_surface

ankith26:
I approved and merged #2670 but immediately realised that there was no addition of a versionchanged note... so here it is


damusss (approved):
https://github.com/user-attachments/assets/778f6bcf-5b6f-481c-b788-5c39d07a0d8c

PR #3485, opened 2025-06-07
Use macro for exception raising

MightyJosip:
In #2371 we added support for one liner for setting exception and returning the value from the functions. If the return value is NULL, we have the same support for 10+ years. In this PR I replaced all the calls to PyErr_SetString (and returning the value) with those macros. There is still work to be done with PyErr_Format, but that would make this commit too big, so I will do that in the future.


ankith26 (commented):
Overall I really like this idea but I think we should only introduce the macro for the return -1 case. The current raise could remain `RAISE` as it is a better shorthand.

PR #3488, opened 2025-06-08
Typing enhancements for the `pygame.sprite` module

### LINKED ISSUE #3460: `__init__` method for the `_SupportsSprite` protocol in `pygame.sprite` does not allow child classes with overriden `__init__`
Pygame Version: 2.5.3 (but the issue persists in 2.5.4)
Python Version: 3.13.1
Pylance extension version: 2025.5.100

On current `pyright` implementations, adding an instance that inherits from Sprite to a group can cause type errors when either specifying the contents of a group at creation or specifying the type of the contained sprites:

```py
from typing import Sequence

import pygame as pg

class Derived(pg.sprite.Sprite):
    def __init__(
        self,
        pos: pg.Vector2 | tuple[float, float],
        groups: Sequence[pg.sprite.AbstractGroup, ...] = (),
    ):
        super().__init__(*groups)
        self.rect = pg.Rect(pos, (100, 100))

derived = Derived((10, 10))
group = pg.sprite.Group(derived)  # type error!
group = pg.sprite.Group[Derived](derived)  # type error!
```
This is the error message that pyright gives you on VSCode:

```text
Type "Derived" cannot be assigned to type variable "_TSprite@Group"
  Type "Derived" is not assignable to upper bound "_SpriteSupportsGroup" for type variable "_TSprite@Group"
    "Derived" is incompatible with protocol "_SpriteSupportsGroup"
      "_SupportsSprite" is not assignable to "Derived"
      "_SupportsSprite" is not assignable to "Derived"
      "_SupportsSprite" is not assignable to "Derived"
      "_SupportsSprite" is not assignable to "Derived"
      "__init__" is an incompatible type
        Type "(pos: Vector2 | tuple[float, float], groups: tuple[AbstractGroup[Unknown], ...] = ()) -> None" is not assignable to type "(*groups: _Group) -> None"
```

And a screenshot of the editor with the specific things that error:
Image

I tried commenting out the `__init__` line in sprites.pyi for the `_SupportsSprite` protocol, and all the errors seem to go away.

I couldn't see any actual use of that line in the rest of the stub file, especially since it's part of a protocol for a class that is meant to be inherited from (and will need different initializer values for each implementation), and not part of the concrete sprite implementation class (which I agree should have it), but I may be missing something else.

Thoughts? Happy to help with a PR for some tweaks to sprite typing (I've been delving a bit deeper with sprite and group types for a project).

Thanks for all the great work as always!

nachomaiz:
This is a consolidation of the personal tweaks that I made to my local `sprites.pyi` file.

I created an issue for one of the simpler ones (#3460), but @oddbookworm encouraged me to make some more suggestions.

There were a few issues with the `pygame.sprite` stub file that I was fixing in my local files or working around with complicated setups to have my code editor understand properly. A couple of things I'd consider bugs, and a few were incomplete or somewhat convoluted implementations.

## Bugs fixed

- **Removing `__init__` from `_SupportsSprite`**: this makes subclasses of `Sprite` that take in any initialization arguments compatible with `AbstractGroup`. Fixes #3460.
- **`DirtySprite` inherits from `Sprite`**: make it explicit that `DirtySprite` inherits from `Sprite` like it does in `sprite.py`.
- **`TypeVar`s in sprite functions**: The `callback` parameter types were not correct. They were all `[_TSprite, TSprite2]` when some of them should have been `[_THasRect, _TSprite]`.

## Improvements

- **Handling of typing features in older versions of Python:** `mypy`, `pyright` and `pyrefly` either support or require `typing-extensions` so it should be safe to use for `Self` and `deprecated`. Imports from `type_extensions` depending on `sys.version_info`.
- **More generics in sprite functions**: made `_HasRect` generic for functions with callbacks.
  ```py
  spritecollide(A(), Group[B](), False, callback)  # callback type: Callable[[A, B], Any], returns list[B]
  ```
- **`AbstractGroup.__class_getitem__` types match `_TSprite`**: making `item` be of type `type[_SupportsSprite]` is similar to `_TSprite` being bound to `_SupportsSprite`.
- **All group types are now generic.**
- **Simplify `__init__`, `add` and `remove` types**: An `AbstractGroup` is an `Iterable`, and in all 3 methods, the logic checks for a `Sprite`, an `Iterable[_SupportsSprite]`, and a `_SupportsSprite` (let's say).
- **Simplify `Protocol` inheritances**: Make `_SupportsSprite` inherit from `_HasImageAndRect` explicitly to remove the extra layer of inheritance with `_SpriteSupportsGroup`.
- **Turned `Sprite` and `DirtySprite` generic**: It is now possible to declare a `Sprite[Group]` type. More below...
  - I had to add `__class_getitem__` to `pygame.sprite.Sprite`, but there are no other changes to `sprite.py`.
  - Maybe it should have been its own PR. Let me know if you'd prefer that.

## Generic `Sprite` and `DirtySprite`

> [!NOTE]
> I'll preface by saying that I didn't land on what I'd consider the ideal solution, bu I would like to know your thoughts on whether this is worth doing. The rest of the changes should be able to work without this one.

It always bothered me that `Sprite.groups()` did not return the type of `AbstractGroup`. My motivation to change this was because a while ago I was trying to change the layer the sprite gets drawn on inside its `update` method:

```py
# contrived example
class Example(Sprite):
    def update(self) -> None:
        # ...
        if move_to_top:
            self.groups()[0].change_layer(self, 9999)
```

But in general, I think it would be better if we can get access to the group types that the sprites have registered. In order to do that, we would need to be able to specify the type of group that the `Sprite` should be contained in:

```py
class MySprite(Sprite[LayeredUpdates]):
    ...

MySprite().groups()  # -> list[LayeredUpdates]
```

Using `Sprite[AbstractGroup]` would allow all groups to be used.

### Some issues with this implementation

I wanted it to provide more safety, so if you say that a sprite inherits from `Sprite[LayeredUpdates]` because it expects the group to draw in layers for example, the type system would not let you add it to simple `Group`s.

This does work with the `Sprite.add` method, but is ignored from the `Group.add` method:

```py
sprite: Sprite[LayeredUpdates]
group: Group
sprite.add(group)  # error
group.add(sprite)  # ok, but should ideally be an error
```

There's a discussion over in the `python/typing` repo on whether `TypeVars` should be able to be generic as well, which would probably enable better type safety with this kind of circular container architecture, but it doesn't look like it's going in a definitive direction at the moment. https://github.com/python/typing/issues/548

The good thing is that, if you don't specify the type of the `AbstractGroup` in the definition of your sprite, it basically behaves the same way as before, except that the type of `sprite.groups()` would be `list[Unknown]`. I think this is the only big drawback of this type definition compared to the previous one.

The simplest way to avoid this drawback is to be explicit and specify the inheritance with:

```py
class MySprite(Sprite[AbstractGroup]):
    pass
```

All of this also means that the implementation is not compatible with `mypy --strict` as it creates a circular type relationship that the language cannot express yet. However, I don't believe the previous implementation was compatible either.

## Testing

I've created this script to test the changes with type checkers.

> [!WARNING]  
> This script will fail to run. Use only with type checkers!

<details><summary>Paste into test_sprite_types.py</summary><p>

```py
from typing import Sequence, reveal_type

import pygame as pg

# SPRITE-GROUP INTERACTIONS


# This works the same as before, except without `Sprite[AbstractGroup]`, group types are erased
class UntypedSprite(pg.sprite.Sprite):
    def __init__(self, *groups: pg.sprite.AbstractGroup) -> None:
        super().__init__(*groups)
        self.image = pg.Surface((50, 50))
        self.rect = self.image.get_rect()


# Expected groups can be specified using type variables.
class MySprite(pg.sprite.Sprite[pg.sprite.Group]):
    # _SupportsSprite does not enforce a signature for the constructor.
    def __init__(
        self, active: bool = False, groups: Sequence[pg.sprite.Group] = ()
    ) -> None:
        super().__init__(*groups)
        self.image = pg.Surface((50, 50))
        self.rect = self.image.get_rect()
        self.active = active


# If a class will use its own groups during updates for example,
# it should be annotated with the type of the group it expects.
# This makes it possible to ensure that groups have a `change_layer` method for example.
class MyDirtySprite(pg.sprite.DirtySprite[pg.sprite.LayeredDirty]):
    def __init__(
        self, active: bool = False, groups: Sequence[pg.sprite.LayeredDirty] = ()
    ) -> None:
        super().__init__(*groups)
        self.image = pg.Surface((50, 50))
        self.rect = self.image.get_rect()
        self.active = active

    def update(self) -> None:
        # Groups are correctly typed when iterating
        for grp in self.groups():
            reveal_type(grp)  # should be LayeredDirty
            if self.active:
                grp.change_layer(self, 1)

untyped = UntypedSprite()
sprite = MySprite()
other = MyDirtySprite()

# Groups are correctly typed when iterating
reveal_type(untyped.groups())  # should be list[Unknown]
reveal_type(sprite.groups())  # should be list[Group]
reveal_type(other.groups())  # should be list[LayeredDirty]

# When initializing a group, the type variable is inferred from the passed sprite type(s).
group = pg.sprite.Group(sprite)
group.add(other)  # Error: Type mismatch, expected MySprite
group.add(sprite)  # ok
other.add(group)  # Error: Type mismatch, expected LayeredDirty
sprite.add(group)  # ok

group2 = pg.sprite.LayeredDirty(other)
group2.add(other)  # ok
group2.add(sprite)  # Error: Type mismatch, expected MyDirtySprite
other.add(group2)  # ok
sprite.add(group2)  # Error: Type mismatch, expected Group

# Mypy complains that `group3` needs an annotation.
group3 = pg.sprite.Group()  # Error: Type variable 'T_Sprite' is unbound
group3.add(sprite)  # ok
sprite.add(group3)  # ok
other.add(group3)  # Error: Type mismatch, expected LayeredDirty

# This one should ideally be an error, because `other` expects a LayeredDirty group,
# but Python does not support Higher Kinded Types, so we cannot express this constraint.
# See: <https://github.com/python/typing/issues/548>
group3.add(other)  # ok, but should be an error

# Declare allowed sprites as a type variable to Group explicitly instead.
# This should be done when initializing an empty group.
# This mimics typing behavior of other containers like list, dict, etc.
group4 = pg.sprite.Group[MySprite]()
group4.add(sprite)  # ok
group4.add(other)  # Error: Type mismatch, expected MySprite

# LayeredDirty will reject MySprite because it expects DirtySprite or a subclass.
pg.sprite.LayeredDirty[MySprite]()  # Error: Type mismatch, expected DirtySprite
pg.sprite.LayeredDirty(sprite)  # Error: Type mismatch, expected DirtySprite

# A LayeredDirty group without declared type variable will accept normal sprites unfortunately.
# It will accept any object in fact, but this is the same behavior as before.
# Mypy does complain that `group5` needs an annotation.
group5 = pg.sprite.LayeredDirty()  # Error: Type variable 'T_Sprite' is unbound
group5.add(other)  # ok
group5.add(sprite)  # ok, but should be an error
group5.add(1)  # ok, but should be an error

group6 = pg.sprite.Group(sprite, other)  # ok
reveal_type(group6)  # should be Group[MySprite | MyDirtySprite]


# GROUP COLLISION DETECTION


def if_collide(a: pg.sprite.Sprite, b: pg.sprite.Sprite) -> None:
    # accepts any two sprites
    pass


def if_collide_specific(a: MySprite, b: MyDirtySprite) -> None:
    # ok when checking if `sprite` collides with `group2`
    # type error when checking if `other` collides with `group`
    pass


pg.sprite.spritecollide(sprite, group, False)  # ok
out = pg.sprite.spritecollide(other, group, False, collided=if_collide)  # ok
reveal_type(out)  # should be list[MySprite]

pg.sprite.spritecollide(sprite, group2, False, collided=if_collide_specific)  # ok
pg.sprite.spritecollide(other, group, False, collided=if_collide_specific)  # error

out2 = pg.sprite.groupcollide(group, group2, False, False, collided=if_collide)  # ok
reveal_type(out2)  # should be dict[MySprite, list[MyDirtySprite]]

out3 = pg.sprite.spritecollideany(sprite, group)  # ok
reveal_type(out3)  # should be MySprite | None


# SPRITE COLLISION DETECTION


class Masked:
    def __init__(self) -> None:
        self.mask = pg.mask.Mask((50, 50))
        self.rect = pg.Rect(0, 0, 50, 50)


class Circle:
    def __init__(self) -> None:
        self.rect = pg.Rect(0, 0, 50, 50)
        self.radius = 25


class NonSprite:
    def __init__(self) -> None:
        self.image = pg.Surface((50, 50))
        self.rect = self.image.get_rect()


masked = Masked()
circle = Circle()
non_sprite = NonSprite()

pg.sprite.collide_rect(sprite, other)  # ok
pg.sprite.collide_rect(sprite, non_sprite)  # ok
pg.sprite.collide_rect_ratio(0.5)(sprite, other)  # ok
pg.sprite.collide_circle(sprite, circle)  # ok
pg.sprite.collide_circle_ratio(0.5)(sprite, circle)  # ok
pg.sprite.collide_mask(sprite, masked)  # ok

```

</p></details>

It should display the following outputs:

> Using `mypy==1.16.0` and the locally built `pygame-ce` on Python `3.13.2`.

```
❯ mypy test_sprite_typing.py
test_sprite_typing.py:43: note: Revealed type is "pygame.sprite.LayeredDirty[Any]"
test_sprite_typing.py:52: note: Revealed type is "builtins.list[Any]"
test_sprite_typing.py:53: note: Revealed type is "builtins.list[pygame.sprite.Group[Any]]"
test_sprite_typing.py:54: note: Revealed type is "builtins.list[pygame.sprite.LayeredDirty[Any]]"
test_sprite_typing.py:58: error: Argument 1 to "add" of "AbstractGroup" has incompatible type "MyDirtySprite"; expected "_SpriteOrIterable[MySprite]"  [arg-type]
test_sprite_typing.py:60: error: Argument 1 to "add" of "Sprite" has incompatible type "Group[MySprite]"; expected "LayeredDirty[Any]"  [arg-type]
test_sprite_typing.py:65: error: Argument 1 to "add" of "LayeredUpdates" has incompatible type "MySprite"; expected "_SpriteOrIterable[MyDirtySprite]"  [arg-type]
test_sprite_typing.py:67: error: Argument 1 to "add" of "Sprite" has incompatible type "LayeredDirty[MyDirtySprite]"; expected "Group[Any]"  [arg-type]
test_sprite_typing.py:70: error: Need type annotation for "group3"  [var-annotated]
test_sprite_typing.py:73: error: Argument 1 to "add" of "Sprite" has incompatible type "Group[Any]"; expected "LayeredDirty[Any]"  [arg-type]
test_sprite_typing.py:85: error: Argument 1 to "add" of "AbstractGroup" has incompatible type "MyDirtySprite"; expected "_SpriteOrIterable[MySprite]"  [arg-type]
test_sprite_typing.py:88: error: Value of type variable "_TDirtySprite" of "LayeredDirty" cannot be "MySprite"  [type-var]
test_sprite_typing.py:89: error: Value of type variable "_TDirtySprite" of "LayeredDirty" cannot be "MySprite"  [type-var]
test_sprite_typing.py:94: error: Need type annotation for "group5"  [var-annotated]
test_sprite_typing.py:100: note: Revealed type is "pygame.sprite.Group[pygame.sprite._SupportsSprite[Any]]"
test_sprite_typing.py:119: note: Revealed type is "builtins.list[test_sprite_typing.MySprite]"
test_sprite_typing.py:122: error: Argument "collided" to "spritecollide" has incompatible type "Callable[[MySprite, MyDirtySprite], None]"; expected "Callable[[MyDirtySprite, MySprite], Any] | None"  [arg-type]
test_sprite_typing.py:125: note: Revealed type is "builtins.dict[test_sprite_typing.MySprite, builtins.list[test_sprite_typing.MyDirtySprite]]"
test_sprite_typing.py:128: note: Revealed type is "Union[test_sprite_typing.MySprite, None]"
Found 11 errors in 1 file (checked 1 source file)
```


oddbookworm:
Wow... I was not expecting this lol. Sprite is one of those modules that everyone wants to see improved, but nobody wants to touch themselves lol. It might take a hot minute for reviews to roll in on this one, but I'm going to specifically request a review from @ankith26 on this one. I'm pretty sure he's the typing guru here

aatle:
Very nice! These are a thorough set of improvements, and some of them are quite smart.
After reviewing, I see there's still a lot of work to be done on the sprite stubs, so I'll probably make another PR for more improvements after this one is merged.

One minor note about the `deprecated` import from `typing_extensions` improvement: version checking isn't really important to do here (it's the same regardless), and the other imports of `deprecated` don't do the version checking so it is a bit inconsistent. Of course, this improvement can be left in.

buildconfig/stubs/pygame/sprite.pyi
```diff
@@ -87,64 +119,23 @@ class DirtySprite(_SupportsDirtySprite):
...
-# b = Group(MySprite())
-
-# typechecker should pass
-# class MySprite(Sprite):
-#     image: Surface
-#     rect: Rect
-#
-# b = Group(MySprite())
+# typevar for sprite or iterable of sprites, used in Group init, add and remove
+_SpriteOrIterable = Union[_TSprite, Iterable[_SpriteOrIterable]]
```
aatle:
```suggestion
_SpriteOrIterable = Union[_TSprite, Iterable[_SpriteOrIterable[_TSprite]]]
```

buildconfig/stubs/pygame/sprite.pyi
```diff
@@ -87,64 +119,23 @@ class DirtySprite(_SupportsDirtySprite):
...
 
 class AbstractGroup(Generic[_TSprite]):
     spritedict: dict[_TSprite, Optional[Union[FRect, Rect]]]
     lostsprites: list[Union[FRect, Rect]]
-    def __class_getitem__(cls, item: Any, /) -> types.GenericAlias: ...
+    def __class_getitem__(
+        cls, item: type[_SupportsSprite], /
```
aatle: I would prefer to still keep the `item` type as `Any`, because all built-in generic types are like that (see typeshed).

aatle:
I do not think that making `Sprite` generic is worth it.

When the subscript type for a generic is omitted, there is actually less type safety than in the original implementation:
- `Sprite` - existing code that does not subscript `Sprite` would now have their `.groups()` giving `list[Any]` instead of `list[AbstractGroup[_SupportsSprite]]`, and all objects would be accepted as a valid group
- `Sprite`'s `Group` - for a `Sprite[Group]`, the `.groups()` would give `list[Group[Any]]`, so the group's contents lack type information (instead of `list[AbstractGroup[_SupportsSprite]]`)

Additionally, having `Sprite` be generic over the type of containers that hold it is generally not intuitive and doesn't seem that useful.
IMO there is little need to statically enforce what type of groups a sprite is placed in. And it's also a bit backwards to be accessing the *group* from the *sprite*, let alone include the group type in the sprite type.

I think there are alternatives and workarounds for your use case:
- perform this functionality elsewhere, iterating over the group instead of using the sprite's update
    - one possible place is overriding the group's `update()`
- add a dedicated sprite attribute for the group, rather than unsafely reaching into the list of all registered groups
    - or, add a method/property that grabs the group with a cast, so that the cast only has to be done in this one place
- implement a layering system without `LayeredUpdates`, such that sprites own their layer/depth rather than the group
- `cast()` or `# type: ignore` the group access everywhere

buildconfig/stubs/pygame/sprite.pyi
nachomaiz: Sounds good, better to keep it consistent. Will set it back to `Any`

buildconfig/stubs/pygame/sprite.pyi
nachomaiz: Ah, good catch! :)

nachomaiz:
> I do not think that making `Sprite` generic is worth it.

Yeah, I see what you mean, and I agree with your points. I was probably trying to have the type hints do too much.

The type erasure of `TypeVar`s even when they are declared with a `bound` is quite annoying. If it were possible to type correctly there would be a stronger argument for this I think.

Ok, I'll drop the generic changes. At least the rest of changes should work on their own.

nachomaiz:
> One minor note about the `deprecated` import from `typing_extensions` improvement: version checking isn't really important to do here (it's the same regardless), and the other imports of `deprecated` don't do the version checking so it is a bit inconsistent. Of course, this improvement can be left in.

Yeah, you're right it's not needed. Though I think it's more in line with the expectations of the typing community. Hopefully it will make it easier to maintain down the line.

nachomaiz:
Ok, I've reverted `Sprite` to non-generic and the types of `_TGroup` to `_Group` as before.

However, once I did I ran into an issue (I remember encountering it in the past) where this example was giving me an error:

```py
sprite = pg.sprite.Sprite()
group = pg.sprite.Group(sprite)
sprite.add(group)  # type error
```

The error message was:
> Type parameter "_TSprite@AbstractGroup" is invariant, but "Sprite" is not the same as "_SupportsSprite"

Changing `_TSprite` to covariant `_TSprite_co` resolves the issue, but `add_internal`, `remove_internal` and `has_internal` had to be typed as `_SupportsSprite`.

<details><summary>New test script</summary>

```py
from typing import Sequence, reveal_type

import pygame as pg

# SPRITE-GROUP INTERACTIONS


class MySprite(pg.sprite.Sprite):
    # _SupportsSprite does not enforce a signature for the constructor.
    def __init__(
        self, active: bool = False, groups: Sequence[pg.sprite.AbstractGroup] = ()
    ) -> None:
        super().__init__(*groups)
        self.image = pg.Surface((50, 50))
        self.rect = self.image.get_rect()
        self.active = active


class MyDirtySprite(pg.sprite.DirtySprite):
    def __init__(
        self, active: bool = False, groups: Sequence[pg.sprite.AbstractGroup] = ()
    ) -> None:
        super().__init__(*groups)
        self.image = pg.Surface((50, 50))
        self.rect = self.image.get_rect()
        self.active = active


# Other sprite classes for testing type inference
class OtherSprite(pg.sprite.Sprite):
    pass


class OtherDirtySprite(pg.sprite.DirtySprite):
    pass

class DerivedSprite(MySprite):
    pass

sprite = MySprite()
other_sprite = OtherSprite()
derived_sprite = DerivedSprite()  # should be compatible with MySprite
dirty = MyDirtySprite()
other_dirty = OtherDirtySprite()

reveal_type(sprite.groups())  # should be list[_Group]
reveal_type(dirty.groups())  # should be list[_Group]
reveal_type(other_sprite.groups())  # should be list[_Group]
reveal_type(other_dirty.groups())  # should be list[_Group]

# When initializing a group, the type variable is inferred from the passed sprite type(s).
group = pg.sprite.Group(sprite)  # group is Group[MySprite]
group.add(dirty, other_sprite, other_dirty)  # Error: Type mismatch, expected MySprite
group.add(derived_sprite)  # ok, derived from MySprite
group.add(sprite)  # ok
dirty.add(group)  # ok
sprite.add(group)  # ok
derived_sprite.add(group)  # ok, derived from MySprite

for grp in sprite.groups():
    reveal_type(grp)  # should be AbstractGroup[_SupportsSprite]
    for spr in grp:
        reveal_type(spr)  # should be _SupportsSprite

group2 = pg.sprite.LayeredDirty(dirty)  # group2 is LayeredDirty[MyDirtySprite]
group2.add(dirty)  # ok
dirty.add(group2)  # ok
group2.add(other_dirty, sprite, other_sprite)  # Error: Type mismatch, expected MyDirtySprite
sprite.add(group2)  # ok, but should this be an error?? (Sprite is not _SupportsDirtySprite)

# When initializing a group with multiple sprite types, the type variable is a union of the types.
# group3 is LayeredDirty[MyDirtySprite | OtherDirtySprite]
group3 = pg.sprite.LayeredDirty(dirty, other_dirty)

# GROUP COLLISION DETECTION


def if_collide(a: pg.sprite.Sprite, b: pg.sprite.Sprite) -> None:
    # accepts any two sprites
    pass


def if_collide_specific(a: MySprite, b: MyDirtySprite) -> None:
    # ok when checking if `sprite` collides with `group2`
    # type error when checking if `other` collides with `group`
    pass


pg.sprite.spritecollide(sprite, group, False)  # ok
out = pg.sprite.spritecollide(dirty, group, False, collided=if_collide)  # ok
reveal_type(out)  # should be list[MySprite]

pg.sprite.spritecollide(sprite, group2, False, collided=if_collide_specific)  # ok
pg.sprite.spritecollide(dirty, group, False, collided=if_collide_specific)  # error

out2 = pg.sprite.groupcollide(group, group2, False, False, collided=if_collide)  # ok
reveal_type(out2)  # should be dict[MySprite, list[MyDirtySprite]]

out3 = pg.sprite.spritecollideany(sprite, group)  # ok
reveal_type(out3)  # should be MySprite | None

# SPRITE COLLISION DETECTION


class Masked:
    def __init__(self) -> None:
        self.mask = pg.mask.Mask((50, 50))
        self.rect = pg.Rect(0, 0, 50, 50)


class Circle:
    def __init__(self) -> None:
        self.rect = pg.Rect(0, 0, 50, 50)
        self.radius = 25


class NonSprite:
    def __init__(self) -> None:
        self.image = pg.Surface((50, 50))
        self.rect = self.image.get_rect()


masked = Masked()
circle = Circle()
non_sprite = NonSprite()

pg.sprite.collide_rect(sprite, dirty)  # ok
pg.sprite.collide_rect(sprite, non_sprite)  # ok
pg.sprite.collide_rect_ratio(0.5)(sprite, dirty)  # ok
pg.sprite.collide_circle(sprite, circle)  # ok
pg.sprite.collide_circle_ratio(0.5)(sprite, circle)  # ok
pg.sprite.collide_mask(sprite, masked)  # ok
```

</details>

Type checker outputs:

```
❯ mypy test_sprite_typing.py
test_sprite_typing.py:46: note: Revealed type is "builtins.list[pygame.sprite.AbstractGroup[pygame.sprite._SupportsSprite]]"
test_sprite_typing.py:47: note: Revealed type is "builtins.list[pygame.sprite.AbstractGroup[pygame.sprite._SupportsSprite]]"
test_sprite_typing.py:48: note: Revealed type is "builtins.list[pygame.sprite.AbstractGroup[pygame.sprite._SupportsSprite]]"
test_sprite_typing.py:49: note: Revealed type is "builtins.list[pygame.sprite.AbstractGroup[pygame.sprite._SupportsSprite]]"
test_sprite_typing.py:53: error: Argument 1 to "add" of "AbstractGroup" has incompatible type "MyDirtySprite"; expected "_SpriteOrIterable[MySprite]"  [arg-type]
test_sprite_typing.py:53: error: Argument 2 to "add" of "AbstractGroup" has incompatible type "OtherSprite"; expected "_SpriteOrIterable[MySprite]"  [arg-type]
test_sprite_typing.py:53: error: Argument 3 to "add" of "AbstractGroup" has incompatible type "OtherDirtySprite"; expected "_SpriteOrIterable[MySprite]"  [arg-type]
test_sprite_typing.py:61: note: Revealed type is "pygame.sprite.AbstractGroup[pygame.sprite._SupportsSprite]"
test_sprite_typing.py:63: note: Revealed type is "pygame.sprite._SupportsSprite"
test_sprite_typing.py:68: error: Argument 1 to "add" of "LayeredUpdates" has incompatible type "OtherDirtySprite"; expected "_SpriteOrIterable[MyDirtySprite]"  [arg-type]
test_sprite_typing.py:68: error: Argument 2 to "add" of "LayeredUpdates" has incompatible type "MySprite"; expected "_SpriteOrIterable[MyDirtySprite]"  [arg-type]
test_sprite_typing.py:68: error: Argument 3 to "add" of "LayeredUpdates" has incompatible type "OtherSprite"; expected "_SpriteOrIterable[MyDirtySprite]"  [arg-type]
test_sprite_typing.py:91: note: Revealed type is "builtins.list[test_sprite_typing.MySprite]"
test_sprite_typing.py:94: error: Argument "collided" to "spritecollide" has incompatible type "Callable[[MySprite, MyDirtySprite], None]"; expected "Callable[[MyDirtySprite, MySprite], Any] | None"  [arg-type]
test_sprite_typing.py:97: note: Revealed type is "builtins.dict[test_sprite_typing.MySprite, builtins.list[test_sprite_typing.MyDirtySprite]]"
test_sprite_typing.py:100: note: Revealed type is "Union[test_sprite_typing.MySprite, None]"
Found 7 errors in 1 file (checked 1 source file)
```

aatle:
Sorry, but it is incorrect for `_TSprite` to be covariant and you'll need to revert that.
This is because `Group` is a mutable collection, not immutable, and so its generic typevar must be invariant (similar to `list`). I can explain further if needed.

The fact that `Group` is invariant is also the cause of your error, of course.
I think the best course of action is to have the group type used in `Sprite` be `AbstractGroup[Any]` instead of `AbstractGroup[_SupportsSprite]`. This will remove all of the false positives, but unfortunately it will create a few false negatives (less type safety). Note that `AbstractGroup[Self]` cannot be used because it leads to undesired strictness too.
However, the return type of `sprite.groups()` should still be `list[AbstractGroup[_SupportsSprite]]`. This is because it is the most correct: the sprite's groups cannot be verified to hold anything more specific than all sprites. So, `list[AbstractGroup[Self]]` is wrong, and `list[AbstractGroup[Any]]` is not useful because a type inside the group other than `_SupportsSprite` is technically wrong.

aatle:
Oh, I forgot to mention, most people don't come across the sprite type error you're experiencing because they use `group.add(sprite)` instead of `sprite.add(group)`. I'm not sure why the latter really exists.

nachomaiz:
Yeah, you're right. I've reverted it to invariant now.

nachomaiz:
Apologies, and set the groups return type to `list[AbstractGroup[_SupportsSprite]`

nachomaiz:
> Oh, I forgot to mention, most people don't come across the sprite type error you're experiencing because they use `group.add(sprite)` instead of `sprite.add(group)`. I'm not sure why the latter really exists.

Yeah, I was just trying to be thorough. 😊

Agree, not sure why or how I found it, or if people use it. I tend to explore the pygame source code when trying things out so I probably saw that it errored and just used `group.add` from then on.

nachomaiz:
> Sorry, but it is incorrect for `_TSprite` to be covariant and you'll need to revert that. This is because `Group` is a mutable collection, not immutable, and so its generic typevar must be invariant (similar to `list`). I can explain further if needed.

Also, thank you for this. I was reading through the typing docs for variance and it wasn't clicking until I saw the right words in the right order.

nachomaiz:
Thank you very much for all your help!

buildconfig/stubs/pygame/sprite.pyi
```diff
@@ -87,59 +118,14 @@ class DirtySprite(_SupportsDirtySprite):
...
-# b = Group(MySprite())
-
-# typechecker should pass
-# class MySprite(Sprite):
-#     image: Surface
-#     rect: Rect
-#
-# b = Group(MySprite())
+# typevar for sprite or iterable of sprites, used in Group init, add and remove
+_SpriteOrIterable = Union[_TSprite, Iterable[_SpriteOrIterable[_TSprite]]]
```
ankith26, line 128: What surprised me is that this is making the definition recursive. I had to double check the implementation and yes, the code does accept arbitrarily nested iterables, with agrees with this definition. This is fine for now but may be something we could rethink in pygame-ce 3

ankith26 (approved):
LGTM, thanks for contributing to pygame-ce :tada: 

Loads of nice enhancements and fixes in this one PR :smile:

PR #3489, opened 2025-06-08
Fix cppcheck skip logic on CI

ankith26:
#3483 introduced a minor CI whoopsie :facepalm: causing cppcheck fails on main. This PR adds a fix and also makes the entire step optional (and on error, the cppcheck logic should run)


oddbookworm (approved):
Pulled it into my mega branch and ran it in a real-world scenario lol. Works fine. LGTM!

PR #3490, opened 2025-06-08
Add missing/Fix versionadded tags in display, mixer and surface

damusss:


oddbookworm (approved):
LGTM! I should have caught that the first time lol

ankith26 (commented):
I took this opportunity to go through all the past new API and I found another versionadded related issue. It would be nice to fix in the same PR:
`Surface.get_frect` and `Channel.id` have versionadded but it doesn't render correctly because of a missing space

PR #3492, opened 2025-06-09
Refactor rect properties

JoKing-1999:
A lot of code used for different rect attributes is repeated. So this PR moves it to macro, and for each attribute only defines unique parts

### DIFF
--- a/src_c/rect_impl.h
+++ b/src_c/rect_impl.h
@@ -2388,487 +2388,135 @@ Unimplemented:
     return Py_NotImplemented;
 }
 
-/*width*/
-static PyObject *
-RectExport_getwidth(RectObject *self, void *closure)
-{
-    return PythonNumberFromPrimitiveType(self->r.w);
-}
-
-static int
-RectExport_setwidth(RectObject *self, PyObject *value, void *closure)
-{
-    PrimitiveType val1;
-
-    if (NULL == value) {
-        /* Attribute deletion not supported. */
-        PyErr_SetString(PyExc_AttributeError, "can't delete attribute");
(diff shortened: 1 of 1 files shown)


src_c/rect_impl.h
```diff
@@ -2388,487 +2388,135 @@ RectExport_richcompare(PyObject *o1, PyObject *o2, int opid)
...
+               PrimitiveType val2, twoPrimitivesFromObj(value, &val1, &val2),
+               self->r.x = val1;
+               self->r.y = val2;)
+
+RECT_ATTRIBUTE(topright,
+               TupleFromTwoPrimitives(self->r.x + self->r.w, self->r.y),
+               PrimitiveType val1;
+               PrimitiveType val2, twoPrimitivesFromObj(value, &val1, &val2),
+               self->r.x = val1 - self->r.w;
+               self->r.y = val2;)
```
zoldalma999, line 2460: Not sure how much the formatter allows it, but it was not clear to me that val1 and val2 are in the same argument, or what each argument does. Maybe adding comments might help?

zoldalma999 (commented):
There's a conflict now, because relcenter was added. Also added a nitpicky comment/question.

Starbuck5:
Making functions inside macros is definitely tempting, especially for repetition like this. However it does make it harder to jump around the code with editor tools and sometimes to read at all. (Although rect is already pretty bad in this regard).

So my idea here would be to leave the function definitions hardcoded, but use macros to take over repetitive parts / the whole of the bodies of the functions.

Anyways marking as draft pending merge conflict resolution.

PR #3494, opened 2025-06-09
pygame.Window.get_info()

gresm:
Alternative to ``pygame.Window._handle`` (+extra).


damusss:
I don't particurly like this, because:
- get_info does not describe what kind of info it is, and looks more important than the returned value
- there are a lot of properties that are basically useless, and they are inconsistent between sdl2/sdl3. this dict-like api feels "old" as pointed out by starbuck in my PR
- we are already having a hard time deciding if the window handle is worth it, whether all the many numbers are worth it would be even harder

gresm:
@damusss That's completely understandable. Just note that this PR is more of "port ``pygame.display.get_wm_info``", only that adding support for ``pygame.Window`` was trivial. The API is intentionally the same. Nonetheless, I'm open for suggestions. 

IMHO, exposing window identifier is useful.

MyreMylar:
To reduce the number of PRs competing to solve the same problem, I think we should try for just the window handle PR for now and then we can see if later on there is any user demand for exposing other, seemingly more obscure, platform specific properties of windows to be exposed in pygame-ce.

PR #3496, opened 2025-06-10
Wasm/py313

pmp-p:
- reordering of static imports,
- adapt surface module to avoid runtime error and typing problems ( sprite.py wants pygame.surface.Surface too early )
- tested pygame-ce 2.5.6.dev1 (SDL 2.30.9, Python 3.13.3) on emsdk 3.1.74 and 3.1.61 : https://pygame-web.github.io/showroom/pythongit.html?cpython3.13#src/test_gfxdraw.py

### DIFF
--- a/src_c/surface.c
+++ b/src_c/surface.c
@@ -4595,7 +4595,7 @@ pgSurface_Blit(pgSurfaceObject *dstobj, pgSurfaceObject *srcobj,
 }
 
 static PyMethodDef _surface_methods[] = {{NULL, NULL, 0, NULL}};
-#if !defined(BUILD_STATIC)
+
 int
 exec_surface(PyObject *module)
 {
@@ -4630,11 +4630,11 @@ exec_surface(PyObject *module)
 
     PyObject *apiobj;
     static void *c_api[PYGAMEAPI_SURFACE_NUMSLOTS];
-
+#ifndef BUILD_STATIC
     if (pg_warn_simd_at_runtime_but_uncompiled() < 0) {
         return -1;
     }
-
+#endif
     if (PyModule_AddObjectRef(module, "SurfaceType",
                               (PyObject *)&pgSurface_Type)) {
         return -1;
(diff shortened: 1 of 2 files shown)


src_c/surface.c
```diff
@@ -4688,3 +4688,63 @@ MODINIT_DEFINE(surface)
 
     return PyModuleDef_Init(&_module);
 }
+#else
```
oddbookworm, line 4687: Is multiphase initialization not compatible at all?

ankith26: This adds code duplication, even if multiphase initialization is not supported on build_static we can simply re-use the exec code something like:

```c
MODINIT_DEFINE(surface)
{
#ifndef BUILD_STATIC
    static PyModuleDef_Slot surf_slots[] = {
        {Py_mod_exec, &exec_surface},
#if PY_VERSION_HEX >= 0x030c0000
        {Py_mod_multiple_interpreters,
         Py_MOD_MULTIPLE_INTERPRETERS_NOT_SUPPORTED},  // TODO: see if this can
                                                       // be supported later
#endif
#if PY_VERSION_HEX >= 0x030d0000
        {Py_mod_gil, Py_MOD_GIL_USED},  // TODO: support this later
#endif
        {0, NULL}};
#endif
    static struct PyModuleDef _module = {PyModuleDef_HEAD_INIT,
                                         "surface",
                                         DOC_SURFACE,
                                         0,
                                         _surface_methods,
#ifndef BUILD_STATIC
                                         surf_slots,
#else
                                         NULL,
#endif
                                         NULL,
                                         NULL,
                                         NULL};
#ifndef BUILD_STATIC
    return PyModuleDef_Init(&_module);
#else
    PyObject *module = PyModule_Create(&_module);
    if (module) {
        if (exec_surface(module) != 0) {
            Py_DECREF(module);
            return NULL;
        }
    }
    return module;
#endif
}
```

pmp-p: good idea i'll try that

ankith26 (approved):
LGTM. I may not 100% understand these changes but I trust you have tested these changes and it works out okay

PR #3498, opened 2025-06-13
Fix event docs grammar

wade-cheng:


oddbookworm:
@wade-cheng thanks for the contribution! 🎉 
If you're in the Pygame Community discord server, let us know what your username is so you can get the contributor role

wade-cheng:
oh wow, thanks! I'm @saffron_8811.

PR #3499, opened 2025-06-14
Use PyModule_AddType

Starbuck5:
Not exhaustive, but gets in most places it's immediately practical to.

### DIFF
--- a/src_c/_sdl2/controller.c
+++ b/src_c/_sdl2/controller.c
@@ -584,14 +584,7 @@ MODINIT_DEFINE(controller)
         return NULL;
     }
 
-    if (PyType_Ready(&pgController_Type) < 0) {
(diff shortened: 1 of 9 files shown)


src_c/_sdl2/controller.c
```diff
@@ -584,13 +584,8 @@ MODINIT_DEFINE(controller)
         return NULL;
     }
 
-    if (PyType_Ready(&pgController_Type) < 0) {
-        return NULL;
-    }
-
     Py_INCREF(&pgController_Type);
```
JoKing-1999: I am pretty sure you can remove INCREF/DECREF here

Starbuck5: Good spot, thanks. Fixed.

PR #3500, opened 2025-06-14
Pull windows deps on update in meson buildconfig

ankith26:
Previously a pull would happen only if the whole prebuilts directory didn't exist. Now the logic has been improved to check for exact expected versions and handle pull accordingly


Starbuck5 (approved):
I'd become pretty quick at deleting them so it would reinstall by default, but this requires less manual work. Thanks Ankith!

I had thought the real solution is to not hardcode SDL versions in the windows build code, because it should be fine installing on an older version. But this way all our Windows contributors will stay synced on a version depending on what commit they're on, which seems good.

PR #3501, opened 2025-06-15
Fix display.set_mode segfault after resizing

### LINKED ISSUE #2571: Calling a second time ``pygame.display.set_mode`` with ``vsync`` and ``vsync`` already enabled doesn't work correctly

MightyJosip:
This should fix the following code that segfaults
```py
import pygame
pygame.display.set_mode([800, 600], flags=pygame.SCALED)
pygame.display.set_mode([800, 598], flags=pygame.RESIZABLE | pygame.SCALED)
```
This also fixes #2571

TLDR: Changing size of the window invalidates surface returned by SDL_GetWindowSurface() as explained in https://wiki.libsdl.org/SDL2/SDL_GetWindowSurface, so this commit makes sure the module always has the reference to the right SDL_Surface (hopefully). There are other possible fixes for this, but this one to me looks the most elegant (unless we rewrite the entire module/function).

And the long explanation is available in our discord server https://discord.com/channels/772505616680878080/772940667231928360/1383591122101604402

Also, if someone knows another place in our code when the resizing could happen, make sure to notify me to add this new helper function there as well

### DIFF
--- a/src_c/display.c
+++ b/src_c/display.c
@@ -707,6 +707,16 @@ _get_video_window_pos(int *x, int *y, int *center_window)
     return 0;
 }
 
+static void
+_check_window_resized(SDL_Window *window)
+{
+    SDL_Surface *sdl_surface = SDL_GetWindowSurface(window);
+    pgSurfaceObject *old_surface = pg_GetDefaultWindowSurface();
+    if (sdl_surface != old_surface->surf) {
+        old_surface->surf = sdl_surface;
+    }
+}
+
 static int SDLCALL
 pg_ResizeEventWatch(void *userdata, SDL_Event *event)
 {
@@ -786,11 +796,7 @@ pg_ResizeEventWatch(void *userdata, SDL_Event *event)
     }
 
     if (event->window.event == SDL_WINDOWEVENT_SIZE_CHANGED) {
-        SDL_Surface *sdl_surface = SDL_GetWindowSurface(window);
-        pgSurfaceObject *old_surface = pg_GetDefaultWindowSurface();
-        if (sdl_surface != old_surface->surf) {
-            old_surface->surf = sdl_surface;
-        }
+        _check_window_resized(window);
     }
     return 0;
 }
@@ -1374,6 +1380,7 @@ pg_set_mode(PyObject *self, PyObject *arg, PyObject *kwds)
             surface = pgSurface_New2(surf, newownedsurf != NULL);
         }
         else {
+            _check_window_resized(win);
             pgSurface_SetSurface(surface, surf, newownedsurf != NULL);
             Py_INCREF(surface);
         }


ankith26 (approved):
LGTM, thanks for the fix! :tada: 

re: memory leak, I am not seeing any noticeable evidence of that. I believe that even if there is a memory leak, it would probably not be a result of this PR.

> Also, if someone knows another place in our code when the resizing could happen, make sure to notify me to add this new helper function there as well
There is some logic in `pg_flip_internal` that is exactly like `_check_window_resized` you may wanna update that snippet as well, just for code quality reasons. Nothing related to this fix per se.

JoKing-1999:
What I mean by the memory leak is the following:
```py
import pygame
import random
import psutil
import os
pygame.display.set_mode((800, 600), flags=pygame.SCALED)
while True:
    print(f"Memory usage: {psutil.Process(os.getpid()).memory_info().rss / 1024 / 1024:.2f} MB")
    pygame.display.set_mode([random.randint(600, 800), random.randint(600, 800)], flags=pygame.RESIZABLE | pygame.SCALED)
```

It is probably happening because with this change the old surface is not freed. However, if I free the surface, then I return to the old problem of segfault 🤣 . Really hard issue

ankith26:
the old surface should be freed already by SDL, it cannot be freed on our side.

src_c/display.c
```diff
@@ -1374,6 +1383,7 @@ pg_set_mode(PyObject *self, PyObject *arg, PyObject *kwds)
             surface = pgSurface_New2(surf, newownedsurf != NULL);
         }
         else {
+            _check_window_resized(win, 0);
```
gresm, line 1386: I'm probably failing to notice something here, but why aren't we freeing the old surface here?

JoKing-1999: Because if we free then we get segmentation fault from the linked issue. You have more explanation in the discord link

ankith26 (changes requested):
I'm not sure if the current freeing logic here is okay.

We essentially have to consider two cases:
1. The display surface is created on our side with a `PG_CreateSurface` call. In this case we are responsible for freeing it, otherwise it leaks
2. The display surface is obtained via a call from `SDL_GetWindowSurface`. In this case calling `SDL_FreeSurface` is a noop because its managed by SDL. The fact that this surface becomes invalidated by SDL is the reason for any segfault we get while accessing it.

This makes our freeing logic complex. We have to now actually keep track of *how* the previous surface was allocated and free it based on that condition.

MyreMylar:
> I'm not sure if the current freeing logic here is okay.
> 
> We essentially have to consider two cases:
> 
> 1. The display surface is created on our side with a `PG_CreateSurface` call. In this case we are responsible for freeing it, otherwise it leaks
> 2. The display surface is obtained via a call from `SDL_GetWindowSurface`. In this case calling `SDL_FreeSurface` is a noop because its managed by SDL. The fact that this surface becomes invalidated by SDL is the reason for any segfault we get while accessing it.
> 
> This makes our freeing logic complex. We have to now actually keep track of _how_ the previous surface was allocated and free it based on that condition.

Almost sounds like we need something like 'PGObtainDisplaySurface' with a 'PGFreeDisplaySurface' and a 'PGDisplaySurface' that tracks who owns the surface.

PR #3508, opened 2025-06-17
Remove unused code in mask/alphablit

Starbuck5:
The mask module C API is undocumented and unused within pygame-ce. Also removes two strange extern declarations in alphablit.c for functions that are not referenced in the codebase.


MightyJosip (approved):
Yea it is good, I would do some more clearing, like remove both headers atm, and then later reorganize the code after we have everything split up (also when we decide how we should organize everything, so entire codebase is unified, right now that mask.h is completely different to the rest of the code base. I would also move pgMaskObject from pygame_mask. to pygame.h like all the other objects.

But there are no problems with this commit so 👍

PR #3513, opened 2025-06-20
iOS support

seekerluke:
For context, I posted a video on Discord showing pygame-ce working as a native iOS app through Briefcase. This has detailed instructions about how it was done, and what iOS support looks like going forward.

## How to build for iOS (no wheel)
While working on this build, I made a lot of super hacky changes to `meson.build` to create a proof of concept, and those files are part of this PR, as well as an `ios-crossbuild.txt` file which is passed into the `--cross-file` argument. Lots of paths here are local paths on my machine (sorry), so you need to modify them to work for you. But basically:
- A `Python.framework` is linked instead of using `py.dependency` to get a dependency to Python itself. This framework can be found in a generated Briefcase iOS project, more on that below.
- SDL2 and its auxillary libs (image, ttf, mixer) need to be downloaded and built with CMake to get iOS frameworks to link with. Be sure to download and link the dependencies that come with SDL_ttf and SDL_mixer. I used the latest stable for each, but I don't think it matters.
- portmidi and porttime are disabled for this build. freetype2 is not needed as it was linked with SDL_ttf.
- All references to `py.extension_module` are replaced with `shared_module`, and all references to `py` are removed.

The build I made for the Discord video works by building the pygame-ce modules with `meson` directly using `--cross-file` to create iOS artifacts, instead of using meson-python to generate a wheel. I used the following commands:
```
meson setup build-ios --cross-file ios-crossbuild.txt --buildtype=release
meson compile -C build-ios
```

Once you have those built, you can create a fresh Briefcase app, specifying no framework (no Toga, etc) so there's no extra Python code in the way. You then create an Xcode project with `briefcase create iOS`. At this point you can attach your downloaded SDL, SDL_image, SDL_ttf, and SDL_mixer Xcode projects to the generated Xcode project through their respective iOS instructions.

Inside that Xcode project are two `app-packages` folders, one for `iphoneos` and one for `iphonesimulator`. You can copy `src_py` into both of these folders, then rename them `pygame` to make them `import`able from your game's code. You can then copy your built C modules from Meson into this folder so the `pygame` module can see them on import (make sure they are named `base.so` instead of `libbase.dylib` or whatever Meson calls them on build, they need to be `.so` for the custom xcode build phases).

Finally, you need to change `main.m` to get rid of the call to `UIApplicationMain`, since SDL calls this internally on init.

Add your game code to your Briefcase app code in `__main__.py`. Run the app on your device and it should work, I'm having issues with simulator builds but maybe you're smarter than me.

## What's probably needed for an iOS wheel
1. cibuildwheel 3.0, this is the only known stable way to build iOS wheels as far as I know.
2. A dev build of meson-python (unless it's stable by the time you read this).
3. Some way to link SDL2 and friends to the wheel, maybe they have to be wheels themselves.
4. Changes to `meson.build` to support the "ios" plat.
5. PyPI support for iOS wheels for releases, which is apparently ready.

### DIFF
--- a/ios-crossbuild.txt
+++ b/ios-crossbuild.txt
@@ -1,0 +1,20 @@
+[binaries]
+c = 'clang'
+cpp = 'clang++'
+ar = 'ar'
+strip = 'strip'
+pkgconfig = 'pkg-config'
+
+[host_machine]
+system = 'ios'
+cpu_family = 'aarch64'
+cpu = 'aarch64'
+endian = 'little'
+
+[properties]
+needs_exe_wrapper = true
+sys_root = '/Applications/Xcode.app/Contents/Developer/Platforms/iPhoneOS.platform/Developer/SDKs/iPhoneOS18.5.sdk'
+c_args = ['-isysroot', '/Applications/Xcode.app/Contents/Developer/Platforms/iPhoneOS.platform/Developer/SDKs/iPhoneOS18.5.sdk', '-target', 'arm64-apple-ios']
+cpp_args = ['-isysroot', '/Applications/Xcode.app/Contents/Developer/Platforms/iPhoneOS.platform/Developer/SDKs/iPhoneOS18.5.sdk', '-target', 'arm64-apple-ios']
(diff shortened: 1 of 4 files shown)


ankith26:
> 3. Some way to link SDL2 and friends to the wheel, maybe they have to be wheels themselves.

This will probably be the hardest part of getting ios wheels. At the moment we don't have the equivalent of the delocate tool (for bundling all the SDL dependencies into the wheel) on ios.

PR #3514, opened 2025-06-20
Simplify key code

Starbuck5:
- Removes unnecessary null checks (PyBool_FromLong always succeeds)
- Stores directly into PyTuple_SET_ITEM (which steals reference)
- Moves the index declaration into the for loop -- we're not on C89 anymore.


oddbookworm (approved):
LGTM! Thanks! 🎉

MyreMylar (approved):
LGTM 👍

PR #3515, opened 2025-06-21
Port more SDL functions (SDL3)

Starbuck5:
While https://github.com/pygame-community/pygame-ce/pull/3435 does get Surface compiling, it's actual functionality is quite limited due to the prevalence of SDL int return code -> bool return code issues.

When combined with #3435 and built on SDL3, this brings the surface test suite from
`FAILED (failures=8, errors=198, skipped=2)` to `FAILED (failures=5, errors=130, skipped=2)`, a significant improvement. A lot of the remaining fails after this are about the lack of display.


ankith26 (approved):
LGTM

PR #3518, opened 2025-06-30
`(F)Rect.inflate(_ip)` FASTCALL

ankith26:
A low hanging fruit for minor performance enhancements, and also makes the code inline with rest of the file.

## Benchmarking results

| Argument   | Before            | After                  | Improvement (%) |
|------------------|---------------------|----------------------|------------------|
| `Vector2`        | 0.000643572          | 0.000483602          | **25.0%**        |
| `list`           | 0.000617076          | 0.000442811          | **28.2%**        |
| `tuple`          | 0.000617972          | 0.000454474          | **26.5%**        |
| `2 args, inline` | 0.000611962          | 0.000431108          | **29.5%**        |

## Benchmarking code

```py
from timeit import repeat
from pygame import Rect, Vector2

r = Rect(0, 0, 100, 100)
a1 = Vector2(10, 10)
a2 = [10, 10]
a3 = (10, 10)

G = globals()
N = 10000
R = 1000

tests = [
    ("Vector2", "r.inflate(a1)"),
    ("list", "r.inflate(a2)"),
    ("tuple", "r.inflate(a3)"),
    ("2 args, inline", "r.inflate(10, 10)"),
]

for name, test in tests:
    print(name + ": ", end="")
    print(min(repeat(test, globals=G, number=N, repeat=R)))
```


damusss (approved):
LGTM, thank you for the speedup

Starbuck5 (approved):
Thanks!

Took special care to test it still works with weirdly recursive inputs, and it does.

PR #3520, opened 2025-06-30
docs: logical_size might be (0, 0)

DickerDackel:
When Renderer.logical_size hasn't been changed before, it will contain (0, 0) and not the size of the window as one might expect.

This bit me, and I only by chance found a remark pointing that out in the SDL2 docs.

This adds a similar remark to the pygame-ce docs.


ankith26 (approved):
LGTM, thanks for the PR! :partying_face:

damusss (approved):
Good idea to specify it

PR #3522, opened 2025-07-01
fixed incredibly small grammar error

tomatophu:
This error irritates me a little bit, so I fixed it.


ankith26 (approved):
LGTM, thanks for contributing :balloon: While the old version was not wrong I think, the updated version is more succinct and therefore better.

aatle (approved):
LGTM, thanks. I agree that the change makes more sense and is consistent with `range`.

PR #3507, opened 2025-06-17
Added Joystick.set_led

oddbookworm:
closes #3506

PR #3524, opened 2025-07-01
Fix local stubtest for `set_led()`

aatle:
Stubtest correctly fails locally because the `color` argument of 2 `set_led()` methods is not typed as positional-only. Fixes the error.
#3507 
```
error: pygame._sdl2.controller.Controller.set_led is inconsistent, stub argument "color" should be positional-only (rename with a leading double underscore, i.e. "__object")
Stub: in file C:\Users\antho\development\pygame-ce\buildconfig\stubs\pygame\_sdl2\controller.pyi:34
def (self: pygame._sdl2.controller.Controller, color: Union[pygame.color.Color, pygame.typing.SequenceLike[builtins.int], builtins.str, builtins.int]) -> builtins.bool
Runtime:
def (self, object, /)

error: pygame.joystick.JoystickType.set_led is inconsistent, stub argument "color" should be positional-only (rename with a leading double underscore, i.e. "__object")
Stub: in file C:\Users\antho\development\pygame-ce\buildconfig\stubs\pygame\joystick.pyi:35
def (self: pygame.joystick.JoystickType, color: Union[pygame.color.Color, pygame.typing.SequenceLike[builtins.int], builtins.str, builtins.int]) -> builtins.bool
Runtime:
def (self, object, /)

Found 2 errors (checked 179 modules)
```
___
Since these uncaught issues are common, I did some more investigation and found that only python 3.13 experiences this error. Python versions from 3.9-3.12 did not catch this problem. (3.9 had an unrelated error but it is small and I won't fix it.)
So these errors could appear in GitHub CI later as python versions update.


ankith26 (approved):
LGTM, good catch :tada:

damusss (approved):
LGTM, thanks!

PR #3527, opened 2025-07-09
Add pypy 3.11 support

oddbookworm:


pyproject.toml
```diff
@@ -75,7 +75,7 @@ install = ['--tags=runtime,python-runtime,pg-tag']
 # dependencies. Here is where uv comes into the picture. It is an "installer" like pip,
 # but faster. It has been observed to save a couple of minutes of CI time.
 build-frontend = "build[uv]"
-build = "cp3{9,10,11,12,13}-* pp3{9,10}-*"
+build = "cp3{9,10,11,12,13}-* pp3{9,10,11}-*"
```
ankith26:
```suggestion
build = "cp3{9,10,11,12,13}-* pp3{10,11}-*"
```
the 9 can be dropped from this selector now

oddbookworm:
> numpy pypy 3.11 manylinux wheels are `manylinux_2_28`, so the wheels cannot be installed on our ` manylinux_2_17` (AKA `manylinux2014`) build environment. The easiest fix here is to simply skip numpy on pypy 3.11 manylinux.

Is updating our build an option?

ankith26:
> Is updating our build an option?

That can be considered in the long term but for now I think that would be the worse option. We would loose support for some older platforms if we did that, and for little benefit (testing numpy specific features on pypy 3.11)

oddbookworm:
> That can be considered in the long term but for now I think that would be the worse option. We would loose support for some older platforms if we did that, and for little benefit (testing numpy specific features on pypy 3.11)

"older platforms"
You mean EOL systems? We already have wheels that they can install, and those wheels aren't going anywhere. We shouldn't prevent ourselves from updating to new things because of EOL systems. We have to stop supporting them eventually

Starbuck5:
> > Is updating our build an option?
> 
> That can be considered in the long term but for now I think that would be the worse option. We would loose support for some older platforms if we did that, and for little benefit (testing numpy specific features on pypy 3.11)

We would **lose** support! @ankith26 It is now my mission in life to make sure you use the right lose/loose, I'm sorry to impose.

Starbuck5 (approved):

Starbuck5:
> > That can be considered in the long term but for now I think that would be the worse option. We would loose support for some older platforms if we did that, and for little benefit (testing numpy specific features on pypy 3.11)
> 
> "older platforms" You mean EOL systems? We already have wheels that they can install, and those wheels aren't going anywhere. We shouldn't prevent ourselves from updating to new things because of EOL systems. We have to stop supporting them eventually

So I did some research. We're using manylinux2014, which is still supported by the pypa. The next step up is manylinux_2_28. (manylinux2014 is basically manylinux_2_17). Making that jump drops Debian (8, 9) and Ubuntu (14.04, 16.04, 18.04).

Ubuntu 20.04 came out on April 23rd 2020, so if someone set up their computer on April 22nd 2020 with Ubuntu they would have 18.04 (I think). Being able to install new software on a computer set up a little bit over 5 years ago sounds reasonable to me. That being said that version of Ubuntu is officially out of support now. See this table for support dates vs manylinux versions: https://github.com/mayeut/pep600_compliance?tab=readme-ov-file#distro-compatibility

oddbookworm:
> Being able to install new software on a computer set up a little bit over 5 years ago sounds reasonable to me

Sure, but that's assuming they haven't updated their distro at all to a newer release version and are actively rocking an EOL version of ubuntu.

ankith26:
I think we can merge this PR as it is and worry about manylinux bumping in a future PR (which will anyways involve a bit more work)

PR #3530, opened 2025-07-13
Add `pygame.display.get_desktop_usable_bounds`

### LINKED ISSUE #3563: Add `pygame.display.get_desktop_workarea()` to retrieve usable display bounds (excluding taskbar / Dock)
I tried to create a borderless window using `pygame.Window` module, but found that the `Window.maximize` function was not working. I suspect that this function uses the system's window controller, which is disabled by the borderless parameter.

So I tried manually achieving maximization by directly setting the `Window.size` and `Window.position` properties. This requires obtaining the screen resolution, but window maximization is different from full-screen mode, it should not include areas where components like the taskbar or Dock are located. I tested the `pygame.display.get_desktop_sizes()` function, which returns the entire screen size, but that's not what I wanted.

In fact, SDL provides a function called `SDL_GetDisplayUsableBounds()`, which retrieves the working area without including the taskbar or Dock. However, pygame does not seem to provide an encapsulation for this function.

I tried using `ctypes` to directly call this function from `SDL2` and successfully obtained the desired rectangular area:

```python
import ctypes
SDL = ctypes.CDLL('SDL2')

SDL.SDL_GetDisplayUsableBounds.argtypes = [ctypes.c_int, ctypes.c_void_p]
SDL.SDL_GetDisplayUsableBounds.restype  = ctypes.c_int

class SDL_Rect(ctypes.Structure):
    _fields_ = [("x", ctypes.c_int),
                ("y", ctypes.c_int),
                ("w", ctypes.c_int),
                ("h", ctypes.c_int)]

r = SDL_Rect()
SDL.SDL_GetDisplayUsableBounds(0, ctypes.byref(r))
print("work-area:", r.x, r.y, r.w, r.h)
```
The SDL2 module is provided directly by pygame. The result is:
```cmd
work-area: 0 0 2560 1528
```
My screen size is 2560×1600, running on Windows 11. This result does not include the taskbar area.

I don't think adding this new API will cause any disruptive changes. This is just a piece of code that I used temporarily, and it may not be universally applicable. If necessary, I can also conduct testing on the MacOS system.

damusss:
As I said on discord, `pygame.display.get_desktop_sizes` exist but doesn't take in consideration the taskbar or any other OS reserved space. The SDL function for it exists, so it makes sense to me to add it to display. It would be very tedious to do this in python cross platform without it. It's a differnt function from the sizes because it returns rects, but the return list uses the same logic so with the same index you get the full size and usable area for the same desktop. Also, it's alredy SDL3 compatible, so there will be no problems when display is ported. There is no conflict with window because this is one of the features that make sense for display, and doesn't concert any windows (display should only hold this functions)


ankith26:
@coderabbitai full review

robertpfeiffer:
This could be used to implement "Do what I mean" window sizing when passing in (0,0) for windowed mode. I always hated how windowed mode overlaps with the task bar in that case!

MyreMylar (commented):
OK, LGTM 👍  Seems like it could be useful for application positioning especially with things like no frame windows.

damusss:
> This could be used to implement "Do what I mean" window sizing when passing in (0,0) for windowed mode. I always hated how windowed mode overlaps with the task bar in that case!

Yeah I agree, useful to manually maximize a window with it being aware of the taskbar or other things occupying space.

Also thanks myre for pushing the memory fixes, have been out of the contributing loop during summer (github even signed me out...)

src_c/display.c
```diff
@@ -2296,6 +2296,65 @@ pg_get_desktop_screen_sizes(PyObject *self, PyObject *_null)
...
 
+static PyObject *
+pg_get_desktop_usable_bounds(PyObject *self, PyObject *_null)
+{
+    int display_count, i;
+    PyObject *result = NULL;
+
+    VIDEO_INIT_CHECK();
+
+#if PG_SDL3
```
ankith26: minor nit: this is technically correct but we use `SDL_VERSION_ATLEAST(3, 0, 0)` consistently throughout the codebase so we should probably just stick to that everywhere.

src_c/display.c
```diff
@@ -2296,6 +2296,65 @@ pg_get_desktop_screen_sizes(PyObject *self, PyObject *_null)
...
+        SDL_free(displays);
+#endif
+        return NULL;
+    }
+
+    for (i = 0; i < display_count; i++) {
+        SDL_Rect bounds;
+#if PG_SDL3
+        SDL_DisplayID display_id = displays[i];
+        if (SDL_GetDisplayUsableBounds(display_id, &bounds) == SDL_FALSE) {
```
ankith26: minor nit: no need to check for equality with boolean, can just do `!`.

test/display_test.py
```diff
@@ -697,6 +697,17 @@ def test_get_set_window_position(self):
...
+        bounds = pygame.display.get_desktop_usable_bounds()
+        sizes = pygame.display.get_desktop_sizes()
+        self.assertIsInstance(bounds, list)
+        self.assertEqual(len(bounds), len(sizes))
+        for i, bound in enumerate(bounds):
+            self.assertIsInstance(bound, pygame.Rect)
+            size = sizes[i]
+            self.assertLessEqual(bound.w, size[0])
+            self.assertLessEqual(bound.h, size[1])
+
```
ankith26, line 723: Can also add a check that checks that all attributes (`x`, `y`, `w`, `h`) are `>= 0`

ankith26 (commented):
left a few minor reviews

test/display_test.py
damusss: All addressed :)

buildconfig/stubs/pygame/display.pyi
```diff
@@ -850,7 +869,8 @@ def message_box(
     :param str message: A message string. If this parameter is set to ``None``, the message will be the title.
     :param str message_type: Set the type of message_box, could be ``"info"``, ``"warn"`` or ``"error"``.
     :param pygame.Window parent_window: The parent window of the message box.
-    :param tuple buttons: An optional sequence of button name strings to show to the user.
+    :param tuple buttons: An 
+    sequence of button name strings to show to the user.
```
Matiiss: Teeny tiny nitpick, it should be "_A_ sequence" 😁

src_c/display.c
```diff
@@ -2821,6 +2821,65 @@ pg_get_desktop_screen_sizes(PyObject *self, PyObject *_null)
...
 
+static PyObject *
+pg_get_desktop_usable_bounds(PyObject *self, PyObject *_null)
+{
+    int display_count, i;
+    PyObject *result = NULL;
+
+    VIDEO_INIT_CHECK();
+
+#if SDL_VERSION_ATLEAST(3, 0, 0)
```
Matiiss, line 2832: The rest of the function uses `#if PG_SDL3`, seems inconsistent.
(Besides, shouldn't we be using `#ifdef PG_SDL3` anyway?)

Matiiss: Apparently the idiomatic approach is to use `#if SDL_VERSION_ATLEAST(3, 0, 0)` (according to @ankith26), so I'd like to see all `#if PG_SDL3` be replaced with it.

damusss: sure!

buildconfig/stubs/pygame/display.pyi
damusss: oops

damusss: Somehow that line was split and went to a new line, losing the word "optional". It wasn't like that before I pulled my local version.

Matiiss:
This is more so coming from @Starbuck5, but pretend it's not so... or don't... anyway

Here's the idea, `pygame.Monitor` or `pygame.Display` (would work better with SDL names), a class and it would have attributes like `.usable_bounds` (at least for the purposes of this PR, so as to not have to implement all of the potential attributes that could go with such a class (and there are many))

So the API would look more like this (if, for example, you wanted a list of usable bounds)
```py
displays = pygame.display.get_displays()
desktop_usable_bounds = [d.usable_bounds for d in displays]
```

`.usable_bounds` in this case would simply be a property getter for the `Display` class


What all of this means for this PR:
- [ ] Create a `Display` type (full qualified name would likely be `pygame.display.Display`)
- [ ] Add `.id` `property` to said type (for SDL2 purposes it could be something like 0 (which in SDL 3 is considered an invalid ID btw, but anyway, maybe keep a counter and increment it to ensure uniqueness or something)
- [ ] Add `.usable_bounds` `property` to said type that will basically query the SDL APIs you are already querying in this PR right now (to get those bounds)
- [ ] Add `pygame.display.get_displays()` with return type `list[pygame.Display]` (that also actually returns a list of `pygame.Display`s)

These are **required changes** by the way :grin:

damusss:
@Matiiss so you think the backbone of the API should be implemented in this specific pull request?
(has starbuck ever talked about this?)

Matiiss:
> @Matiiss so you think the backbone of the API should be implemented in this specific pull request? (has starbuck ever talked about this?)

Yeah, I think it might as well be the starting point, just rename the PR, rename the branch, I mean, there's a lot of code here that I think is still quite relevant. But you need to include all of the `Display` type code as well... Or maybe a new PR for a plain `Display` object would be better and this just adds the `.usable_bounds` property. :thinking:

damusss:
@Matiiss yeah I also think a new PR for the start of the API would be better. I'll mark this as draft, and re open it when the direction for display is enstablished.

damusss:
https://github.com/pygame-community/pygame-ce/pull/3720 implements this out of the box, and is the only way this feature should be added. I'm therefore closing this pull request.

PR #3536, opened 2025-07-15
Fix FRect contains error message

### LINKED ISSUE #3533: Incorrect contains error message in FRect
I was looking at the relevant code for @Hato1's proposal in #3531 and I noticed that FRects and Rects have the same error message.

```py
>>> import pygame
pygame-ce 2.5.5 (SDL 2.32.6, Python 3.12.3)
>>> r = pygame.Rect(10,20,30,40)
>>> fr = pygame.FRect(10,20,30,40)
>>>
>>> 10 in r
True
>>> 10.0 in r
TypeError: 'in <pygame.rect.Rect>' requires rect style object or int as left operand
>>>
>>> 10 in fr
TypeError: 'in <pygame.rect.FRect>' requires rect style object or int as left operand
>>> 10.0 in fr
True
```

FRect requires a float as the left operand but the error message says it requires an int.

```
TypeError: 'in <pygame.rect.FRect>' requires rect style object or int as left operand
```

Hato1:
FRect requires a float as the left operand but the error message says it requires an int.

Before this PR:
```
>>> fr = pygame.FRect(10,20,30,40)
>>> 10 in fr
TypeError: 'in <pygame.rect.FRect>' requires rect style object or int as left operand
```

After:
```
>>> fr = pygame.FRect(10,20,30,40)
>>> 10 in fr
TypeError: 'in <pygame.rect.FRect>' requires rect style object or float as left operand
```

Resolves #3533

### DIFF
--- a/src_c/rect_impl.h
+++ b/src_c/rect_impl.h
@@ -1999,6 +1999,16 @@ RectExport_containsSeq(RectObject *self, PyObject *arg)
         PyErr_SetString(PyExc_TypeError, "'in <" ObjectName
                                          ">' requires rect style object"
                                          " or int as left operand");
+        if (strcmp(ObjectName, "pygame.rect.FRect") == 0) {
+            PyErr_SetString(PyExc_TypeError, "'in <" ObjectName
+                                             ">' requires rect style object"
+                                             " or float as left operand");
+        }
+        else {
+            PyErr_SetString(PyExc_TypeError, "'in <" ObjectName
+                                             ">' requires rect style object"
+                                             " or int as left operand");
+        }
     }
     return ret;
 }
--- a/src_c/rect_impl.h
+++ b/src_c/rect_impl.h
@@ -1994,21 +1994,16 @@ RectExport_containsSeq(RectObject *self, PyObject *arg)
         return coord == self->r.x || coord == self->r.y ||
                coord == self->r.w || coord == self->r.h;
     }
+
     int ret = RectExport_contains_internal(self, (PyObject *const *)&arg, 1);
     if (ret < 0) {
-        PyErr_SetString(PyExc_TypeError, "'in <" ObjectName
-                                         ">' requires rect style object"
-                                         " or int as left operand");
-        if (strcmp(ObjectName, "pygame.rect.FRect") == 0) {
-            PyErr_SetString(PyExc_TypeError, "'in <" ObjectName
-                                             ">' requires rect style object"
-                                             " or float as left operand");
-        }
-        else {
-            PyErr_SetString(PyExc_TypeError, "'in <" ObjectName
-                                             ">' requires rect style object"
-                                             " or int as left operand");
-        }
+        const char *operand_type =
+            strcmp(ObjectName, "pygame.rect.FRect") == 0 ? "float" : "int";
+
+        PyErr_Format(
+            PyExc_TypeError,
+            "'in %s' requires rect style object or %s as left operand",
+            ObjectName, operand_type);
     }
     return ret;
 }


src_c/rect_impl.h
```diff
@@ -1994,11 +1994,16 @@ RectExport_containsSeq(RectObject *self, PyObject *arg)
...
                coord == self->r.w || coord == self->r.h;
     }
+
     int ret = RectExport_contains_internal(self, (PyObject *const *)&arg, 1);
     if (ret < 0) {
-        PyErr_SetString(PyExc_TypeError, "'in <" ObjectName
-                                         ">' requires rect style object"
-                                         " or int as left operand");
+        const char *operand_type =
+            strcmp(ObjectName, "pygame.rect.FRect") == 0 ? "float" : "int";
```
JoKing-1999: Hey, congrats on your first PR 🎉 

While this indeed works, can you change it to
```c
        const char *operand_type =
            pgFRect_Check((PyObject *)self) ? "float" : "int";
```
It gives a bit better performance (although if you are raising the error then you don't really care about the performance since the program is supposed to crash). But most importantly, it follows our conventions for checking the type of the object in C code.

Thanks again

Hato1: Applied.

Starbuck5:
The rect_impl template is compiled separately for Rect and FRect, it shouldn't need a runtime check inside itself to determine which type is which. I was thinking a new define could be added to rect.c to communicate the type name as a string to the implementation. Or perhaps even there's a way to treat the C type that's already communicated as a string.

JoKing-1999:
Starbuck what do you think about something like this
```diff
diff --git a/src_c/rect.c b/src_c/rect.c
index 155f97faf..c9097096a 100644
--- a/src_c/rect.c
+++ b/src_c/rect.c
@@ -49,6 +49,7 @@ static int
 four_floats_from_obj(PyObject *obj, float *val1, float *val2, float *val3,
                      float *val4);

+#define IS_RECT
 #define RectExport_init pg_rect_init
 #define RectExport_subtypeNew4 _pg_rect_subtype_new4
 #define RectExport_new pg_rect_new
@@ -165,7 +166,9 @@ four_floats_from_obj(PyObject *obj, float *val1, float *val2, float *val3,
 #define RectOptional_Freelist_Num pg_rect_freelist_num
 #endif /* PYPY_VERSION */
 #include "rect_impl.h"
+#undef IS_RECT

+#define IS_FRECT
 #define RectExport_init pg_frect_init
 #define RectExport_subtypeNew4 _pg_frect_subtype_new4
 #define RectExport_new pg_frect_new
@@ -282,6 +285,7 @@ four_floats_from_obj(PyObject *obj, float *val1, float *val2, float *val3,
 #define RectOptional_Freelist_Num pg_frect_freelist_num
 #endif /* PYPY_VERSION */
 #include "rect_impl.h"
+#undef IS_FRECT

diff --git a/src_c/rect_impl.h b/src_c/rect_impl.h
index 9018f9522..51b474637 100644
--- a/src_c/rect_impl.h
+++ b/src_c/rect_impl.h
@@ -1997,13 +1997,16 @@ RectExport_containsSeq(RectObject *self, PyObject *arg)

     int ret = RectExport_contains_internal(self, (PyObject *const *)&arg, 1);
     if (ret < 0) {
-        const char *operand_type =
-            pgFRect_Check((PyObject *)self) ? "float" : "int";
-
+#if defined(IS_RECT)
+#define OperandType "int"
+#elif defined(IS_FRECT)
+#define OperandType "float"
+#endif
         PyErr_Format(
             PyExc_TypeError,
             "'in %s' requires rect style object or %s as left operand",
-            ObjectName, operand_type);
+            ObjectName, OperandType);
+#undef OperandType
     }
     return ret;
 }
```

Starbuck5:
I think it would be simpler if you defined operandtype directly on the outer layer without the ISRECT/ISFRECT indirection.

Also I'd think should be able to be compiled into the string that goes into PyErr_Format.

Starbuck5:
Since this PR has been stalled for months, I handled it myself in https://github.com/pygame-community/pygame-ce/pull/3632.

PR #3542, opened 2025-07-27
2.5.5.post1

Starbuck5:
A version of 2.5.5 with Python 3.14 and Pypy 3.11 wheels.


Starbuck5:
I think this is going to fail because it's doing dep rebuilds and it needs to be on a repo branch rather than a fork branch to do that properly.

Issue #3546, opened 2025-08-02
Implement `Color.__bytes__`


Starbuck5:
The below is written by @Matiiss, @Starbuck5 helped with opening the ticket

```py
import pygame

c = pygame.Color(1, 2, 3)
print(bytes(c))
```
Does not exactly produce the output one might expect, Python treats the `Color` as an integer object (it is sort of) and creates a `bytes` object of that length filled with 0s, more explanation here: https://docs.python.org/3/library/stdtypes.html#bytes
<img width="642" height="82" alt="Image" src="https://github.com/user-attachments/assets/7584d177-69ec-49cf-b54b-3153b99982de" />

---

Full conversation on discord: https://discord.com/channels/772505616680878080/772940667231928360/1401287406635843784
Copied here for persistence

#### Big Whoop — Yesterday at 22:35
hey whats up with this: 
```
from pygame import Color
c = Color(1, 2, 3)
b = bytes(c)
b
```
 ?
#### Matiiss — Yesterday at 23:05
curiously the length of those bytes translates to the color values
```py
from itertools import batched

from pygame import Color

c = Color(5, 2, 3, 7)
b = bytes(c)

length = len(b)

# 0b...
length_in_binary = bin(length)
# 0b... -> [2:] -> ...
length_in_binary_no_prefix = length_in_binary[2:]
length_in_binary_no_prefix_padded = length_in_binary_no_prefix.zfill(32)

print(c)
print(length)
print(length_in_binary)
print(length_in_binary_no_prefix_padded)
print(["".join(batch) for batch in batched(length_in_binary_no_prefix_padded, 8)])
print([int("".join(batch), base=2) for batch in batched(length_in_binary_no_prefix_padded, 8)])
```
```
pygame-ce 2.5.3 (SDL 2.30.12, Python 3.12.8)
Color(5, 2, 3, 7)
84017927
0b101000000100000001100000111
00000101000000100000001100000111
['00000101', '00000010', '00000011', '00000111']
[5, 2, 3, 7]
```
#### NotMEE12 — Yesterday at 23:15
_length of the bytes has the data you need..._
_Okay but why not let the data be in bytes?_
#### Starbuck — Yesterday at 23:19
Holy shit
#### Big Whoop — Yesterday at 23:24
yeah fun times
#### Andrew — Yesterday at 23:58
Output? I’m on mobile right now lol
#### Starbuck — Yesterday at 23:59
16909311 bytes of 0
[...]

PR #3547, opened 2025-08-03
Add `Color.__bytes__`

Matiiss:
Fixes https://github.com/pygame-community/pygame-ce/issues/3546


oddbookworm (changes requested):
Where docs?

Starbuck5:
I was thinking docs aren't necessary for this change, since it makes color behave like a normal sequence would.

However I do think it would be a good idea to put a comment in the code explaining why it's necessary to have the method explicitly written out.

oddbookworm:
> I was thinking docs aren't necessary for this change, since it makes color behave like a normal sequence would.

I disagree. A `versionadded` note at least, that way if someone does use this functionality, they at least have some way to reference that it's broken pre-2.5.whatever

test/color_test.py
```diff
@@ -773,6 +773,17 @@ def test_long(self):
...
+    def test_bytes(self):
+        c = pygame.Color(0x00012345)
+        self.assertEqual(c.r, 0x00)
+        self.assertEqual(c.g, 0x01)
+        self.assertEqual(c.b, 0x23)
+        self.assertEqual(c.a, 0x45)
+
+        as_bytes = bytes(c)
+        self.assertEqual(as_bytes, bytes([0x00, 0x01, 0x23, 0x45]))
+        self.assertEqual(len(as_bytes), 4)
```
oddbookworm, line 785: Does endianness not affect this?

oddbookworm:
> > I was thinking docs aren't necessary for this change, since it makes color behave like a normal sequence would.
> 
> I disagree. A `versionadded` note at least, that way if someone does use this functionality, they at least have some way to reference that it's broken pre-2.5.whatever

A user-side implementation could look like:
```py
if pygame.version.vernum < (2, 5, 6):
    color_bytes = bytes([color.r], [color.g], [color.b], [color.a])
else:
    color_bytes = bytes(color)
```

test/color_test.py
oddbookworm: Nvm, I think it'll be consistent between the two

aatle:
> > > I was thinking docs aren't necessary for this change, since it makes color behave like a normal sequence would.
> > 
> > 
> > I disagree. A `versionadded` note at least, that way if someone does use this functionality, they at least have some way to reference that it's broken pre-2.5.whatever
> 
> A user-side implementation could look like:
> 
> ```python
> if pygame.version.vernum < (2, 5, 6):
>     color_bytes = bytes([color.r], [color.g], [color.b], [color.a])
> else:
>     color_bytes = bytes(color)
> ```

Yeah, a `versionadded` at least.

Also that snippet doesn't work, here are two ways to convert colors currently:
```py
bytes(list(color))
int(color).to_bytes(4)  # over 2x faster
```

src_c/color.c
```diff
@@ -1753,6 +1758,16 @@ _color_float(pgColorObject *color)
...
 }
 
+/**
+ * bytes(color)
+ */
+static PyObject *
+_color_bytes(pgColorObject *color)
+{
+    return PyBytes_FromFormat("%c%c%c%c", color->data[0], color->data[1],
+                              color->data[2], color->data[3]);
```
ankith26:
```suggestion
    return PyBytes_FromStringAndSize((char *)color->data, 4);
```

I wonder if something like this can work. Would result in cleaner and a little bit faster code too I would guess.

ankith26:
Also I'm +1 for andrews suggestion for adding this in a `versionadded` note. People need to know to not use this if they are targetting older pygame-ce versions

src_c/color.c
Matiiss: Cleaner? Don't know about that, it does look slicker though and I like it (and it seems to work)
A little bit faster? Maybe, though then again this is C... so does it matter?

src_c/color.c
```diff
@@ -256,6 +259,17 @@ static PyMethodDef _color_methods[] = {
...
      DOC_COLOR_PREMULALPHA},
     {"update", (PyCFunction)_color_update, METH_FASTCALL, DOC_COLOR_UPDATE},
+
+    /**
+     * While object.__bytes__(self) is listed in the Data Model reference (see:
+     * https://docs.python.org/3/reference/datamodel.html#object.__bytes__) it
+     * does not appear to have a PyTypeObject struct analog (see:
+     * https://docs.python.org/3/c-api/typeobj.html), so we declare it for the
+     * type as any other custom method.
+     */
```
oddbookworm, line 269: I'm not sure this giant comment is strictly necessary in the middle of this large array definition

docs/reST/ref/color.rst
```diff
@@ -92,6 +92,11 @@
    :returns: a newly created :class:`Color` object
    :rtype: Color
 
+   .. versionchanged:: 2.5.6
```
oddbookworm, line 95: I still think `versionadded` is more appropriate because `versionchanged` indicates that we changed behavior on our end, instead of overriding default python behavior by adding a new method.

Matiiss: I'm not so sure, this is essentially a bugfix, there was unintended behaviour (a bug) and we changed the behaviour to fix it.

src_c/color.c
Matiiss: Not to be overly pedantic, but it's basically at the very end of that array. Can you suggest a better place for it? I think this placement is fine because it's the closest it can be to the item it refers to.
I suppose it could be moved to right before the implementation (the function implementing it).

Starbuck5 (approved):
The code works, has doc and tests. I can hit green button.

---
I think the note in the docs about assuming that `bytes` means the built in bytes class is ridiculously pedantic, and I sorta think the comment in color.c is unnecessary. But these are not merge blocking things. And if you agree you can change these things after my approval.

docs/reST/ref/color.rst
ankith26: I agree, I think `versionchanged` is fine for now.

src_c/color.c
oddbookworm: IMO this comment isn't really even necessary

oddbookworm: But I'm not going to hold up the merge for it

oddbookworm (approved):

PR #3548, opened 2025-08-03
Update python versions in `run-ubuntu-checks` workflow

oddbookworm:


src_c/rect_impl.h
```diff
@@ -1929,12 +1929,9 @@ RectExport_clipline(RectObject *self, PyObject *const *args, Py_ssize_t nargs)
     }
 
     if (!RectImport_IntersectRectAndLine(rect, &x1, &y1, &x2, &y2)) {
-        Py_XDECREF(rect_copy);
         return PyTuple_New(0);
     }
 
-    Py_XDECREF(rect_copy);
```
ankith26, line 1936: This fix still doesn't solve the memory leak that the code was intending to avoid. The result of the `RectExport_RectNew` call is still being leaked.

The actual issue here seems to be that the code is trying to decref the SDL_Rect struct instead of its parent python rect object.

oddbookworm: `Rect_copy` isn't even a `PyObject`! It's a pointer to a raw SDL struct. This was segfaulting locally for me, so I fixed it.

ankith26: Yes I pointed out the same thing. `rect_copy` is a part of a pyobject and that pyobject has to be freed.

oddbookworm: Maybe now?

ankith26: Yeah the fix looks good now

ankith26 (approved):

Matiiss (approved):
LGTM

PR #3544, opened 2025-07-30
Fix some Surface and Window tests on SDL3

Starbuck5:
Tested in combination with #3428 and #3435.

Before
Surface: FAILED (failures=6, errors=98, skipped=2)
Window: FAILED (errors=7)

After
Surface: FAILED (failures=10, errors=19, skipped=2)
Window: FAILED (errors=5)

PR #3549, opened 2025-08-04
Fix font test segfault on SDL3

Starbuck5:
Tested in combination with https://github.com/pygame-community/pygame-ce/pull/3428 and https://github.com/pygame-community/pygame-ce/pull/3544

EDIT: removed extremely misleading AI summary

### DIFF
--- a/src_c/font.c
+++ b/src_c/font.c
@@ -1210,6 +1210,17 @@ font_dealloc(PyFontObject *self)
 {
     TTF_Font *font = PyFont_AsFont(self);
     if (font && font_initialized) {
+        // In SDL3_ttf, it seems that closing a font after its library was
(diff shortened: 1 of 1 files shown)


ankith26:
@coderabbitai full review

Starbuck5:
For reviewers, this PR is still ready for review. I have looked over the coderabbit comments and none of them are valid or need to be done.

ankith26 (approved):
For now I think this is good enough, fixing a segfault should be a priority and this PR does that. Once our SDL3 support is more mature we could investigate into the edge case (quit not in the same general) and see if there's a memory leak and then plan a fix (probably keeping track of all open fonts on our side and closing stuff manually on quit might be needed?)

PR #3550, opened 2025-08-04
Fix rwops and some image stuff on SDL3

ankith26:
- Fixes image saving return code handling.
- `SDL_INIT_INTERFACE` needs to be used on a `SDL_IOStreamInterface` interface.
- As a result of the above issue we were going into the error handling path where `_pg_rw_close` already does `PyMem_Free(helper);`, resulting in a double free.

With these changes all `imageext` tests pass and some failing `image` tests are also passing now.


JoKing-1999:
Can you do it like this https://github.com/pygame-community/pygame-ce/pull/3515 (without adding new #if SDL_VERSION)

ankith26:
I can do that but I don't see the need for doing it for functions that are used only once.

PR #3556, opened 2025-08-10
Add Sound.copy and Sound.__copy__

oddbookworm:
I added https://github.com/Borishkof as a co-author because I took the test and tweaked the docs from their old pull request


oddbookworm:
image

ankith26:
@coderabbitai full review

src_c/mixer.c
```diff
@@ -801,6 +803,38 @@ snd_get_raw(PyObject *self, PyObject *_null)
...
+    if (sound_init((PyObject *)newSound, NULL, kwargs) != 0) {
+        Py_DECREF(kwargs);
+        Py_DECREF(newSound);
+        return RAISE(pgExc_SDLError, "Failed to initialize copied sound");
+    }
+
+    Py_DECREF(kwargs);
+    Py_INCREF(newSound);
+    return (PyObject *)newSound;
+}
```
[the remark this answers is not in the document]
oddbookworm: The extra `Py_INCREF` is needed, otherwise we end up undercounting the copy compared to the original (thanks Python 3.14.0rc1 with debug symbols)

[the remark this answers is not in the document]
ankith26: I'm with the clanker on this one. Conceptually it is right, `tp_new` should be returning a new reference anyways so I don't get why that extra incref is needed. I also think the patch it suggested has decent changes.

oddbookworm:
@coderabbitai please review the changes now

src_c/mixer.c
```diff
@@ -801,6 +809,77 @@ snd_get_raw(PyObject *self, PyObject *_null)
...
+        return NULL;
+    }
+
+    PyObject *dict = PyDict_New();
+    if (!dict) {
+        if (!PyErr_Occurred()) {
+            PyErr_SetString(PyExc_MemoryError,
+                            "Failed to create internal dictionary to create "
+                            "copy of Sound");
+        }
```
ankith26: We never do this in any of our other `PyDict_New` calls. As per python API the error must be already set if `dict == NULL`. The same goes for above `tp_new` call.

src_c/mixer.c
```diff
@@ -801,6 +809,77 @@ snd_get_raw(PyObject *self, PyObject *_null)
...
+        return NULL;
+    }
+
+    if (PyDict_SetItemString(dict, "buffer", bytes) < 0) {
+        // exception set already
+        PG_SAVE_EXCEPTION
+        Py_DECREF(bytes);
+        Py_DECREF(dict);
+        Py_DECREF(newSound);
+        PG_UNSAVE_EXCEPTION
```
ankith26: I think this save and unsave exception thing is redundant, and we never do this at any other place either like this

oddbookworm: But according to the CPython devs, you're not supposed to call ***any*** C-API with an exception already set. And since they have a habit of throwing those gotchas in the form of asserts, I figured I would be safe rather than sorry

src_c/mixer.c
oddbookworm: The docs say nothing about there being an exception already set. Just that it returns NULL

src_c/mixer.c
ankith26: A lot of stuff would break if incref/decref weren't allowed when exceptions are set. Not just in our code but in CPython itself. Yes, it is generally true that API must not be called with exceptions already set, but I'm sure incref/decref is a pretty intentional exception to this.

I don't want this PR adding these extra checks and thus make it a standard for the rest of our code, even if we want to go that way it needs to be done in a separate PR with its own discussion

src_c/mixer.c
ankith26: Unless explicitly documented otherwise, every function always sets error and returns NULL, atleast thats what [this block of the documentation implies](https://docs.python.org/3/c-api/intro.html#exceptions)

I also went through the `PyDict_New` CPython code and have verified that its safe to assume this for now. If this changes in the future we can handle it then, for now I don't want our code to be overly/inconsistently defensive.

oddbookworm:
@ankith26 comments addressed

src_c/mixer.c
```diff
@@ -517,8 +519,14 @@ mixer_quit(PyObject *self, PyObject *_null)
...
-                Py_XDECREF(channeldata[i].sound);
-                Py_XDECREF(channeldata[i].queue);
+                if (channeldata[i].sound) {
+                    Mix_HaltGroup(
+                        (int)(intptr_t)pgSound_AsChunk(channeldata[i].sound));
+                }
+                if (channeldata[i].queue) {
+                    Mix_HaltGroup(
+                        (int)(intptr_t)pgSound_AsChunk(channeldata[i].queue));
+                }
```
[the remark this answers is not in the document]
ankith26: How about not changing this block of code, but doing `Mix_ChannelFinished(NULL)` (disable the callback and let this function do the cleaning as intended) just when mixer quit is called, does that solve this issue? I believe that would be the cleanest fix to this problem.

src_c/mixer.c
```diff
@@ -517,8 +519,14 @@ mixer_quit(PyObject *self, PyObject *_null)
```
MyreMylar, line 520:
```suggestion
Mix_ChannelFinished(NULL);

```

From @ankith26 & coderabbit suggestion

src_c/mixer.c
```diff
@@ -514,11 +516,18 @@ mixer_quit(PyObject *self, PyObject *_null)
         Py_BEGIN_ALLOW_THREADS;
         Mix_HaltMusic();
         Py_END_ALLOW_THREADS;
+Mix_ChannelFinished(NULL);
```
MyreMylar:
```suggestion
        Mix_ChannelFinished(NULL);
```

MyreMylar (approved):
OK LGTM 👍

src_c/mixer.c
```diff
@@ -514,11 +516,18 @@ mixer_quit(PyObject *self, PyObject *_null)
...
-                Py_XDECREF(channeldata[i].sound);
-                Py_XDECREF(channeldata[i].queue);
+                if (channeldata[i].sound) {
+                    Mix_HaltGroup(
+                        (int)(intptr_t)pgSound_AsChunk(channeldata[i].sound));
+                }
+                if (channeldata[i].queue) {
+                    Mix_HaltGroup(
+                        (int)(intptr_t)pgSound_AsChunk(channeldata[i].queue));
+                }
```
ankith26: Now these changes would need to be reverted too I think.

PR #3570, opened 2025-08-28
Fix failing tests on non-libpng backends

ankith26:
These changes are relevant for when SDL_image isn't using the libpng backend (so not really relevant for our release builds, but useful for when we compiled against system SDL_image on some linux distro like fedora)

precursor to #3562 and SDL3 runtime support work considering we don't use libpng backend on those builds as of now.


Starbuck5 (approved):
I haven't tested or anything, but the code makes sense.

PR #3572, opened 2025-08-30
Remove `None` initialization of `Sprite.rect` and `Sprite.image`, and remove property accessors 

aatle:
Revert the changes of #1942 (initialize `rect` and `image` to `None`, and have those attributes be properties).

Reasons:
The `None` values aren't actually compatible with the sprite functions and methods; they will error for a `None` sprite `rect` or `image`. It isn't really useful to initialize them to `None`.
By having them uninitialized instead, the remaining sprite stub issues involving optionals can be solved. -> makes sprite module comfortably usable and safe with typing.
The property accessors for these attributes seem to be redundant. Also, they are slightly slower than normal attributes.
The revert will also make sprite behavior in line with upstream `pygame`, as #1942 didn't apply to it.

Additionally, this PR fixes typing for `DirtySprite.source_rect`.


yoonthegoon:
@MyreMylar thoughts?

I've just come across this issue now and having the `image` property of a `Sprite` always be annotated a `Surface | None` is really troublesome with type checkers. I would hate to litter my code with asserts or ignore comments.

If you truly wish to annotate that `image` and `rect` on `Sprite` can be `None`, the getters and setters are unnecessary and the following can be done. 

```diff
-    def __init__(self, *groups):
+    def __init__(self, *groups: "Group") -> None:
         self.__g = {}  # The groups the sprite is in
-        self.__image: Optional[pygame.surface.Surface] = None
-        self.__rect: Optional[pygame.rect.Rect] = None
+        self.image: Surface | None = None
+        self.rect: FRect | Rect | None = None
         if groups:
             self.add(*groups)
```

This way, when subclassing, one can simply do this:

```python
class Character(Sprite):
    def __init__(self, group: Group, pos: Vector2 = Vector2(0, 0)) -> None:
        self.image: Surface = PRELOADED_IMAGE
        self.rect: Rect = self.image.rect(topleft=pos)
```

Though I prefer at least `image` to be a field like done in this PR because you'll only need to load the image once (if the user defines the field not in `__init__`).

yoonthegoon:
Also I'm just noticing https://github.com/pygame-community/pygame-ce/blob/f80f75f9b10faf3859862bd014b2f6388718e62a/src_py/sprite.py#L150 is an infinitely recursive loop if `type(group) is not Group`

PR #3574, opened 2025-08-31
Port freetype to SDL3

Starbuck5:
We *may* end up removing freetype, but this wasn't too difficult to port, and it strengthens our SDL3 foundations.


ankith26 (approved):

PR #3576, opened 2025-09-01
[Part 1] SDL3: event: runtime issue fixes

ankith26:
- Vendors in `SDL_gesture.h`
- windowevent fixes


src_c/event.c
```diff
@@ -25,6 +25,7 @@
  */
 #define PYGAMEAPI_EVENT_INTERNAL
 
+#define SDL_GESTURE_IMPLEMENTATION 1
```
Starbuck5, line 32: This could use a comment, I wasn't sure what I was looking at while reviewing the PR before doing a code search for this.

src_c/event.c
```diff
@@ -732,14 +738,15 @@ pg_event_filter(void *_, SDL_Event *event)
...
     }
-    /* TODO:
+    /*
      * Any event that gets blocked here will not be visible to the event
      * watchers. So things like WINDOWEVENT should never be blocked here.
-     * This is taken care of in SDL2 codepaths already but needs to also
-     * be verified in SDL3 porting.
      * If the user requests a block on WINDOWEVENTs we are going to handle
      * it specially and call it a "pseudo-block", where the filtering will
      * happen in a _pg_filter_blocked_events call. */
```
Starbuck5, line 750: In the future would it make sense to implement all event blocking the same way as late in the process as possible? That way we don't need to have 2 types of events that get blocked differently.

ankith26: Good point. I don't want to take that on for this PR but will keep in mind for future PRs.

src_c/event.c
ankith26: Sure, I added a comment.

ankith26:
I rebased this PR on main, and also updated the SDL_gesture header.

PR #3577, opened 2025-09-01
[Part 2] SDL3: display+window: runtime fixes

ankith26:
- display index compat code
- fullscreen compat code unified
- syncwindow calls so that changes take immediate affect like sdl2

### DIFF
--- a/src_c/display.c
+++ b/src_c/display.c
@@ -99,6 +99,19 @@ _display_state_cleanup(_DisplayState *state)
 #endif
 }
 
+/* This macro defines the pixel format we use for pg_renderer.
+ * On our SDL2 code we used to use SDL_PIXELFORMAT_ARGB8888 but on SDL3 it
+ * gives incorrect results because it includes alpha when we don't need it, so
+ * we use the XRGB variant.
+ * TODO: can consider using XRGB unconditionally on both SDL2 and SDL3
+ * codepaths
+ */
+#if SDL_VERSION_ATLEAST(3, 0, 0)
+#define PG_TEXTURE_FMT SDL_PIXELFORMAT_XRGB8888
+#else
+#define PG_TEXTURE_FMT SDL_PIXELFORMAT_ARGB8888
+#endif
+
 // prevent this code block from being linked twice
 // (this code block is copied by window.c)
 #ifndef BUILD_STATIC
@@ -464,7 +477,12 @@ pg_GetVideoInfo(pg_VideoInfo *info)
     }
     else {
 #if SDL_VERSION_ATLEAST(3, 0, 0)
-        if ((mode_ptr = SDL_GetCurrentDisplayMode(0))) {
+        SDL_DisplayID primary_display = SDL_GetPrimaryDisplay();
+        if (primary_display == 0) {
+            PyErr_SetString(pgExc_SDLError, SDL_GetError());
+            return (pg_VideoInfo *)NULL;
+        }
+        if ((mode_ptr = SDL_GetCurrentDisplayMode(primary_display))) {
             info->current_w = mode_ptr->w;
             info->current_h = mode_ptr->h;
             formatenum = mode_ptr->format;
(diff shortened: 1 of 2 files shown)


Starbuck5:
> syncwindow calls so that changes take immediate affect like sdl2

I don't feel any particular need to add this, does this impact any of our unit tests? I don't think people are relying on the synchronicity. Plus when we're only on SDL3 I feel like we shouldn't have this.

src_c/display.c
```diff
@@ -1152,11 +1171,24 @@ PG_SetWindowFullscreen(SDL_Window *window, bool fullscreen,
...
+static bool
+PG_SetWindowFullscreen(SDL_Window *window, bool fullscreen,
+                       bool non_desktop_fullscreen)
+{
+    int flags = 0;
+    if (fullscreen) {
+        flags = non_desktop_fullscreen ? SDL_WINDOW_FULLSCREEN
+                                       : SDL_WINDOW_FULLSCREEN_DESKTOP;
+    }
+    return (SDL_SetWindowFullscreen(window, flags) == 0) ? true : false;
```
Starbuck5: The ternary here is unnecessary, you already have the boolean after the `== 0`

src_c/display.c
```diff
@@ -3447,7 +3507,7 @@ pg_toggle_fullscreen(PyObject *self, PyObject *_null)
                 if (win == NULL) {
                     return RAISE(pgExc_SDLError, SDL_GetError());
                 }
-                if (0 != SDL_SetWindowFullscreen(win, SDL_WINDOW_FULLSCREEN)) {
+                if (!PG_SetWindowFullscreen(win, 1, 1)) {
```
Starbuck5: Extremely minor nit: wouldn't this read nicer with trues and falses instead of 1s and 0s?

src_c/display.c
```diff
@@ -1530,9 +1577,9 @@ pg_set_mode(PyObject *self, PyObject *arg, PyObject *kwds)
                  * changes if the window is fullscreen
                  * See https://github.com/pygame/pygame/issues/2711 */
 #if SDL_VERSION_ATLEAST(3, 0, 0)
-                if (0 != PG_SetWindowFullscreen(
-                             win, sdl_flags & SDL_WINDOW_FULLSCREEN,
-                             non_desktop_fullscreen)) {
+                if (!PG_SetWindowFullscreen(win,
+                                            sdl_flags & SDL_WINDOW_FULLSCREEN,
+                                            non_desktop_fullscreen)) {
```
Starbuck5: This block and the SDL2 block below could be consolidated together with PG_SetWindowFullscreen.

ankith26:
> I don't feel any particular need to add this, does this impact any of our unit tests? I don't think people are relying on the synchronicity. Plus when we're only on SDL3 I feel like we shouldn't have this.

Without doing this a lot of our tests fail. I feel like people may rely on synchronicity, especially given that without this call, SDL makes no guarantee how early or late the sync may actually happen (and it may also be a very platform dependent thing). I believe calling this explicitly keeps stuff more inter-platform consistent and is a good thing to do.

src_c/display.c
ankith26: I think the current code is cleaner. Consolidating this is hard because `non_desktop_fullscreen` variable is only defined in the SDL3 path - and not on SDL2, whereas `SDL_WINDOW_FULLSCREEN_DESKTOP` is only defined on SDL2 and not SDL3. The consolidation would require refactoring to the SDL2 codepath, which I have tried to avoid to the best extent in my porting PRs.

src_c/display.c
ankith26: I don't really have a preference for either way. But most of our C code uses 0-1 bool so I think this shouldn't be very out of place. Using `true`/`false` requires including `<stdbool.h>` (which SDL3 does implicitly, but SDL2 doesn't)

EDIT: I quickly tested and somehow `true`/`false` is still available on the SDL2 codepaths, so I decided to make the change.

src_c/display.c
ankith26: fair point

src_c/display.c
```diff
@@ -99,6 +99,14 @@ _display_state_cleanup(_DisplayState *state)
 #endif
 }
 
+/* TODO: Check if doing SDL_PIXELFORMAT_XRGB8888 in SDL2 codepath is fine.
+ * For now do it only on SDL3 codepath to fix SCALED issues */
```
Starbuck5: Can you elaborate what this means?

ankith26: I elaborated on the comment and force pushed.

PR #3578, opened 2025-09-01
[Part 3] SDL3: surface+pixelarray+font+draw: runtime fixes

ankith26:
- A few abstractions were improved. `PG_ConvertSurface` now takes a surface as its second argument so it can copy the palette from it, to compat for the SDL2 path already doing it because there palette was part of the format. `PG_GetSurfacePalette` now creates a palette when the surface is indexed, SDL2 did it implicitly.
-  Added compat code for `PG_BlitSurface` because SDL3 doesn't modify dstrect anymore.

### DIFF
--- a/src_c/_pygame.h
+++ b/src_c/_pygame.h
@@ -77,8 +77,16 @@
 
 #define PG_CreateSurface SDL_CreateSurface
 #define PG_CreateSurfaceFrom SDL_CreateSurfaceFrom
-#define PG_ConvertSurface SDL_ConvertSurface
 #define PG_ConvertSurfaceFormat SDL_ConvertSurface
+
+/* Convert surface using palette and format of dst */
+static inline SDL_Surface *
+PG_ConvertSurface(SDL_Surface *surface, SDL_Surface *dst)
+{
+    return SDL_ConvertSurfaceAndColorspace(surface, dst->format,
+                                           SDL_GetSurfacePalette(dst),
+                                           SDL_GetSurfaceColorspace(dst), 0);
+}
 
 #define PG_PixelFormatEnum SDL_PixelFormat
 
@@ -144,11 +152,23 @@ PG_SURF_BitsPerPixel(SDL_Surface *surf)
 
 #define PG_PixelFormat const SDL_PixelFormatDetails
 
+static inline SDL_Palette *
+PG_GetSurfacePalette(SDL_Surface *surf)
+{
+    SDL_Palette *ret = SDL_GetSurfacePalette(surf);
+    if (!ret && SDL_ISPIXELFORMAT_INDEXED(surf->format)) {
+        /* Palette doesn't exist but is expected, so create it.
+         * SDL will associate the newly created palette with the surface */
+        ret = SDL_CreateSurfacePalette(surf);
+    }
+    return ret;
+}
+
 static inline bool
 PG_GetSurfaceDetails(SDL_Surface *surf, PG_PixelFormat **format_p,
                      SDL_Palette **palette_p)
 {
-    *palette_p = SDL_GetSurfacePalette(surf);
+    *palette_p = PG_GetSurfacePalette(surf);
     *format_p = SDL_GetPixelFormatDetails(surf->format);
     return *format_p != NULL;
 }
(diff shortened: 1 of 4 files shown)


src_c/_pygame.h
```diff
@@ -143,11 +151,23 @@ PG_SURF_BitsPerPixel(SDL_Surface *surf)
...
 
+static inline SDL_Palette *
+PG_GetSurfacePalette(SDL_Surface *surf)
+{
+    SDL_Palette *ret = SDL_GetSurfacePalette(surf);
+    if (!ret && SDL_ISPIXELFORMAT_INDEXED(surf->format)) {
+        /* Palette doesn't exist but is expected, so create it.
+         * SDL will associate the newly created palette with the surface */
+        ret = SDL_CreateSurfacePalette(surf);
+    }
```
Starbuck5, line 163: I don't think this is how we should be doing this. If our surfaces are missing palettes we deal with that when the surface is not created, not randomly at any point as a side effect of a getter function. And/or make the consumers okay with index formats without palettes, since that's a thing in SDL3 and therefore we will want to support it anyway.

src_c/surface.c
```diff
@@ -1544,24 +1544,21 @@ surf_set_alpha(pgSurfaceObject *self, PyObject *args)
...
+        sdlrect.h = 1;
+        sdlrect.w = 1;
 
         SDL_Surface *surface =
             PG_CreateSurface(1, 1, PG_SURF_FORMATENUM(surf));
 
         SDL_LowerBlit(surf, &sdlrect, surface, &sdlrect);
         SDL_FreeSurface(surface);
     }
-#endif
```
Starbuck5: This is some extremely confusing historic code to be changing without a comment or a note in the description or anything about how this makes sense and doesn't change SDL2 behavior.

src_c/_pygame.h
ankith26: I think this being part of the getter makes sense considering that the getter is responsible for returning a palette, and we must ensure it doesn't return `NULL` when it should actually return a palette. But sure, I suppose it being part of `PG_CreateSurface` could also work (but then we'd also have to think about all the other ways a surface is created, which means more places to modify)

src_c/surface.c
ankith26: It doesn't change SDL2 behaviour at all because it is changing the `srcrect` argument of a blit whose destination is a temporary surface that is immediately destroyed after the said blit.

On SDL3 the same code was segfaulting because of the 0,0,0,0 rect, and hence I thought the fix is worth doing on both SDL2 and SDL3 codepaths because it is likely to segfault under SDL2-compat as well.

And yes, I have verified that this hack is still needed on SDL3, it was needed to get a unit test to pass.

src_c/_pygame.h
Starbuck5: In SDL3's readme migration, it says that "Indexed format surfaces no longer have a palette by default. Surfaces without a palette will copy the pixels untranslated between surfaces." We should aim to support that as well, which is why I don't want to get code in here that undermines that.

src_c/_pygame.h
```diff
@@ -77,9 +77,17 @@
 
 #define PG_CreateSurface SDL_CreateSurface
 #define PG_CreateSurfaceFrom SDL_CreateSurfaceFrom
-#define PG_ConvertSurface SDL_ConvertSurface
 #define PG_ConvertSurfaceFormat SDL_ConvertSurface
 
+/* Convert surface using palette and format of dst */
+static inline SDL_Surface *
+PG_ConvertSurface(SDL_Surface *surface, SDL_Surface *dst)
```
Starbuck5, line 84: In these headers I like to have the SDL3 version just be a rename, so that when we're SDL3 only we can just change PG_ to SDL_ in the call sites and it'll be off to the races. So maybe if this function needs to be doing different / more stuff than convert surface (copying palette ... copying other settings?), maybe it should be called something else. In the transform module there's a function `newsurf_fromsurf` that does a bunch of things. This could also be an addition to the Surface C API, pgSurface_NewFromSurf ?

Or maybe this is a new PG_ConvertSurfaceAndPalette method? That would be better in my opinion than having one of these PG_ compat functions that look like an SDL3 function but without presenting the same API.

PR #3579, opened 2025-09-01
[Part 4] SDL3: mouse+key+rect: runtime fixes

ankith26:
- Add deprecation warning for calling `set_relative_mode` without `set_mode` on SDL2, on SDL3 it errors.
- Use the `PG_IntersectFRectAndLine` compat code on SDL3 for now, for complete SDL2 `FRect` compat (this can be changed in the future)

### DIFF
--- a/src_c/key.c
+++ b/src_c/key.c
@@ -389,7 +389,11 @@ static const struct {
     {1073742054, "right alt"},       /* K_RALT */
     {1073742055, "right meta"},      /* K_RGUI, K_RMETA, K_RSUPER */
     {1073742081, "alt gr"},          /* K_MODE */
-    {1073742094, "AC Back"},         /* K_AC_BACK */
+#if SDL_VERSION_ATLEAST(3, 0, 0)
+    {1073742106, "AC Back"}, /* K_AC_BACK */
+#else
+    {1073742094, "AC Back"}, /* K_AC_BACK */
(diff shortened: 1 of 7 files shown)


ankith26:
@coderabbitai review

src_c/key.c
```diff
@@ -389,7 +389,11 @@ static const struct {
     {1073742054, "right alt"},       /* K_RALT */
     {1073742055, "right meta"},      /* K_RGUI, K_RMETA, K_RSUPER */
     {1073742081, "alt gr"},          /* K_MODE */
-    {1073742094, "AC Back"},         /* K_AC_BACK */
+#if SDL_VERSION_ATLEAST(3, 0, 0)
```
Starbuck5, line 392: Why not say `{SDLK_AC_BACK, "AC Back"}`, I believe that would work on both versions.

ankith26: Yes that would indeed work on both versions, but it would not be consistent with what is going on here. We hardcode integers so that when we need to support a newly added constant, we can just copy paste the integer value and still support older SDL versions.
Though I suspect that doing this is not very relevant anymore since we've raised our minimum supported SDL version significantly since this code was written probably.
Replacing all integers here with SDL constants would be a good cleanup PR for the future but I don't want to do that in this PR.

PR #3580, opened 2025-09-01
[Part 5] SDL3: image+scrap+transform+debug: runtime fixes

ankith26:
probably the easiest pr of the lot to review

### DIFF
--- a/src_c/image.c
+++ b/src_c/image.c
@@ -454,17 +454,10 @@ tobytes_surf_32bpp(SDL_Surface *surf, SDL_PixelFormat *format_details,
 {
     int w, h;
 
-#if SDL_VERSION_ATLEAST(3, 0, 0)
-    Uint32 Rloss = format_details->Rbits;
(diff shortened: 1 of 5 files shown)


ankith26:
@coderabbitai review

src_c/image.c
```diff
@@ -565,19 +558,15 @@ image_tobytes(PyObject *self, PyObject *arg, PyObject *kwarg)
     if (!format_details) {
```
Starbuck5, line 558: This whole SDL_VERSION_ATLEAST could be removed by using PG_GetSurfaceDetails, same with the block below (+PG_PixelFormatEnum) for that one, if you're reducing redundancy on these.

ankith26: Good point. But the main reason for me doing these loss changes is not reducing redundancy, it is fixing behavior because `bits != loss` and our compat macro handles the conversion.
I want to keep the diff minimal in these runtime fixes PRs, cleanups and stuff can always be done in future PRs.

PR #3588, opened 2025-09-14
Support emscripten (pygbag + pyodide) in the meson buildconfig

ankith26:
This is my take on @pmp-p 's PR at https://github.com/pmp-p/pygame-ce-wasm/pull/12

- pygbag path: `dev.py` updated to handle this. It uses `python-wasm-sdk` and makes a custom wheel with a single shared library build of `src_c` and the pure python contents of `src_py`.
- pyodide path: `pyodide build` now works out of box (provided dependencies are installed as done in `before-build` attribute in `[tool.cibuildwheel.pyodide]` section of `pyproject.toml`). Added CI to test building, however running unit tests on CI is still out of scope of this PR, I may look at it in a future PR.

@pmp-p I would appreciate your review on this PR + help with testing it. If this PR works out, pygbag would need to be updated accordingly to handle the wheel file this PR can build.

### DIFF
--- a/.github/workflows/build-emsdk.yml
+++ b/.github/workflows/build-emsdk.yml
@@ -32,39 +32,26 @@ concurrency:
   cancel-in-progress: true
 
 jobs:
-  build:
+  build-pygbag:
     runs-on: ubuntu-22.04
     env:
       # pin SDK version to the latest, update manually
-      SDK_VERSION: 3.1.32.0
-      SDK_ARCHIVE: python3.11-wasm-sdk-Ubuntu-22.04.tar.lz4
+      SDK_VERSION: 3.1.61.12bi
+      SDK_ARCHIVE: python3.13-wasm-sdk-Ubuntu-22.04.tar.lz4
       SDKROOT: /opt/python-wasm-sdk
+      PYBUILD: 3.13
 
     steps:
     - uses: actions/checkout@v5.0.0
 
-    - name: Regen with latest cython (using system python3)
-      run: |
-        pip3 install cython==3.0.10
-        python3 setup.py cython_only
-
     - name: Install python-wasm-sdk
       run: |
         sudo apt-get install lz4
-        echo https://github.com/pygame-web/python-wasm-sdk/releases/download/$SDK_VERSION/$SDK_ARCHIVE
         curl -sL --retry 5 https://github.com/pygame-web/python-wasm-sdk/releases/download/$SDK_VERSION/$SDK_ARCHIVE | tar xvP --use-compress-program=lz4
-        # do not let SDL1 interfere
-        rm -rf /opt/python-wasm-sdk/emsdk/upstream/emscripten/cache/sysroot/include/SDL
       working-directory: /opt
 
     - name: Build WASM with emsdk
-      run: |
-        ${SDKROOT}/python3-wasm setup.py build -j$(nproc)
-
-    - name: Generate libpygame.a static binaries archive
-      run: |
-        mkdir -p dist
-        SYS_PYTHON=python3 /opt/python-wasm-sdk/emsdk/upstream/emscripten/emar rcs dist/libpygame.a $(find build/temp.wasm32-*/ | grep o$)
+      run: ${SDKROOT}/python3-wasm dev.py build --wheel
 
     # Upload the generated files under github actions assets section
     - name: Upload dist
@@ -72,3 +59,18 @@ jobs:
       with:
         name: pygame-wasm-dist
         path: ./dist/*
+
+  build-pyodide:
+    name: Pyodide build
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v5.0.0
+
+    - uses: pypa/cibuildwheel@v3.1.4
+      env:
+        CIBW_PLATFORM: pyodide
+
+    - uses: actions/upload-artifact@v4
+      with:
+        name: pyodide-wheels
+        path: wheelhouse/*.whl
--- a/dev.py
+++ b/dev.py
@@ -11,6 +11,7 @@ import os
 import re
 import subprocess
 import sys
+import sysconfig
 from enum import Enum
 from pathlib import Path
 from typing import Any, Union
@@ -34,6 +35,13 @@ COVERAGE_ARGS = ["-Csetup-args=-Dcoverage=true"]
 
 # We assume this script works with any pip version above this.
 PIP_MIN_VERSION = "23.1"
+
+# we will assume dev.py wasm builds are made for pygbag.
+host_gnu_type = sysconfig.get_config_var("HOST_GNU_TYPE")
+if isinstance(host_gnu_type, str) and "wasm" in host_gnu_type:
+    wasm = "wasi" if "wasi" in host_gnu_type else "emscripten"
+else:
+    wasm = ""
 
 
 class Colors(Enum):
@@ -187,9 +195,62 @@ def check_module_in_constraint(mod: str, constraint: str):
     return mod.lower().strip() == constraint_mod[0]
 
 
+def get_wasm_cross_file(sdkroot: Path):
+    """
+    This returns a meson cross file for pygbag wasm sdk (pygame-web/python-wasm-sdk)
+    as a string.
+    Here we set paths to the compiler tooling and include/library paths to ensure that
+    meson can pick up the compiler and build dependencies from the sdk.
+    """
+    emsdk_dir = sdkroot / "emsdk"
+    bin_dir = emsdk_dir / "upstream" / "emscripten"
+
+    node_matches = sorted(emsdk_dir.glob("node/*/bin/node"))
+    node_path = node_matches[-1] if node_matches else Path("node")
+
+    sysroot_dir = bin_dir / "cache" / "sysroot"
+    inc_dir = sysroot_dir / "include"
+    lib_dir = sysroot_dir / "lib" / "wasm32-emscripten" / "pic"
+
+    c_args = [
+        f"-I{x}"
+        for x in [
+            inc_dir / "SDL2",
+            inc_dir / "freetype2",
+            sdkroot / "devices" / "emsdk" / "usr" / "include" / "SDL2",
+        ]
+    ]
+    c_link_args = [f"-L{lib_dir}"]
+    return f"""
+[host_machine]
+system = 'emscripten'
+cpu_family = 'wasm32'
+cpu = 'wasm'
+endian = 'little'
+
+[binaries]
+c = {str(bin_dir / 'emcc')!r}
+cpp = {str(bin_dir / 'em++')!r}
+ar = {str(bin_dir / 'emar')!r}
+strip = {str(bin_dir / 'emstrip')!r}
+exe_wrapper = {str(node_path)!r}
+
+[project options]
+emscripten_type = 'pygbag'
+
+[built-in options]
+c_args = {c_args!r}
+c_link_args = {c_link_args!r}
+"""
+
+
 class Dev:
     def __init__(self) -> None:
-        self.py: Path = Path(sys.executable)
+        self.py: Path = (
+            Path(os.environ["SDKROOT"]) / "python3-wasm"
+            if wasm
+            else Path(sys.executable)
+        )
         self.args: dict[str, Any] = {}
 
         self.deps: dict[str, set[str]] = {
@@ -227,12 +288,24 @@ class Dev:
             build_suffix += "-sdl3"
         if coverage:
             build_suffix += "-cov"
+        if wasm:
+            build_suffix += "-wasm"
+
+        build_dir = Path(f".mesonpy-build{build_suffix}")
         install_args = [
             "--no-build-isolation",
-            f"-Cbuild-dir=.mesonpy-build{build_suffix}",
+            f"-Cbuild-dir={build_dir}",
         ]
 
         if not wheel_dir:
+            if wasm:
+                pprint(
+                    "Editable builds are not supported on WASM as of now. "
+                    "Pass --wheel to do a regular build",
+                    Colors.RED,
+                )
+                sys.exit(1)
+
             # editable install
             if not quiet:
                 install_args.append("-Ceditable-verbose=true")
@@ -258,6 +331,19 @@ class Dev:
 
         if sanitize:
             install_args.append(f"-Csetup-args=-Db_sanitize={sanitize}")
+
+        if wasm:
+            wasm_cross_file = build_dir / "meson-cross-wasm.ini"
+            build_dir.mkdir(exist_ok=True)
+            wasm_cross_file.write_text(get_wasm_cross_file(self.py.parent))
+            install_args.append(
+                f"-Csetup-args=--cross-file={wasm_cross_file.resolve()}"
+            )
+            if not debug:
+                # sdk uses this environment variable for extra compiler arguments.
+                # So here we pass optimization flags. If this isn't set, sdk will
+                # build for debug by default and we don't want that for release builds.
+                os.environ["COPTS"] = "-Os -g0"
 
         info_str = (
             f"with {debug=}, {lax=}, {sdl3=}, {stripped=}, {coverage=} and {sanitize=}"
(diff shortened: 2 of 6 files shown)


WebReflection:
PyScript maintainer here, we were surprised by latest Pyodide inability to use pygame-ce so we do hope this can be merged sooner than later and see this wonderful package usable again as our users got confused (as much as ourselves) too, thank you!

ankith26:
If things work out I would also like to get this in by our next release 2.5.6 which should be happening soon. My goal for this PR is to lay down the buildconfig support so that both pygbag and pyodide can benefit from this.

I just updated this PR and got compilation to work with pyodide locally. I am facing some runtime issues with the generated wheels on pyodide, and I don't know much about emscripten/pyodide to fix those errors. The current error is `pygame.error: The video driver did not add any displays` when `pygame.display.set_mode` runs as part of a simple test script.

I have also started work on adding this to our CI though there is some compilation error I don't see locally: `wasm-ld: error: unable to find library -lfreetype-legacysjlj`

@ryanking13 hello! I believe you maintain pygame-ce for pyodide. I would appreciate your help with moving this PR ahead.

ankith26:
Some more explanations about the changes in this PR:

All `dev.py` changes are for handling builds with pmp-p's `pygame-web/python-wasm-sdk` (pygbag), these changes should be irrelevant to the pyodide side of things. While for pygbag we do a static build of all our sources to generate one single static library, I'm thinking we should go with our usual shared library build strategy that we do on every other platform for pyodide.

pmp-p:
Actually pygbag since 0.9.0 uses a shared library in a wheel, but the intermediate fully statically linked is important in CI to avoid any runtime link errors that were quite common in emscripten early days when enabling various size optimisations / DCE.

It is quite easy to make a shared lib from the full static one without recompiling anything.

Afaik the major difference beetween python-wasm / pyodide wheels is that pygbag link SDL2/3 and OpenGL/ES into the pygame dynamic library directly and force JS glue code of pymain to embed full support for SDL/GL/GLES/OpenAL/GLFW calls without any sort of code elimination. Pyodide main must be quite different since it cannot load eg panda3d / raylib wheels without missing some symbols.

@ankith26 I will test that PR coming week.

.github/workflows/build-emsdk.yml
```diff
@@ -32,43 +32,45 @@ concurrency:
...
+      run: ${SDKROOT}/python3-wasm dev.py build --wheel
 
     # Upload the generated files under github actions assets section
     - name: Upload dist
       uses: actions/upload-artifact@v4
       with:
         name: pygame-wasm-dist
         path: ./dist/*
+
+  build-pyodide:
```
ryanking13, line 63: Pyodide has a slightly different approach than pygbag. pygbag statically links pygame to python binary, so it does not have `.so` files in the package, but Pyodide uses dynamic linking, which requires linking shared libraries dynamically, which is why the [meta.yaml](https://github.com/pyodide/pyodide-recipes/blob/main/packages/pygame-ce/meta.yaml) file has complex logic. cibuildwheel does not handle those things, so it is probably not easy to make cibuildwheel work here.

pyproject.toml
```diff
@@ -97,6 +97,14 @@ setup-args = [
     "-Derror_docs_missing=true",
 ]
 
+[tool.cibuildwheel.pyodide]
+build = "cp313-*"  # build only for the latest python version.
+
+# taken from pyodide recipe at https://github.com/pyodide/pyodide-recipes/blob/main/packages/pygame-ce/meta.yaml
+before-build = "embuilder build sdl2 sdl2_ttf sdl2_image sdl2_mixer sdl2_gfx libjpeg libpng giflib harfbuzz vorbis mpg123 libmodplug freetype libhtml5 --pic"
```
ryanking13: > wasm-ld: error: unable to find library -lfreetype-legacysjlj

Could you try building the `legacysjlj` version as well?

```
embuilder freetype-legacysjlj libpng-legacysjlj
```

Pyodide>=0.28 uses wasm exception, and it requires differnet version of library which is built with wasm-exception enabled.

.github/workflows/build-emsdk.yml
ankith26: Well, in this PR I am trying to make it so that pyodide builds get built just like the rest of our platforms (dynamically linked with per module shared object files). If it works out, there won't be a need to have complex logic in `meta.yml`, `pyodide build` working out of box would be ideal.

pyproject.toml
ankith26: Thanks, that fixed the compile issue on CI. https://github.com/pygame-community/pygame-ce/actions/runs/18042044641?pr=3588 now we have a pyodide wheel compiling on our CI which is another step done! :partying_face:

.github/workflows/build-emsdk.yml
ryanking13: I see. I didn't read the comment carefully... If it works, it would be great for Pyodide.

ankith26: Right now I am still stuck on `pygame.error: The video driver did not add any displays` error. I tried adding a couple of linker flags and things but nothing really helped. If you have faced this issue before, how did you solve it?

ryanking13:
Currently the pyodide wheel in the CI artifact has dynamic linking issue.

```
>>> pygame.init()
Pyodide has suffered a fatal error. Please report this to the Pyodide mai
ntainers.
The cause of the fatal error was:
Error: Dynamic linking error: cannot resolve symbol FT_Init_FreeType
```

The `font.cpython-313-wasm32-emscripten.so` requires libfreetype, but it doesn't seem linking the libfreetype static library during the linking step.

ryanking13:
The same issue exists in the `mixer.cpython-313-wasm32-emscripten.so` and `mixer_music.cpython-313-wasm32-emscripten.so`. It requires `SDL_mixer` but it doesn't seem linked.

.github/workflows/build-emsdk.yml
ryanking13: When I tested the wheel locally in Pyodide, pygame.init() fails due to a dynamic linking error (see my comments below). Any other errors may happen because the initialization has failed.

ankith26:
Can you retry with the latest CI artifact? I think those linker errors should be gone now.

ryanking13:
@ankith26 

Thanks. The linking errors are gone, but now I am experiencing `The video driver did not add any displays` that you mentioned. This also happened when I was building pygame-ce for Pyodide for the first time. But I didn't figure out the exact reason or the situation when it happens. I basically tried some random flags and someday the error was gone.

I'll try to debug with the wheel and the code changes in this PR more.

ryanking13:
So, I did some debugging, and here is what I found so far.

The problem happens in resolving the global `_this` variable that SDL video use:

SDL uses a [static global variable `_this`](https://github.com/libsdl-org/SDL/blob/cc016b0046d563287f0aa9f09b958b5e70d43696/src/video/SDL_video.c#L336) which is initialized and used by several functions. But I noticed that some functions read this `_this` variable from different addresses.

I think it happens because of three different reasons combined:

1. Static global variables have a different memory address per binary.
2. SDL2 is built as a static library and linked to each module separately.
3. Emscripten's GOT (global offset table) overwrites a symbol when there are multiple symbols with the same name.

Because of this, the following situation happens.

---

1. Inside pygame wheel, there are `controller.so` and `video.so` (of course, there are other .so files too) that are linked to SDL and expose `Emscripten_bootstrap` symbol.
2. `Emscripten_bootstrap` is a function that is used inside SDL2 to initialize emscripten video driver ([code pointer](https://github.com/libsdl-org/SDL/blob/cc016b0046d563287f0aa9f09b958b5e70d43696/src/video/SDL_video.c#L125-L127)).
3. When the pygame wheel is loaded, because there are multiple `Emscripten_bootstrap` symbols exported, Emscripten selects one of them to GOT and discard others. Let's say that the one in the `controller.so` is stored in GOT.
4. pygame tries to initialize the video driver in `video.so`, it first initializes the global `_this` pointer, and calls `Emscripten_bootstrap` to do the rest of the bootstrap.
5. However, since `Emscripten_bootstrap` is from `controller.so`, the `_this` memory address  is different from `video.so`, and it dereferences the different value.
6. Error!

---

... So, how to fix? Here are some possible options that I can think of.

1. Patching SDL

1.a. removing `static` from _this

```diff
- static SDL_VideoDevice *_this = NULL;
+ SDL_VideoDevice *_this = NULL;
```

This seems to make the memory address shared by multiple modules.

1.b. hiding probablematic symbols

[Not verified] Hiding symbols such as `Emscripten_bootstrap` will probably resolve the GOT overwrite issue.

2. Build libsdl as a shared library

Building SDL as a shared library will avoid exporting symbols from multiple binaries. (probably related issue: https://github.com/pyodide/pyodide/discussions/5887)

ankith26:
@ryanking13

Thanks for investigating and finding out the root cause of that issue. Well, I suppose the easiest solution is to go back to doing a single (shared library) build on our side. That way there is only one `_this` in the globals. I implemented that in our meson code and now finally its working!

I tested `examples.stars` and `examples.aliens` and both worked as intended (for aliens I had to tweak some resource loading code, in this PR the examples directory is no longer bundled into the wheel to save space and therefore reduce loading time for users)

I suppose the only thing left to do here is to fix the png loading issue. For now libpng is skipped: when I tried adding it there were some issues related to its setjmp/longjmp usage. This PR currently links against `-lSDL2_image-bmp-gif-jpg-qoi-svg-tga` but I figured I needed it to link against the `-wasm-sjlj` variant of `sdl2_image`. I got stuck on how to install this with `embuilder`, do you know how to do this?

ryanking13:
> Thanks for investigating and finding out the root cause of that issue. Well, I suppose the easiest solution is to go back to doing a single (shared library) build on our side. That way there is only one _this in the globals. I implemented that in our meson code and now finally its working!

Yes, that's also fine as there will not be the same symbols exported from multiplie modules. Thanks for handling it!

> in this PR the examples directory is no longer bundled into the wheel to save space and therefore reduce loading time for users)

That's awesome. In pyodide, we've been removing the docs directory as well to reduce the size further.

ryanking13:
> I suppose the only thing left to do here is to fix the png loading issue. For now libpng is skipped: when I tried adding it there were some issues related to its setjmp/longjmp usage. This PR currently links against -lSDL2_image-bmp-gif-jpg-qoi-svg-tga but I figured I needed it to link against the -wasm-sjlj variant of sdl2_image. I got stuck on how to install this with embuilder, do you know how to do this?

Passing `-sSUPPORT_LONGJMP=wasm` when building sdl2_image with embuilder would work I think. The resulting library would be something like `-lSDL2_image-bmp-gif-jpg-qoi-svg-tga-wasm-sjlj`

ankith26:
But `embuilder` doesn't accept `-sSUPPORT_LONGJMP` as an argument. It throws `embuilder.py: error: unrecognized arguments: -sSUPPORT_LONGJMP=wasm`. The [port file](https://github.com/emscripten-core/emscripten/blob/7d0322e51b8ea2444436630d0ad78ca4f9947cce/tools/ports/sdl2_image.py) does however check for `settings.SUPPORT_LONGJMP` attribute but there is no documentation on how to modify this attribute from embuilder commandline or any env var.

EDIT: I got a fix for it

meson_options.txt
```diff
@@ -40,3 +40,6 @@ option('coverage', type: 'boolean', value: false)
 
 # Controls whether to use SDL3 instead of SDL2. The default is to use SDL2
 option('sdl_api', type: 'integer', min: 2, max: 3, value: 2)
+
+# Specify the type of emscripten build being done.
+option('emscripten_type', type: 'combo', choices: ['pygbag', 'generic'], value: 'generic')
```
pmp-p: why not  wasm1 / wasm2 / wasm3 instead ?

where maybe : 

wasm1 = mvp+bigint  or wasi
wasm2 = mvp+bigint+ intrinsics   pyodide <= 0.27 
wasm3 = mvp+bigint+ intrinsics + native EH    pyodide >= 0.28

dev.py
```diff
@@ -187,9 +195,51 @@ def check_module_in_constraint(mod: str, constraint: str):
     return mod.lower().strip() == constraint_mod[0]
 
 
+def get_wasm_cross_file(sdkroot: Path):
+    emsdk_dir = sdkroot / "emsdk"
+    bin_dir = emsdk_dir / "upstream" / "emscripten"
+    node_bin_dir = emsdk_dir / "node" / "22.16.0_64bit" / "bin"
```
pmp-p: should use a glob on node version bits there, as emsdk may use different custom node builds for each sdk version

meson_options.txt
ankith26: For now I chose to go with `pygbag` and `pyodide` for specific and clear naming. We can add more generic labels in the future if possible/required.

dev.py
ankith26: fixed.

ryanking13:
> But `embuilder` doesn't accept `-sSUPPORT_LONGJMP` as an argument. It throws `embuilder.py: error: unrecognized arguments: -sSUPPORT_LONGJMP=wasm`. The [port file](https://github.com/emscripten-core/emscripten/blob/7d0322e51b8ea2444436630d0ad78ca4f9947cce/tools/ports/sdl2_image.py) does however check for `settings.SUPPORT_LONGJMP` attribute but there is no documentation on how to modify this attribute from embuilder commandline or any env var.
> 
> EDIT: I got a fix for it

Oh yeah, that's a good point. Probably it should be reported to Emscripten so we can pass those flags to embuilder.

The current patch looks okay to me. Another workaround that I use in that situation is to build a dummy source file with all the ports enabled. It will trigger building all the ports.

```sh
# create dummy c file
touch temp.c 
# build dummy file, which ports enabled.
emcc temp.c -sSUPPORT_LONGJMP="wasm" -fpic --use-port=sdl2_image
```

ankith26:
My fix is also pretty hacky but I guess it'll do for now...

Anyways, I should be done with this PR from my end. I have decided to push running tests on our CI to a future PR in the interest of getting this PR merged quicker (+ I see the infra to run tests is already available on pyodide recipe repo).

ryanking13:
I can confirm that the current Pyodide wheel works well with Pyodide 0.28.3. Thanks! There seems like a little bit of performance issue in Chrome, but I think it is a problem in Pyodide side, not Pygame, let me handle that.

Thanks for fixing the Emscripten build!

Starbuck5:
@ankith26 Thanks for working on this, you're amazing, as always.

Thoughts from looking through the diff:
- The `get_wasm_cross_file` could use a comment / docstring explaining what that is. I figured it out, https://mesonbuild.com/Cross-compilation.html, but when I looked at the diff I was like what the heck is that.
- `os.environ["COPTS"] = "-Os -g0"` could use a comment too.
- `if not portmidi_dep.found() and not plat.startswith('emscripten')` is this just to be safe, or does portmidi exist in emscripten environments and we don't want it for some reason?
- `if not get_option('stripped') and not plat.startswith('emscripten')`. Philosophically I'm wondering if it's better to have the system set the stripped argument when emscripten is determined, rather than making emscripten the option pull rank everywhere else. I was wondering the same with src_c/meson.build. Like all it's doing is a flat build, we could test that bit out locally if it was its own setting. I'm not necessarily advocating to change this right now, just wanted to get the thought out there.
- In pyproject.toml, in tool.cibuildwheel.pyodide, you're using 3.13. Should that be updated to 3.14 now that it's out?

ankith26:
I have added comments to explain the wasm cross file and the `COPTS` bits.

As for overriding stripped on emscripten, I think the build system should choose to do the "right thing" by default.  There's no way (that I know of) where we can set `stripped` option to default to `true` on emscripten and remain to be `false` on every other platform right in `meson_options.txt`.

emscripten doesn't have a portmidi port as of now. Without this change, an error would be raised because it cannot find portmidi. Yes, one can pass `-Dmidi=false` to achieve the same affect as doing this, but that again goes against my aim of making the build do the right thing by default and work out of box.

As I understand, each pyodide version is built to work with exactly one python version (and also exactly one emscripten version). The latest pyodide version is based on python 3.13. When they release an update based on 3.14, we can update python then.

Starbuck5:
Okay, that makes sense. I was thinking only dev.py, you'd have dev.py specify the specific options needed/wanted for emscripten. But not all builds will happen through dev.py, theoretically. I see in the meson docs it doesn't allow setting options programmatically.

I don't have any capacity to test this, but I did test the non emscripten bits still work, so I'm comfortable enough to approve.

ankith26:
@ryanking13 

This PR has been merged in for 2.5.6.dev2 (which is our pre-release before the final 2.5.6 release). It would be cool if you could update the pyodide side recipe to use this release and report issues if any.

ryanking13:
> @ryanking13
> This PR has been merged in for 2.5.6.dev2 (which is our pre-release before the final 2.5.6 release). It would be cool if you could update the pyodide side recipe to use this release and report issues if any.

Yep, let me try! Thanks again!

Issue #3585, opened 2025-09-08
Some Pygame types do not support weak referencing


cherez:
Some Pygame types do not support weak references, which can be an obstacle in some use cases such as  caching:

```
  File "/usr/lib/python3.13/weakref.py", line 348, in __new__
    self = ref.__new__(type, ob, callback)
TypeError: cannot create weak reference to 'pygame._sdl2.video.Texture' object
```
```
  File "/usr/lib/python3.13/weakref.py", line 348, in __new__
    self = ref.__new__(type, ob, callback)
TypeError: cannot create weak reference to 'pygame.freetype.Font' object
```

Similar classes (`pygame.Surface` and `pygame.Font` for example) do support weak referencing, so I would suggest more consistently supporting it. Would a PR for this be welcome?

ankith26:
Closing this issue now that you've added weakref support for both classes you mentioned. Feel free to reopen if there's any other class you find that you need weakref support for.

cherez:
That seems perfectly fair; if I find more I'll likely just open PRs for them directly.

PR #3590, opened 2025-09-20
Add weak reference support to Texture

cherez:
Per #3585, pygame-ce is kind of inconsistent in that some of its types support weak references, while others do not. As weak references have value in some cases such as asset caching, I believe it is worthwhile to support them.

This PR aims to bring `_render.Texture` to parity with `Surface` with regard to this feature.

### DIFF
--- a/src_c/include/_pygame.h
+++ b/src_c/include/_pygame.h
@@ -556,6 +556,7 @@ struct pgTextureObject {
     pgRendererObject *renderer;
     int width;
     int height;
+    PyObject *weakreflist;
 };
 
 typedef struct {
--- a/src_c/render.c
+++ b/src_c/render.c
@@ -682,6 +682,7 @@ texture_from_surface(PyObject *self, PyObject *args, PyObject *kwargs)
     Py_XINCREF(new_texture->renderer);
     new_texture->width = surf->w;
     new_texture->height = surf->h;
+    new_texture->weakreflist = NULL;
     return (PyObject *)new_texture;
 }
 
@@ -1169,6 +1170,9 @@ texture_dealloc(pgTextureObject *self, PyObject *_null)
     if (self->texture) {
         SDL_DestroyTexture(self->texture);
     }
+    if (self->weakreflist) {
+        PyObject_ClearWeakRefs((PyObject *)self);
+    }
     Py_TYPE(self)->tp_free(self);
 }
 
@@ -1289,6 +1293,7 @@ static PyTypeObject pgTexture_Type = {
     .tp_basicsize = sizeof(pgTextureObject),
     .tp_dealloc = (destructor)texture_dealloc,
     .tp_doc = DOC_SDL2_VIDEO_TEXTURE,
+    .tp_weaklistoffset = offsetof(pgTextureObject, weakreflist),
     .tp_methods = texture_methods,
     .tp_init = (initproc)texture_init,
     .tp_new = PyType_GenericNew,
--- a/test/render_test.py
+++ b/test/render_test.py
@@ -1,4 +1,6 @@
+import gc
 import unittest
+import weakref
 
 import pygame
 import pygame._render as _render
@@ -464,6 +466,13 @@ class TextureTest(unittest.TestCase):
         result = self.renderer.to_surface()
         for x in range(64, 82):
             self.assertEqual(pygame.Color(80, 120, 160, 255), result.get_at((x, 50)))
+
+    def test_garbage_collection(self):
+        reference = weakref.ref(self.texture)
+        self.assertTrue(reference() is self.texture)
+        del self.texture
+        gc.collect()
+        self.assertIsNone(reference())
 
     def test_update(self):
         surface = pygame.Surface((100, 100))


Starbuck5:
Hello, thanks for your PR, this looks good -- gets it up to standard of the rest of the codebase.

There is a cost to adding weakref support, in that it makes every instance of the object bigger.

```py
import sys

import pygame
from pygame import _render

win = pygame.Window()
renderer = _render.Renderer(win)

print(sys.getsizeof(_render.Texture(renderer, (16, 16))))

# Before: 40
# After: 48
```

The docs (https://docs.python.org/3/c-api/typeobj.html#c.PyTypeObject.tp_weaklistoffset) recommend using Py_TPFLAGS_MANAGED_WEAKREF instead. But of course that only works on 3.12 and above, so I wonder if it's worth the additional code complexity to use `#if PY_VERSION_HEX` checks to use the new stuff on 3.12 and above and the old stuff on the older versions of Python we support. I believe it would get rid of the object size increase.

---------------------------

Also, as a new contributor, how did you find getting set up to compile with our documentation? Anything misleading or lacking would be great feedback.

cherez:
> The docs (https://docs.python.org/3/c-api/typeobj.html#c.PyTypeObject.tp_weaklistoffset) recommend using Py_TPFLAGS_MANAGED_WEAKREF instead. But of course that only works on 3.12 and above, so I wonder if it's worth the additional code complexity to use `#if PY_VERSION_HEX` checks to use the new stuff on 3.12 and above and the old stuff on the older versions of Python we support. I believe it would get rid of the object size increase.

I may have misunderstood how to do this, but making the weakreflist conditional on Python < 3.12:
```python
#if PY_VERSION_HEX < 0x030c0000
    PyObject *weakreflist;
#endif
```
And setting the flag on >= 3.12:
```python
#if PY_VERSION_HEX >= 0x030c0000
    .tp_flags = Py_TPFLAGS_DEFAULT | Py_TPFLAGS_MANAGED_WEAKREF,
#else
    .tp_weaklistoffset = offsetof(pgTextureObject, weakreflist),
#endif
```

Actually made the object size larger on 3.13.7:
```python
>>> print(sys.getsizeof(_render.Texture(renderer, (16, 16))))
56
```

I'm not sure what's going on under the hood, but it looks like Py_TPFLAGS_MANAGED_WEAKREF implicitly adds 2 pointers versus this implementation's 1.


> Also, as a new contributor, how did you find getting set up to compile with our documentation? Anything misleading or lacking would be great feedback.

I'm relatively experienced with Python development, and already had all the necessary libraries installed, so a `python setup.py build`  just worked right away. I spent a little while trying to figure out a way to run tests without installing the package more globally before I just accepted `pip install --user .` would be easier. A script to run tests versus your local changes would be nice, but I didn't see anything like that in a quick perusal of the docs.

Starbuck5:
That's very interesting that you were able to get the size at all, because I just tried to do this on pygame.Rect (reducing the size of a commonly used object would be awesome), and I just got segfault after segfault.

Apparently, Py_TPFLAGS_MANAGED_WEAKREF has an undocumented constraint that it can only be used on GC types. https://github.com/python/cpython/issues/134786 . But Texture isn't a GC type either.

Starbuck5:
> I'm relatively experienced with Python development, and already had all the necessary libraries installed, so a python setup.py build just worked right away. I spent a little while trying to figure out a way to run tests without installing the package more globally before I just accepted pip install --user . would be easier. A script to run tests versus your local changes would be nice, but I didn't see anything like that in a quick perusal of the docs.

I mean if you pip install while within a virtualenv you could run the tests off of that?

We've largely moved away from setup.py with the demise of distutils, `pip install .` no longer will run it iirc, and we also support editable installs / code formatting / developer things with our own custom "dev.py": https://github.com/pygame-community/pygame-ce/wiki/Advanced-compilation-guide-%28for-developers%29

cherez:
> I mean if you pip install while within a virtualenv you could run the tests off of that?

True, I was being quick and dirty, but a virtualenv would work fine.

> We've largely moved away from setup.py with the demise of distutils, `pip install .` no longer will run it iirc, and we also support editable installs / code formatting / developer things with our own custom "dev.py": https://github.com/pygame-community/pygame-ce/wiki/Advanced-compilation-guide-%28for-developers%29

That does look like exactly what I was looking for! In that case maybe linking that advanced compilation guide in the How to Contribute section of the README would help. I kind of breezed past the Building From Source section, since I had it compiling and didn't expect testing information to be behind the links there.

cherez:
> That's very interesting that you were able to get the size at all, because I just tried to do this on pygame.Rect (reducing the size of a commonly used object would be awesome), and I just got segfault after segfault.
> 
> Apparently, Py_TPFLAGS_MANAGED_WEAKREF has an undocumented constraint that it can only be used on GC types. [python/cpython#134786](https://github.com/python/cpython/issues/134786) . But Texture isn't a GC type either.

I was getting segfaults whenever a texture was deallocated, but I was able to measure the size before the segfault, so I treated that as a separate issue.

It sounds like this might be a more complex upstream issue, so I'd probably move to defer supporting `Py_TPFLAGS_MANAGED_WEAKREF` until things are a bit more clear on the CPython side.

Starbuck5 (approved):
Alright, this looks good to me. Thanks for the PR!

PR #3596, opened 2025-09-29
Add weak reference support to freetype.Font

cherez:
Following the template of #3590, adding weak reference support to freetype.Font.

This aims to bring freetype.Font to parity with pygame.Font with respect to weak references.


MightyJosip (approved):
Same as your last commit. Thank you. +1

ankith26 (approved):
LGTM, thanks for contributing! 🎉

PR #3597, opened 2025-10-01
Added outline property to pygame.font.Font

cherez:
This is my first 'meaty' PR, so let me know if I missed anything.

Adds wrappers for [TTF_SetFontOutline](https://wiki.libsdl.org/SDL2_ttf/TTF_SetFontOutline) and [TTF_GetFontOutline](https://wiki.libsdl.org/SDL2_ttf/TTF_GetFontOutline). As-is I don't believe there's a clean way to outline fonts in pygame.


docs/reST/ref/font.rst
```diff
@@ -562,6 +578,32 @@ solves no longer exists, it will likely be removed in the future.
...
+      | :sl:`set the point size of the font`
+      | :sg:`set_outline(size, /) -> None`
+
+      Sets the outline value of the font.
+
+      .. note:: This is the same as the :attr:`outline` attribute.
+
+      .. versionadded:: 2.5.6
+
+      .. ## Font.set_outline ##
```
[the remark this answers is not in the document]
cherez: Good bot, updated with [8c13923].

docs/reST/ref/font.rst
```diff
@@ -562,6 +578,32 @@ solves no longer exists, it will likely be removed in the future.
...
+      | :sg:`get_outline() -> int`
+
+      Returns the outline value of the font.
+
+      .. note:: This is the same as the :attr:`outline` attribute.
+
+      .. versionadded:: 2.5.6
+
+      .. ## Font.get_outline ##
+
```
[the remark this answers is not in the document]
cherez: Good bot, updated with [8c13923].

docs/reST/ref/font.rst
```diff
@@ -306,6 +306,22 @@ solves no longer exists, it will likely be removed in the future.
...
+      | :sl:`Gets or sets the font's outline value`
+      | :sg:`outline -> int`
+
+      The outline value of the font.
+
+      When set to 0, the font will be drawn normally. When positive,
+      the text will be drawn as a hollow outline. This can be drawn
+      underneath unoutlined text to create a text outline effect.
+
+      .. versionadded:: 2.5.6
```
oddbookworm: You don't describe how setting it to different values affects it. How would 2 differ from 1? You also specifically call out positive as opposed to nonzero, so what about setting it to a negative number?

src_c/font.c
```diff
@@ -901,6 +901,96 @@ font_set_ptsize(PyObject *self, PyObject *arg)
...
+    if (val < 0) {
+        return RAISE(PyExc_ValueError, "outline must be >= 0");
+    }
+    TTF_SetFontOutline(font, (int)val);
+    Py_RETURN_NONE;
+#else
+    return RAISE(pgExc_SDLError,
+                 "pygame.font not compiled with a new enough SDL_ttf version. "
+                 "Needs SDL_ttf 2.0.12 or above.");
+#endif
```
oddbookworm: You've got code duplication. I suggest moving the common code into a helper function that gets called by both the dedicated setter and the property setter

test/font_test.py
```diff
@@ -685,6 +685,70 @@ def test_point_size_method(self):
...
+        )
+        f = pygame_font.Font(pathlib.Path(font_path), 25)
+
+        ttf_version = pygame_font.get_sdl_ttf_version()
+        if ttf_version < (2, 0, 12):
+            with self.assertRaises(pygame.error):
+                f.outline = 0
+            with self.assertRaises(pygame.error):
+                _ = f.outline
+            return
```
oddbookworm: I think it would be cleaner to just skip this test in this case using [unittest.skipIf](https://docs.python.org/3/library/unittest.html#unittest.skipIf)

test/font_test.py
```diff
@@ -1120,18 +1198,22 @@ def query(
...
         f.set_strikethrough(strikethrough)
+        f.set_outline(outline)
         s = f.render(text, antialiase, (0, 0, 0))
         screen.blit(s, (offset, y))
         y += s.get_size()[1] + spacing
         f.set_bold(False)
         f.set_italic(False)
         f.set_underline(False)
         f.set_strikethrough(False)
+        f.set_outline(0)
```
oddbookworm: This seems like it needs to guard against SDL_ttf version as well

oddbookworm (changes requested):
Overall, excellent add! I do have some comments, but they're all nice and simple. Thanks for the contribution! 🎉

docs/reST/ref/font.rst
cherez: That's a fair point. The [upstream docs](https://wiki.libsdl.org/SDL2_ttf/TTF_SetFontOutline) are quite vague about what the value does, and I didn't want to offer a more specific explanation that isn't accurate.

src_c/font.c
cherez: I concur; I considered deduplicating the functions, but since the rest of the file seemed to use this sort of duplication I erred toward consistency. Is there a convention for this kind of deduplication elsewhere in the project?

test/font_test.py
cherez: That would be cleaner, but that would also mean the case of trying to use the property on an unsupported version is not being tested. Is that a concern?

docs/reST/ref/font.rst
oddbookworm: I think I'd like for it to at least mention that negative values are disallowed and larger values mean a thicker outline

src_c/font.c
oddbookworm: It's scattered from what I remember. Personally, I'm a huge advocate for not duplicating code when possible

test/font_test.py
oddbookworm: You can have a dedicated test for it that skips if the SDL_ttf version is too new. I'm not a huge fan of having this weird logic in the middle of the test

test/font_test.py
cherez: Added guards with 935ab5c.

test/font_test.py
cherez: Changed to use skipIf, and added tests for the stubs with  935ab5c.

src_c/font.c
cherez: I rewrote font_get_outline to just call font_getter_outline (We could probably even tell python to call font_getter_outline instead with a type cast?), and font_set_outline to call font_setter_outline and massage the return value.

Common helper functions would work too, but I don't see a reason to define 6 functions when 4 (3?) will give the same result.

docs/reST/ref/font.rst
cherez: Fair points, I expanded the explanation with 0c1d1ca.

Starbuck5:
I haven't looked through the code, but just reading for interface I see you've added this functionality as a getter/setter and as a property. While most things in font are duplicated in this way, it's because things are being off boarded from being getter/setters. So I see no reason to add getter/setters in a new API, if it's going to be available as a property too.

Although, in the SDL3 documentation for this function it specifies that this can clear a cache of generated glyphs: https://wiki.libsdl.org/SDL3_ttf/TTF_SetFontOutline. So maybe a getter/setter is better because it implies more action taken by the backend?

My thought is we should only support 1 modern thing with a new API, and that it should be a property instead of getter/setter.

oddbookworm (approved):
Alright, this looks good to me now. Thanks for the contribution! 🎉

Starbuck5:
Hey, I noticed that SDL_ttf 2.0.12 is below our minimum required version, so we don't need to do any version checking on it (https://github.com/pygame-community/pygame-ce?tab=readme-ov-file#dependencies). However, there should be version checking to handle error checking with SDL3: https://wiki.libsdl.org/SDL3_ttf/TTF_SetFontOutline can return an error, unlike the SDL2 version.

Another thing I see is that you're doing a PyLong_Check before PyLong_AsLong. PyLong_AsLong will do the check itself, and you're already handling errors off of that, so the check is unnecessary. I tested it out, the error looks like `TypeError: an integer is required (got type str)`, which is perfectly suitable.

Lastly, I'm concerned about the usability of this. Your docs say "This can be drawn underneath unoutlined text to create a text outline effect." But you have to know exactly how to position the output surfaces to do, which is not specified here. had to figure it it out with experimentation.

What happens if you take both Surfaces and blit at same coordinates
image

Versus with the standard text offset by the outline width in x and y
image

I wrote a function that does what I think users will want:
```py
def render_outlined(
    font: pygame.Font,
    text: str,
    text_color: pygame.typing.ColorLike,
    outline_color: pygame.typing.ColorLike,
    outline_width: int,
) -> pygame.Surface:
    old_outline = font.outline
    if old_outline != 0:
        font.outline = 0
    base_text_surf = font.render(text, True, text_color)
    font.outline = outline_width
    outlined_text_surf = font.render(text, True, outline_color)

    outlined_text_surf.blit(base_text_surf, (outline_width, outline_width))
    font.outline = old_outline
    return outlined_text_surf
```
This could be included in the docs as an example. It's precedented. But then the question is that if this is what people are actually going to do with it, why not have a built in function to do it?

There was a prior attempt to port this functionality by nari in https://github.com/pygame-community/pygame-ce/pull/2828, it got bogged down in complexity because we were trying to make the end result more usable and that was complex to do in C. Your implementation is very simple but harder to use.

Starbuck5:
Also btw this PR fixes #2452, my issue from 2023.

I think we should go with your current API and not get perfect-is-the-enemy-of-the-good-ed by me. Also in the current day I've been more in favor of wrapping SDL functionality directly instead of trying to dress it up.

But in your docs it would definitely be helpful to give some detail about how the outline value is in pixels, and if drawing an outline around text, the outline can be drawn and then the un-outlined text can be positioned on top of it by the outline value. Perhaps including my function as an example. And then maybe in the future we can just that function onto the font object directly.

Oh and also here is my test script:
```py
import time

import pygame

pygame.init()

clock = pygame.time.Clock()

# fontname = "comicsans"
fontname = "timesnewroman"
outline = 10

font = pygame.font.SysFont(fontname, 150)
print(font.name)

text = "PYGAME-CE"
text = "pygame-ce"


def render_outlined(
    font: pygame.Font,
    text: str,
    text_color: pygame.typing.ColorLike,
    outline_color: pygame.typing.ColorLike,
    outline_width: int,
) -> pygame.Surface:
    old_outline = font.outline
    if old_outline != 0:
        font.outline = 0
    base_text_surf = font.render(text, True, text_color)
    font.outline = outline_width
    outlined_text_surf = font.render(text, True, outline_color)

    outlined_text_surf.blit(base_text_surf, (outline_width, outline_width))
    font.outline = old_outline
    return outlined_text_surf


start = time.time()
for _ in range(1):
    surf = render_outlined(font, text, "#aaeebb", "dodgerblue4", outline)
print(time.time() - start)

rect = surf.get_rect(center=(1280 / 2, 720 / 2))

while True:
    screen.fill("mediumpurple1")
        if event.type == pygame.QUIT:
            pygame.quit()
            raise SystemExit

    screen.blit(surf, rect)
    pygame.display.flip()
    clock.tick(144)
```

cherez:
Good feedback, pushed up changes removing the TTF < 2.0.12 checks, testing for errors on TTF >= 3, removing the redundant PyLong_Check, and expanding the docs with your example function.

I wouldn't be opposed to adding some render_outlined methods, though I can think of a few possible approaches and concerns, and they could be added in future versions.

Starbuck5 (approved):
Ok, I think this is good. Thanks for working it up.

The versionchanged should be updated before merge, we're now on 2.5.7.

I'd also like to wait another week to merge after this approval to let other contributors weigh in if they'd like, even though my approval gets us to the minimum required approvals.

cherez:
Updated the version change note.

Since I don't expect 2.5.7 to release in the near future, I don't see much problem with giving this more time for review.

cherez:
Rebased onto the latest main to resolve a merge conflict in `font.pyi`;  no other conflicts. (The outline property was added next to the changed `__init__` signature from dropping Python 3.9 support.)

ankith26 (approved):
The API and PR LGTM, thanks for contributing! 🎉

PR #3598, opened 2025-10-06
Pin astroid version below 4.0.0 to fix docs

oddbookworm:
Docs currently broken due to astroid 4.0.0 installed by sphinx. Pinning below that to fix the docs generation for now


Starbuck5:
CC @zoldalma999

PR #3600, opened 2025-10-06
Remove usage of deprecated `ByteString`

ankith26:
`ByteString` is deprecated and scheduled for removal in python 3.17

https://docs.python.org/3.14/whatsnew/3.14.html#pending-removal-in-python-3-17


oddbookworm (approved):
LGTM!

PR #3604, opened 2025-10-08
Clean Surface docs 2: Rewrite object, convert, flags docstrings

damusss:
I'm splitting https://github.com/pygame-community/pygame-ce/pull/3455 . This PR contains the four docstrings that are closely related to surface format internals and that have been rewritten almost entirely, requiring more attention.


buildconfig/stubs/pygame/surface.pyi
```diff
@@ -293,23 +288,27 @@ class Surface:
...
-        unless subclass specific instance attributes also need copying.
+        Creates a new copy of the Surface with the pixel format changed. The
+        format can be controlled in different ways:
+           * If no arguments are passed the new Surface will have the same pixel
+             format as the display Surface which is always the fastest format
+             for blitting. It is a good idea to convert all Surfaces before
+             blitting them. In this case the converted Surface won't have any
+             alpha. See :meth:`convert_alpha()` for preserving or creating per-pixel alpha.
+             A :class:`pygame.error` exception is thrown if the display Surface
+             was not initialized.
```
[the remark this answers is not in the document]
damusss: You're AI can you not notice this improvements one at a time so I don't spam commits 🙏

buildconfig/stubs/pygame/surface.pyi
```diff
@@ -46,77 +46,72 @@ class Surface:
...
+         bit depth and masks as default.
+
+    An indexed format stores an index up to 255 instead of a color for each pixel,
+    therefore requires an associated palette. While pygame provides a default palette,
+    it can be controlled with the appropriate methods.
+
+    Surfaces additionally have alpha, colorkey and rectangle clipping. These characteristics
+    mainly affect how the Surface is blitted to other Surfaces. The blit routines will
+    attempt to use hardware acceleration when possible, otherwise they will use highly
+    optimized software blitting methods.
```
zoldalma999: I know that the old docs mention hardware acceleration but I don't think there's a way to use the GPU with surfaces. I would omit this sentence completely.

buildconfig/stubs/pygame/surface.pyi
```diff
@@ -46,77 +46,72 @@ class Surface:
...
+       * Global alpha: A single alpha value that applies to the whole Surface.
+       * Colorkey: A color flagged to be considered transparent in operations.
+
+    Per-pixel alpha allows the greatest flexibility while being generally
+    slower than global alpha or colorkey. Pre-multiplied alpha is generally
+    faster than regular per-pixel alpha.
+
+    All types of transparency can be used together and correctly mix, except
+    pre-multiplied alpha does not consider colorkey nor global alpha.
+    Also, an alpha of 255 is opaque while an alpha of 0 is fully transparent.
```
zoldalma999: I don't think we need to mention this (we don't explain how colors work either) but even if we do, this should be mentioned earlier imo. If a person reads through all of the types of alpha they probably know how alpha values are stored, and I doubt that the people who need this information would get this far.

buildconfig/stubs/pygame/surface.pyi
```diff
@@ -293,23 +288,27 @@ class Surface:
...
+             format as the display Surface which is always the fastest format
+             for blitting. It is a good idea to convert all Surfaces before
+             blitting them. In this case the converted Surface won't have any
+             alpha. See :meth:`convert_alpha()` for preserving or creating per-pixel alpha.
+             A :class:`pygame.error` exception is raised if the display Surface
+             was not initialized.
+           * If another Surface is provided, the format of that Surface will
+             be used.
+           * Passing the bit depth of a pixel that will result in the same formats
+             explained in the :class:`Surface` constructor. Flags can also be
```
zoldalma999: Not sure what this explanation is referring to? Either link to it (with for example [labels](https://www.sphinx-doc.org/en/master/usage/referencing.html#ref-role)) or mention them here again.

buildconfig/stubs/pygame/surface.pyi
```diff
@@ -800,28 +797,22 @@ class Surface:
...
+           * ``RLEACCEL``: Surface has run-length encoding
+           * ``PREALLOC``: Surface uses preallocated memory
 
         See :func:`pygame.display.set_mode()` for flags exclusive to the
         display Surface.
 
-        Used internally (read-only)
-
-        ::
+        Obsolete flags:
```
zoldalma999: Since these are obsolete since pygame 2 and are also probably very rarely used, I think we can remove this section

buildconfig/stubs/pygame/surface.pyi
```diff
@@ -46,77 +46,72 @@ class Surface:
...
+       * Providing another ``Surface`` after the flags argument. That Surface's
+         format will be used and other arguments will be ignored.
+       * Advanced users can provide a bit depth and masks to precisely control how a
+         pixel's channels are organized in memory. If no masks are provided they will
+         be selected optimally from the bit depth. Normal usage of Surfaces is to leave
+         bit depth and masks as default.
+
+    An indexed format stores an index up to 255 instead of a color for each pixel,
+    therefore requires an associated palette. While pygame provides a default palette,
+    it can be controlled with the appropriate methods.
```
zoldalma999: We should link to these "appropriate methods". This goes for the transparency and clipping sections too.

buildconfig/stubs/pygame/surface.pyi
damusss: I also have no idea how could surfaces ever work with the GPU. It's probably best not to mention hardware acceleration with Surfaces, Renderer and the future GPU exist for that.

buildconfig/stubs/pygame/surface.pyi
damusss: In the original version of this PR I explained what each depth value would correspond to, so I was referring to that, but I was advised not to go in such detail. I will update this mysterious reference, thank you.

buildconfig/stubs/pygame/surface.pyi
damusss: I also removed the mention of the obsolete flag in the constructor, and moved the SRCALPHA mention.

buildconfig/stubs/pygame/surface.pyi
damusss: I changed to briefly mentioning it when alpha is introduced. I see the mention useful only to clarify alpha is not stored on a 0-1 range, nothing more than that is needed.

buildconfig/stubs/pygame/surface.pyi
```diff
@@ -46,77 +46,72 @@ class Surface:
...
-    :mod:`pygame.surfarray` module (It's quite quick, but requires NumPy.)
+    When creating a new Surface it will be cleared to all black or all transparent
+    depending on the format. The only required argument is the size, a sequence
+    containing the width and height of the Surface. With no additional arguments
+    the format will best match the display Surface if one was initialized, otherwise
+    the most suitable format is chosen for the current platform.
+
+    Different ways to control the pixel format are available:
+       * Providing a bitmask of flags. The flag ``SRCALPHA`` will incide a per-pixel
+         alpha channel (transparency).
```
aatle: "incide" seems to be a typo.
```suggestion
       * Providing a bitmask of flags. For example, the flag ``SRCALPHA`` gives a
         per-pixel alpha channel (transparency).
```

buildconfig/stubs/pygame/surface.pyi
```diff
@@ -46,77 +46,72 @@ class Surface:
...
+       * Providing another ``Surface`` after the flags argument. That Surface's
+         format will be used and other arguments will be ignored.
+       * Advanced users can provide a bit depth and masks to precisely control how a
+         pixel's channels are organized in memory. If no masks are provided they will
```
aatle: Somewhat unclear what "they" is. Also, a comma could be added after "they" to make the sentence structure clearer.

buildconfig/stubs/pygame/surface.pyi
```diff
@@ -46,77 +46,72 @@ class Surface:
...
+         be selected optimally from the bit depth. Normal usage of Surfaces is to leave
+         bit depth and masks as default.
```
aatle: Nitpick: IMO, the last sentence is implied by "Advanced users".
```suggestion
         be selected optimally from the bit depth.
```

aatle: Maybe bit depth and masks could be explained in a paragraph or in the `get_masks()` docstring. This could be added in a later PR.

buildconfig/stubs/pygame/surface.pyi
```diff
@@ -46,77 +46,72 @@ class Surface:
...
+    Use :meth:`convert` and :meth:`convert_alpha` to convert the Surface to a different
+    format.
```
aatle: I think it should be clarified that these methods are intended for existing surfaces.
I've seen many times where users do `Surface(...).convert_alpha()` which is unnecessary.

A note about converting loaded assets, a common performance tip, is also probably appropriate.
```suggestion
    Use :meth:`convert` or :meth:`convert_alpha` to convert an existing Surface to a
    different format, especially Surfaces loaded by :mod:`pygame.image` functions.
```

buildconfig/stubs/pygame/surface.pyi
```diff
@@ -46,77 +46,72 @@ class Surface:
...
+
+    An indexed format stores an index up to 255 instead of a color for each pixel,
+    therefore requires an associated palette. While pygame provides a default palette,
+    it can be controlled with the appropriate get/set palette methods.
```
aatle: IMO, I feel like the explanation could be clearer to unfamiliar users if it led with the concept of a color palette instead of suggesting that the palette is secondary.

```suggestion
    An indexed format uses a Surface color palette, a sequence of colors indexed from 
    0 to 255. Each pixel stores an index instead of a color. While pygame provides a
    default palette, it can be controlled with the appropriate get/set palette methods.
```

buildconfig/stubs/pygame/surface.pyi
```diff
@@ -46,77 +46,72 @@ class Surface:
...
+
+    Surfaces additionally have alpha (0-255), colorkey and rectangle clipping. These
```
aatle: I recommend specifying that the alpha is global to avoid confusion.
```suggestion
    Surfaces additionally have global alpha, colorkey and rectangle clipping. These
```
The range of values (`0-255`) can be stated down in the transparency list (`A single alpha value (0-255) that...`), maybe also for `Per-pixel alpha`.

buildconfig/stubs/pygame/surface.pyi
```diff
@@ -46,77 +46,72 @@ class Surface:
...
+         :meth:`Surface.premul_alpha` for more.
+       * Global alpha: A single alpha value that applies to the whole Surface.
+       * Colorkey: A color flagged to be considered transparent in operations.
+
+    Per-pixel alpha allows the greatest flexibility while being generally
+    slower than global alpha or colorkey. Pre-multiplied alpha is generally
+    faster than regular per-pixel alpha.
+
+    All types of transparency can be used togerher and correctly mix, except
+    pre-multipled alpha does not consider colorkey nor global alpha.
```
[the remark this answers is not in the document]
aatle: The typo fixes seem to have gotten reverted somehow.

buildconfig/stubs/pygame/surface.pyi
```diff
@@ -46,77 +46,72 @@ class Surface:
...
+         every pixel has its own alpha value.
+       * Pre-multiplied alpha: An advanced usage of per-pixel alpha. See
+         :meth:`Surface.premul_alpha` for more.
+       * Global alpha: A single alpha value that applies to the whole Surface.
+         Managed with :meth:`get_alpha` and :meth:`set_alpha`.
+       * Colorkey: A color flagged to be considered transparent in operations.
+         See :meth:`get_colorkey` and :meth:`set_colorkey` for more.
+
+    Per-pixel alpha allows the greatest flexibility while being generally
+    slower than global alpha or colorkey. Pre-multiplied alpha is generally
```
aatle: I'm not sure that this is generally true in modern pygame-ce, excluding web platform.

I found this with a benchmark on my desktop:
```
---------------------------------------------
Testing 'blank' surface:

NO_TRANSPARENCY:
0.0880 seconds for 1,000 operations, 0.0880 ms per operation
PER_PIXEL_ALPHA:
0.1099 seconds for 1,000 operations, 0.1099 ms per operation
PREMULTIPLIED_ALPHA:
0.0878 seconds for 1,000 operations, 0.0878 ms per operation
GLOBAL_ALPHA:
3.8913 seconds for 1,000 operations, 3.8913 ms per operation
COLORKEY:
0.1644 seconds for 1,000 operations, 0.1644 ms per operation
```

Ping on Discord for benchmark code.

buildconfig/stubs/pygame/surface.pyi
```diff
@@ -46,77 +46,72 @@ class Surface:
...
+    pre-multipled alpha does not consider colorkey nor global alpha.
+
+    Surfaces can be created that reference the pixel data of other Surfaces.
+    These are called subsurfaces and are created with the :meth:`subsurface()`
+    method. Changing the pixels referenced by either the original Surface or
+    the subsurface will have an effect on both.
+
+    You can use the :meth:`get_at()` and :meth:`set_at()` functions to access
+    pixels of Surfaces but while they are fine for simple access they will be
+    slow when doing pixel work with them.
```
aatle: nitpick: fix run-on sentence flow
```suggestion
    You can use the :meth:`get_at()` and :meth:`set_at()` methods to access
    pixels of Surfaces. While fine for simple access, they are very slow for
    any heavy or frequent pixel work.
```

buildconfig/stubs/pygame/surface.pyi
```diff
@@ -293,23 +288,27 @@ class Surface:
...
+             for blitting. It is a good idea to convert all Surfaces before
+             blitting them. In this case the converted Surface won't have any
+             alpha. See :meth:`convert_alpha()` for preserving or creating per-pixel alpha.
+             A :class:`pygame.error` exception is raised if the display Surface
+             was not initialized.
+           * If another Surface is provided, the format of that Surface will
+             be used.
+           * Passing the bit depth of a pixel will result in the most reasonable
+             format compatible with it. Flags can also be provided, for example
+             ``SRCALPHA`` will request an alpha channel.
```
aatle, line 303: I wasn't able to get a Surface with per-pixel alpha using `convert()` and flags or surfaces.
Could you check this? If the surface never has per-pixel alpha then that should also be specified in one single place in the docstring.

aatle (commented):
Some final remarks for the rewrite.
Otherwise, the changes look great and will help clear up one of the most misunderstood and technical parts of pygame!

buildconfig/stubs/pygame/surface.pyi
damusss: Yeah, considering bit depth and masks aren't supposed to be used often the constructor isn't the right place. get_masks wouldn't be wrong, but I remember about plans to have a unified API representing a "PixelFormat". If that's true, it sounds like the right place for it in the future.

buildconfig/stubs/pygame/surface.pyi
damusss: Perhaps our SIMD paths are more efficient for per pixel alpha? I updated it slightly, tell me what you think.

buildconfig/stubs/pygame/surface.pyi
damusss: I tested, and I was able to get per-pixel alpha, and it follows what I expected when looking at the C implementation.
```py
import pygame

surf = pygame.Surface((100, 100))

# use each method one at a time

# method one
surf = surf.convert(32, pygame.SRCALPHA)
# method two
copy_format = pygame.Surface((10, 10), pygame.SRCALPHA)
surf = surf.convert(copy_format)

surf.fill(0) # ! important

pygame.draw.circle(surf, (0, 255, 0), (50, 50), 40)
pygame.image.save(surf, "test.png")
```

Perhaps you didn't add the `surf.fill(0)` call after conversion. without it, the surface retains the black background it had before even if it now has per pixel alpha.

oddbookworm:
@coderabbitai resume and fresh review

oddbookworm:
@coderabbitai resume

oddbookworm:
@coderabbitai fresh review

buildconfig/stubs/pygame/surface.pyi
aatle: Sorry for the late reply.

Testing the given code, it seems like per-pixel alpha only works when both methods are applied in conjunction, with method one applied and then method two.

The way I check if the per-pixel alpha worked is not by saving a PNG. I believe that image saving uses the alpha channel even if the surface doesn't have per-pixel alpha enabled, making it an inaccurate way to test.

Instead, I check how it is blitted to other surfaces, as well as checking the `get_flags() & SRCALPHA` (these two methods seem to tell the same result).

Test code:
```py
"""

If a bright red circle appears, then no per-pixel alpha.
If a dark red circle appears, then per-pixel alpha is enabled.
"""

import pygame

surf = pygame.Surface((100, 100))

# method one
# surf = surf.convert(32, pygame.SRCALPHA)

# method two
copy_format = pygame.Surface((10, 10), pygame.SRCALPHA)
surf = surf.convert(copy_format)

surf.fill(0)

pygame.draw.circle(surf, (255, 0, 0, 128), (50, 50), 40)

print("Alpha flag:", surf.get_flags() & pygame.SRCALPHA)


window = pygame.Window("test", (100, 100))


def main() -> None:
    while True:
        window.get_surface().fill("black")
        window.get_surface().blit(surf)
        window.flip()
            if event.type == pygame.QUIT:
                return


main()

# pygame.image.save(surf, "test.png")
```

oddbookworm:
@coderabbitai fresh review

buildconfig/stubs/pygame/surface.pyi
aatle: Seems good.
But I personally feel like people will want a whole paragraph on performance and its nuances; noting the platforms, etc.

PR #3605, opened 2025-10-08
Clean Surface docs 3: minor clarifications, typos and movements

damusss:
I'm splitting https://github.com/pygame-community/pygame-ce/pull/3455 . This PR handles the remaining small clarifications and typos. The locking functions and the _pixels_address property are also moved at the end of the file with no changes.

### DIFF
--- a/buildconfig/stubs/pygame/surface.pyi
+++ b/buildconfig/stubs/pygame/surface.pyi
@@ -271,7 +271,7 @@ class Surface:
         :param special_flags: the flag(s) representing the blend mode used for each Surface.
                             See :doc:`special_flags_list` for a list of possible values.
 
-        :returns: ``None`` (unlike regular blitting)
+        :returns: ``None`` (unlike `blit()` and `blits()`)
 
         .. note:: This method only accepts a sequence of (source, dest) pairs and a single
             special_flags value that's applied to all Surfaces drawn. This allows faster
@@ -341,7 +341,7 @@ class Surface:
         attributes then it should override ``copy()``. Shallow copy and deepcopy
         are supported, Surface implements __copy__ and __deepcopy__ respectively.
 
-        If the Surface was a subsurface, the returned Surface will *not* retain
+        If the Surface is a subsurface, the returned Surface will *not* retain
         the parent and will be a regular Surface with its own pixel data.
 
         .. versionadded:: 2.3.1


buildconfig/stubs/pygame/surface.pyi
```diff
@@ -447,80 +451,19 @@ class Surface:
...
-        Returns ``True`` when the Surface is locked. It doesn't matter how many
-        times the Surface is locked.
-        """
-
-    def get_locks(self) -> tuple[Any, ...]:
-        """Gets the locks for the Surface.
-
-        Returns the currently existing locks for the Surface.
+        Return the current alpha value for the Surface. If the blendmode of the
+        Surface is not ``pygame.BLENDMODE_NONE``, this method will always return
```
Starbuck5: The blendmode of a surface is not an accessible detail to the readers of this documentation.

buildconfig/stubs/pygame/surface.pyi
```diff
@@ -447,80 +451,19 @@ class Surface:
         The optional flags argument can be set to ``pygame.RLEACCEL`` to provide
         better performance on non accelerated displays. An ``RLEACCEL`` Surface
         will be slower to modify, but quicker to blit as a source.
+
+        .. versionchangedold:: 2.0 per-Surface alpha can be combined with per-pixel
+            alpha.
```
Starbuck5, line 450: How do you know this?

Starbuck5:
Can you explain why you moved the locking functions / pixel address around?

buildconfig/stubs/pygame/surface.pyi
damusss: that's a very good question I'll see if I find clues in my a-few-months old code

buildconfig/stubs/pygame/surface.pyi
damusss: Makes sense

buildconfig/stubs/pygame/surface.pyi
damusss: Ok @Starbuck5 apparently my source are the existing docs. That line existed before my cleanup, I just moved it to the appropriate location within the docstring. But it looks like I copied the docstring wrong, and now there's a duplicate line.

damusss:
> Can you explain why you moved the locking functions / pixel address around?

This PR just does what I did in the original, so I'll try to explain my past self:
All the locking methods seemed kind of useless so I thought of slapping them to the bottom. But I'm pretty sure things are added in the docs in chronological order, so it might not be the best idea looking at it now, although there are other parts of the docs where methods are added where they make sense and not just at the end.
The property, it's because inside classes you usually put properties with properties and methods with methods. made sense to put it among other properties, at the bottom since it's private. If we want to keep a chronological order this change wouldn't make sense.
I'll move the locking methods back and will wait your opinion on the property.

buildconfig/stubs/pygame/surface.pyi
```diff
@@ -436,23 +439,23 @@ class Surface:
...
         """
 
     def get_alpha(self) -> Optional[int]:
         """Get the current Surface transparency value.
 
-        Return the current alpha value for the Surface.
+        Return the current alpha value for the Surface which is an integer in the
```
Starbuck5: I think this needs a comma.

`Return the current alpha value for the Surface, which is an integer...`

damusss: Added this while fixing the merge conflicts, good call

ankith26 (approved):
LGTM, thanks! 🎉

PR #3610, opened 2025-10-12
Added do_clamp to vector lerping

oddbookworm:
Someone on discord asked why we force clamp with vector lerping, but have an option to disable clamping on vector lerping. This PR brings the vector lerp into parity with the scalar lerp


oddbookworm:
Here's a piece of test code I wrote:
```py3
import pygame

pygame.init()
win = pygame.Window()
screen = win.get_surface()

font = pygame.font.SysFont("Arial", 20)

vec1 = pygame.Vector2(screen.get_rect().center)
vec2 = pygame.Vector2(screen.get_rect().midright) / 1.5

def generate_lerped_values(vec1, vec2, do_clamp):
    points = []

    x = -1.0
    DX = 0.01

    while x < 2.0:
        try:
            points.append(vec1.lerp(vec2, x, do_clamp))
        except ValueError:
            pass
        x += DX
    return points

did_update = True
do_clamp = True

running = True
while running:
        if event.type == pygame.QUIT:
            running = False

            if event.button == 3:
                vec1.xy = event.pos
                did_update = True
            elif event.button == 1:
                vec2.xy = event.pos
                did_update = True

                do_clamp = not do_clamp
                did_update = True

    screen.fill("black")
    if (did_update):
        points = generate_lerped_values(vec1, vec2, do_clamp)
        did_update = False
    for point in points:
        pygame.draw.circle(screen, "blue", point, 2)
    pygame.draw.line(screen, "red", (0, 0), vec1)
    pygame.draw.line(screen, "green", (0, 0), vec2)
    screen.blit(font.render(f"{do_clamp = }", True, "white"))

    win.flip()
```

Left click will move the target, right click will move the source, and spacebar will toggle `do_clamp`

Starbuck5:
Is lerping between two vectors without respecting the endpoints / clamping useful?

aatle:
I think I remember wanting to use such a feature before. But in general, it is just a math operation that you can use to multiply and then apply the offset between two vectors, so it would be good to allow any multiplier, not just within `[0, 1]`.

oddbookworm:
I've run into my own use case for it: extrapolation, like if you're subdividing a triangular mesh

Matiiss:
Why `do_clamp` instead of just `clamp`?

PR #3614, opened 2025-10-13
Update README install instructions for linux

### LINKED ISSUE #3611: pip & pip3 issue with work-around
Someone within the python community in their infinite wisdom decided to change how python packages are installed. I am not sure if this is unique to Linux distros or what. A simple pip install <package-name> or pip3 install <package-name> will no longer work. 

I use Linux Mint Cinnamon and this is the response I get:

```
 This environment is externally managed
╰─> To install Python packages system-wide, try apt install
    python3-xyz, where xyz is the package you are trying to
    install.
    
    If you wish to install a non-Debian-packaged Python package,
    create a virtual environment using python3 -m venv path/to/venv.
    Then use path/to/venv/bin/python and path/to/venv/bin/pip. Make
    sure you have python3-full installed.
    
    If you wish to install a non-Debian packaged Python application,
    it may be easiest to use pipx install xyz, which will manage a
    virtual environment for you. Make sure you have pipx installed.
    
    See /usr/share/doc/python3.12/README.venv for more information.

note: If you believe this is a mistake, please contact your Python installation or OS distribution provider. You can override this, at the risk of breaking your Python installation or OS, by passing --break-system-packages.
hint: See PEP 668 for the detailed specification.
```
I have checked this in Arch and Kali with the same results. See PEP 668: https://peps.python.org/pep-0668/

PEP 668 provides several ways to work around this but the simplest is to use: --break-system-packages like so:
pip install --break-system-packages <package-name> or pip3 install --break-system-packages <package-name>

Given the events around the forking of pygame to pygam-ce I know for certain that Linux Mint repositories do not have pygame-ce but they do have pygame. This is likely the same for Arch based distros as well.

It saddens me that such toxicity within the dev community exists. But I recommend that this work around is incorporated in the install instructions for Linux users.

ankith26:
fixes #3611

Add a section to workaround the infamous Linux "Breaking System Packages" issue.

I cannot update the instructions for the READMEs that are in other languages, I suppose they would have to be left outdated for now.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -30,10 +30,49 @@ New contributors are welcome!
 Installation
 ------------
 
+To install pygame-ce, first make sure you have Python (and pip) installed and available on your PATH.
+Then, run the following command in your terminal or command prompt:
+
 ::
 
    pip install pygame-ce
 
+Note that on some platforms you may need to use ``pip3`` instead of ``pip``.
+
+Linux Note: "Breaking System Packages" Error
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+On some Linux distributions (like Ubuntu or Debian), installing Python packages with ``pip`` can trigger a message like the following:
+
+::
+
+   error: externally-managed-environment
+   This environment is externally managed when system packages are installed.
+
+This means your system Python is managed by the OS, and using ``pip3`` globally might interfere with system packages.
+
+Recommended Solutions
+^^^^^^^^^^^^^^^^^^^^^
+
+1. Use ``venv`` (Virtual Environment):
+
+   You can create a virtual environment and install pygame-ce within it.
+   This way you still use system Python but create a separate environment for installed packages so that they do not conflict with or break system ones.
+
+   .. code-block:: bash
+
+      python3 -m venv venv      # create a venv named 'venv'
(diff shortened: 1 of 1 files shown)


README.rst
```diff
@@ -30,10 +30,57 @@ New contributors are welcome!
...
+On some Linux distributions (like Ubuntu or Debian), installing Python packages with ``pip`` can trigger a message like the following:
+
+.. code-block:: bash
+
+   error: externally-managed-environment
+   This environment is externally managed when system packages are installed.
+
+This means your system Python is managed by the OS, and using ``pip3`` globally might interfere with system packages.
+
+Recommended Workarounds
```
Starbuck5: I don't like calling it what everybody has to do on Linux a "workaround".

How about recommended solutions?

Also in your list you have an entry that says not recommended, which is contradictory to the title.

Also also do you actually have experience with pyenv or uv? This list doesn't need to be comprehensive, if we haven't used a tool we can refrain from recommending it.

ankith26: I did workarounds -> solutions and removed the not recommended option.

I actively use `pyenv` in my workflow (this is how I use python 3.14). As for `uv` I do have some experience using it, while I don't actively use it at the moment I think it's definitely worth recommending.

Starbuck5 (approved):

PR #3615, opened 2025-10-13
[Fix] Detect homebrew directory

ninjaguardian:
No longer assumes homebrew is at `/opt/homebrew` and attempts to find its prefix. Resorts to the default install directory (`/opt/homebrew`) if it cannot be found or an error occurs.

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -289,7 +289,10 @@ else
     if brew_cmd.found()
         result = run_command(brew_cmd, '--prefix', check: false)
         if result.returncode() == 0
-            homebrew_prefix = result.stdout().strip()
+            detected_prefix = result.stdout().strip()
+            if detected_prefix != '' and fs.is_dir(detected_prefix)
+                homebrew_prefix = detected_prefix
+            endif
         endif
     endif
 


buildconfig/config_darwin.py
```diff
@@ -136,16 +137,23 @@ def main(auto_config=False):
...
 
     print('Hunting dependencies...')
-    incdirs = ['/usr/local/include', '/opt/homebrew/include']
-    incdirs.extend(['/usr/local/include/SDL2', '/opt/homebrew/include/SDL2', '/opt/local/include/SDL2'])
+
+    homebrew_prefix = '/opt/homebrew'
+    try:
+        homebrew_prefix = check_output(['brew', '--prefix'], text=True, stderr=DEVNULL).strip()
+    except (FileNotFoundError, CalledProcessError):
+        pass
```
oddbookworm, line 145: I'd like to see a message that it was unable to find the prefix and it's defaulting to something specific for debugging purposes

ankith26: I think changing things here inconsequential given the new meson based build config takes precedence over the old.

PR #3616, opened 2025-10-15
Drop python 3.9

ankith26:
Python 3.9 will be EOL by October 31st. In addition to dropping python 3.9 support, this PR modernises our stubs with python 3.10 specific features.


buildconfig/stubs/pygame/surface.pyi
```diff
@@ -196,15 +196,15 @@ class Surface:
...
+                | tuple[Surface, Point | RectLike, RectLike | int]
+                | tuple[Surface, Point | RectLike, RectLike, int]
+            )
         ],
-        doreturn: Union[int, bool] = 1,
-    ) -> Union[list[Rect], None]:
-        """Draw many Surfaces onto this Surface at their corresponding location.
+        doreturn: int | bool = 1,
+    ) -> list[Rect] | None:
+        """Draw many Surfaces onto this surface at their corresponding location.
```
zoldalma999: Seems like it is undoing the capitalization from #3603, missed from a merge conflict resolution?

ankith26: True, my bad. I have fixed it now.

buildconfig/stubs/pygame/sprite.pyi
```diff
@@ -143,23 +141,23 @@ class LayeredDirty(LayeredUpdates[_DirtySpriteT]):
     def draw(
         self,
         surface: Surface,
-        bgd: Optional[Surface] = None,
-        special_flags: Optional[int] = None,
-    ) -> list[Union[FRect, Rect]]: ...
+        bgd: Surface | None = None,
+        special_flags: int | None = None,
+    ) -> list[FRect | Rect]: ...
```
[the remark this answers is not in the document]
ankith26: please stop suggesting implementation fixes in this PR, this is merely a porting PR. Implementation fixes will be done in future PRs

src_py/typing.py
```diff
@@ -42,24 +42,27 @@ def __len__(self) -> int: ...
...
 
 
 class _HasRectAttribute(Protocol):
     # An object that has a rect attribute that is either a rect, or a function
     # that returns a rect conforms to the rect protocol
+    # We still use old style Union[...] syntax here because RectLike is a
+    # forward reference and on python < 3.14 new style unions would result in a
+    # syntax error.
     @property
     def rect(self) -> Union["RectLike", Callable[[], "RectLike"]]: ...
```
aatle: The best way to do this, error-free, is to use `from __future__ import annotations`. Equivalently, stringizing the entire type expression is an alternative, but I would prefer the `__future__` import as it is cleaner and easier to migrate for 3.14.
Add a `from __future__ import annotations` after the comment about duplicate, before the `__all__` assignment. Remove quotes from `RectLike`. (Remove `Union` import and regenerate `typing.pyi`.)

ankith26: I think I would prefer the stringizing approach, given that [future import annotations are deprecated](https://docs.python.org/3/whatsnew/3.14.html#from-future-import-annotations).

ankith26:
I have fixed the missing `TypeAlias` on the types you mentioned, thanks for catching those.

src_py/typing.py
oddbookworm: I think I agree with using the future import still, at least until 3.14 is our lowest supported version. At that point, we just have to remove an unneeded import statement.

aatle: Both are reasonable.
Technically, future annotations is more modern even though it is deprecated and string literals aren't. It's a lot more convenient to use and migrate.

src_py/typing.py
```diff
@@ -42,24 +42,24 @@ def __len__(self) -> int: ...
...
-ColorLike = Union[Color, SequenceLike[int], str, int]
+ColorLike: TypeAlias = Color | SequenceLike[int] | str | int
 
 
 class _HasRectAttribute(Protocol):
     # An object that has a rect attribute that is either a rect, or a function
     # that returns a rect conforms to the rect protocol
     @property
-    def rect(self) -> Union["RectLike", Callable[[], "RectLike"]]: ...
+    def rect(self) -> "RectLike | Callable[[], RectLike]": ...
```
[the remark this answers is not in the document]
ankith26: this is a false flag

src_py/typing.py
ankith26: I disagree with future import annotations being modern now that 3.14 has been released:
- The behaviour of future import annotations (which is basically "implicitly stringize all annotations") is different from python 3.14 behaviour (where a special `__annotate__` function is internally defined).
- By stringizing the annotation ourself we are limiting scope of the change to that specific instance. Considering `typing.py` is public API we have to consider what change would be the least disruptive. And I imagine that if we were to add from future import annotations now and then remove it inevitably, there could be some disruption.

damusss: I also prefer the specific stringizing approach.

aatle (approved):
LGTM!

note: once this is merged, other open PRs will need to update their stubs

Starbuck5:
What is the advantage of declaring things a TypeAlias over the status quo?

aatle:
It is recommended to use `TypeAlias` to explicitly annotate type aliases. It removes ambiguity, and separates global variables from type aliases.
See https://peps.python.org/pep-0613/.
When pygame-ce reaches python 3.12 minimum, these should be converted to cleaner `type` statements.

Starbuck5:
According to the PEP, 

> These implicit type alias declaration rules create confusion when type aliases involve forward references, invalid types, or violate other restrictions enforced on type alias declaration

But this is not the case for ours? We're not using forward references or violating other rules.

aatle:
Even so, `TypeAlias` should be used, it is standard for type aliases. Type aliases are not normal global variables, so they should be separated in some way.
Once the `type` syntax is reached in 3.12, there are more benefits (which reflects how type aliases are not really traditional variables):
- Lazy evaluation
- Better generic type parameter syntax e.g. `_SpriteOrSprites[T: Sprite] = ...`. This will be needed to fully integrate into the modern 3.12 typing syntax.
- The value of the alias is now a `TypeAliasType` instance, which cannot be used as a normal class

ankith26:
+1 for eventually using the `type` syntax when we get to python 3.12. The current `TypeAlias` changes in this PR can be thought of as a precursor to that change (when we get there it would be easy to find+replace `TypeAlias`), it is making the code more explicit and therefore more clear.

Issue #3581, opened 2025-09-02
SDL3: Mixer changes, what to do?


Starbuck5:
As many of you may know, SDL3_mixer features substantial changes over SDL2_mixer. See https://wiki.libsdl.org/SDL3_mixer/README-migration. It has yet to be formally released, and also relies on an unreleased SDL version.

I've spent quite a bit of time this weekend going through the API and writing a stub file that exposes all SDL3_mixer functionality the way I would do it with a fresh port. I also had to do the same for SDL3's audio module, because SDL3_mixer relies on input/output with audio types like formats, specs, and devices, to various degrees.

<details><summary>mixer2.pyi</summary>

```py
from typing import Type, TypeVar, TypedDict
from collections.abc import Callable

from pygame.typing import FileLike
import audio

from typing_extensions import Buffer


def init() -> None: ...
def quit() -> None: ...
def get_sdl_mixer_version(linked: bool = True) -> tuple[int, int, int]: ...
def ms_to_frames(sample_rate: int, ms: int) -> int: ...
def frames_to_ms(sample_rate: int, frames: int) -> int: ...
def get_decoders() -> list[str]: ...

T = TypeVar("T")
track_stopped_callback = Callable[[T, Track], None]
track_mix_callback = Callable[[T, Track, audio.AudioSpec, Buffer], None]
group_mix_callback = Callable[[T, Group, audio.AudioSpec, Buffer], None]
post_mix_callback = Callable[[T, Mixer, audio.AudioSpec, Buffer], None]

class Mixer:
    def __init__(self, device: audio._AudioDevice = audio.DEFAULT_PLAYBACK_DEVICE, spec: audio.AudioSpec | None = None) -> None: ...
    @property
    def gain(self) -> float: ...
    @gain.setter
    def gain(self, value: float): ...
    def play_tag(self, tag: str, **kwargs) -> None: ...
    def stop_tag(self, tag: str, fade_out_ms: int) -> None: ...
    def pause_tag(self, tag: str) -> None: ...
    def resume_tag(self, tag: str) -> None: ...
    def set_tag_gain(self, tag: str, gain: float) -> None: ...
    def play_audio(self, audio: Audio)-> None: ...
    def stop_all_tracks(self, fade_out_ms: int) -> None: ...
    def pause_all_tracks(self) -> None: ...
    def resume_all_tracks(self) -> None: ...
    @property
    def format(self) -> audio.AudioSpec: ...
[...]

PR #3617, opened 2025-10-17
Support devicename=None in _sdl2.AudioDevice

rhendric:
The docstring already claimed that this was supported, but it didn't work.

### DIFF
--- a/buildconfig/stubs/pygame/_sdl2/audio.pyi
+++ b/buildconfig/stubs/pygame/_sdl2/audio.pyi
@@ -1,4 +1,5 @@
 from collections.abc import Callable
+from typing import Optional
 
 AUDIO_U8: int
 AUDIO_S8: int
@@ -25,7 +26,7 @@ def get_audio_device_names(iscapture: bool = False) -> list[str]: ...
 class AudioDevice:
     def __init__(
         self,
-        devicename: str,
+        devicename: Optional[str],
         iscapture: bool,
         frequency: int,
         audioformat: int,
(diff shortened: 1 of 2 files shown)


Starbuck5:
@rhendric Can you provide some details about your use of the _sdl2.audio API? I'm curious. I was unaware anyone used it, besides from the list device names function.

rhendric:
I'm only using it for the async audio API. I'm playing with generating music dynamically in real time, so I don't have pre-recorded audio files to play, and so getting a callback invoked when SDL needs audio data is much more straightforward than racing to proactively queue up Sounds through the mixer API.

rhendric:
I should probably also mention that I'm in the very early stages of experimenting with this, and while my first demo script showed promise, in the context of a larger application I'm currently struggling with some skipping issues that might force me to switch to synchronous after all (though there seems to be a different bug in the mixer API that is getting in the way of queuing dynamic Sounds repeatedly). So don't move any mountains based solely on my use case just yet.

Starbuck5:
@rhendric You might be interested in what I've been working on for SDL3 audio support then, where it will hopefully not be hidden away in a private undocumented module. https://github.com/pygame-community/pygame-ce/issues/3581

> though there seems to be a different bug in the mixer API that is getting in the way of queuing dynamic Sounds repeatedly

Are you running into the default limit of 8 sounds playing at once? That can be raised.

rhendric:
> SDL3 audio support

Looks good! The `open_stream` function in particular would be a match for my use case, I think, provided that something like the overhead of waiting for the GIL on every callback isn't the source of my current woes.

> Are you running into the default limit of 8 sounds playing at once?

No, I think I'm running into #531. There's a race condition between the main thread rapidly queuing sound data and the SDL thread moving sounds from the queue field to the sound field for the channel. The root issue seems to be that `endsound_callback` in `mixer.c` tests whether `queue` is NULL before locking on anything. I have a patch that I'm in the process of stress testing.

PR #3625, opened 2025-10-31
Make universal2 wheels on macos

### LINKED ISSUE #3535: MacOS: start shipping only a single universal2 build instead of separate x86_64 and arm64 builds
When we added arm64 support keeping it separate from our x86_64 wheels made sense because we had to support old python/pip/packaging versions that did not handle `universal2`. Now it's been a few years, and it looks like the time we drop python 3.9 support will be the best time to do this change.

`universal2` support was [added in `packaging` v20.9](https://github.com/pypa/packaging/blob/main/CHANGELOG.rst#209---2021-01-29) which was then [vendored into `pip` v21.0.1](https://pip.pypa.io/en/latest/news/#v21-0-1) and python 3.10.0 is the first major release to vendor a newer `pip` at the time of its release. So we can safely assume all our users have new enough tooling to be able to install `universal2` wheels.

## Why?

- Most of our wheel size is basically stuff that is not architecture dependent. Using one wheel instead of two saves our PyPI space.
- Should hopefully save a bit of CI time.
- [Github is dropping their `macos-13` runner](https://github.blog/changelog/2025-07-11-upcoming-changes-to-macos-hosted-runners-macos-latest-migration-and-xcode-support-policy-updates/). This is their last runner that runs on x86_64. Supporting x86_64 beyond that is going to require buildconfig changes on our end anyways. This is a good opportunity to do this change.

ankith26:
closes #3535

### DIFF
--- a/.github/workflows/build-macos.yml
+++ b/.github/workflows/build-macos.yml
@@ -70,15 +70,7 @@ jobs:
     env:
       MAC_ARCH: ${{ matrix.macarch }}
 
-      # Explicitly tell CIBW what the wheel arch deployment target should be
-      # There seems to be no better way to set this than this env
-      # We need this because our minimum is 10.11, different from default
-      # of 10.9 on x86s
-      # Related issue: https://github.com/pypa/cibuildwheel/issues/952
-      _PYTHON_HOST_PLATFORM: ${{ matrix.macarch == 'x86_64' && 'macosx-10.11-x86_64' || 'macosx-11.0-arm64'}}
-
-      # Similarly, we need to tell CIBW that the wheel's linking steps
-      # should be for 10.11 on x86
+      # We need to tell CIBW that the wheel's linking steps should be for 10.11 on x86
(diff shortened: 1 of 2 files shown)


Starbuck5:
Very cool! I like this strategy. We already depend on delocate through CIBW iirc, so we're just depending on delocate working here too.

Is it worth pinning the delocate version?

I would like to get someone on MacOS to confirm this works before merging, as an additional sanity check. Edit: asked on discord - https://discord.com/channels/772505616680878080/772505616680878083/1455098234631622761

ankith26:
I updated the commit by adding a version pin to delocate

Starbuck5:
I got a test of this on Mac from Notenlish on discord: https://discord.com/channels/772505616680878080/772505616680878083/1455903851793354823
(intel mac)

They report no issues.

ankith26:
I added the comment to explain the step at high level.

PR #3627, opened 2025-11-01
Fix no palette 8 bit surface gfxdraw segfault

### LINKED ISSUE #586: SDL2: gfxdraw.aa* functions segfault with 8-bit surfaces (849)

ankith26:
Fixes #586

Yes, gfxdraw is scheduled for eventual removal and is generally unmaintained, but we should still not segfault.

### DIFF
--- a/src_c/SDL_gfx/SDL_gfxPrimitives.c
+++ b/src_c/SDL_gfx/SDL_gfxPrimitives.c
@@ -293,17 +293,13 @@ _putPixelAlpha(SDL_Surface *dst, Sint16 x, Sint16 y, Uint32 color, Uint8 alpha)
                     *((Uint8 *)dst->pixels + y * dst->pitch + x) = color;
                 }
                 else {
+                    /* Patched on pygame-ce end to fix segfault when no palette
+                     */
                     Uint8 *pixel = (Uint8 *)dst->pixels + y * dst->pitch + x;
-                    SDL_Palette *palette = format->palette;
-                    SDL_Color *colors = palette->colors;
-                    SDL_Color dColor = colors[*pixel];
-                    SDL_Color sColor = colors[color];
-                    Uint8 dR = dColor.r;
-                    Uint8 dG = dColor.g;
-                    Uint8 dB = dColor.b;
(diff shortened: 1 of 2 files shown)


Starbuck5:
I was unaware that it was possible to create 8 bit surfaces without palettes in pygame-ce running on SDL2. I think a lot of our code assumes 8 bit == palette, probably more changes needed this for SDL3 support given they don't have palettes on their surfaces by default.


Verified with test script, compared with and without palette.
```py
import pygame
import pygame.gfxdraw


pygame.init()


screen = pygame.display.set_mode((512, 512))
clock = pygame.time.Clock()
running = True

WEIRD = True

if WEIRD:
    s = pygame.Surface(size=(512, 512), flags=0, depth=8, masks=(0xE0, 0x1C, 0x3, 0x0))
else:
    s = pygame.Surface(size=(512, 512), depth=8)


pygame.gfxdraw.aacircle(
    s,
    256,
    50,
    64,
    pygame.Color("red"),
)

pygame.gfxdraw.box(s, [50, 50, 300, 100], (255, 0, 0, 150))


while running:
        if event.type == pygame.QUIT:
            running = False

    screen.fill("purple")
    screen.blit(s)

    pygame.display.flip()

    clock.tick(60)

pygame.quit()
```

Starbuck5 (approved):

PR #3629, opened 2025-11-02
Document VIDEORESIZE different OS behavior

### LINKED ISSUE #3448: VIDEORESIZE Event Differs Between Linux and Windows
I found an inconsistency between Windows and Linux.

On Linux, the resize event is called every frame during the resize but on Windows it's only called once, upon releasing the drag. This creates issues if you have code inside the if event block that should run only once.

For example, if you use set_mode to re-create the display, the window will keep re-opening forever.

If there is no fix to this inconsistency, it would be nice to have it as a note in the display docs. Right now it says:  `pygame.VIDEORESIZE events will be sent when the user adjusts the window dimensions` but this is only correct for Windows.
(possible related [issue](https://github.com/pygame-community/pygame-ce/issues/1071))

UPDATE: The same applies to WINDOWRESIZED

https://github.com/user-attachments/assets/2e7de6c0-3ae6-4e85-a84c-a2150d5886aa

https://github.com/user-attachments/assets/65c4d604-05d1-49b9-b68e-c9d56b80211c

Starbuck5:
Fixes #3448 for @AntonisDevStuff

### DIFF
--- a/buildconfig/stubs/pygame/display.pyi
+++ b/buildconfig/stubs/pygame/display.pyi
@@ -34,11 +34,12 @@ automatically switch to operate on the new display.
 When the display mode is set, several events are placed on the pygame event
 queue. ``pygame.QUIT`` is sent when the user has requested the program to
 shut down. The window will receive ``pygame.ACTIVEEVENT`` events as the display
-gains and loses input focus. If the display is set with the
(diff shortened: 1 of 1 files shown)


ankith26 (approved):
LGTM, good docs enhancement 👍

PR #3633, opened 2025-11-02
Fix segfault and stub problem with camera module

### LINKED ISSUE #3022: Segfault and other problems when initializing a camera without opencv
**Environment:**

```
pygame-ce 2.5.1.dev1 (SDL 2.30.5, Python 3.12.0)
Processor:              AMD64 Family 23 Model 104 Stepping 1, AuthenticAMD      SSE2: Yes       AVX2: Yes       NEON: No

Python:                 CPython 3.12.0 (tags/v3.12.0:0fb18b0, Oct  2 2023, 13:03:39) [MSC v.1935 64 bit (AMD64)]
pygame version:         2.5.1.dev1
SDL versions:           Linked: 2.30.5  Compiled: 2.30.5

Display Driver:         windows
```

**Current behavior:**

Imagining you don't have ``opencv-python`` installed, and you don't force ``OpenCV`` backend with environment variables , you get a segfault when initialising a camera object.

**Expected behavior:**

No segfault, and a descriptive message of what is actually hapenning.

**Test code**
```py
import pygame
import os
import pygame.camera
# os.environ["PYGAME_CAMERA"] = "opencv"

screen = pygame.display.set_mode([512, 512])

pygame.print_debug_info()
pygame.camera.init()

print(pygame.camera.list_cameras())

camera = pygame.camera.Camera(0)
camera.start()

version = True

while True:
        if event.type == pygame.QUIT:
            pygame.quit()
            exit()

    screen.blit(camera.get_image())
    pygame.display.flip()
```

**Other things to mention :**

- It would be interesting to tell in the wiki for compilation from source for Windows, how to get access to ``pygame.camera``, by explaining how to compile with MSVC.
- ``pygame.camera.list_cameras()`` is broken, sometimes it returns you a list of string (when I don't force opencv backend), sometimes a list of one integer (when I force opencv backend) and ultimately a ``None`` value.
- When I don't force ``opencv`` backend, I get this error which is not logical :
```py
  File "\\wsl.localhost\Ubuntu-24.04\home\rakoon\pygame-ce\a.py", line 13, in <module>
    camera = pygame.camera.Camera()
             ^^^^^^^^^^^^^^^^^^^^^^
TypeError: function takes at least 1 argument (0 given)
```

Starbuck5:
Fixes #3022 

Stubs: 
- list_cameras might return the list [0], so it could be a list of int. 
- remove the default value from the Camera.init device parameter. It's not present in the _camera backends.

Segfault:
- 4 years ago I forgot to error check the function :(

### DIFF
--- a/src_c/_camera.c
+++ b/src_c/_camera.c
@@ -1886,6 +1886,9 @@ camera_init(pgCameraObject *self, PyObject *arg, PyObject *kwargs)
 
     /* needs to be freed with PyMem_Free later */
     dev_name = PyUnicode_AsWideCharString(name_obj, NULL);
+    if (dev_name == NULL) {
+        return -1;
+    }
 
     p = windows_device_from_name(dev_name);
 


buildconfig/stubs/pygame/camera.pyi
```diff
@@ -7,7 +7,7 @@ from pygame.typing import IntPoint
 def get_backends() -> list[str]: ...
 def init(backend: str | None = None) -> None: ...
 def quit() -> None: ...
-def list_cameras() -> list[str]: ...
+def list_cameras() -> list[str | int]: ...
```
ankith26: I wonder if `list[str] | list[int]` is better here? unless `str` and `int` are allowed to mix, which I don't think is the case?

Starbuck5: You're right that it would be more accurate, force pushed.

ankith26 (approved):
LGTM, thanks!

I trust that you have tested the windows segfault fix, as I didn't verify it.

ankith26:
Some general notes that are related to this PR but could/should be future PRs?
- I think having a default camera makes sense. if no arguments are passed maybe all implementations must just pick `list_cameras()[0]`, only erroring if it's empty.
- Should this be somehow unit-tested for? `camera` doesn't have any tests at all. I get that it is something that's hard to test, but even some light testing for API consistency is better than having no tests.

Starbuck5:
SDL has default audio devices, those make sense. Even though a computer could lack speakers/microphones. So I guesssss that would make sense.

We'll take a fresh look at the whole thing with SDL3's camera API. Same with testing.

But those are future things.

Starbuck5:
Yes, I did test it with my webcam.

PR #3637, opened 2025-11-05
Add Event Watchers and Filters

SuperRedingBros:
Adds `pygame.event.add_event_watcher`, `pygame.event.remove_event_watcher`, `pygame.event.add_event_filter`, and `pygame.event.remove_event_filter` as loose wrappers around the corresponding SDL functions as a potential new way to handle events.

This has two primary rationales.
Firstly, to allow using an event handler paradigm if desired while having no impact on the standard event loop structure.
As this PR does nothing without explicitly opting into it opens up more options while not having any impact on other methods of handling events.

Secondarily, event watchers can be used as an effective solution for the quirk of the Windows version of SDL of the main thread freezing while the OS is sending resize or window movement events. This allows for more advanced users who may be impacted by the issue to resolve it without having to mess with the windows APIs directly. An example of the issue from upstream pygame can be found [here](https://github.com/pygame/pygame/issues/4407). 

A simple example that will print out all events that are added to the queue could be done like this:
```python
@pygame.event.add_event_watcher
def event_logger(ev):
     print(ev)
```

and an example of a filter that will block only left click events:
```python
@pygame.event.add_event_filter
def left_click_filter(ev):
     if ev.type == pygame.MOUSEBUTTONDOWN and ev.button == BUTTON_LEFT:
         return True # Returns true to filter out the event
     return False
```

I have also created a more complex demo implementation showcasing the event watchers as well as demonstrating both the windows issue and an example solution for the issue [here](https://github.com/user-attachments/files/23347341/EventWatcherDemo.py).


SuperRedingBros:
> * Ideally it would have been better if you opened a new issue to discuss the API before implementing it.

Fair enough, I'll try to keep that in mind for the future. Should I make a new issue to discuss this there?
> * SDL3, which is what we are porting pygame-ce to, has the new [main callbacks](https://wiki.libsdl.org/SDL3/README-main-functions) API. Ideally, any new API we add must follow SDL3 style/behaviour even if implemented in the current SDL2 codebase.

I do remember reading that somewhere, would you suggest making the event watchers behave more like SDL_AppEvent?
> * For this PR, I am not opposed to adding event callbacks but I believe there should be only one thing added, and that should be equivalent to SDL3's `SDL_AppEvent`. Adding both watchers and filters is exposing too much details to the users and is potentially more confusing IMHO.

Also pretty fair, I mostly just added the filters because I considered them an adjacent feature that I figured I would add while I was adding the event watchers. If you think that they are making things more confusing I have no problem with removing them.

ankith26:
> would you suggest making the event watchers behave more like SDL_AppEvent?

yes that is what I would personally prefer.

> Should I make a new issue to discuss this there?

That would be appreciated, I would also like to hear other contributors opinions on this API.

aatle:
A registerable callback like `app_event` would be nice, avoiding some boilerplate.
These are pretty large changes though, as it alters the main game loop that usually gives users all of the control; maybe it should be added alongside other large API changes (such as the rest of the main callbacks API).

Also see another large event API change proposal, #2759.

SuperRedingBros:
> These are pretty large changes though, as it alters the main game loop that usually gives users all of the control; maybe it should be added alongside other large API changes (such as the rest of the main callbacks API).

That's actually why I implemented it the way I did, it allows for the normal event loop to be used as-is, while offering an alternative completely optionally. My goal for this particular bit of API was to add the functionality without having any impact on anything else.

> Also see another large event API change proposal, #2759.

That's actually pretty cool, maybe the function's signature could support specifying event types to target? 
i.e.
```python
@pygame.add_event_watcher(KeydownEvent)
def keydownListener(event: KeydownEvent):
      ... # use the event somehow
```

SuperRedingBros:
For the time being, I think that a simple general-purpose interface to the underlying SDL function is best. 
Then a higher-level interface written in python that implements more of the suggestions could potentially be provided as a separate module or in later PR.
A possible implementation of this module could be [this gist](https://gist.github.com/SuperRedingBros/7cd452529841d6a5d6ba80d16de8983d)

PR #3639, opened 2025-11-08
Use SDL_AllocFormat instead of creating it manually

marcin-serwin:
According to the docs, `SDL_PixelFormat` is a read-only structure. Creating it manually leaves out some important fields like `format` and `next` pointer to be undefined.

--------------------------------------------------------------
EDIT Starbuck5:
This was needed for SDL2 compat support to pass certain tests, now it's unclear whether it's needed.

### DIFF
--- a/src_c/surface.c
+++ b/src_c/surface.c
@@ -1844,9 +1844,8 @@ surf_convert(pgSurfaceObject *self, PyObject *args)
              */
             int bpp = 0;
             SDL_Palette *palette = SDL_AllocPalette(default_palette_size);
-            SDL_PixelFormat format;
+            Uint32 format_enum = 0;
 
-            memcpy(&format, surf->format, sizeof(format));
             if (pg_IntFromObj(argobject, &bpp)) {
                 Uint32 Rmask, Gmask, Bmask, Amask;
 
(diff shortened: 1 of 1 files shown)


marcin-serwin:
I've run into the same issue in `surf_convert`. @ankith26 could you re-review?

src_c/surface.c
```diff
@@ -1904,66 +1903,49 @@ surf_convert(pgSurfaceObject *self, PyObject *args)
...
                 }
-                mask =
-                    format.Rmask | format.Gmask | format.Bmask | format.Amask;
-                for (bpp = 0; bpp < 32; ++bpp) {
-                    if (!(mask >> bpp)) {
-                        break;
-                    }
-                }
+                format_enum = SDL_MasksToPixelFormatEnum(bpp, Rmask, Gmask,
+                                                         Bmask, Amask);
```
ankith26, line 1929: I think that the `bpp` isn't actually defined to the correct value in this codepath, the for loop that you just removed was doing that. I feel like that bit should be re-added.

marcin-serwin: You're right, restored.

src_c/surface.c
```diff
@@ -1904,66 +1903,62 @@ surf_convert(pgSurfaceObject *self, PyObject *args)
...
-                 * then SDL_ConvertSurface checks that the palette is not
-                 * empty-- that at least one entry is not black.
-                 */
-                format.palette = NULL;
+            SDL_PixelFormat *format = SDL_AllocFormat(format_enum);
+            if (!format) {
+                SDL_SetError(
+                    "Failed to allocate format for converted surface");
+                SDL_FreePalette(palette);
+                return NULL;
```
ankith26: Minor nit: A python exception should be set here, not an SDL exception. It's usually done with the `RAISE` macro, check out other fail returns in this function.

marcin-serwin: Done.

ankith26 (approved):
LGTm, thanks again! 🎉

Starbuck5 (commented):
I feel like I wanted to do stuff like this when porting SDL3 but I (probably with Ankith's encouragement), decided not to, in order to just leave the old code functioning how it is.

What is the justification for changing this code?

marcin-serwin:
> What is the justification for changing this code?

Since https://github.com/libsdl-org/sdl2-compat/pull/521 sdl2-compat relies on the `format` and `next` fields to be set correctly. The old code doesn't set them which causes some of the surface tests (e.g., `GeneralSurfaceTests.test_convert_palettize`, `SurfaceSelfBlitTest.test_colorkey`) to fail with the new 2.32.58 version.

ankith26:
It is interesting to note that the PR has been (partially) reverted as of https://github.com/libsdl-org/sdl2-compat/pull/545 which was merged just a few days ago, which may change things.

However I think this PR is still worth merging because it is making the code more robust.

PR #3640, opened 2025-11-09
docs: `math.Vector2|3` zero vector normalisation raises ValueError

elliot-100:
This PR documents that `ValueError` is raised when attempting to normalize a zero vector with `Vector2|3.normalize()`, `Vector2|3.normalize_ip()`.

Wording is copied from existing text for `Vector2|3.scale_to_length()`.


oddbookworm (approved):
LGTM! Thanks for the update! We've had a few complaints about this over the past couple of years that I can remember, but I never noticed that it's not documented, so I'm in favor of this docs addition. 🎉

PR #3641, opened 2025-11-09
Add more system info

AntonisPylos:
This PR adds **get_cpu_cores** and **get_platform** functions to the system module, providing basic system information to reduce the need for external libraries.


Starbuck5:
I agree with Ankith.

AntonisPylos:
I think wrapping GetCPUCount and GetPlatform could be a good addition to the system module, since they both provide similar functionality to the other system functions. I trust your judgment, so I’ll close this PR.

PR #3647, opened 2025-11-20
Remove unavailable and deprecated name_forindex function from documentation

BatteRaquette581:
`name_forindex(index)` is deprecated, but now unavailable in `pygame._sdl2.controller` (it now resides in `pygame._sdl2.controller_old`), and it seems like it should not be used.


oddbookworm:
I would like to get @zoldalma999 to chime in here. Is a better alternative adding this function into the new controller module?

PR #3657, opened 2025-12-09
SDL3 `_audio` module

Starbuck5:
For #3581 mega-issue.

To fully implement next generation mixer, I realized I needed parts of SDL3's audio subsystem to be exposed. Not for a ton of API surface, but it is necessary for full implementation and necessary for a half-decent compatibility layer with the old mixer API. And it will need to be written anyway for full SDL3 support.

So I spent some time and wrote a whole new module, SDL3 only, starting from the type stubs mocked up in #3581.

I experimented with a hybrid approach with C level PyObjects as well as Python level classes. I'm not sure I'm in love with the exact way I did it, but I think it definitely shows promise as an approach. Some of the classes are only in Python, some are wrappers over low-level calls and "state objects". I really like how convenient it is to add error checking or stuff like the AudioDevice weakref dict at the Python level.

Example script:
```py
import pygame
from pygame import _audio as audio
pygame.audio = audio

WAV_PATH = r"C:\Users\charl\Desktop\pygame-ce\examples\data\house_lo.wav"

audio.init()

print("==AUDIO METADATA==")
print("Available drivers:", pygame.audio.get_drivers())
print("Current driver:", pygame.audio.get_current_driver())

print("Playback devices--")

devices = pygame.audio.get_playback_devices()
for device in devices:
    print(device.name)

print("==PLAYING AUDIO==")

print("Opening default playback device--")
logical_device = pygame.audio.DEFAULT_PLAYBACK_DEVICE.open()
print(logical_device.name)

print("Loading audio--")
audio_spec, audio_bytes = pygame.audio.load_wav(WAV_PATH)
print(audio_spec, "|", len(audio_bytes), "bytes")

print("Creating stream")
stream = pygame.audio.AudioStream(audio_spec, audio_spec)
print(stream)
stream.frequency_ratio = 1.4
stream.gain = 0.4
stream.bind(logical_device)
stream.put_data(audio_bytes)
stream.flush()

# It can change after binding to a device
print("Stream output spec now:", stream.dst_spec)

# Wait for stream to be exhausted
while stream.num_queued_bytes:
    pygame.time.wait(100)

print("Done.")
```

Next steps (future PRs):
- Docs
- Tests
- Complete implementation (quit support if we want it, callbacks, other unimplemented API)
- Next gen mixer
- Old mixer compat layer on next gen mixer


Starbuck5:
@coderabbitai review

src_c/_base_audio.c
```diff
@@ -0,0 +1,1126 @@
...
+static void
+adevice_state_dealloc(PGAudioDeviceStateObject *self)
+{
+    SDL_CloseAudioDevice(self->devid);
+    PyObject_GC_UnTrack(self);
+    PyTypeObject *tp = Py_TYPE(self);
+    freefunc free = PyType_GetSlot(tp, Py_tp_free);
+    free(self);
+    Py_DECREF(tp);
+}
```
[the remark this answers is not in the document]
Starbuck5: FWIW I did test that this actually works, but it's easy enough to check whether it's physical or not, and that protects us if SDL changes something in the future.

src_c/meson.build
```diff
@@ -420,10 +420,21 @@ if sdl_ttf_dep.found()
...
+if sdl_api == 3
+    base_audio = py.extension_module(
+        '_base_audio',
+        '_base_audio.c',
+        c_args: warnings_error,
+        dependencies: pg_base_deps,
+        install: true,
+        subdir: pg,
+    )
+endif
```
ankith26, line 433: minor nit: this should be a part of the `sdl_mixer_dep.found()` if block.

src_c/_base_audio.c
```diff
@@ -0,0 +1,1137 @@
...
+    SDL_AudioDeviceID devid = ((PGAudioDeviceStateObject *)arg)->devid;
+    if (SDL_IsAudioDevicePlayback(devid)) {
+        Py_RETURN_TRUE;
+    }
+    Py_RETURN_FALSE;
+}
+
+static PyObject *
+pg_audio_get_audio_device_name(PyObject *module, PyObject *const *args,
+                               Py_ssize_t nargs)
```
ankith26: if this always expects exactly one arg why not just use `METH_O` convention? As of now this just silently ignores extra args, and must have ub for no arg?

I also observe other places in this file where argument checking is non existent or non sufficient. I get why it is the case though, these functions are internal. But still, better error checking does no harm and catches any current/future issues in our python level more gracefully rather than segfaulting

buildconfig/stubs/pygame/_audio.pyi
```diff
@@ -0,0 +1,164 @@
...
+    def is_signed(self) -> bool: ...
+    @property
+    def is_unsigned(self) -> bool: ...
+    @property
+    def name(self) -> str: ...
+    @property
+    def silence_value(self) -> bytes: ...
+    def __index__(self) -> int: ...
+    def __repr__(self) -> str: ...
+
```
ankith26, line 57: I think implementing `__eq__` would be handy too, so users could say `spec.format == U8` or something like that.

buildconfig/stubs/pygame/_audio.pyi
```diff
@@ -0,0 +1,164 @@
...
+S16BE: AudioFormat
+S32LE: AudioFormat
+S32BE: AudioFormat
+F32LE: AudioFormat
+F32BE: AudioFormat
+S16: AudioFormat
+S32: AudioFormat
+F32: AudioFormat
+
+class AudioSpec:
```
ankith26, line 71: this could be a `dataclass`?

ankith26 (commented):
Thanks for taking this up and working on this 🎉 , I'm sure it was a lot of work 😅 

I've given everything a quick scroll and left a few minor reviews, nothing important.

One more point of discussion I wanted to bring up since you mentioned implementing "next gen mixer" on top of this. I feel like having 3 layers of implementation is overcomplicating things to the end user. We don't want to create another draw/gfxdraw situation (this is not the best analogy but you get the idea). Ideally, the python level `audio` implementation should be easy enough to use that it is essentially the "next gen mixer". Maybe this means that we should consider adding more convenience methods/classes to the current `audio` implementation. Of course, old mixer compat could be the sole 3rd layer, but that would just be a compat layer with no new features added.

buildconfig/stubs/pygame/_audio.pyi
Starbuck5: This is handled by Python’s default identity comparison, because AudioFormats are singletons.

buildconfig/stubs/pygame/_audio.pyi
Starbuck5: I believe in the first draft it was, but when I actually went to write it I wanted understandable full control over all of the runtime consequences. Like I made the fields immutable. So for the typing and implementation I don’t see an advantage to making it into a dataclass.

Starbuck5:
> One more point of discussion I wanted to bring up since you mentioned implementing "next gen mixer" on top of this. I feel like having 3 layers of implementation is overcomplicating things to the end user. We don't want to create another draw/gfxdraw situation (this is not the best analogy but you get the idea). Ideally, the python level audio implementation should be easy enough to use that it is essentially the "next gen mixer". Maybe this means that we should consider adding more convenience methods/classes to the current audio implementation. Of course, old mixer compat could be the sole 3rd layer, but that would just be a compat layer with no new features added

Good points to discuss. I believe I did say that (sounds like me), and I’m sorry to mislead, but my plan for the next generation mixer is to build it on top of SDL3_mixer, not on top of this. The audio module here is supposed to be a faithful port of SDL3’s audio module (built in). I worked up this PR so that a future mixer PR would be able to implement stuff like accepting an AudioDevice or AudioSpec parameter. This is why I didn’t put the meson build code for this under mixer-dep-found, because for this it doesn’t matter if it was found.

In terms of module proliferation pygame-ce already has an audio module, _sdl2.audio, that gives some of this enhanced functionality, like device enumeration, voice recording or streaming audio. And even though it’s completely undocumented some people still found it and used it, as we saw with a recent new contributor PR to fix something with it.

Advantages of having an audio module rather than limping into mixer:
- Builds could exist without SDL_mixer and a lot of functionality (this module’s) would still work
- It’s closer to SDL (when in doubt, follow SDL API)
- It delineates nicely imo? Streams and devices vs tracks and effects.

With regards to next generation vs existing mixer APIs I was thinking they would be in the same namespace to the user, probably just implemented in a separate file and imported.

ankith26:
oh yeah fair enough, I always confuse audio and mixer mb. Anyways, to elaborate and correct myself, what I propose is:
- `_base_audio.c` for wrapping SDL3 audio as closely as possible (as you have already implemented)
- `_base_mixer.c` for wrapping `SDL3_mixer` as closely as possible
- `pygame.audio` to be a unified interface for accessing everything (tracks/effects/devices/streams/whatever) audio related (extending the current PR in the future by encorporating mixer features in the same audio python code).

Yes, I see that this implies that a `SDL3_mixer`-less build would have an incomplete `pygame.audio` implementation but I think this is an edge case we need not worry about too much given we can assume SDL3_mixer is almost always available.

src_c/_base_audio.c
Starbuck5: Yep that should be METH_O. (It’s one of the first functions I wrote and I didn’t adjust it to match the later style). I’m away from my computer for a few days so I will make the change when I am back or you/anyone could push that.

I intentionally omitted all argument error handling, although I used safe/stable ABI operations for all Pyobjects once received. The idea is that if checks can be moved into the Python VM that JIT Python might be smart enough to eliminate them and lead to higher performance. Not that performance is particularly important for this module. 

Maybe a simple and cheap check to do would be to check the number of args is as expected?

Starbuck5:
The future module organization doesn’t need to be settled in this PR, this code would work for either strategy.

I would like to bring some of this discussion over to #3581 though.

src_c/_base_audio.c
Starbuck5: I've switched to METH_O in two places that were fastcall with 1 arg.

src_c/_base_audio.c
```diff
@@ -0,0 +1,1137 @@
...
+
+    state->audio_stream_state_type =
+        PyType_FromModuleAndSpec(module, &astream_state_spec, NULL);
+    if (state->audio_stream_state_type == NULL) {
+        return -1;
+    }
+    if (PyModule_AddType(module,
+                         (PyTypeObject *)state->audio_stream_state_type) < 0) {
+        return -1;
+    }
```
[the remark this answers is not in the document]
Starbuck5: It does not leak, because the reference is Py_CLEAR-ed in pg_audio_clear, which is called by pg_audio_free, which is automatically called on error as the module object gets cleaned up.

src_c/_base_audio.c
```diff
@@ -0,0 +1,1137 @@
...
+        (PGAudioDeviceStateObject *)adevice_state_type->tp_alloc(
+            adevice_state_type, 0);
+    if (device == NULL) {
+        SDL_CloseAudioDevice(logical_id);
+        return NULL;
+    }
+    device->devid = logical_id;
+
+    return (PyObject *)device;
+}
```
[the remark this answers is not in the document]
Starbuck5: This is part of how it's written, it's the responsibility of the Python layer above to pass correct arguments. This is not publicly exposed.

src_c/_base_audio.c
```diff
@@ -0,0 +1,1141 @@
...
+    audio_state *state = GET_STATE(module);
+    if (state->audio_initialized) {
+        SDL_QuitSubSystem(SDL_INIT_AUDIO);
+        state->audio_initialized = false;
+    }
+    Py_RETURN_NONE;
+}
+
+static PyObject *
+pg_audio_get_init(PyObject *module, PyObject *_null)
```
damusss, line 755: `audio.get_init` could become misleading in a scenario where another module initializes the audio submodule (meaning `state->audio_initialized` is still false) and a function like `audio.get_current_buffer` is used since `AUDIO_INIT_CHECK` will fail erroring "audio not initialized" while `audio.get_init` returns True.

Therefore either the macro should check for the subsystem being initialized before erroring (updating the state) or get_init should only return the state of this module, depends on how much the internals rely on this module's specific init function.

ankith26 (approved):
I am approving this after a quick review, I agree that it would be better to get this PR in now and make changes later as needed instead of stalling this PR.

 Thanks a lot again starbuck, for working on this mammoth task! 🎉

damusss (approved):
I had already read through it all and liked it, no reason to stall it. It's currently private anyways. I like the hybrid C and Python approach especially.

Thank you! 💯

PR #3658, opened 2025-12-10
Update pythoncapi_compat.h

Starbuck5:
Just to keep this file up to date. It's been 7 months since last update.


Starbuck5:
Since this is already approved, lets have that be dealt with in another PR.

PR #3659, opened 2025-12-10
Move SDL hint so it always triggers on import

Starbuck5:
Rather than relying on pygame.init() being called, since not all workflows would do that.

This just moves where the code added in https://github.com/pygame-community/pygame-ce/pull/3654 triggers. Thanks to reviewers on that for reviewing it so quickly, I'm sort of embarrassed to already want to change it.

But I was thinking about what Ankith said in his review, that we wouldn't want SDL to ever override signal handlers from Python. And my previous PR didn't do that. It just turned it off if one happened to call pygame.init(). Like in my audio example script in my other recent PR I don't call pygame.init(), I just call audio.init().

### DIFF
--- a/src_c/base.c
+++ b/src_c/base.c
@@ -217,15 +217,6 @@ pg_init(PyObject *self, PyObject *_null)
         /* IMPPREFIX "_sdl2.controller", Is this required? Comment for now*/
         NULL};
 
-#if SDL_VERSION_ATLEAST(3, 0, 0)
-    // In SDL3, specify that signal handlers should not be enabled.
-    // By default, unlike SDL2, these signal handlers convert into QUIT
-    // events. However, if QUIT events / events aren't being handled,
-    // this leaves people unable to quit their script. Plus it's different
-    // than SDL2 behavior.
-    SDL_SetHint(SDL_HINT_NO_SIGNAL_HANDLERS, "1");
-#endif
-
     /*nice to initialize timer, so startup time will reflec pg_init() time*/
 #if defined(WITH_THREAD) && !defined(MS_WIN32) && defined(SDL_INIT_EVENTTHREAD)
     pg_sdl_was_init = PG_InitSubSystem(SDL_INIT_EVENTTHREAD | PG_INIT_TIMER |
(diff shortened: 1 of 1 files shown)


src_c/base.c
```diff
@@ -2130,6 +2121,15 @@ MODINIT_DEFINE(base)
...
     PyObject *module, *apiobj;
 
+#if SDL_VERSION_ATLEAST(3, 0, 0)
+    // In SDL3, specify that signal handlers should not be enabled.
+    // By default, unlike SDL2, these signal handlers convert into QUIT
+    // events. However, if QUIT events / events aren't being handled,
+    // this leaves people unable to quit their script. Plus it's different
+    // than SDL2 behavior.
+    SDL_SetHint(SDL_HINT_NO_SIGNAL_HANDLERS, "1");
+#endif
```
ankith26, line 2131: Now that this logic has been moved here, have you tested that something like say, quit+init doesn't reset this hint? I feel like something like this must be done at one of the init functions, say `display.init` or the internal `event` init function.

Starbuck5: Dang I didn't test that because I expected hints to be global. But no, I guess quitting SDL resets them.

It needs to get called on every possible event path that could initialize event? mixer.init, display.init, others?

Starbuck5: Test program

```py
import pygame

pygame.display.init()
clock = pygame.time.Clock()
running = True
try:
    while running:
            if event.type == pygame.QUIT:
                print("RECEIVED QUIT EVENT OMG 1")
                running = False

        clock.tick(60)
except KeyboardInterrupt:
    print("KEYBOARD INTERRUPTED 1")

pygame.display.quit()



pygame.init()
print("initialized 1")
clock = pygame.time.Clock()
running = True

try:
    while running:
            if event.type == pygame.QUIT:
                print("RECEIVED QUIT EVENT OMG 1")
                running = False

        clock.tick(60)
except KeyboardInterrupt:
    print("KEYBOARD INTERRUPTED 1")

pygame.quit()

pygame.init()
print("initialized 2")

try:
    while running:
            if event.type == pygame.QUIT:
                print("RECEIVED QUIT EVENT OMG 2")
                running = False

        clock.tick(60)
except KeyboardInterrupt:
    print("KEYBOARD INTERRUPTED 2")

pygame.quit()
```

ankith26: I think the hint should be set in `pgEvent_AutoInit`, any other init function that needs event to be init must call `pg_mod_autoinit(IMPPREFIX "event")` (like `display.init` already does it)

Starbuck5: Some other SDL things initialize event internally, so it can be potentially non obvious. Like I ran into this problem with SDL_INIT_AUDIO.

zoldalma999: As far as I can tell, only a full SDL_Quit resets the hints, so I tried setting it right after quitting (in base.c _pg_quit, right after calling pg_atexit_quit), and it worked for me with the test script above.

PR #3666, opened 2025-12-21
Use TemporaryDirectory in unit tests for temp dirs

ankith26:
When I was testing out the [pygame-ce AUR package](https://aur.archlinux.org/packages/python-pygame-ce) I found out that some test fails arise because our test code assumes directories like `examples/data` and `tests/fixtures/fonts` are writeable. In a package manager managed install we cannot assume this, and in general writing to our install location is a bad idea.

I replaced all instances where we were using these directories with `tempfile.TemporaryDirectory` API.

As a nice bonus, the context manager API of `TemporaryDirectory` ensures the directory is cleaned up at exit so we need not do it explicitly

### DIFF
--- a/test/font_test.py
+++ b/test/font_test.py
@@ -5,6 +5,7 @@ import pathlib
 import platform
 import sys
 import unittest
+from tempfile import TemporaryDirectory
 
 import pygame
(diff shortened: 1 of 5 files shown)


Starbuck5 (approved):
Cool!

PR #3673, opened 2026-01-05
Update meson, cython, fix meson warning

Starbuck5:
I ran the test suite locally and everything checked out.

The warning:
```
  ..\meson_options.txt:27: WARNING: Project does not target a minimum version but uses feature deprecated since '1.1.0': "boolean option" keyword argument "value" of type str. use a boolean, not a string
  ..\meson_options.txt:31: WARNING: Project does not target a minimum version but uses feature deprecated since '1.1.0': "boolean option" keyword argument "value" of type str. use a boolean, not a string
  ..\meson_options.txt:35: WARNING: Project does not target a minimum version but uses feature deprecated since '1.1.0': "boolean option" keyword argument "value" of type str. use a boolean, not a string
```

There are also 2 other mesons warnings, one about using fs.relative_to (a relatively recent feature) without declaring a minimum version, and the other about using the built in warning compilation options, but I have not touched those in this PR.

### DIFF
--- a/meson_options.txt
+++ b/meson_options.txt
@@ -24,15 +24,15 @@ option('midi', type: 'feature', value: 'enabled')
 # Controls whether to make a "stripped" pygame install. Enabling this disables
 # the bundling of docs/examples/tests/stubs in the wheels.
 # The default behaviour is to bundle all of these.
-option('stripped', type: 'boolean', value: 'false')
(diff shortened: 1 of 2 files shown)


ankith26 (approved):
LGTM, thanks! 🎉

PR #3675, opened 2026-01-11
Move Color docs to stub file

Starbuck5:
For #2757 

- The "Note" in the Color main doc about color formats wouldn't format right so I had to spaces between each, it changes how it displays slightly but doesn't look broken.
- I had some problem with the type-ignored unified docs for overloaded methods, specifically overloaded classmethods. Stubtest complained unless I removed the "\@\classmethod" on the unified ones, but in the doc it still generates as a "classmethod" luckily.


aatle (commented):
The `@classmethod` decorator should still be added to the implementation definition; from testing, one can just put the `# type: ignore` on that `@classmethod` line instead of the `def` line to ignore the error.

Also, I think that the return type should still be typed for the implementation definition (but not `*args`).

Starbuck5:
> The @\classmethod decorator should still be added to the implementation definition; from testing, one can just put the # type: ignore on that @\classmethod line instead of the def line to ignore the error.

Yes, good point. Changed.

> Also, I think that the return type should still be typed for the implementation definition (but not *args).

I did not do that because I followed zoldalma's lead, the existing ones are not like that. Like in surface.pyi. And it makes sense, it's a type-ignored line  - why type-annotate it?

aatle:
The line is indeed type ignored, but that does not mean that all of the rest of the typing should be thrown out.
Implementation definitions are generally still typed. If the "disallow untyped definitions" was turned on, it would flag an untyped implementation definition still.
For our stubs, the arguments are a hassle to do so `*args` works fine, but the return type is generally same across overloads so it's not bad to just put in.
Surface stubs don't have the return type, but the mouse and key stubs do, so there is not a consistent precedent yet.

Starbuck5:
> If the "disallow untyped definitions" was turned on, it would flag an untyped implementation definition still.

I was curious, so I tested this, and this is not true. The below script has 2 untyped definitions but only a single error, because type ignoring the definition causes mypy to ignore it.

image

aatle:
By that, I meant that generally, programmers should type implementation definitions in code.
Using a type: ignore specifically for the stub implementation definition warning does not mean that all other typing at it is null. Typically, it is narrowed to `# type: ignore[misc]` for example. Also, mypy still recognizes and registers the return type of a type ignored definition.

However, with stubs, the overload implementation definition is not visible. The one main reason I'd prefer to have it typed is for consistency, as all other things in the stubs are typed.

Starbuck5:
Ok. I have added the return types as requested.

aatle:
Is a `# type: ignore` still needed on the `def` overload implementation definition line when there is an ignore on the classmethod line?

Starbuck5:
>  Is a # type: ignore still needed on the def overload implementation definition line when there is an ignore on the classmethod line?

~~I gave this a test, and apparently not.~~

NVM, fails stubcheck.

PR #3676, opened 2026-01-11
Remove _sdl2.video.messagebox

Starbuck5:
pygame.display.message_box has been around to replace it since pygame-ce 2.4.0


Starbuck5:
The vision is to publicize features out of _sdl2 into permanent API locations with rewrites out of Cython. This is what happened with the Window module, this is what should happen with the SDL rendering API stuff.

Since the message box functionality has been available outside of _sdl2 for some time now, it should be removed from _sdl2.

Precedent PR from when Window was migrated: https://github.com/pygame-community/pygame-ce/pull/2294

PR #3680, opened 2026-01-16
Disable mouse rect test on null backend

marcin-serwin:
The SDL_SetWindowMouseRect function returns false on unsupported backends since <https://github.com/libsdl-org/SDL/commit/2a873be9cdb35972dc8bacd12653aafc02e835de>.

### DIFF
--- a/test/window_test.py
+++ b/test/window_test.py
@@ -114,8 +114,9 @@ class WindowTypeTest(unittest.TestCase):
         self.assertFalse(self.win.always_on_top)
 
     @unittest.skipIf(
-        SDL < (2, 0, 18),
-        "requires SDL 2.0.18+",
+        SDL < (2, 0, 18)
+        or os.environ.get("SDL_VIDEODRIVER") == pygame.NULL_VIDEODRIVER,
+        "requires SDL 2.0.18+ and SDL_VIDEODRIVER to be a non-null value",
     )
     def test_mouse_rect(self):
         self.win.mouse_rect = None


marcin-serwin:
> This is something I had already noticed in my SDL3 porting PRs (and already implemented a similar fix sometime back at #3562). I'm curious how you found this, I'm presuming this fail come up with sdl2-compat?

Yes, I help with maintaining SDL libraries in nixpkgs and this came up after the latest sdl3/sdl2-compat bump.

Since you're also working with SDL3, there are two more tests that started failing after the bump:
```
FAIL: test_aafunctions_depth_segfault (pygame.tests.draw_test.DrawModuleTest.test_aafunctions_depth_segfault)
Ensure future commits don't break the segfault fixed by pull request
  File "/nix/store/1iv40fycqb9d1v8lcw4dm8v48mn9bds1-python3.14-pygame-ce-2.5.6/lib/python3.14/site-packages/pygame/tests/draw_test.py", line 7540, in test_aafunctions_depth_segfault
    self.assertEqual(surf.get_at(pixel), pixel_colors_16[i])
    ~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
AssertionError: Color(0, 0, 214, 255) != Color(0, 0, 213, 255)
```
```
FAIL: test_fill_rle (pygame.tests.surface_test.SurfaceTypeTest.test_fill_rle)
Test RLEACCEL flag with fill()
  File "/nix/store/1iv40fycqb9d1v8lcw4dm8v48mn9bds1-python3.14-pygame-ce-2.5.6/lib/python3.14/site-packages/pygame/tests/surface_test.py", line 303, in test_fill_rle
    self.assertTrue(blit_surf.get_flags() & pygame.RLEACCEL)
    ~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
AssertionError: 0 is not true
```
I haven't had time to look into it yet, but perhaps you've also run across them and know the cause.

ankith26:
I noticed these fails too when I was testing out something in an arch box. I didn't investigate it yet but I plan to do it when I get the time. All I can tell is that its something that broke very recently, as I do not get these issues on my ubuntu 25.10 machine which is on an older SDL3 and sdl2-compat.

ankith26:
For the `test_aafunctions_depth_segfault` fail, I believe it is a trivial fail so modifying the test to allow for small delta variations in the pixel should probably be an acceptable fix.

I looked in into the `test_fill_rle` fail and came up with a C level SDL MRE and reported the issue over at https://github.com/libsdl-org/sdl2-compat/issues/575

Merging this PR for now. Once there is a resolution on the SDL side I will update both tests accordingly in a new PR.

Issue #3693, opened 2026-02-01
`get_grab()` not working


robotsnh:
**Environment:** MacOS Tahoe 26.2

- **Relevant hardware** I'm on a Macbook Air from 2020, though I doubt that that's relevant.

**Current behavior:**  All tests run fine up to this one.

When I try to run `python3 dev.py test`, I get an error about a test not working. The traceback is provided below.

**Expected behavior:** I expected at least one test to fail, so this didn't come as much of a surprise.

**Screenshots**

If applicable, copy and paste screenshots to help explain your problem.

**Steps to reproduce:**

Please explain the steps required to duplicate the issue, especially if you are able to provide a sample application.
If the bug is caused by a specific file (image, font, sound, level ...etc) please upload it as an attachment.

1. Clone the repo
2. Run `python3 dev.py test`

**Stack trace/error output/other error logs**  

Here's the error traceback:

```
loading pygame.tests.window_test
..............................................................................................................................................................sssssss.................ss..................sF..............ss..................................................................................................................................................................................................................................................................................F.................F.................F.................................................................................................................F................................................................................................................................................................................................................................................................................s.....s.........................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................F.s...F.................F......................................................................................................................................................................E.........................................................s......................................................s.......s..........................................s.............s.s.s..........................F............F..F........................................................................................
======================================================================
FAIL: test_toggle_fullscreen (pygame.tests.display_test.DisplayModuleTest.test_toggle_fullscreen)
Test for toggle fullscreen
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/Library/Frameworks/Python.framework/Versions/3.14/lib/python3.14/site-packages/pygame/tests/display_test.py", line 697, in test_toggle_fullscreen
    self.assertEqual(
    ~~~~~~~~~~~~~~~~^
        (test_surf.get_width(), test_surf.get_height()), width_height
        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    )
    ^
AssertionError: Tuples differ: (800, 600) != (640, 480)

First differing element 0:
800
640

- (800, 600)
+ (640, 480)

Stderr:
[...]

PR #3694, opened 2026-02-01
Fixed an error in `test/display_test.py`

robotsnh:
Updated line 698 to use `test_surf.width` and `test_surf.height` instead of `test_surf.get_width()` and `test_surf.get_height()`, respectively.

When built, this works but the same file creates another error, ~300 lines later.

Fixes issue #3693

### DIFF
--- a/test/display_test.py
+++ b/test/display_test.py
@@ -695,7 +695,7 @@ class DisplayModuleTest(unittest.TestCase):
             # if not original width/height should be preserved
             else:
                 self.assertEqual(
-                    (test_surf.get_width(), test_surf.get_height()), width_height
+                    (test_surf.width, test_surf.height), width_height
                 )
 
     @unittest.skipIf(is_wayland, "not supported on wayland")


Starbuck5:
Hello,

Width and height are the same as get_width and get_height, so I don't see how this could impact the behavior you pulled up in your issue in any way. So I'm going to close this PR.

On a completely separate note, since you said you're on MacOS, you could potentially try out wheels from https://github.com/pygame-community/pygame-ce/pull/3625 ? Download is available under pygame-wheels-macos on this page: https://github.com/pygame-community/pygame-ce/actions/runs/20690590039?pr=3625. This PR switches us from having separate binaries for ARM vs x64 Macs to one unified binary, I've been trying to get people to test it out.

robotsnh:
I will try to find the root cause of my error and post reports in #3693.

robotsnh:
@Starbuck5 I will try to test the wheel if I can find the time.

robotsnh:
@Starbuck5 Which of these (presumably architecture-agnostic) wheels should I test:   
cp310  
cp311  
cp312   
cp313  
cp314  
pp311  
Or should I test the non-architecture-agnostic ones?

ankith26:
@robotsnh you only need to install one wheel for the python version you use, for instance if you are on Python 3.14 the corresponding tag would be cp314.
You only need to test the architecture agnostic "universal2" wheel, the name of the file would contain the word universal2 somewhere near the end. Once you have identified the wheel simply do `pip install <path_to_file>`.
The zip file you download should be from the `pygame-wheels-macos` artifact.

I hope that clears it up, if there's still any issues feel free to let me know, would be happy to help! :smile:

robotsnh:
@ankith26 It built, is there anything else I should do to it?

ankith26:
Oh cool, could you run tests (with the command `python -m pygame.tests`) and report the results? You can also try out a few games to see if the install is all right.

Thanks in advance.

robotsnh:
All the tests worked up to where it failed previously, which is documented in #3693. Is that all? (The aliens and drop file tests worked.)

Issue #3682, opened 2026-01-22
The `pygame.draw.aacircle` function cannot draw arcs with a thickness of 2.


Fantastair:
**Environment:**
```
Platform:               Linux-6.12.63+deb13-amd64-x86_64-with-glibc2.41
System:                 Linux
System Version:         #1 SMP PREEMPT_DYNAMIC Debian 6.12.63-1 (2025-12-30)
Processor:                      SSE2: Yes       AVX2: Yes       NEON: No
Architecture:           Bits: 64bit     Linkage: ELF

Python:                 CPython 3.13.5 (main, Jun 25 2025, 18:55:22) [GCC 14.2.0]
GIL Enabled:            True
pygame version:         2.5.6
SDL versions:           Linked: 2.32.10 Compiled: 2.32.10
SDL Mixer versions:     Linked: 2.8.1   Compiled: 2.8.1
SDL Font versions:      Linked: 2.24.0  Compiled: 2.24.0
SDL Image versions:     Linked: 2.8.8   Compiled: 2.8.8
Freetype versions:      Linked: 2.13.3  Compiled: 2.13.3

Display Driver:         Display Not Initialized
Mixer Driver:           Mixer Not Initialized
```

- **Relevant hardware**: NVIDIA GeForce RTX 4060 Laptop GPU (Driver Version: 550.163.01), 2560×1600 high-DPI display, 150 % system scaling.

**Current behavior:**

The `pygame.draw.aacircle` function accepts a `width` parameter. When `width > 0`, it controls the thickness of the arc line. However, I've discovered this function cannot draw arcs with a width of 2. When `width` = 1, it draws an arc with a width of 1; when `width > 1`, it draws an arc with a width of `width + 1`. This means the actual widths that can be drawn are “1, 3, 4, 5, ...”.

Incidentally, I'm using the `pygame.Window` class to manage the window and have set `allow_high_dpi=True` for my high-DPI display, though I don't believe this should affect the issue.

Fantastair:
Here are some additional screenshots.

> I initially thought it was a bug in my own program, so I tried to fix it and discovered this relationship between the thickness of the arc and the `width`.

`bw` refers to the border width. The horizontal and vertical lines were drawn using `pygame.Surface.fill`, while the arc lines were drawn using `pygame.draw.aacircle`.

- `width = 1`
  The arc and straight line share the same thickness, both measuring 1 pixel, with seamless alignment.
[...]

PR #3699, opened 2026-02-03
fix width calculation for `draw.aacircle()`

Fantastair:
For more details, please refer to the discussion in #3682 .

### DIFF
--- a/src_c/draw.c
+++ b/src_c/draw.c
@@ -1011,7 +1011,7 @@ aacircle(PyObject *self, PyObject *args, PyObject *kwargs)
         }
         else {
             draw_circle_xiaolinwu(surf, surf_clip_rect, surf_format, posx,
-                                  posy, radius, width, color, 1, 1, 1, 1,
+                                  posy, radius, width - 1, color, 1, 1, 1, 1,
                                   drawn_area);
         }
     }
@@ -1029,7 +1029,7 @@ aacircle(PyObject *self, PyObject *args, PyObject *kwargs)
         }
         else {
             draw_circle_xiaolinwu(surf, surf_clip_rect, surf_format, posx,
-                                  posy, radius, width, color, top_right,
+                                  posy, radius, width - 1, color, top_right,
                                   top_left, bottom_left, bottom_right,
                                   drawn_area);
         }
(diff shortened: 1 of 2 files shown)


Fantastair:
As for why I ultimately chose to modify the parameter passing on the calling side rather than the underlying implementation, it's because `draw_circle_xiaolinwu()` also handles drawing solid circles. Modifying the width here would cause the center pixels of solid circles to be left empty. Therefore, passing `width - 1` when drawing the circle border in `draw.aacircle()` represents the smallest possible change.

Additionally, I added an explanatory comment inside the ‘draw_circle_xiaolinwu()’ function to serve as a reminder for future developers.

src_c/draw.c
```diff
@@ -3489,6 +3489,9 @@ draw_circle_xiaolinwu(SDL_Surface *surf, SDL_Rect surf_clip_rect,
                       int thickness, Uint32 color, int top_right, int top_left,
                       int bottom_left, int bottom_right, int *drawn_area)
 {
+    /* The final drawn thickness will be 1 pixel greater than the value of the
+     * “thickness” parameter, as the iteration is performed over a closed
+     * interval. */
     for (int layer_radius = radius - thickness; layer_radius <= radius;
```
ankith26, line 3495: I wonder if it would be better to just fix the logic here instead, maybe we should do a `radius - thickness + 1` initialization?

EDIT: Just saw your comment about centre of solid circles, but I'm still not convinced. This function should ideally do the "right thing" and the centre special case could be handled separately? I need to do a bit more testing though to understand things in more detail.

ankith26: Okay I played around a bit with this, and now I have a clearer understanding.

Basically the main issue is that `draw.aacircle` always draws an extra pixel at the centre, so a 1 radius circle needs a 3x3 box, a 2 radius circle needs a 5x5 box and so on (yes, just like #1487). So there's always going to be a jump somewhere and this PR is just shifting that jump from `1 -> 3` to `(r - 2) -> r`. I personally think this is a reasonable compromise given the status quo.

ankith26 (approved):
I've come around to being okay with this approach for now. Thanks for contributing to pygame-ce 🎉 

I'd like to see a couple of unit tests added to test for this fix, but that won't be blocking the approval.

damusss (approved):
Tested it locally, now the function renders the correct width every time, and I'm fine with the proposed approach.
Thank you!

oddbookworm:
Thanks for the contribution @Fantastair!
If you're in the pygame community discord server, let us know what your discord username is so that you can receive the appropriate role for being a contributor

Fantastair:
Thanks for the invite @oddbookworm ! My Discord username is “Fantastair” too. Glad to contribute to pygame-ce.

PR #3709, opened 2026-02-09
Move `Credits` section in the `README.rst` to `docs/LEGACY_CREDITS.rst`

Matiiss:
These credits are very very old and don't get updated and probably never will get updated.
They also take up a significant portion of the README.


damusss:
What about the credits in different languages of the README?

damusss:
@Starbuck5 makes sense, and I suppose since they're legacy anyways it's not an issue if they're only available in English.

PR #3716, opened 2026-02-17
Delete README translations

### LINKED ISSUE #3646: Inconsistency in README translations
As I’m writing this, the README has been translated into 10 different languages, and most of them are currently outdated.

I can think of three ways to address this problem, both now and in the future:

1. Add the updated text untranslated until someone provides the translation.
2. Add a note each time a translation becomes outdated.
3. Use online tools to provide a temporary translation.

This should be an easy way to handle the issue. The only other solution would be to assign a person to each translation, but I find the other options much simpler and an outdated README can still be useful.

Matiiss:
They are outdated and mostly unmaintained
And they don't add much value to the project

Closes https://github.com/pygame-community/pygame-ce/issues/3646


robotsnh:
@Matiiss I personally think that this allows those who don't speak English to add value to this project.

oddbookworm:
@robotsnh sure, but what about the ongoing maintenance requirements of the translations? They're all in different levels of "up-to-date" with the English README because someone hasn't happened to swing by and drop an update

oddbookworm:
Plus, all of the documentation is in English, so if someone doesn't have the English skills to read the readme, then I've got some unfortunate news in that the docs are going to be nearly impossible to read

robotsnh:
@oddbookworm Fair enough. In addition, people can use Google Translate.

oddbookworm (approved):
I tend to agree with this move for the reasons mentioned already in the comments. But I think the decision to move forward with this should be left to the new steering council after the election

aatle (approved):
I agree with this change, pygame-ce's language translation does not currently extend past the README, and maintenance is very difficult.

PR #3721, opened 2026-03-01
Fixed segfault in Controller.get_mapping

oddbookworm:
…. strtok_r doesn't allocate memory

### DIFF
--- a/src_c/_sdl2/controller.c
+++ b/src_c/_sdl2/controller.c
@@ -260,7 +260,6 @@ controller_get_mapping(pgControllerObject *self, PyObject *_null)
         token = strtok_r(NULL, ",", &saveptr);
     }
 
-    SDL_free(saveptr);
     SDL_free(mapping);
     return dict;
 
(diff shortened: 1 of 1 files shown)


Starbuck5:
I can't replicate a segfault, but you're right that the code is not correct.

```py
import pygame
import pygame._sdl2.controller

pygame._sdl2.controller.init()
pygame.init()

print(pygame._sdl2.controller.get_count())

c = pygame._sdl2.controller.Controller(0)
print(c.get_mapping())

while pygame.time.get_ticks() < 5000:
        print(event)

    print(c.get_button(pygame.CONTROLLER_BUTTON_A))
    pygame.time.wait(30)

print("Done")
```

Install some virtual controller program to try to exercise this. It does seem to see the virtual controller, but I couldn't get any input through.

ankith26 (approved):
LGTM, thanks! 🎉

oddbookworm:
I could only reproduce the segfault on linux with an actual controller connected

PR #3730, opened 2026-03-06
Move non deprecated scrap function docs to stubs

Starbuck5:
For https://github.com/pygame-community/pygame-ce/issues/2757

Removes the "rtype" and explicit parameter typings b/c that is redundant to the other visible content in the stub.

image

### DIFF
--- a/buildconfig/stubs/pygame/scrap.pyi
+++ b/buildconfig/stubs/pygame/scrap.pyi
@@ -3,6 +3,41 @@ from typing_extensions import (
     deprecated,  # added in 3.13
 )
 
+def put_text(text: str, /) -> None:
+    """
+    Places text into the clipboard.
+
+    Places the input text into the clipboard. The data should be a string.
+    This is the same clipboard as the legacy scrap API when using ``SCRAP_TEXT``.
+
+    :raises pygame.error: if video mode has not been set_mode
+
+    .. note:: ``pygame.display.set_mode()`` should be called before using the ``scrap`` module
+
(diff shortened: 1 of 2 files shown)


oddbookworm:
> Will pygame keep supporting just strings or also implement the new SDL3 functions allowing arbitrary clipboard data? (I'm not sure how legacy pygame was doing that without the proper SDL functions). Supporting those functions doesn't seem easy, but I happen to copy and paste images from one application to another frequently, and considering pygame can load and save images to bytes easily it might be worth considering. (for more information: https://wiki.libsdl.org/SDL3/CategoryClipboard) Should we start aliasing this module with "clipboard" (the name I think would fit better for pygame 3)?

I don't know about everyone else, but eventually I planned on making that update... when I have time

oddbookworm:
> While looks good to me, I think you could copy the module docstrings (well... the non-deprecated part) into the stubs as well, then have an rst like this:
> 
> ```
> .. include:: common.txt
> 
> :mod:`pygame.scrap`
> =======================
> 
> .. autopgmodule:: pygame.scrap
>    :members:
> 
> [Rest of the docs with the deprecated functions]
> ```
> 
> but I'll approve anyway, this can be done in a later pr too.
> 
> > Removes the "rtype" and explicit parameter typings b/c that is redundant to the other visible content in the stub.
> 
> autoapi can read type hints, so in the future we could automatically pull typing information and add it to the docs.

I'll open a new issue to track this

PR #3736, opened 2026-03-10
Fix incorrect behavior of `_render.Texture.alpha`

damusss:
I've addressed the following issues with `pygame._render.Texture.alpha`
- It was not raising a TypeError for the wrong type (rather a SystemError, error return with no exception set)
- It was not accepting floats, unlike its predecessor `pygame._sdl2.Texture.alpha` and `Surface.set_alpha` (the stubs still ask for ints since the range is 0-255).
- It was not clamping the input before casting it to a `Uint8` (issue present with the predecessor as well, but doesn't matter considering it will be forgotten). Before, Texture.alpha = 400 would result in an alpha of 144.

I additionally added tests for the new behavior. A negative alpha input currently (like before) raises a ValueError, tho it can be changed in the future if needed.

MRE of the wrong behaviors:
```py
import pygame
from pygame._render import Renderer, Texture

pygame.init()
win = pygame.Window()
renderer = Renderer(win)
tex = Texture(renderer, (10, 10))
tex.alpha = 400
print(tex.alpha)
# run each assignment separately
tex.alpha = "10"  # SystemError
tex.alpha = 20.5  # SystemError
tex.alpha = -2  # SystemError inside a thread if there's no print after
print(tex.alpha)  # SystemError
```


test/render_test.py
```diff
@@ -267,6 +267,14 @@ def test_alpha(self):
         self.assertEqual(255, self.texture.alpha)
         self.texture.alpha = 128
         self.assertEqual(128, self.texture.alpha)
+        self.texture.alpha = 64.8
+        self.assertEqual(64, self.texture.alpha)
```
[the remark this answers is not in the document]
damusss: I'm not sure why this is flagged as a "Major issue". If I had to choose, I'd be in favor of keeping int. The underlying alpha is still an integer in range 0-255. A floating point alpha is usually associated with a 0-1 range, and I don't like the confusion. If a user is using a type checker, it's not that bad to get an error, as they get a chance to convert it to an integer in their own terms. I'm mainly basing off the fact Surface.set_alpha expects an integer but also supports floats (same as the old _sdl2.Texture.alpha).
I suppose I'll leave this "opinion" to the people reviewing the pull request.

src_c/render.c
```diff
@@ -1016,13 +1016,27 @@ texture_get_alpha(pgTextureObject *self, void *closure)
...
 {
-    if (PyLong_Check(arg)) {
-        unsigned long longval = PyLong_AsUnsignedLong(arg);
+    PyObject *intobj = NULL;
+    Uint8 alpha;
+    if (PyNumber_Check(arg) && (intobj = PyNumber_Long(arg))) {
+        unsigned long longval = PyLong_AsUnsignedLong(intobj);
+        Py_DECREF(intobj);
+        if (longval == (unsigned long)-1 && PyErr_Occurred()) {
+            RAISERETURN(PyExc_ValueError, "Texture alpha must be positive",
```
oddbookworm: Technically speaking, this error message is not correct. It's entirely possible someone ***does*** give a positive value, just larger than what an `unsigned long` can hold

test/render_test.py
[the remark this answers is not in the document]
oddbookworm: I'll have to agree with @damusss here. Ultimately, we want to make it as clear as possible what the *intended* behavior is, not strictly speaking what is *allowed*

test/render_test.py
```diff
@@ -267,6 +267,14 @@ def test_alpha(self):
...
+        self.texture.alpha = 512
```
oddbookworm, line 272: Try testing with a value larger than an `unsigned long`

oddbookworm (changes requested):
Some small comments, but enough that I think a couple changes are warranted

test/render_test.py
damusss: do you mean, locally to "verify" what you said (which I trust), or an actual additional test? I don't think we usually test for conversion overflow.

src_c/render.c
damusss: Fair enough, I'll find a way to make the error more generic

src_c/render.c
```diff
@@ -1016,13 +1016,28 @@ texture_get_alpha(pgTextureObject *self, void *closure)
...
+        else {
+            alpha = (Uint8)longval;
+        }
         RENDERER_PROPERTY_ERROR_CHECK(
-            SDL_SetTextureAlphaMod(self->texture, (Uint8)longval))
+            SDL_SetTextureAlphaMod(self->texture, alpha))
         return 0;
     }
-    return -1;
+    RAISERETURN(PyExc_TypeError, "Texture alpha must be a number", -1);
```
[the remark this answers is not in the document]
damusss: I seriously don't believe all this exception overengeneering is necessary. I can just make the error message slightly more generic by saying "Texture alpha must be a valid number", and it's good. The problem mentioned doesn't sound concerning enough to bloat it. Most of pygame's exception raising consists of "parameter is invalid" with no details whatsoever. This PR is already improving incorrect behavior, if whatever benefit is gained by branching the error again is really needed it can be improved more.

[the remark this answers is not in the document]
damusss: I'll let @oddbookworm be the judge of this, since he requested the changes.

test/render_test.py
```diff
@@ -267,6 +267,14 @@ def test_alpha(self):
...
+        self.assertEqual(255, self.texture.alpha)
+        with self.assertRaises(TypeError):
+            self.texture.alpha = "128"
+        with self.assertRaises(ValueError):
```
oddbookworm: This should now be an `OverflowError`

oddbookworm (changes requested):
This should fix the slew of failing builds

test/render_test.py
damusss: All fixed :)

oddbookworm (approved):

Starbuck5:
The checking, converting to int, then finally converting to C adds a few lines of code here. Could it use pg_IntFromObj to be a little simpler on this side? I guess then it needs to deal with negative values.

Or could PyArg_Parse with format "b" work?

damusss:
@Starbuck5 I checked both options, but to keep the same functionality they don't seem viable.
- pg_UIntFromObj: apart from forcing UInt32, which isn't a big deal, the error set by python is cleared before returning, forcing a generic error that was criticized in this PR.
- PyArg_Parse with K for unsigned long would not provide the same amount of error checking (quoting "For unsigned integer formats, no range checking is done").
Either option would reduce the code but has "drawbacks", I'm open to hearing your thoughts.

PR #3750, opened 2026-03-19
Fix segfault in `transform.solid_overlay` when the width of the given surface is 0

### LINKED ISSUE #3747: Segmentation fault when using solid_overlay on empty rendered text
**Environment:**

```
pygame-ce 2.5.7 (SDL 2.32.10, Python 3.13.7)
Platform:		Windows-10-10.0.19045-SP0
System:			Windows
System Version:		10.0.19045
Processor:		AMD64 Family 23 Model 24 Stepping 1, AuthenticAMD	SSE2: Yes	AVX2: Yes	NEON: No
Architecture:		Bits: 64bit	Linkage: WindowsPE

Python:			CPython 3.13.7 (tags/v3.13.7:bcee1c3, Aug 14 2025, 14:15:11) [MSC v.1944 64 bit (AMD64)]
GIL Enabled:		True
pygame version:		2.5.7
SDL versions:		Linked: 2.32.10	Compiled: 2.32.10
SDL Mixer versions:	Linked: 2.8.1	Compiled: 2.8.1
SDL Font versions:	Linked: 2.24.0	Compiled: 2.24.0
SDL Image versions:	Linked: 2.8.8	Compiled: 2.8.8
Freetype versions:	Linked: 2.11.1	Compiled: 2.11.1

Display Driver:		windows
Mixer Driver:		wasapi
```

**Current behavior:**

Python crashes with a segmentation fault.

**Expected behavior:**

The code should make a small window, and then close.

**Test code**

```
import pygame
pygame.init()
win = pygame.display.set_mode((1,1)) # Resolution doesn't seem to matter
Font = pygame.font.Font("Fonts/CloisterBlack.ttf", 1) # Font and size doesn't seem to matter
RenderedText = Font.render("", 0, (0, 0, 0)).convert_alpha() # Text has to be empty
pygame.transform.solid_overlay(RenderedText,(255,255,255))
```

```
pygame-ce 2.5.7 (SDL 2.32.10, Python 3.13.7)
Platform:		Windows-10-10.0.19045-SP0
System:			Windows
System Version:		10.0.19045
Processor:		AMD64 Family 23 Model 24 Stepping 1, AuthenticAMD	SSE2: Yes	AVX2: Yes	NEON: No
Architecture:		Bits: 64bit	Linkage: WindowsPE

Python:			CPython 3.13.7 (tags/v3.13.7:bcee1c3, Aug 14 2025, 14:15:11) [MSC v.1944 64 bit (AMD64)]
GIL Enabled:		True
pygame version:		2.5.7
SDL versions:		Linked: 2.32.10	Compiled: 2.32.10
SDL Mixer versions:	Linked: 2.8.1	Compiled: 2.8.1
SDL Font versions:	Linked: 2.24.0	Compiled: 2.24.0
SDL Image versions:	Linked: 2.8.8	Compiled: 2.8.8
Freetype versions:	Linked: 2.11.1	Compiled: 2.11.1

Display Driver:		windows
Mixer Driver:		wasapi
Fatal Python error: pygame_parachute: (pygame parachute) Segmentation Fault

Current thread 0x000006ec (most recent call first):
  File "[filepath removed for privacy]/Pygame Smoothscale Reproduction.py", line 7 in <module>

Extension modules: pygame.base, pygame.constants, pygame.rwobject, pygame.rect, pygame.bufferproxy, pygame.math, pygame.surflock, pygame.surface, pygame.window, pygame.display, pygame.draw, pygame.joystick, pygame.event, pygame.imageext, pygame.image, pygame.key, pygame.mouse, pygame.pixelcopy, pygame.mask, pygame.time, pygame.transform, pygame.font, pygame.mixer_music, pygame.mixer, pygame.scrap, pygame.system, pygame._freetype (total: 27)
```

Matiiss:
Fixes https://github.com/pygame-community/pygame-ce/issues/3747

### DIFF
--- a/src_c/transform.c
+++ b/src_c/transform.c
@@ -2391,6 +2391,9 @@ solid_overlay(pgSurfaceObject *srcobj, Uint32 color, pgSurfaceObject *dstobj,
             PyExc_ValueError,
             "Destination surface must be the same size as source surface."));
     }
+    else if (newsurf->w == 0 || newsurf->h == 0) {
+        return newsurf;
+    }
 
     PG_PixelFormat *fmt, *newsurf_format;
     SDL_Palette *src_palette, *newsurf_palette;
--- a/src_c/transform.c
+++ b/src_c/transform.c
@@ -2454,68 +2454,34 @@ solid_overlay(pgSurfaceObject *srcobj, Uint32 color, pgSurfaceObject *dstobj,
         const int dst_skip = newsurf->pitch / 4 - newsurf->w;
         int n, height = src->h;
 
-        if (srcobj == dstobj) {
-            if (!keep_alpha) {
-                while (height--) {
-                    LOOP_UNROLLED4(
-                        {
-                            if (*srcp) {
-                                *dstp = color;
-                            }
-                            srcp += 4;
-                            dstp++;
-                        },
-                        n, src->w);
-                    srcp += src_skip;
-                    dstp += dst_skip;
-                }
-            }
-            else {
-                while (height--) {
-                    LOOP_UNROLLED4(
-                        {
-                            if ((a = *srcp)) {
-                                *dstp = color | (a << dst_ashift);
-                            }
-                            srcp += 4;
-                            dstp++;
-                        },
-                        n, src->w);
-                    srcp += src_skip;
-                    dstp += dst_skip;
-                }
+        if (!keep_alpha) {
+            while (height--) {
+                LOOP_UNROLLED4(
+                    {
+                        if (*srcp) {
+                            *dstp = color;
+                        }
+                        srcp += 4;
+                        dstp++;
+                    },
+                    n, src->w);
+                srcp += src_skip;
+                dstp += dst_skip;
             }
         }
         else {
-            if (!keep_alpha) {
-                while (height--) {
-                    LOOP_UNROLLED4(
-                        {
-                            if (*srcp) {
-                                *dstp = color;
-                            }
-                            srcp += 4;
-                            dstp++;
-                        },
-                        n, src->w);
-                    srcp += src_skip;
-                    dstp += dst_skip;
-                }
-            }
-            else {
-                while (height--) {
-                    LOOP_UNROLLED4(
-                        {
-                            if ((a = *srcp)) {
-                                *dstp = color | (a << dst_ashift);
-                            }
-                            srcp += 4;
-                            dstp++;
-                        },
-                        n, src->w);
-                    srcp += src_skip;
-                    dstp += dst_skip;
-                }
+            while (height--) {
+                LOOP_UNROLLED4(
+                    {
+                        if ((a = *srcp)) {
+                            *dstp = color | (a << dst_ashift);
+                        }
+                        srcp += 4;
+                        dstp++;
+                    },
+                    n, src->w);
+                srcp += src_skip;
+                dstp += dst_skip;
             }
         }
     }
--- a/test/transform_test.py
+++ b/test/transform_test.py
@@ -1,3 +1,4 @@
+import itertools
 import os
 import platform
 import unittest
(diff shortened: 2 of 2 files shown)


src_c/transform.c
```diff
@@ -2451,68 +2454,34 @@ solid_overlay(pgSurfaceObject *srcobj, Uint32 color, pgSurfaceObject *dstobj,
         const int dst_skip = newsurf->pitch / 4 - newsurf->w;
         int n, height = src->h;
 
-        if (srcobj == dstobj) {
```
ankith26, line 2454: this block of changes is interesting, I understand you are eliminating redundant code but this has gotten me wondering whether we are supposed to handle the `srcobj == dstobj` case specially? Probably not but having a testcase added to check for this would be a good thing I think.

oddbookworm: I decided to write my own pair of tests to check this. It doesn't segfault on windows when the pointers are equal, but it also doesn't do the expected thing either
```py
    def test_solid_overlay_dest_is_not_src(self):
        src = pygame.Surface((50, 50))
        src.fill("blue")

        pygame.image.save(src, "is_not_pre.png")

        dest = pygame.Surface((50, 50))
        dest.fill("red")

        self.assertNotEqual(pygame.transform.solid_overlay(src, "white", dest_surface=dest), src)
        pygame.image.save(dest, "is_not_out.png")

    def test_solid_overlay_dest_is_src(self):
        src = pygame.Surface((50, 50))
        src.fill("blue")

        pygame.image.save(src, "is_pre.png")

        self.assertIs(pygame.transform.solid_overlay(src, "white", dest_surface=src), src)
        pygame.image.save(src, "is_out.png")
```

oddbookworm: In the case of `test_solid_overlay_dest_is_not_src`, I get `is_not_pre.png` as a blue square and `is_not_out.png` is a red square.

In the case of `test_solid_overlay_dest_is_src`, both pngs are identical

Matiiss: I'm pretty sure in your tests any processing is skipped (i.e., it doesn't even get to `if (srcobj == dstobj) {}`) because of this guard clause https://github.com/pygame-community/pygame-ce/blob/965c7a6c8d580f4c383e040f5290b6bc3dcc1cd5/src_c/transform.c#L2408-L2412
Which arguably is a bit incorrect, since it should probably fill the (destination) surface with the given overlay color before returning it in this case.

Matiiss: > this block of changes is interesting, I understand you are eliminating redundant code but this has gotten me wondering whether we are supposed to handle the `srcobj == dstobj` case specially? Probably not but having a testcase added to check for this would be a good thing I think.

This is already being tested https://github.com/pygame-community/pygame-ce/blob/965c7a6c8d580f4c383e040f5290b6bc3dcc1cd5/test/transform_test.py#L555-L565

If a destination surface is not given, it will create a new surface, so the addresses are not the same, if, however, the destination surface is the same as the input, then naturally those addresses will be the same, hence testing both paths.

oddbookworm: If the surfaces need to support alpha for anything to happen, then I would consider the documentation flawed, as it says nothing about actually needing alpha. Granted, if you don't have an alpha surface, then the better solution would be `fill`, but still

Matiiss: Well, it could be that the implementation is flawed instead and we might want to conform to the docs and change the behavior. Or both. We could deprecate current behavior in code, by emitting a warning in that code path with a plan to eventually either raise an exception or do a simple fill + update docs that the behavior is broken for non-alpha surfaces (if a dest surface is given, it just returns that, otherwise it will just return a blank ((0, 0, 0, 0)-filled?) surface (which arguably is like the worst possible behavior this function could have, but oh well)). We can eventually fix this behavior for 3.0 probably.

Matiiss: Do you want those docs and/or behavior changes in this PR though? So far I think what Ankith originally asked about can be marked as resolved, since the tests he asked for already exist (unless he wants to extend those (again, in this PR though?)).

ankith26: I think any docs/behaviour changes related to this should be a different PR, yes. I only bought it up because you made this change in this PR rn.

ankith26 (approved):
This PR LGTM, thanks 🎉

Starbuck5 (approved):
Thanks!

PR #3759, opened 2026-03-28
Harden gh actions workflows

ankith26:
After learning about the recent [litellm security incident](https://docs.litellm.ai/blog/security-update-march-2026) (TL;DR: supply chain attack via github actions, the PyPI token of a popular library was leaked to attackers who released bad releases that then compromised user creds and so on) I was prompted into further scrutinizing our own actions for security.

The good news is that we are already pretty secure in terms of PyPI publishing. We don't use a static PyPI token and instead use the recommended [Trusted PyPI Publisher](https://docs.pypi.org/trusted-publishers/) approach for managing our releases. So an attack like the above mentioned case wouldn't be possible with us.

But still, to be cautious (using [zizmor](https://docs.zizmor.sh/) for github actions static analysis) this PR identifies and fixes a couple of security gotchas, mainly under three categories:
- Pins all actions to hashes instead of versions (this secures us against potential malicious upstream releases that re-use old tags). Dependabot is capable of keeping these hashes updated as hashes.
- Explicit content read-only permissions for all actions (this secures us against actions making writes to the repo)
- Checkout without persisting git credentials

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -5,3 +5,5 @@ updates:
     schedule:
       # Check for updates to GitHub Actions every week
       interval: "weekly"
+    cooldown:
+      default-days: 7
--- a/.github/workflows/build-debian-multiarch.yml
+++ b/.github/workflows/build-debian-multiarch.yml
@@ -35,6 +35,10 @@ concurrency:
   group: ${{ github.workflow }}-${{ github.ref }}-debian-multiarch
   cancel-in-progress: true
 
+permissions:
+  contents: read
+  packages: write
+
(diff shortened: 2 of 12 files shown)


oddbookworm (approved):
This seems fine to me

Starbuck5 (approved):
Did you mean to add permissions to build-manylinux as well?

Otherwise happy to go with this, thanks for checking it out Ankith.

ankith26:
build-manylinux already has permissions defined but at the job level not workflow level. This is fine for now because the workflow has only one job defined but this can be changes in the future.

PR #3760, opened 2026-04-05
Ignore non string values from fonts registry

### LINKED ISSUE #3742: Windows sysfont.py splittext() TypeError when enumerating fonts
System:Windows
Python: 3.13.3
pygame version:  2.5.3

When calling pygame.font.SysFont() on windows, a TypeError occurs in sysfont.py:
TypeError: expected str, bytes or os.PathLike object, not int

Image

Font environment: Some font entries have interget values of string filenames
Excetped behavior: SysFont() skip non-string registry entries and not crash.
To fix: Add a type check before calling splittext() to skip non-string font values.

Starbuck5:
(On Windows)

Closes #3742

### DIFF
--- a/src_py/sysfont.py
+++ b/src_py/sysfont.py
@@ -77,6 +77,12 @@ def initsysfonts_win32():
                 except OSError:
                     break
 
+                # https://github.com/pygame-community/pygame-ce/issues/3742
+                # Registry values could be other types besides strings.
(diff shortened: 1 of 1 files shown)


ankith26 (approved):
LGTM, thanks! 🎉

damusss (approved):
Fair enough, LGTM!

PR #3765, opened 2026-04-12
Math doc updates

Starbuck5:
- Vector3 update was incorrectly typed to take ints
- Smoothstep param name mismatched between docs and stub
- Swizzling out 3 elements returns a Vector3 actually
- Smooth interpolations formula was flipped
- No need to have a sentence that says "general information about vector{x}" and then nothing else, removed


damusss (approved):
Thank you for fixing it!

ankith26 (approved):
LGTM, thanks 🥳

PR #3766, opened 2026-04-12
Font doc updates

Starbuck5:
- match_font can return None (description in docs said so, source said so, but stubs and signature line did not)
- metrics can include Nones for unknown characters
- Returns use more proper type names
- point size extra descriptions seemed irrelevant to me. The single line description gives enough. We don't need vague notices about things that apply to a single legacy codepath for the default font.


damusss (approved):
LGTM, thank you!

docs/reST/ref/font.rst
```diff
@@ -564,10 +555,6 @@ solves no longer exists, it will likely be removed in the future.
       | :sl:`Get the point size of the font.`
       | :sg:`get_point_size() -> int`
 
-      Returns the point size of the font. Will not be accurate upon
-      initializing the font object when the font name is initialized
-      as ``None``.
```
ankith26, line 569: Sure, this is a legacy edge case thing, but I feel like there's no harm in leaving it as it is. It could potentially help someone who is confused by the legacy behaviour?

Starbuck5: The AI review pointed this out as a potentially confusing block, and it suggested some edit that just restructured it a bit to be clearer, and then I was like I think this should be removed entirely.

I think that to document this in a good way would be to add a whole new paragraph to Font init about this specific case where the font name is None the font size is slightly smaller for some reason. But I don't want to add docs for behavior I want to remove in pygame-ce 3. Plus there's cost to adding more and more details to the docs, makes it harder to understand.

I will revert this change for this PR.

docs/reST/ref/font.rst
aatle, line 558: It should be documented, like the other properties, that these two methods are the same as the `point_size` property.

Hmm, maybe the `linesize` and potentially also the read-only `ascent`, `descent`, `height` methods should also have properties.

docs/reST/ref/font.rst
```diff
@@ -552,9 +549,6 @@ solves no longer exists, it will likely be removed in the future.
       | :sl:`Set the point size of the font.`
       | :sg:`set_point_size(size, /) -> None`
 
-      Sets the point size of the font, which is the value that was used to
-      initialize this font.
-
```
aatle, line 557: I think you forgot to revert the removal of this docstring.

Starbuck5: This is a separate change, I just feel like the text here is not relevant, so I removed it to simplify the doc.

set_point_size: Sets the point size. Who would've guessed?
"which is the value that was used to initialize this font" -- Uhhhh, what? Yes font init takes a size, but the value being set has nothing to do with what value was used to initialize.

aatle (approved):
Cool

PR #3767, opened 2026-04-12
Ignore docs changes more on workflows ignoring docs

Starbuck5:
I think other builds could probably ignore docs changes as well, but they aren't right now.

This PR makes it so that docs changes like I just put up a few PRs for should be ignored by at least a few workflows.


damusss (approved):
Makes sense!

ankith26 (approved):
LGTM, thanks 🎉

PR #3771, opened 2026-04-14
Use noexcept in cython recording_cb to fix warning

ankith26:
This PR fixes cython compiler warnings that pop up during build time. These functions shouldn't raise exceptions (handling is done by printing the exception) so they must be marked with `noexcept`


Starbuck5:
Given we're moving to 3.0 and this code will be deleted, I'm thinking we should just close this?

Thanks for submitting it.

ankith26:
I'd like to defer closing this PR till we actually remove the cython files. If we do end up reuse some of this logic down the line this PR may still be needed

PR #3772, opened 2026-04-15
Remove mask/freetype warning supressions

ankith26:
This PR:
- updates all our structs to use [FAM](https://en.wikipedia.org/wiki/Flexible_array_member) where needed. This fixes some mask warnings
- cleans up some `ft_cache` logic. The new logic is (almost) equivalent but easier to understand and does not trip the compiler into raising warnings.
- Updates the `bitcount` code to use macro checks instead of if-else checks

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -544,15 +544,9 @@ summary(
 # stores werror
 warnings_error = []
 
-# module specific warnings (TODO remove these by fixing code)
-warnings_temp_freetype = []
-warnings_temp_mask = []
-
 if cc.get_argument_syntax() == 'msvc'
     # base warnings on msvc
     add_global_arguments(['/W3', '/wd4142', '/wd4996'], language: 'c')
-
-    warnings_temp_mask += ['/wd6385', '/wd6386']
 
     if get_option('error_on_warns') and not is_pypy
         warnings_error += '/WX'
@@ -561,19 +555,6 @@ if cc.get_argument_syntax() == 'msvc'
 else
     # base warnings on gcc/clang
     add_global_arguments(['-Wall', '-Wno-error=unknown-pragmas'], language: 'c')
-
-    # check for exact gcc (flags not needed on clang)
-    if cc.get_id() == 'gcc'
-        warnings_temp_freetype += [
-            '-Wno-error=unused-but-set-variable',
-            '-Wno-error=array-bounds',
-        ]
-        warnings_temp_mask += '-Wno-error=array-bounds'
-    endif
-
-    if cc.sizeof('long') != 8
-        warnings_temp_mask += '-Wno-error=shift-count-overflow'
-    endif
 
     if get_option('error_on_warns') and not is_pypy
         warnings_error += '-Werror'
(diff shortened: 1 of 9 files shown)


Starbuck5 (commented):
Took some time to wrap my head around these changes. I'm fully on board with bitmask and base changes, but I'm not sure about freetype. The previous implementation in get_hash went through dwords backwards, this one goes through it forwards.

I tried taking away the warnings things only, with no other changes from this PR, and surprisingly dev.py successfully compiled on Windows.

I went down a mini-rabbithole about bitcount cpu optimizations. There is an instruction for it, popcnt, that seems pretty well supported on x86, like as common as SSE2. But getting to it in C in a portable way is only available in C23, https://en.cppreference.com/c/numeric/bit_manip (stdc_count_ones), which MSVC seems to have no plans to support anytime soon.

ankith26:
> The previous implementation in get_hash went through dwords backwards, this one goes through it forwards.

Yes I did realise this while making the PR, but it shouldn't really affect things anyways because all the loop is doing is computing hashes of data. As long as we are covering the same memory, the order of the loop shouldn't matter. I can of course, reverse the loop and retain the same order if you'd like. My main goal here was to fix the weird `- 1` offset on a temporary loop variable which only triggers warnings on older GCC compilers (like the one manylinux uses) and not on other compilers like modern MSVC.

> There is an instruction for it, popcnt, that seems pretty well supported on x86, like as common as SSE2.

It would be technically possible to use popcnt, we'd just have to write compiler specific code. On MSVC we can use something like `__popcnt64`/`__popcnt` and `__builtin_popcountl` on GCC/Clang. These have been supported for a long time. We would need to apply similar runtime detection logic as our SIMD code, because popcnt is not a part of x86_64 baseline and older CPUs would need software fallback.

Starbuck5:
The code comments say it's based on MurmurHash3, and when I look up MurmurHash3 the implementation given on wikipedia is also using a downwards loop. https://en.wikipedia.org/wiki/MurmurHash . So it does seem like it matters. I don't think all the operations performed are commutative.

ankith26:
I updated the PR to do reverse looping, the compiler should now produce identical assembly in theory.

Starbuck5:
I asked Claude to take a look, out of curiosity, and it came up with this:

```
Minor — NUM_DWORDS macro is missing outer parentheses:

// Current (ft_cache.c:36)
#define NUM_DWORDS (sizeof(KeyFields) + 3) / 4

// Should be:
#define NUM_DWORDS ((sizeof(KeyFields) + 3) / 4)
As written, uses in arithmetic contexts like (NUM_DWORDS - 1) expand to (sizeof(KeyFields) + 3) / 4 - 1, which is fine due to C operator precedence (/ before -). But it's fragile — a future use like 8 / NUM_DWORDS would expand to 8 / (sizeof(KeyFields) + 3) / 4, giving wrong results. Standard practice is to wrap the whole expression.
```

Starbuck5 (approved):

ankith26:
I force pushed with just the macro parenthesis fix.

PR #3773, opened 2026-04-19
Image/draw doc updates

Starbuck5:
- Some slight wording/formatting stuff
- In draw, the two aacircle overloads are actually compatible, since all the extra arguments of the second one are default arguments. So this removes the first overload.
- Image.frombytes only said it takes arguments similar to `tobytes`, but it didn't list which format codes would actually work. Not all the tobytes formats work.
- Clarifies that image.load_extended will specifically raise a NotImplementedError if there is no extended image support.


robertpfeiffer (approved):

ankith26 (approved):
LGTM, thanks 🎉

PR #3780, opened 2026-04-26
Update image dependencies

Starbuck5:
Updates SDL_image to 2.8.10
- Ankith said he wants to change how Windows deps works, I am skeptical this is a good release to do that in, but I generated a 2.8.10 release in PGCE SDL_image already, and this preserves the status quo. If later in this release's lifespan, the Windows builds change, that is fine by this PR-- the Linux side, the meson version selection will be kept.

Updates image loading libraries
- It was very tempting to go ham updating a bunch of stuff, like freetype, brotli, glib, but decided to stay focused on just image stuff in this PR


ankith26 (approved):
LGTM, thanks 🎉 

Agreed, we can do the other deps in future PRs, and even the windows part of this PR is good to get in for now.

PR #3783, opened 2026-04-29
`pos` attribute added to `MOUSEWHEEL` event

DickerDackel:
Up to now, if the position where a MOUSEWHEEL event happends is required, a user has to either fall back on the additionally created MOUSEBUTTON events, or poll the mouse position manually.

Position information in the MOUSEWHEEL event was added with SDL 2.0.26. In case pygame was built with an SDL version older than 2.0.26, the `pos` value will be `None`.

### DIFF
--- a/docs/reST/ref/event.rst
+++ b/docs/reST/ref/event.rst
@@ -160,7 +160,7 @@ These ``TEXT*`` events are useful to implement inputs/languages that require com
 provided by the system's IME (Input Method Editor). The ``TEXTINPUT`` event is only
 fired when the text is confirmed.
 
-(``5``): Position information was added to ``MOUSEWHEEL`` events in SDL 2.0.26.  If
+(``5``): Position information was added to ``MOUSEWHEEL`` events in SDL 2.26.0  If
 pygame was build on a version lower than that, the ``pos`` attribute is ``(0, 0)``.
 
 |
--- a/docs/reST/ref/event.rst
+++ b/docs/reST/ref/event.rst
@@ -88,7 +88,7 @@ You can also find a list of constants for keyboard keys
     MOUSEBUTTONUP              pos, button, touch(2), clicks(3), window(1)
     MOUSEBUTTONDOWN            pos, button, touch(2), clicks(3), window(1)
     MOUSEWHEEL                 which, flipped, x, y, touch(2), precise_x,
-                                precise_y, window(1)
+                                precise_y, window(1), pos(5)
 
     JOYAXISMOTION              instance_id, axis, value (deprecated: joy)
     JOYBALLMOTION              instance_id, ball, rel (deprecated: joy)
@@ -160,6 +160,9 @@ These ``TEXT*`` events are useful to implement inputs/languages that require com
 provided by the system's IME (Input Method Editor). The ``TEXTINPUT`` event is only
 fired when the text is confirmed.
 
+(``5``): Position information was added to ``MOUSEWHEEL`` events in SDL 2.0.26.  If
+pygame was build on a version lower than that, the ``pos`` attribute is ``(0, 0)``.
+
 |
 
 Here is a list of the new window events.
@@ -222,6 +225,8 @@ On Android, the following events can be generated:
 
 .. versionadded:: 2.5.7 The ``clicks`` attribute was added to ``MOUSEBUTTONDOWN``
    and ``MOUSEBUTTONUP`` events.
+
+.. versionadded:: 2.5.8 The ``pos`` attribute was added to ``MOUSEWHEEL`` events.
 
 |
 
--- a/docs/reST/ref/event.rst
+++ b/docs/reST/ref/event.rst
@@ -161,7 +161,7 @@ provided by the system's IME (Input Method Editor). The ``TEXTINPUT`` event is o
 fired when the text is confirmed.
 
 (``5``): Position information was added to ``MOUSEWHEEL`` events in SDL 2.26.0  If
-pygame was build on a version lower than that, the ``pos`` attribute is ``(0, 0)``.
+pygame was build on a version lower than that, the ``pos`` attribute is ``None``.
 
 |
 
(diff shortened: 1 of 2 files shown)


src_c/event.c
```diff
@@ -1380,6 +1380,17 @@ dict_from_event(SDL_Event *event)
...
                 PyBool_FromLong((event->wheel.which == SDL_TOUCH_MOUSEID)));
 
+#if SDL_VERSION_ATLEAST(3, 0, 0)
+            obj = pg_tuple_couple_from_values_int((int)event->wheel.mouse_x,
+                                                  (int)event->wheel.mouse_y);
+#elif SDL_VERSION_ATLEAST(2, 26, 0)
+            obj = pg_tuple_couple_from_values_int((int)event->wheel.mouseX,
+                                                  (int)event->wheel.mouseY);
+#else
+        obj = pg_tuple_couple_from_values_int(0, 0);
```
oddbookworm: I don't know about giving it a default value here.... I could see people reading the docs but not bothering to check their pygame-ce version and then reporting issues that it doesn't work.

DickerDackel: I was contemplating this, but took sdl2's `logical_size` as guide, which does something similar in returning `(0, 0)` on read when it had never been set.

https://pyga.me/docs/ref/sdl2_video.html#pygame._sdl2.video.Renderer.logical_size

But I'm completely open to removing it.

oddbookworm: I think there's a difference between this situation and `logical_size`, where `logical_size` is something that is available, but unused so far. In this case, it's a feature that's actually just unavailable at compile-time, so we'd just be lying to the user and relying on them having read the docs and checking their compiled SDL2 version

DickerDackel: How common is the use case of having a too old sdl2 version in the first place?  I know you bundle your own with the official wheels (at least on linux).

If I remove the whole attribute on versions smaller than 2.26.0, the user would need to catch an `AttributeError` or do a probe with `hasattr` first.

I can raise a `RuntimeError` warning about the too old SDL2 version instead of a default, but that would still put the burden of catching that on the user.

I'm not a friend of removing the whole #if clause, since in the same event, there's already another attribute checking for version 2.0.18, but that one has an easy usable fallback value.

The final option would be to request the mouse coordinates in older versions.  Looking at mouse.get_pos, that's a whole different can of worms, and since that functionality is not extracted into a dedicated C function that would either mean doing a call through the pythonn layer, duplicating a bunch of code, or factoring the actual functionality out of mouse_get_pos, so I could call it directly.

Looking at all these, either my solution with the defaults or a `RuntimeError` feel the cleanest to me.  What would you prefer?

DickerDackel: Ooops, forgot to answer here.  We agreed on `None` instead of silently ignoring the error or creating an exception.  Code is updated accordingly.

src_c/event.c
```diff
@@ -1380,6 +1380,17 @@ dict_from_event(SDL_Event *event)
...
+#if SDL_VERSION_ATLEAST(3, 0, 0)
+            obj = pg_tuple_couple_from_values_int((int)event->wheel.mouse_x,
+                                                  (int)event->wheel.mouse_y);
+#elif SDL_VERSION_ATLEAST(2, 26, 0)
+            obj = pg_tuple_couple_from_values_int((int)event->wheel.mouseX,
+                                                  (int)event->wheel.mouseY);
+#else
+        obj = Py_None;
+#endif /* SDL_VERSION_ATLEAST(2, 26, 0) */
+            _pg_insobj(dict, "pos", obj);
```
oddbookworm, line 1393: Hmmmm, this seems like it doesn't handle the reference count... but CPython also changed how singleton refcounts worked in 3.14 I think?

docs/reST/ref/event.rst
```diff
@@ -160,6 +160,9 @@ These ``TEXT*`` events are useful to implement inputs/languages that require com
 provided by the system's IME (Input Method Editor). The ``TEXTINPUT`` event is only
 fired when the text is confirmed.
 
+(``5``): Position information was added to ``MOUSEWHEEL`` events in SDL 2.26.0  If
+pygame was build on a version lower than that, the ``pos`` attribute is ``(0, 0)``.
```
oddbookworm: Update docs to the new default value on an invalid SDL version

oddbookworm (changes requested):
I have one definitely actionable comment, and one I would like to see justified if behavior is correct, or adjusted if incorrect

src_c/event.c
Starbuck5: Good call, the reference counting in this function is incorrect.

pg_tuple_couple_from_values_int returns a tuple with a ref count of 1, then in pg_insobj the dictionary adds a reference to the tuple, and the object is decref-ed. Now the only reference to the tuple is from the dictionary.

In the Py_None case, pg_insobj "gives up the reference" to the dictionary the same way, but the caller does not have a reference to give up.

It should be `obj = Py_NewRef(Py_None)`, or use `_pg_insobj` for the other two paths and directly call `PyDict_SetItem` for the None path (to avoid incref-ing just to decref).

Even if None happens to be an immortal object in recent Python versions, we should treat it like any other object.

docs/reST/ref/event.rst
DickerDackel: I'm stupid :sob:

src_c/event.c
DickerDackel: Thanks, I "stole" the Py_None from another part and missed that the incref only happened 2 lines down.
Fixing...

oddbookworm (approved):
LGTM! Thanks Dackel! 🎉

Starbuck5 (approved):
Thanks!

docs/reST/ref/event.rst
```diff
@@ -160,6 +160,9 @@ These ``TEXT*`` events are useful to implement inputs/languages that require com
 provided by the system's IME (Input Method Editor). The ``TEXTINPUT`` event is only
 fired when the text is confirmed.
 
+(``5``): Position information was added to ``MOUSEWHEEL`` events in SDL 2.26.0  If
```
damusss: I think there should be a period separating 2.26.0 and "If", and I believe "build" should instead be "built".

damusss (approved):
I pointed out a few typos, but other than that, it looks good to me, thank you!

docs/reST/ref/event.rst
DickerDackel: Correct, fixed.  Thank you!

PR #3785, opened 2026-05-03
Further optimized `mask.from_surface()` when converting 32 bit alpha surfaces

itzpr3d4t0r:
Follows #2895 

In the previous implementation I circumvented SDL pixel getters by rolling in some manual alpha-getting procedures. That sped it up quite a bit but the code still wasn't parallel enough and clogged itself with unnecessary calculations while also not exploiting the way the bitmask is stored.

As documented in the code, bitmasks are stored column-group-major, with each group being 32 or 64 bits depending on the platform.

Take a 128X2 surface 64-bit platform, the bitmap will be laid out as follows.
```
Index:     [0]         [1]          [2]         [3]
        row0,col0   row1,col0    row0,col1   row1,col1
        px 0-63      px 0-63      64-127     px 64-127
```
We exploit this fact and do:
- Still read all source surface's alpha channel only by jumping between the alphas horizontally (so we stick to the same row-scanline as before).
- write all of the bitmask's group of pixels (32 or 64) by row so:
    1.  Do bits[0] in groups of 8 in parallel up to the nearest multiple of 8 possible
    2.  Do remaining 1-7 pixels one by one up to 32 or 64
    3.  Jump to the next group on the same row (index 2 from the example)
    4. Go to the next row and repeat the process.
    
I've manually unrolled the loop so there aren't as many data dependencies and removed the duff's device.
From my testing all this makes the algorithm 60-80% faster.

Figure_2

### DIFF
--- a/src_c/mask.c
+++ b/src_c/mask.c
@@ -43,24 +43,6 @@
 #define M_PI 3.14159265358979323846
 #endif
 
-/* Pretty good idea from Tom Duff :-). */
-#ifndef LOOP_UNROLLED4
-#define LOOP_UNROLLED4(code, n, width) \
-    n = (width + 3) / 4;               \
-    switch (width & 3) {               \
-        case 0:                        \
-            do {                       \
-                code;                  \
-                case 3:                \
-                    code;              \
-                case 2:                \
-                    code;              \
-                case 1:                \
-                    code;              \
-            } while (--n > 0);         \
-    }
-#endif
-
 /* Macro to create mask objects. This will call the type's tp_new and tp_init.
  * Params:
  *     w: width of mask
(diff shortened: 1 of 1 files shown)


Starbuck5:
@itzpr3d4t0r the man the myth the legend! Cool stuff.

One thing noticed is that it is not possible to have a 24 bit surface with per pixel alpha, so the 24 bit codepath is not reachable.

Also for bonus points, it also is theoretically possible to reach the 4 BPP codepath with a format that would be unsupported, like SDL_PIXELFORMAT_ARGB2101010. So maybe safest patch would be instead of (bpp < 3) for the generic codepath, (SDL_PIXELLAYOUT(surf->format) == SDL_PACKEDLAYOUT_8888) or something to that end.

itzpr3d4t0r:
Bunny was correct, brought a 15-20% perf improvement so now we're comfortably above 110% better than before this PR.
image

For some reason my pc doesn't like the branchless logic suggestion. I guess it's because my CPU has a very advanced two-ahead branch predictor (AMD 9800X3D). So i suggest to whoever wants to check it out with maybe older hardware, just substitute the inner loop with the 8 pixels at a time with this 
```C
const int bp = x & BITMASK_W_MASK;
Uint8 byte = 0;

byte |= (srcp[0]  > u_threshold) << 0;
byte |= (srcp[4]  > u_threshold) << 1;
byte |= (srcp[8]  > u_threshold) << 2;
byte |= (srcp[12] > u_threshold) << 3;
byte |= (srcp[16] > u_threshold) << 4;
byte |= (srcp[20] > u_threshold) << 5;
byte |= (srcp[24] > u_threshold) << 6;
byte |= (srcp[28] > u_threshold) << 7;

block |= ((BITMASK_W)byte) << bp;
srcp += 32; 
x += 8;
```

Starbuck5 (approved):
Alright, I'm cool with this.

ankith26 (approved):
LGTM on a quick scan, thanks for working on this 🥳

PR #3786, opened 2026-05-03
Start using Py_NewRef in a few modules

Starbuck5:
When looking at ref counting https://github.com/pygame-community/pygame-ce/pull/3783, I was surprised to see we don't use Py_NewRef at all throughout our codebase.

Docs: https://docs.python.org/3/c-api/refcounting.html#c.Py_NewRef

Instead of assigning something and then incref-ing it, we can use newref inline to return a new reference to the object. Which is the same thing, just more concise. I checked the generated assembly with my `add_global_arguments('/FAcsu', language: 'c')` trick for MSVC, I did not see any meaningful change in the generated output.

It seems that this is the preferred way of doing things in CPython internally now: https://github.com/python/cpython/issues/99300


ankith26 (approved):
LGTM, thanks 🎉

oddbookworm (approved):
LGTM!

PR #3790, opened 2026-05-10
Fix incorrect 24bit writes

Starbuck5:
Closes #3702

Originally assigned to Matt, but it was dormant with no activity for 3 months so I took it over.

This PR fixes incorrect handling of certain 24 bit surfaces when written to by transform.threshold, PixelArray.replace, and PixelArray[...] = ...

I did research in the original issue (thanks again to Kyuchumimo for reporting), and then I used Claude to implement the code fixes and tests, along with asking it to find other instances of the problem (pixelarray stuff).

I then wrote a verification script by hand as well, to demonstrate the changes.

<details><summary>Script</summary>

```py
import pygame
pygame.init()

surfs = []

# RGB/BGR names are confusing here. I believe what we would intuitively call 
# BGR would be RGB24 in SDL pixelformat parlance

surfs.append((pygame.Surface((100,100), depth=24, masks=(0xFF, 0xFF00, 0xFF0000, 0)), "rgb24"))
surfs.append((pygame.Surface((100,100), depth=24, masks=(0xFF0000, 0xFF00, 0xFF, 0)), "bgr24"))
surfs.append((pygame.Surface((100,100), depth=32), "32bit"))

def test_pixelarray_replace():
    print("Testing replace()")

    for surf, name in surfs:
        print(name, surf, surf.get_shifts())

        surf.fill("red")

        with pygame.PixelArray(surf) as px:
            px.replace(pygame.Color("red"), (40,60,80))

        print(surf.get_at((0,0)), surf.get_at((0,0)) == (40,60,80))
    print()


def test_pixelarray_assign_sequence():
    print("Testing assign_sequence")

    for surf, name in surfs:
        print(name, surf, surf.get_shifts())

        surf.fill("red")

        with pygame.PixelArray(surf) as px:
            px[0:20] = [(40,60,80) for _ in range(20)]

        print(surf.get_at((0,0)), surf.get_at((0,0)) == (40,60,80))
    print()


def test_pixelarray_assign_slice():
    print("Testing assign_slice")

    for surf, name in surfs:
        print(name, surf, surf.get_shifts())

        surf.fill("red")

        with pygame.PixelArray(surf) as px:
            px[0:20] = (40,60,80)

        print(surf.get_at((0,0)), surf.get_at((0,0)) == (40,60,80))
    print()


def test_pixelarray_assign_item():
    print("Testing assign_item")

    for surf, name in surfs:
        print(name, surf, surf.get_shifts())

        surf.fill("red")

[...]

PR #3791, opened 2026-05-10
Fix laplacian dst format and simplify 24bit writes

Starbuck5:
This is a follow-up to https://github.com/pygame-community/pygame-ce/pull/3790

CodeRabbit said the bugfix in that PR was lacking from laplacian, and it was wrong about that. However, going through laplacian found that it also has a bug with writing pixel data-- it is using the wrong surface's information about what to write.

For laplacian--
Previously, the color gathered in acolor was mapped using the source surface's information, and then written into the destination surface. Since destination surfaces are restricted to the same depth, this did work most of the time, but would not work if two surfaces had the same depth but different pixel layouts. Except for in the 24 bit path, which would map RGB, then unmap with source format, then remap into channels in destination format.

Simplified 24 bit writes--
These are using mapped colors, they can be stored directly into the pixel buffer. There is no need to "unmap" them using the pixelformat and store them in the red, green, and blue channels separately.
(This is related because simplifying the 24 bit writes broke the laplacian test, because of the aforementioned bug)

### DIFF
--- a/test/transform_test.py
+++ b/test/transform_test.py
@@ -1269,6 +1269,39 @@ class TransformModuleTest(unittest.TestCase):
         finally:
             pygame.display.quit()
 
+    def test_laplacian__cross_format(self):
+        # Regression: laplacian used to map the filtered RGB with the source
+        # format and write the bytes directly into the destination, which
+        # swapped channels when the surfaces had the same bytes-per-pixel
+        # (so the size check passed) but different channel layouts.
+        SIZE = 32
+        # RGBA layout for the source (R at the low byte).
+        s1 = pygame.Surface(
+            (SIZE, SIZE),
+            0,
+            32,
+            masks=(0xFF, 0xFF00, 0xFF0000, 0xFF000000),
+        )
+        s1.fill((10, 10, 70))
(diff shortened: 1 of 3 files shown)


test/transform_test.py
```diff
@@ -1269,6 +1269,39 @@ def test_laplacian__24_big_endian(self):
...
+        pygame.draw.line(s1, (255, 0, 0), (0, 31), (31, 31))
+
+        # BGRA layout for the destination — different channel order.
+        s2 = pygame.Surface(
+            (SIZE, SIZE),
+            0,
+            32,
+            masks=(0xFF0000, 0xFF00, 0xFF, 0xFF000000),
+        )
+        pygame.transform.laplacian(s1, s2)
```
oddbookworm, line 1296: I kinda agree with the 🤖 here in that it would be nice to verify that the two surfaces have different channel formats before and that they're the same afterward, but I'm not going to hold the pull for that tiny little thing

oddbookworm (approved):
One small nitpick comment, but LGTM! Thanks! 🎉

Matiiss (approved):
LGTM!

PR #3804, opened 2026-05-17
Remove red or black tutorial

zoldalma999:
This tutorial has pretty bad formatting issues (code is in images instead of text, and then every image has the same random code snippet next to it for some reason), is not accessible in the documentation other than a direct URL (for the english version) and is probably not used much. Removing it reduces the wheel size by ~650kb for me (from 10264 on current main to 9610 on this branch).


oddbookworm (approved):
LGTM! Thanks @zoldalma999 🎉

Matiiss (approved):
LGTM!
I was actually planning on doing the same (among other things) in my currently-draft docs PR (https://github.com/pygame-community/pygame-ce/pull/3751), too, for similar reasons.

PR #3809, opened 2026-05-25
Improve mixer stub/docs accuracy

Starbuck5:
Fixes a few issues where the stubs and docs do not properly match the implementation.

I noticed some of these disparities when working on an SDL3 compat layer.

I did something fancy with find_channel so that if someone sets force=True, the type hinter will know that they get a Channel back, rather than a Channel or None.


damusss (approved):
LGTM, thanks!

aatle (approved):
Verified the changes. LGTM!

PR #3811, opened 2026-05-25
Remove obsolete mousewheel flip compat

Starbuck5:
According to SDL2 whatsnew (https://github.com/libsdl-org/SDL/blob/SDL2/WhatsNew.txt), the direction field has existed since SDL 2.0.4. (Our minimum SDL version is 2.0.14)

So, there is no need to test for the existence of the SDL_MOUSEWHEEL_FLIPPED enum value, and have a codepath if it is not found.

- ancient SDL: FLIPPED is not defined, so NO_SDL_MOUSEWHEEL_F is defined, so it hardcodes false.
- SDL: FLIPPED is defined, so NO_SDL_MOUSEWHEEL_F is not defined, so it checks direction.

### DIFF
--- a/src_c/event.c
+++ b/src_c/event.c
@@ -1345,13 +1345,9 @@ dict_from_event(SDL_Event *event)
             break;
         case SDL_MOUSEWHEEL:
             /* https://wiki.libsdl.org/SDL_MouseWheelEvent */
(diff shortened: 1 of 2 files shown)


oddbookworm (approved):
Seems reasonable to me

damusss (approved):
Sounds good!

PR #3812, opened 2026-05-25
Fix Sound.play return type to Channel | None (#3810)

### LINKED ISSUE #3810: Sound.play docs/types are wrong
The docs:

> Begin playback of the Sound (i.e., on the computer's speakers) on an available Channel. This will forcibly select a Channel, so playback may cut off a currently playing sound if necessary.

The stubs:

```py
    def play(
        self,
        loops: int = 0,
        maxtime: int = 0,
        fade_ms: int = 0,
    ) -> Channel: ...
```

The source:
```c
    Py_BEGIN_ALLOW_THREADS;
    if (fade_ms > 0) {
        channelnum =
            Mix_FadeInChannelTimed(-1, chunk, loops, fade_ms, playtime);
    }
    else {
        channelnum = Mix_PlayChannelTimed(-1, chunk, loops, playtime);
    }
    Py_END_ALLOW_THREADS;
    if (channelnum == -1) {
        Py_RETURN_NONE;
    }
```

If there are no free channels, SDL_mixer returns -1, and None is returned.

I would've handled this in https://github.com/pygame-community/pygame-ce/pull/3809 but it seems more fundamental to me, so I wanted to raise it separately. The documented description is how I always thought this worked... is it how it **should** work? I guess implementation should reign supreme, and no one has complained about this before.

Also maybe this will all be made irrelevant by SDL3 soon.

ivandebugr:
Fixes #3810.

Sound.play() can return None when no free channel is available, but the stub said `-> Channel` and the docs claimed it 'forcibly' selects one. The C implementation in `src_c/mixer.c` returns None when `Mix_PlayChannelTimed` / `Mix_FadeInChannelTimed` returns -1.

## Changes
- `mixer.pyi`: `Sound.play` return type → `Channel | None`
- `mixer.rst`: drop the inaccurate 'forcibly select' line; document the None case and point users to `find_channel(force=True)`
- `mixer_doc.h`: regenerated to match

## Note on PR #3809
This touches the same three files as #3809 but at different lines (#3809 fixes `find_channel`, `get_init`, `get_sound`, `get_queue`). No semantic overlap; minor textual rebase only.

### DIFF
--- a/docs/reST/ref/mixer.rst
+++ b/docs/reST/ref/mixer.rst
@@ -384,11 +384,10 @@ The following file formats are supported
    .. method:: play
 
       | :sl:`begin sound playback`
-      | :sg:`play(loops=0, maxtime=0, fade_ms=0) -> Channel`
+      | :sg:`play(loops=0, maxtime=0, fade_ms=0) -> Channel | None`
 
       Begin playback of the Sound (i.e., on the computer's speakers) on an
-      available Channel. This will forcibly select a Channel, so playback may
-      cut off a currently playing sound if necessary.
+      available Channel.
 
       The loops argument controls how many times the sample will be repeated
       after being played the first time. A value of 5 means that the sound will
@@ -404,7 +403,10 @@ The following file formats are supported
       fade up to full volume over the time given. The sample may end before the
       fade-in is complete.
 
-      This returns the Channel object for the channel that was selected.
+      Returns the Channel object for the channel that was selected, or
+      ``None`` if no Channel was available to play the Sound (i.e. every
+      channel was already busy). Use :func:`pygame.mixer.find_channel` with
+      ``force=True`` first if you need a guaranteed Channel.
 
       .. ## Sound.play ##
 
--- a/buildconfig/stubs/pygame/mixer.pyi
+++ b/buildconfig/stubs/pygame/mixer.pyi
@@ -56,7 +56,7 @@ class Sound:
         loops: int = 0,
         maxtime: int = 0,
         fade_ms: int = 0,
-    ) -> Channel: ...
+    ) -> Channel | None: ...
     # possibly going to be deprecated/removed soon, in which case these
     # typestubs must be removed too
     @property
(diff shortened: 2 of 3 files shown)


docs/reST/ref/mixer.rst
```diff
@@ -404,7 +403,10 @@ The following file formats are supported
       fade up to full volume over the time given. The sample may end before the
       fade-in is complete.
 
-      This returns the Channel object for the channel that was selected.
+      Returns the Channel object for the channel that was selected, or
+      ``None`` if no Channel was available to play the Sound (i.e. every
+      channel was already busy). Use :func:`pygame.mixer.find_channel` with
+      ``force=True`` first if you need a guaranteed Channel.
```
oddbookworm, line 409: I don't see how `find_channel` does anything here. It doesn't stop playback if there are no playing channels, it'll just return the channel that's been playing the longest. It does not guarantee you can use that channel unless you manually stop playback on that channel

Starbuck5: According to the channel.play documentation, "If the Channel is currently playing any other Sound it will be stopped."

Starbuck5:
Hello,

The purpose I had opening #3810 was to start a discussion on what we should do, not to prompt a PR immediately. We should decide what to do, then do it -- and it doesn't need to happen instantly, it seems likely that this behavior has been misstated for 20 years (someone could do that research and add it to 3810!). So I'm somewhat tempted to close this PR.

(Even though the solution will probably just end up being this)

docs/reST/ref/mixer.rst
oddbookworm: Sure, if you play on the channel directly. But that's not what this change is implying

ivandebugr:
Thanks for the feedback! I understand now that #3810 was meant to open a discussion first. I'm happy to keep this PR as a draft for when the approach is decided, or close it entirely if that's preferable. I'll wait for consensus on #3810 before pushing further changes!

PR #3814, opened 2026-05-27
Add docstrings to SDL3 _audio module

Starbuck5:
This PR advances the readiness of our SDL3 _audio module by adding docstrings to most functions/methods/etc. This is so in the future we can generate docs from these docstrings.

In several places I copied SDL docs word for word, because it is well written and very relevant. Although all my school training was saying "plagiarism bad." 

--------------

Why does this not generate docs? 

In my local workspace I added _audio.rst in docs/reST/ref:
```
.. include:: common.txt

:mod:`pygame._audio`
====================

.. autopgmodule:: pygame._audio
    :members:
```

But it wouldn't generate anything because it sees `_audio` as a private module. So I also had to add `
autoapi_options = ['members', 'undoc-members', 'private-members']` in conf.py. I also was unsure about adding docs to the build that are not relevant to SDL2 builds, even hidden docs.

--------------

However, it is probably way easier to review as HTML then going through the code, so I've attached 
[generated.zip](https://github.com/user-attachments/files/28299899/generated.zip). This is a copy of my generated docs on this branch. So docstrings here + additional changes to get html generating not committed. To check it out, unzip it and click on ref/_audio.html.

### DIFF
--- a/src_c/_base_audio.c
+++ b/src_c/_base_audio.c
@@ -463,6 +463,12 @@ pg_audio_get_audio_stream_data(PyObject *module, PyObject *const *args,
         return PyBytes_FromStringAndSize("", 0);
     }
 
+    // Allocates working space, then SDL copies into it, then we copy into a
+    // bytes object.
+    // It would be more efficient to create the bytes object and then write
+    // into it directly, but I didn't want to use _PyBytes_Resize (soft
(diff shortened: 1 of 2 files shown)


damusss (approved):
The docs do an amazing job at explaining how the system works, thank you!

Starbuck5:
Thanks for taking a look, Damus.

PR #3816, opened 2026-05-29
Fixed certain render tests failing when using specific macos versions

AndreyVarvar:
Some render tests were failing due to the fact that the renderer was not cleared before performing said tests. This causes the renderer to contain garbage data on some macos computers, and since some tests expect the untouched renderer data to be pure black, they obviously fail.

### DIFF
--- a/test/render_test.py
+++ b/test/render_test.py
@@ -47,6 +47,8 @@ class RendererTest(unittest.TestCase):
             self.renderer.to_surface(small_surf)
 
     def test_blit(self):
+        self.renderer.draw_color = "BLACK"
+        self.renderer.clear()
         surface = pygame.Surface((10, 10))
         surface.fill((80, 120, 160, 200))
         texture = _render.Texture.from_surface(self.renderer, surface)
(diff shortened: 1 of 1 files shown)


test/render_test.py
```diff
@@ -47,6 +47,8 @@ def test_to_surface(self):
             self.renderer.to_surface(small_surf)
 
     def test_blit(self):
+        self.renderer.draw_color = "BLACK"
+        self.renderer.clear()
```
oddbookworm: What happens if you put this code in `setUp` instead, so that it runs before every single test?

AndreyVarvar: huh, yeah, I think that's actually better... I thought setUp was ran only once and not before every single test

oddbookworm: [It runs before every test, and tearDown runs after every test](https://docs.python.org/3/library/unittest.html#unittest.TestCase.setUp)

oddbookworm (approved):
This fixes the tests, which is what you intended to do, so LGTM! I would like at some point for there to be a C-side fix if there is an actual issue here about renderers having weird data when created, but that's for another day

Thanks! 🎉

ankith26 (approved):
LGTM, thanks! 🎉

PR #3817, opened 2026-05-29
Added web as imcompatible with Briefcase

### LINKED ISSUE #3805: Add Web as a Briefcase incompatibility
https://github.com/pygame-community/pygame-ce/blob/d999e346109f2702ee8db8223c2ea75058b6ec4d/src_py/__briefcase/pygame_ce.py#L5

Should specify Web here as an incompatibility. Yes, we can deploy to web, but not through Briefcase

scooggins:
Changed display_name_annotation in src_py/__briefcase/pygame_ce.py to include web as incompatible, matching previous style guides.

### DIFF
--- a/src_py/__briefcase/pygame_ce.py
+++ b/src_py/__briefcase/pygame_ce.py
@@ -2,7 +2,7 @@ from briefcase.bootstraps.base import BaseGuiBootstrap
 
 
 class PygameCEGuiBootstrap(BaseGuiBootstrap):
-    display_name_annotation = "does not support iOS/Android deployment"
+    display_name_annotation = "does not support iOS/Android/Web deployment"
 
     def app_source(self):
         return """\


oddbookworm (approved):
I mean, it would be really hard to get this wrong lol. Thanks! 🎉 
LGTM!

Starbuck5 (approved):

ankith26 (changes requested):
In theory this is something we could potentially support, but not recommend. Briefcase uses pyscript and that runs on top of pyodide which is something that does work right now.

ankith26 (approved):
So it turns out that briefcase uses an old pyscript/pyodide version and there an `import pygame` fails. To achieve briefcase support on web we would need to either wait for briefcase to update or pin to a newer version ourselves. And probably a bit more work will be required from our side as well.

Given the status quo this PR is appropriate to merge right now, thanks for contributing

PR #3827, opened 2026-06-17
Added function normalize_or_zero

andy-mf:
This feature is mentioned in #2269 which suggests that there should be method to normalise vectors automatically returning zero if a zero-vector is passed to it, as in other engines like Godot and Unity. This is useful as it allows to avoid manually checking if the vector is zero outside of the normalisation function (this may be too much of an abstraction, people were split in the issue thread).

Note, this implementation repeats the normalisation code, so it may be a more modular implementation.  An alternative and cleaner solution mentioned in the thread discussion is passing an argument to the `normalize` function (i.e. `vec.normalize(zero=True)`) although this would require unpacking arguments.


oddbookworm:
Thank you for the pull request, but if you read the comments in the issue, it's not clear that there was a consensus that this is even a feature we should add or what the actual API should be. There should be some kind of general agreement before we even think about implementing it.

oddbookworm:
@andy-mf thanks for the pull request, but the path forward for this issue right now is going to be that we won't support it in 2.x versions of pygame-ce, but 3.x versions will support the zero vector natively in `normalize`.

PR #3829, opened 2026-06-21
Fix typo in vector __str__ NotImplementedError message

### LINKED ISSUE #3828: Vector `__str__` implementation has a typo about it being a `__repr__` method
While answering a question on the PGC discord, I stumbled on this typo in the generic vector `__str__` method where it refers to itself as `__repr__`. Almost certainly a copy/paste issue. And ideally, this error will never be thrown in the first place because you cannot construct a vector not of size 2 or 3, but it's still something I'd like to see fixed. Excellent first issue for someone to take.

https://github.com/pygame-community/pygame-ce/blob/26dd14d2e9226c468e0b056f64df30a6a1447d97/src_c/math.c#L1971

MeGaurav4:
While reviewing issue #3828, I noticed the `NotImplementedError` raised by
`vector_str` in `src_c/math.c` (line 1971) incorrectly referred to itself as
`repr()`. The function is registered as `tp_str` (Python's `__str__`), so the
message should say `str()`.

The corresponding `repr()` function on line 1940 correctly mentions
`repr()`, so this only touches the str branch.

Closes #3828

## Testing

- Verified the diff only changes the str function (line 1971)
- Did not run the C build since this is a one-character text fix in an error
  message, no logic change

### DIFF
--- a/src_c/math.c
+++ b/src_c/math.c
@@ -1968,7 +1968,7 @@ vector_str(pgVector *self)
     else {
         return RAISE(
             PyExc_NotImplementedError,
-            "repr() for Vectors of higher dimensions are not implemented yet");
+            "str() for Vectors of higher dimensions are not implemented yet");
     }
 
     if (!_vector_check_snprintf_success(tmp, STRING_BUF_SIZE_STR)) {


ankith26 (approved):
LGTM, thanks for contributing! 🎉

oddbookworm (approved):
LGTM!

PR #3833, opened 2026-06-28
Update to CIBW 4.1.0

Starbuck5:
This PR contains patches on top of dependabot's https://github.com/pygame-community/pygame-ce/pull/3832.

The pyodide build was broken because we're manually editing a specific emscripten settings file that is now in a different location, I re-pointed the path.

The Windows build was broken because they are now using delvewheel to try to fix up deps like auditwheel or delocate on Mac and Linux. I turned this off, to preserve the status quo of our builds.

### DIFF
--- a/.github/workflows/build-emsdk.yml
+++ b/.github/workflows/build-emsdk.yml
@@ -77,7 +77,7 @@ jobs:
       with:
         persist-credentials: false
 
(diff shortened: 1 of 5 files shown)


ankith26 (approved):
LGTM, thanks!

PR #3841, opened 2026-07-07
Small stub adjustment for mixer.Sound.play()

kristiansroberts:
I did read where this could be rendered irrelevant soon, but I wanted to go ahead and address a tiny change I noticed from issue #3810 in the stubs and the docs for the mixer.Sound.play() method. It currently asserts that it returns a Channel, however it is possible for the method to return None, for instance if a channel is exceeding the set limit. So I just wanted to adjust the stubs and docs to indicate a None return is indeed possible.


ankith26 (commented):
Thanks for taking time to contribute to pygame-ce!

While this change is technically sound, what we also need to do is discuss (as the issue intends) whether we need to revaluate this behaviour. Changing signature and stubs makes it inconsistent with the actual doc description which states that the channel will be selected forcibly.

kristiansroberts:
Oh okay, I understand, so is the discussion about whether the c code should be adjusted so that it always finds a channel instead of ever returning None so it matches the current documentation? I could try a potential solution for that as well.

PR #3843, opened 2026-07-09
mixer/music module SDL3

Starbuck5:
This is at least an initial version of a compatibility layer for these modules running on SDL3 primitives ported in prior PRs. It's not perfect, but I really think it is good enough to support a lot of pygame-ce/SDL3 testing. Like we can run examples with audio now, and real games, and they don't just crash because pygame.mixer ImportError. This should be very helpful to identifying problems with the SDL3 port.

~~This PR is based on https://github.com/pygame-community/pygame-ce/pull/3753, so it includes those changes as well (that's why the diff line count is so high)~~

This comes in 3 parts:
Implement _sdl3_mixer.AudioDecoder class 
- I needed this for Sound.get_raw. SDL3_mixer provides to method to get the raw data out of an Audio object, so I need to decode the audio first, store it in a buffer, then send that buffer to SDL3_mixer.

Implement _sdl3_mixer.Track.set_stopped_callback
- I needed this for set_endevent on Channel/music + queued musics

Add initial version of pygame2 mixer compat shim SDL3

I'm thinking it would make sense to split this PR into just the additional _sdl3_mixer functionality as one, and then the Python stuff as another?


src_py/mixer.py
```diff
@@ -0,0 +1,829 @@
...
+
+    return (mix_spec.frequency, realform, mix_spec.channels)
+
+
+def get_driver() -> str:
+    MixerInternals.init_check()
+    return _audio.get_current_driver()
+
+
+def set_num_channels(count: int, /) -> None:
```
damusss, line 225: In a few early functions in this file `MixerInternals.init_check` is not used and the check that that method replaces is used instead (I noticed `set_num_channels`, `get_num_channels`, `set_reserved`, and `Channel.__new__`).

src_py/mixer.py
```diff
@@ -0,0 +1,829 @@
...
+        self._track: _sdl3_mixer.Track | None = None
+
+        self._queued_audio: _sdl3_mixer.Audio | None = None
+        self._queued_loops = 0
+
+        self._end_event = 0
+
+    def _stopped_callback(self, _: _sdl3_mixer.Track, __: None) -> None:
+        if self._end_event and pygame.display.get_init():
+            pygame.event.post(pygame.Event(0, {}))
```
damusss: I was about to specify the missing end event in the post, but AI already caught that. I'll still leave this comment and mark it as resolved later.

src_py/mixer.py
```diff
@@ -0,0 +1,829 @@
...
+
+        return self._track.playing
+
+    def set_pos(self, pos: float, /) -> None:
+        if self._track is None:
+            raise pygame.error("mixer not initialized")
+
+        self._track.set_playback_position(self._track.ms_to_frames(round(pos)))
+
+    def get_pos(self) -> int:
```
damusss, line 825: `MusicImplementation.get_pos` is correct, but I think it's not a real compat method, in the sense that from what I know `pygame.mixer.music.get_pos` doesn't truly return the playback position, but instead the time from when `play()` was called (which does not take into account an playback position sets that happened afterwards), because `pygame.mixer.music.get_pos` was added _before_ the corresponding SDL music function was created. I'm not 100% confident of my words, as I may be forgetting some later change, but after checking the current C implementation I notice that it's doing calculation with ticks instead of calling the proper SDL function, making me think it's still currently wrong. Should the compat layer keep the error, or just not care and return the correct value? (I prefer the latter!)

src_py/mixer.py
```diff
@@ -0,0 +1,829 @@
...
+            self._audio = self._queued_audio
+            self._queued_audio = None
+
+            self._track.set_audio(self._audio)
+            self._track.play(loops=self._queued_loops)
+
+            self._queued_loops = 0
+
+    def load(self, filename, namehint: str = "") -> None:
+        if self._track is None:
```
damusss, line 742: The `if self._track is None: raise ...` check is repeated very often. Wouldn't it be better to add a `MusicImplementation._init_check` in a similar fashon to `MixerInternals`?

src_py/mixer.py
```diff
@@ -0,0 +1,829 @@
...
+            raise pygame.error("mixer not initialized")
+
+        self._audio = _sdl3_mixer.Audio(
```
damusss, line 745: Why does `MusicImplementation.load` load an audio? shouldn't it do `Track.set_filestream`, to prevent the music from being preloaded in RAM? Unless the way the audio is being created already prevents it of course.

damusss (changes requested):
I specified a few things I noticed in the comments below.

Thank you for working on the compat layer, it's amazing how close we are to full SDL3 compatibility, especially with your contributions!

src_py/mixer.py
Starbuck5: I need it an audio for get_metadata. Audio objects have a `predecode` parameter (default False), which I've not overridden, so it is telling SDL_mixer to not preload it.

src_py/mixer.py
Starbuck5: Resolved, I believe.

src_py/mixer.py
Starbuck5: I think the compat layer should keep the error, but right now I just added a note about it into the code. Good spot.

src_py/mixer.py
Starbuck5: Fixed. I did not have time tonight to fix the tests to make that be caught programmatically.

src_py/mixer.py
```diff
@@ -0,0 +1,829 @@
...
+        if self._track is None:
+            raise pygame.error("mixer not initialized")
+
+        return self._track.playing
+
+    def set_pos(self, pos: float, /) -> None:
+        if self._track is None:
+            raise pygame.error("mixer not initialized")
+
+        self._track.set_playback_position(self._track.ms_to_frames(round(pos)))
```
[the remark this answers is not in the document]
Starbuck5: Fixed

src_py/mixer.py
```diff
@@ -0,0 +1,829 @@
...
+        z = math.cos(angle) * distance
+        self._track.set_3d_position((x, 0, z))
+
+    def set_volume(self, arg1: float, arg2: float | None = -1.0, /):
+        if arg2 == -1.0:
+            self._track.gain = arg1
+            return
+
+        if arg2 == None:
+            arg2 = arg1
```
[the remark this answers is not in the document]
Starbuck5: Fixed

src_py/mixer.py
```diff
@@ -0,0 +1,829 @@
...
+            for i, v in enumerate(samples):
+                # promote 8 bit to 16 bit in special case
+                ov[i] = (v << 8) & 0xFFFF
+        else:
+            out = bytearray(len(samples) * 2)
+            ov = memoryview(out).cast("H")
+            for i, v in enumerate(samples):
+                # convert 2/4 byte values to 2 bytes
+                ov[i] = v & 0xFFFF
+
```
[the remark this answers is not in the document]
Starbuck5: It's the status quo in mixer.c

src_py/mixer.py
```diff
@@ -0,0 +1,829 @@
...
+        if potential_devices:
+            mixer_device = potential_devices[0]
+
+    MixerInternals.initialized = True
+    MixerInternals.mixer_buf_spec = mixer_spec
+    MixerInternals.mixer = _sdl3_mixer.Mixer(mixer_device, mixer_spec)
+    MixerInternals.channels = [Channel._allocate(i) for i in range(8)]
+    MixerInternals.reserved_channels = 0
+
+    music._init()
```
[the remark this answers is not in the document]
Starbuck5: Fixed

test/mixer_test.py
```diff
@@ -18,6 +18,11 @@
 FREQUENCIES = [11025, 22050, 44100, 48000]
 SIZES = [-16, -8, 8, 16]  # fixme
 # size 32 failed in test_get_init__returns_exact_values_used_for_init
+
+# SDL3 does not support unsigned 16 bit audio
+if pygame.get_sdl_version()[0] > 2:
+    SIZES.remove(16)
+
```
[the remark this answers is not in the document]
Starbuck5: I'm not worried about reimports at all, but easy to fix, so whatever.

src_py/mixer.py
```diff
@@ -0,0 +1,829 @@
...
+        self._track: _sdl3_mixer.Track | None = None
+
+        self._queued_audio: _sdl3_mixer.Audio | None = None
+        self._queued_loops = 0
+
+        self._end_event = 0
+
+    def _stopped_callback(self, _: _sdl3_mixer.Track, __: None) -> None:
+        if self._end_event and pygame.display.get_init():
+            pygame.event.post(pygame.Event(0, {}))
```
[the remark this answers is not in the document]
Starbuck5: Fixed

src_c/_sdl3_mixer_c.c
```diff
@@ -0,0 +1,2030 @@
...
+pg_mixer_obj_dealloc(PGMixerObject *self)
+{
+    MIX_DestroyMixer(self->mixer);
+    self->mixer = NULL;
+    PyObject_GC_UnTrack(self);
+    PyTypeObject *tp = Py_TYPE(self);
+    freefunc free = PyType_GetSlot(tp, Py_tp_free);
+    free(self);
+    Py_DECREF(tp);
+}
```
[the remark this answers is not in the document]
Starbuck5: What? there's no Py_BEGIN_ALLOW_THREADS here.

I'm not worried about threads with this, pygame-ce is not thread safe in general.

src_c/_sdl3_mixer_c.c
```diff
@@ -0,0 +1,2030 @@
...
+
+    /* Now that the SDL track is gone, it's safe to drop the Python references
+     * it depended on. */
+    Py_CLEAR(self->stopped_callback);
+    Py_CLEAR(self->stopped_callback_userdata);
+    Py_CLEAR(self->mixer_obj);
+    Py_CLEAR(self->source_obj);
+
+    return 0;
+}
```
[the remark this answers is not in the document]
Starbuck5: I don't want to have other threads run in the middle of my object deconstruction. I have left a TODO at the top of the file to make sure track object deallocation and the callback don't interfere.

src_py/mixer.py
damusss: Fair enough then, I felt like that predecode was doing the magic but I wasn't 100% sure :)

Starbuck5:
Hey @damusss I am on work travel for the next 2 weeks so I will not be able to put appreciable time into this for a second.

If you have time and desire I’d be happy to see you continue this work. It seems like there’s a piece of work on improving the mixer tests so that entire methods can’t be missing without failing. I started a bit of that in this branch but now some of those are failing on SDL2, which didn’t happen on my system? 

There also a few mixer methods I’ve missed because of the aforementioned lack of tests.

Then there is stuff here that could be broken out and merged, like the AudioDecoder class would probably be straightforward as a separate change.

anyways feel free to push to this branch if you’d like.

PR #3844, opened 2026-07-13
docs(surfarray): document float array rounding in blit_array/make_surface

### LINKED ISSUE #3683: `surfarray.blit_array` performs undocumented rounding of float arrays
```
def blit_array(surface, array):
    """pygame.surfarray.blit_array(Surface, array): return None

    Blit directly from an array of values.

    Directly copy values from an array into a Surface. This is faster than
    converting the array into a Surface and blitting. The array must be the
    same dimensions as the Surface and will completely replace all pixel
    values. Only integer, ascii character and record arrays are accepted.

    This function will temporarily lock the Surface as the new values are
    copied.
    """
    if isinstance(array, numpy_ndarray) and array.dtype in numpy_floats:
        array = array.round(0).astype(numpy_uint32)
    return array_to_surface(surface, array)
```

it's explicitly mentioned that `Only integer, ascii character and record arrays are accepted.`, but pygame-ce will also accept floats and implicitly run a costly conversion instead, which is surprising because this is meant to be a more performance oriented API. I can't seem to find anywhere where this behaviour is documented, I discovered it by accident myself when debugging a path that mistakenly outputted a float array, and performance tanked implicitly as a result.

jryyangjy:
## Fixes

Fixes #3683

## Problem

`surfarray.blit_array` and `surfarray.make_surface` both silently accept NumPy float arrays:

```python
if isinstance(array, numpy_ndarray) and array.dtype in numpy_floats:
    array = array.round(0).astype(numpy_uint32)
```

...even though both docstrings/reference docs state only integer, ASCII character, or record arrays are accepted. This is undocumented anywhere, and as the issue reporter found, it's a silent perf trap: passing a float array by mistake causes an implicit rounding-and-copy pass, defeating the whole point of these being the "fast path" functions.

Confirmed the current (2.5.7) behavior directly:

```python
arr = np.zeros((4, 4), dtype=np.float64)
arr[:] = 123.7
surfarray.blit_array(surf, arr)   # succeeds silently, no warning
surf.get_at((0, 0))               # -> Color(0, 0, 124, 255), i.e. rounded
```

## Fix

Doc-only change, no behavior change:
- Updated the `blit_array` and `make_surface` docstrings in `src_py/surfarray.py` to state that float arrays are accepted, rounded, and copied -- and that this defeats the performance benefit of `blit_array`.
- Mirrored the same wording in `docs/reST/ref/surfarray.rst`.

## Test plan

- Verified current behavior against the installed 2.5.7 wheel (float array silently accepted and rounded, as shown above).
- Parsed the edited `.rst` with `docutils` to confirm no new syntax errors were introduced (pre-existing Sphinx-only directives like `.. function::`/`:sl:`/`:mod:` predictably show as "unknown" under bare docutils across the whole file, not just the edited sections -- this is expected without a full Sphinx build).
- `ast.parse()` on the edited `surfarray.py` to confirm valid Python syntax.
- I don't have a full local build of the pygame-ce C extensions on this machine, so I did not attempt `dev.py docs --full`; this change doesn't touch any C code or runtime behavior.

### DIFF
--- a/docs/reST/ref/surfarray.rst
+++ b/docs/reST/ref/surfarray.rst
@@ -267,6 +267,9 @@ will be locked during the lifetime of the array.
    make_surface uses the array struct interface to acquire array properties,
    so is not limited to just NumPy arrays. See :mod:`pygame.pixelcopy`.
 
+   NumPy float arrays are also accepted, but are first rounded to the
+   nearest integer and copied into a new array.
+
    New in pygame 1.9.2: array struct interface support.
 
    .. ## pygame.surfarray.make_surface ##
@@ -280,6 +283,11 @@ will be locked during the lifetime of the array.
    converting the array into a Surface and blitting. The array must be the same
    dimensions as the Surface and will completely replace all pixel values. Only
    integer, ASCII character and record arrays are accepted.
+
+   NumPy float arrays are also accepted, but are first rounded to the
+   nearest integer and copied into a new array. This extra copy defeats
+   the performance benefit this function is meant to provide, so pass an
+   integer array directly whenever possible.
 
    This function will temporarily lock the Surface as the new values are
    copied.
--- a/docs/reST/ref/surfarray.rst
+++ b/docs/reST/ref/surfarray.rst
@@ -267,8 +267,9 @@ will be locked during the lifetime of the array.
    make_surface uses the array struct interface to acquire array properties,
    so is not limited to just NumPy arrays. See :mod:`pygame.pixelcopy`.
 
-   NumPy float arrays are also accepted, but are first rounded to the
-   nearest integer and copied into a new array.
+   ``float32``, ``float64`` and (where available) ``float96`` NumPy arrays
+   are also accepted, but are first rounded to the nearest integer and
+   copied into a new array before creating the surface.
 
    New in pygame 1.9.2: array struct interface support.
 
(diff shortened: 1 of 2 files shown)


Starbuck5 (approved):
Thanks for taking a look at this, I'm good to go with it.

However, I don't think it fully resolves #3683, just makes it better. Since it doesn't do a warning or decisively figure we don't need a warning. But that could make it more complicated, this is good.

I put a comment on the parent issue as well about going to "any numpy float array" instead of certain hardcoded values.

docs/reST/ref/surfarray.rst
```diff
@@ -278,8 +282,15 @@ will be locked during the lifetime of the array.
...
-   integer, ASCII character and record arrays are accepted.
+   dimensions as the Surface and will completely replace all pixel values.
+   Accepted array types are integer, ASCII character and record arrays, as
+   well as ``float32``, ``float64`` and (where available) ``float96`` NumPy
+   arrays.
+
+   Float arrays are first rounded to the nearest integer and copied into
+   a new array before blitting. This extra copy defeats the performance
+   benefit this function is meant to provide, so pass an integer array
+   directly whenever possible.
```
oddbookworm: > This extra copy defeats the performance benefit this function is meant to provide

Do you have an actual benchmark or analysis to back this up? This is a claim that cannot be put into documentation without rock-solid justification

oddbookworm: Ok, so I tested this myself. I found that passing an array of ints is 1.5-3x faster than passing an array of floats. But, I found that passing an array of floats is still almost 300x faster than a native python implementation, so this claim seems like it's trying to imply the float array would be as slow as the python implementation, which just isn't true in my own testing
@jryyangjy

oddbookworm (approved):
LGTM with that quick fix in! 🎉

PR #3852, opened 2026-07-25
Raise minimum SDL dependency versions

### LINKED ISSUE #3794: Raise minimum SDL dependency versions
In the README, it specifies that we rely on

SDL >= 2.0.14
SDL_mixer  >= 2.0.4
SDL_image >= 2.0.4
SDL_ttf >= 2.0.15

This is because we were using the versions present as Debian Bullseye packages. However, we've dropped support for Python 3.9 (the version Bullseye comes with), so we should move our minimum dependencies to align with a distro that ships with 3.10, Ubuntu Jammy.

Jammy comes with [SDL 2.0.20](https://packages.ubuntu.com/jammy/libsdl2-2.0-0), [SDL_mixer 2.0.4](https://packages.ubuntu.com/jammy/libsdl2-mixer-2.0-0), [SDL_image 2.0.5](https://packages.ubuntu.com/jammy/libsdl2-image-2.0-0), and [SDL_ttf 2.0.18](https://packages.ubuntu.com/jammy/libsdl2-ttf-2.0-0).

See https://github.com/pygame-community/pygame-ce/pull/3213 for an example of what a dependency raise looks like.

ochderek:
## Summary

- raise the minimum SDL version from 2.0.14 to 2.0.20
- raise SDL_image from 2.0.4 to 2.0.5
- raise SDL_ttf from 2.0.15 to 2.0.18
- remove compatibility code for older SDL versions

## Testing

- built and installed pygame-ce successfully
- verified pygame-ce 2.5.8.dev1 with SDL 2.0.20
- ran `test/render_test.py` and `test/window_test.py` using SDL's dummy video driver and software renderer
- 55 passed, 7 skipped


src_c/cython/pygame/_sdl2/video.pxd
```diff
@@ -374,7 +340,6 @@ cdef extern from "SDL.h" nogil:
                                    float *lx,
                                    float *ly);
 
-    int SDL_VERSION_ATLEAST(int major, int minor, int patch)
```
oddbookworm, line 377: Nope. I don't know why you're messing in places unrelated to the pull request, but you should not randomly be removing things just because you see them (at least not without a full explanation as to why it matters in this pull request)

README.rst
```diff
@@ -125,13 +125,13 @@ Dependency versions:
...
-| SDL      | >= 2.0.14              |
+| SDL      | >= 2.0.20              |
 +----------+------------------------+
 | SDL_mixer| >= 2.0.4               |
 +----------+------------------------+
-| SDL_image| >= 2.0.4               |
+| SDL_image| >= 2.0.5               |
 +----------+------------------------+
-| SDL_ttf  | >= 2.0.15              |
+| SDL_ttf  | >= 2.0.18              |
```
oddbookworm, line 134: There's absolutely nowhere that we check for SDL_TTF_VERSION_ATLEAST(2, 0, 18)? I have a hard time believing that

oddbookworm (dismissed):
This is a more complete attempt than the previous pull request that went stale after changes were requested. But I still have some comments that need to be addressed.

src_c/cython/pygame/_sdl2/video.pxd
Starbuck5: They seem to have removed every instance of this function being used in cython code, so it is being removed from the pxd file because it no longer needs to be defined to cython.

README.rst
Starbuck5: I see a few for 2.0.18 and one for 2.0.15

src_c/cython/pygame/_sdl2/video.pxd
oddbookworm: Oh, oops... should have read that a bit better

ochderek:
Thanks for pointing these out. I understand what I missed now. I’ll restore the declaration, check the remaining SDL_ttf version checks, and update the PR.

Starbuck5 (approved):
Looks good to me! Thanks

ankith26 (approved):
LGTM, thanks for contributing to pygame-ce! 🥳

PR #3853, opened 2026-07-25
Simplify ptr -> Python int code

Starbuck5:
I was looking at the 3.15 release notes, and it said that PY_LONG_LONG is now soft deprecated. So I found some instances of us using it, and switched to using PyLong_FromVoidPtr instead of different size variants for different pointer sizes


ankith26 (approved):
LGTM, thanks!

PR #3854, opened 2026-07-25
Remove a tiny bit of obsolete code

Starbuck5:
Many other places in the codebase, including in color.c, are using C round() without issues, so this must not be necessary anymore.

Nowhere in the codebase was using pg_qd, nor is macintosh a thing that ever gets defined, I believe.

### DIFF
--- a/src_c/color.c
+++ b/src_c/color.c
@@ -47,17 +47,6 @@
 
 #include <ctype.h>
 
-static inline double
-pg_round(double d)
-{
-#if (!defined(__STDC_VERSION__) || __STDC_VERSION__ < 199901L) && \
-    !defined(round)
-    return (((d < 0) ? (ceil((d)-0.5)) : (floor((d) + 0.5))));
-#else
-    return round(d);
-#endif
-}
-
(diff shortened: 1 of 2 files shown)


ankith26 (approved):
LGTM, thanks! 🎉

src_c/color.c
```diff
@@ -859,8 +848,7 @@ _color_lerp(pgColorObject *self, PyObject *args, PyObject *kw)
     }
 
     for (int i = 0; i < 4; i++) {
-        new_rgba[i] =
-            (Uint8)pg_round(self->data[i] * (1 - amt) + rgba[i] * amt);
+        new_rgba[i] = (Uint8)round(self->data[i] * (1 - amt) + rgba[i] * amt);
```
oddbookworm, line 851: This isn't *technically* the same... C's `round` function does the same thing as Python's, which is banker's rounding (away from zero on the half-integers), while `pg_round` rounds up on half integers

oddbookworm: Nvm, ignore that

oddbookworm (approved):
LGTM!

PR #3857, opened 2026-07-25
Remove legacy safe pickle system

Starbuck5:
It is something to difficult figure out when this became irrelevant, but modern pickle-able classes do not need to have a \_\_safe_for_unpickling\_\_ property / method. PEP 303 from Python 2.3 talks about it in past tense (https://peps.python.org/pep-0307/#security-issues), but it is documented as needed until Python 2.7 (https://docs.python.org/release/2.7.18/library/pickle.html#pickling-and-unpickling-extension-types). By 3.1 at the latest, it was removed (https://docs.python.org/3.1/library/pickle.html#pickle.object.__reduce__)


ankith26 (approved):
I checked cpython sources and yes, I found this line
> NOTE:  checks for __safe_for_unpickling__ went away in Python 2.3.

And of course, no line of code actually dealing with this attribute. We should be good to go with this cleanup (which is long overdue, over 23 years late xD)

oddbookworm:
> I checked cpython sources and yes, I found this line

[PEP-307 covered this, no need for source digging lol](https://peps.python.org/pep-0307/#security-issues)

oddbookworm (approved):
Thought I approved this before.... oops. LGTM!

PR #3858, opened 2026-07-25
Math module simplifications

Starbuck5:
I was looking for ways to clean up code / simplify the math module.

Changes
- Unify scalar/vector path of vector_elementwiseproxy_richcompare
    - This is the largest change, and an individual commit. Previously, the code paths were exactly the same except one compared to an an indexed value and one compared to a single scalar value. Now the single scalar is broadcast into the indexed value memory so the same code path can be used on both. This saves ~50 lines of code, decreases code size somewhat, and makes the output slightly slower (which I don't think is a big deal).
- vector_getAttr_swizzle had a redundant goto swizzle, swizzle label.
- vector2|3_project not necessary, they just call vector_project_onto
- math_invlerp raise is impossible to trigger because argument macros will throw before it gets there
- Remove stray #

### DIFF
--- a/src_c/math.c
+++ b/src_c/math.c
@@ -2073,9 +2073,6 @@ vector_getAttr_swizzle(pgVector *self, PyObject *attr_name)
             case 'y':
             case 'z':
                 idx = attr[i] - 'x';
-                goto swizzle_idx;
-
-            swizzle_idx:
                 if (idx >= self->dim) {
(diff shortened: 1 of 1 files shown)


damusss (approved):
LGTM, thank you!

ankith26 (approved):
LGTM, thanks! 🎉

PR #3860, opened 2026-07-26
docs: Document historical copy-paste bugs in deprecated math functions

ShivamKR12:
This PR documents long-standing copy-paste bugs in deprecated math functions. The C execution logic is deliberately left untouched to preserve backward compatibility.

**Changes made:**
* **`math.c`:** Added internal comments explaining that `vector3_rotate_y_ip_rad` and `vector3_rotate_z_ip_rad` incorrectly apply `x` rotation, and `vector_getsafepickle` accepts a `pgRectObject`.
* **`docs/reST/ref/math.rst`:** Added public `.. warning::` directives to the affected `Vector3` methods, informing users of the axis bug and directing them to the correct `_rad_ip` functions.

### DIFF
--- a/docs/reST/ref/math.rst
+++ b/docs/reST/ref/math.rst
@@ -1101,6 +1101,11 @@ Conversion can be combined with swizzling or slicing to create a new order
 
       DEPRECATED: Use rotate_y_rad_ip() instead.
 
+      .. warning::
+         Due to a historical bug, this deprecated function actually
+         applies rotation around the X-axis. Please use
+         :meth:`rotate_y_rad_ip` instead.
+
       .. versionaddedold:: 2.0.0
       .. deprecatedold:: 2.1.1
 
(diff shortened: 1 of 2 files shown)


Starbuck5:
I'm handling the pickle one in https://github.com/pygame-community/pygame-ce/pull/3857

ShivamKR12:
> I'm handling the pickle one in #3857

Ok then, i think i will revert those changes. Alright ?

ShivamKR12:
@Starbuck5 Done 👍 
Good Now ?

Starbuck5:
@ShivamKR12 when you started working on this, I did not have an opinion on the best solution. I've thought about it a bit now, I feel that the bug should be fixed.

Imagine someone is writing code without an editor that supports the deprecated decorator, and they write rotate_y_ip_rad instead of rotate_y_rad_ip. It won't crash, it won't inform them of anything, it will just produce wrong results in a confusing way.

My preferred now would be a fix, a note about the affected versions in the documentation, and tests for the fix,

PR #3866, opened 2026-07-30
Move some more font docs to stubs

Starbuck5:
https://github.com/pygame-community/pygame-ce/issues/2757

Only content changes were to get_sdl_ttf_version. I added the (major, minor, linked) explainer text from pygame.image.get_sdl_image_version because going with the pyi type line removed that context, and I added a period to the end of the final sentence.


Matiiss:
Why not move all of the docs for the module at once?

oddbookworm:
> Why not move all of the docs for the module at once?

Makes it easier to review probably

Starbuck5:
> Why not move all of the docs for the module at once?

Because my heart despairs to move the misshapen blob that is the font.render docstring into poor user's editors

Because I'm doing this manually and I do what I have time for and then submit it

buildconfig/stubs/pygame/font.pyi
```diff
@@ -70,18 +120,123 @@ class Font:
...
     ) -> Surface: ...
-    def size(self, text: str | bytes, /) -> tuple[int, int]: ...
-    def set_underline(self, value: bool, /) -> None: ...
-    def get_underline(self) -> bool: ...
-    def set_strikethrough(self, value: bool, /) -> None: ...
-    def get_strikethrough(self) -> bool: ...
-    def set_bold(self, value: bool, /) -> None: ...
-    def get_bold(self) -> bool: ...
-    def set_italic(self, value: bool, /) -> None: ...
+    def size(self, text: str | bytes, /) -> tuple[int, int]:
```
oddbookworm, line 123: My biggest concern with this change is that the old docs specify the order of width and height, but this version loses that information

oddbookworm (commented):
I just have one concern here. I generated locally on both main and this branch, and this was the only difference in the html worth noting

buildconfig/stubs/pygame/font.pyi
Starbuck5: I did think of this, but I thought it would be self evident with the rest of pygame-ce style.

But it's better to be explicit, it should be possible to enter the docs at any point and understand things.

Added (width, height) context to the text blob below.

oddbookworm (approved):
LGTM! 🎉

PR #3868, opened 2026-07-31
Move sndarray docs to stubs

Starbuck5:
#2757


oddbookworm (approved):
LGTM!

Matiiss (approved):
LGTM!

Though I'm noticing that `sndarray.py` also has some of these docstrings: https://github.com/pygame-community/pygame-ce/blob/eb496f68cbccb1dfffe67822250e335891a3fcd6/src_py/sndarray.py
Should those be removed from there? They're not as complete, for one, but also it just currently means we have the same information in two places. Either way can be done in a different PR.

PR #3869, opened 2026-08-01
Use more ideal PyLong_ functions

Starbuck5:
This one is super minor, but I noticed that instead of casting to long in base.c, it can just use PyLong_FromSsize_t. This is more correct on large values. Py_ssize_t is 8 bytes, long is 4 bytes, so huge values would be truncated(?).

And in rwobject.c, instead of assuming long long is the same as int64 (which it probably always is though), we can use the fancy new sized API PyLong_AsInt64, a new function from CPython that no longer needs to call PyErr_Occurred.


ankith26 (approved):
LGTM, thanks! 🎉

Issue #3871, opened 2026-08-01
Rwobject issues


Starbuck5:
# `src_c/rwobject.c`

Dump from Claude on the Python-file-object `SDL_RWops` / `SDL_IOStream` system, informed by my testing and observations from my investigation into a potential performance gain using memoryviews.

- **Scope:** `fetch_object_methods`, `pgRWops_FromFileObject`, and the
  `_pg_rw_size` / `_pg_rw_seek` / `_pg_rw_read` / `_pg_rw_write` / `_pg_rw_close`
  callbacks.

---

### Unbounded `memcpy` into SDL's buffer

[`src_c/rwobject.c:636-638`](src_c/rwobject.c#L636-L638)

```c
retval = PyBytes_GET_SIZE(result);
if (retval) {
    memcpy(ptr, PyBytes_AsString(result), retval);
```

Nothing checks `retval <= size * maxnum`. A file object whose `read(n)` returns
more than `n` bytes overflows the destination buffer, which is frequently a
stack buffer in the SDL caller.

`read()` is only *documented* to return at most `n` bytes — nothing enforces it,
and pygame hands arbitrary user-supplied objects to this path via
`pygame.image.load`, `pygame.mixer.Sound`, `pygame.font.Font`, etc.

Fix: clamp, or treat an oversized return as an error.

---

### `return -1` from `size_t` functions is `SIZE_MAX`

[`:617`](src_c/rwobject.c#L617), [`:416`](src_c/rwobject.c#L416),
[`:425`](src_c/rwobject.c#L425), plus the `retval = -1` assignments in both
`_pg_rw_read` and `_pg_rw_write`.

Both SDL versions signal failure by returning *less than requested*:

- SDL2 `read`/`write` return the number of **objects** transferred; less than
  `num` means error/EOF.
- SDL3 returns **bytes** transferred, `0` on error/EOF, with `*status` set.

`(size_t)-1` is an enormous success count. Whether it is caught depends entirely
on how the caller checks — `!= num` happens to trip, `< num` reads as success,
and any `total += n` accumulator corrupts. Should be `0`.

### SDL2 partial reads are silently discarded

[`:640`](src_c/rwobject.c#L640)

```c
retval /= size;
```

Integer truncation. Called with `size=80, num=1` and a 40-byte short read, this
reports `0` objects — indistinguishable from EOF — and the 40 bytes are dropped.
Any short read with `size > 1` loses data.

[...]

PR #3873, opened 2026-08-01
Set python exception as SDL error in rwops code

ankith26:
In our rwops code we currently print exceptions when the underlying python file-like object raises errors. Printing exceptions isn't a good way of dealing with them because the user cannot easily handle what's printed, and may not expect any stdout/stderr. The simplest solution is to convert the python exception to an SDL error.

It is upto SDL to decide what to do with this error: it may be forwarded to the caller which generally results in the same error message resurfacing back as a `pygame.error` exception for the pygame user; or SDL may just rewrite this message entirely in which case the pygame user would see a more generic error message.

This PR also addresses `-1` return issue from unsigned functions, of #3871

### DIFF
--- a/src_c/rwobject.c
+++ b/src_c/rwobject.c
@@ -82,10 +82,21 @@ _pg_sdl_error_from_py_exc()
             value ? PyUnicode_FromFormat("Got %S: %S", type, value)
                   : PyUnicode_FromFormat("Got %S", type);
 
-        SDL_SetError("%s", error_str
-                               ? PyUnicode_AsUTF8(error_str)
-                               : "An unknown error occured, probably some "
-                                 "memory allocation failed");
+        const char *message =
+            "An unknown error occurred, a memory allocation probably failed";
+        if (error_str) {
+            const char *utf8_message = PyUnicode_AsUTF8(error_str);
+            if (utf8_message) {
+                message = utf8_message;
+            }
+            else {
+                PyErr_Clear();
+            }
+        }
+        else {
+            PyErr_Clear();
+        }
+        SDL_SetError("%s", message);
         Py_XDECREF(error_str);
     }
     Py_XDECREF(type);
@@ -647,9 +658,10 @@ _pg_rw_read(SDL_RWops *context, void *ptr, size_t size, size_t maxnum)
         goto end;
     }
 
+    /* TODO: Handle other bytes-like objects here */
     if (!PyBytes_Check(result)) {
         Py_DECREF(result);
-        _pg_sdl_error_from_py_exc();
+        SDL_SetError("read returned non-bytes object when bytes was expected");
         retval = 0;
         goto end;
     }
--- a/src_c/rwobject.c
+++ b/src_c/rwobject.c
@@ -70,6 +70,29 @@ static int
 _pg_rw_close(SDL_RWops *);
 #endif
 
+/* If a python exception is set, it is cleared and a corresponding SDL error is
+ * set */
+static void
+_pg_sdl_error_from_py_exc()
+{
+    PyObject *type, *value, *traceback;
+    PyErr_Fetch(&type, &value, &traceback);
+    if (type) {
+        PyObject *error_str =
+            value ? PyUnicode_FromFormat("Got %S: %S", type, value)
+                  : PyUnicode_FromFormat("Got %S", type);
+
+        SDL_SetError("%s", error_str
+                               ? PyUnicode_AsUTF8(error_str)
+                               : "An unknown error occured, probably some "
+                                 "memory allocation failed");
+        Py_XDECREF(error_str);
+    }
+    Py_XDECREF(type);
+    Py_XDECREF(value);
+    Py_XDECREF(traceback);
+}
+
 /* Converter function used by PyArg_ParseTupleAndKeywords with the "O&" format.
  *
  * Returns: 1 on success
@@ -339,7 +362,7 @@ _pg_rw_size(SDL_RWops *context)
      */
     pos = PyObject_CallNoArgs(helper->tell);
     if (!pos) {
-        PyErr_Print();
+        _pg_sdl_error_from_py_exc();
         goto end;
     }
 
@@ -347,7 +370,7 @@ _pg_rw_size(SDL_RWops *context)
      */
     tmp = PyObject_CallFunction(helper->seek, "ii", 0, SEEK_END);
     if (!tmp) {
-        PyErr_Print();
+        _pg_sdl_error_from_py_exc();
         goto end;
     }
     Py_DECREF(tmp);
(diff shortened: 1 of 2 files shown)


src_c/rwobject.c
```diff
@@ -70,6 +70,40 @@ static int
 _pg_rw_close(SDL_RWops *);
 #endif
 
+/* If a python exception is set, it is cleared and a corresponding SDL error is
+ * set */
+static void
+_pg_sdl_error_from_py_exc()
+{
+    PyObject *type, *value, *traceback;
+    PyErr_Fetch(&type, &value, &traceback);
```
Starbuck5, line 79: Hey I haven't played with this yet or anything, but I did notice that PyErr_Fetch is deprecated, per https://docs.python.org/3.16/c-api/exceptions.html#c.PyErr_Fetch

oddbookworm: While I agree that it's not great form to use it here, I feel like we can push that off to a dedicated PR to replace `PyErr_Fetch` with something else across the codebase. Probably a version-based macro for `PyErr_Fetch` for versions below 3.12 and PyErr_GetRaisedException for 3.12+

Starbuck5: I was going to say we can just use pythoncapicompat, but surprisingly `PyErr_GetRaisedException` isn't available there!

oddbookworm (approved):
I don't see any issues here that need to be resolved in this pull, so LGTM!

Starbuck5:
Andrew, since you and Ankith are both good with it, I think you should feel comfortable to merge. 2 members of the SC is enough to merge 😄

My biggest question right now is what this will actually look like for the user when it happens. (like an example of the error going through this system)

oddbookworm:
> My biggest question right now is what this will actually look like for the user when it happens. (like an example of the error going through this system)

I commented out the raise assertions in the tests and here's what I see:
```
loading test.rwobject_test
...................EEE
ERROR: test_read_error (test.rwobject_test.RWopsUnderlyingFileErrorTest.test_read_error)
  File "C:\Users\andre\Projects\pygame-ce\test\rwobject_test.py", line 180, in test_read_error
    pygame.image.load(FailingFileObject("read", error))
    ~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
pygame.error: Unsupported image format

ERROR: test_seek_error (test.rwobject_test.RWopsUnderlyingFileErrorTest.test_seek_error)
  File "C:\Users\andre\Projects\pygame-ce\test\rwobject_test.py", line 200, in test_seek_error
    pygame.image.load(FailingFileObject("seek", error))
    ~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
pygame.error: Unsupported image format

ERROR: test_write_error (test.rwobject_test.RWopsUnderlyingFileErrorTest.test_write_error)
  File "C:\Users\andre\Projects\pygame-ce\test\rwobject_test.py", line 186, in test_write_error
    pygame.image.save(
    ~~~~~~~~~~~~~~~~~^
        surface,
        ^^^^^^^^
        FailingFileObject("write", error),
        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    )
    ^
pygame.error: Got <class 'RuntimeError'>: write operation failed

Ran 22 tests in 0.006s

FAILED (errors=3)
```

Starbuck5:
Ah that's what Ankith was saying about sometimes SDL will replace the error. If that's the case, should we continue to print it then? To give users some clue about what the issue on the backend is?

ankith26:
I think that printing exceptions is a bad idea in general, I am open to any other alternatives of dealing with this

Starbuck5:
I think we need to keep the prints, at least relative to this change. Putting the error through SDL is cool, but it gets swallowed too often.

Which is better:
```
  File "C:\Users\charl\Desktop\projects\test_files\test_rwerror.py", line 22, in read
    raise self._exception
RuntimeError: I'm important runtime context
  File "C:\Users\charl\Desktop\projects\test_files\test_rwerror.py", line 52, in <module>
    x = pygame.Sound(fo)
pygame.error: Couldn't read first 4 bytes of audio data
```
(status quo)

or

```
  File "C:\Users\charl\Desktop\projects\test_files\test_rwerror.py", line 52, in <module>
    x = pygame.Sound(fo)
pygame.error: Couldn't read first 4 bytes of audio data
```
(after this PR)

I also tested with image loading, the context also gets hidden there. If I'm a user and something is going wrong with my IO, I want to know what it is, not get a generic error message through SDL. 

<details>
<summary>Code</summary>

```py
import os
import io

import pygame
pygame.init()

def example_path(path):
    return os.path.join(r"C:\Users\charl\Desktop\pygame-ce", "examples", path)

class FailingFileObject:
    """File-like object that raises a Python exception on a selected operation."""

    def __init__(self, operation, exception):
        with open(example_path("data/alien1.png"), "rb") as f:
            self._file = io.BytesIO(f.read())

        self._operation = operation
        self._exception = exception

    def read(self, *args, **kwargs):
        if self._operation == "read":
            raise self._exception
        return self._file.read(*args, **kwargs)

    def write(self, *args, **kwargs):
        if self._operation == "write":
            raise self._exception
        return self._file.write(*args, **kwargs)

    def seek(self, *args, **kwargs):
        if self._operation == "seek":
            raise self._exception
        return self._file.seek(*args, **kwargs)

    def tell(self):
        return self._file.tell()

    def flush(self):
        return self._file.flush()

    def close(self):
        return self._file.close()



z = RuntimeError("I'm important runtime context")

fo = FailingFileObject("read", z)

#x = pygame.image.load(fo)
x = pygame.Sound(fo)
```
</details>

Starbuck5:
There are more complicated things we could do out of band. RW_SetError, RW_GetError, RW_ClearError?

Or instead of mallocing a pgRWHelper inside rwobject, we could do a stack allocated struct at the caller level and have an error field on that? Plus could replace the way we get the file extension out currently.

But I wonder if SDL3 would be better with this with the ability to return a status (https://wiki.libsdl.org/SDL3/SDL_IOStatus) through.

Anyways in terms of steps forward I would endorse this PR if the prints were retained alongside the SDL error set.

ankith26:
I am also considering checking out [`PyException_SetContext`](https://docs.python.org/3/c-api/exceptions.html#c.PyException_SetContext) and/or [`PyException_SetCause`](https://docs.python.org/3/c-api/exceptions.html#c.PyException_SetCause) API so it would trace the original exception but in the standard pythonic way (this approach would be equivalent to something like `raise pygame.error("error message from SDL here") from original_exception`). This approach however creates a requirement on `PyErr_GetRaisedException` API which is 3.12+ and doesn't exist in c-api compat as you pointed out. I'll dig more into how to deal with this, consider this PR a draft for now.

PR #3874, opened 2026-08-01
Remove support prompt (3.0.0)

oddbookworm:
Following up from a poll on Discord, the vote was to remove the support prompt


docs/reST/ref/pygame.rst
```diff
@@ -416,6 +416,9 @@ console that tells you which version of python,
 pygame & SDL you are using. Must be set before
 importing pygame.
 
+.. note::
+   Support prompt entirely removed in pygame-ce 2.5.8.
+
```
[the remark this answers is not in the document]
oddbookworm: No, this is how I want to leave it because people use versions older than the most recent release.

Starbuck5:
Hey Andrew, thanks for working on this.

Sorry to be unclear about it, I was thinking this would be a 3.0 item. It’s such a part of the experience right now, 3.0 makes sense as the time to re-evaluate things like this and forge the best path forward.

oddbookworm:
> Hey Andrew, thanks for working on this.
> 
> Sorry to be unclear about it, I was thinking this would be a 3.0 item. It’s such a part of the experience right now, 3.0 makes sense as the time to re-evaluate things like this and forge the best path forward.

Oops lol, will update

Starbuck5:
Since this is the first change for 3.0, does it make sense to start our equivalent of https://github.com/libsdl-org/SDL/blob/main/docs/README-migration.md here?

oddbookworm:
Do you have a summary of expected differences that I can reference to write it here?

Starbuck5:
Well right now the only expected difference would be this, and then we'd add more as more stuff lands. But I'm not sure I understand the question.

I think docs/README-migration.md would be a great place for it, I like how SDL does it by module as well, even though this change specifically doesn't really belong to any module.

oddbookworm:
Ah, I see. Yeah, that's fine. I'll create it here. I was thinking you had a more concrete list of changes that the user would see for now

oddbookworm:
@coderabbitai resume

oddbookworm:
@coderabbitai fresh review

src_py/__init__.py
```diff
@@ -453,22 +453,24 @@ def __color_reduce(c):
 copyreg.pickle(Color, __color_reduce, __color_constructor)
 
 if "PYGAME_HIDE_SUPPORT_PROMPT" not in os.environ:
```
Matiiss, line 455: Is it worth it to maybe print a warning if this environment variable is still used in a 3.x version?
Also, if this PR is to land with 3.0 or some such, can it include a docs update that removes this as a documented env var? Or should a new issue be created to track that?

oddbookworm: Won't remove it completely from the docs because 2.x versions will still be used by some, that's why I put the note in the docs just under it. Also, the the `docs/README-migration.md` file is intended to document expected changes like this between 2.x and 3.x

shaurya-sharma-dev:
If this is the first change for 3.0.0, wouldn't it make more sense to make a 3.0.0 branch so we can start working on basic 3.0.0 features like this? Or would we make a 3.0.0 branch after 2.5.8 is released?

oddbookworm:
When we start properly getting ready for 3.x, we'll cut a 2.x branch and main will be 3.x. This is just some easy preliminary stuff to start getting some of the stuff we'll need in place

Matiiss (approved):

Starbuck5 (approved):
I think I'm going to miss it.

Starbuck5:
This should also remove the platform import in init.py

ankith26 (approved):
I am also lowkey sad to see this go, but it must be done.

damusss (approved):
My heart aches

oddbookworm:
Made a quick push to pull up-to-date with main and remove the platform import in `src_py/__init__.py`

PR #3875, opened 2026-08-01
Implement _sdl3_mixer.AudioDecoder class

Starbuck5:
https://wiki.libsdl.org/SDL3_mixer/MIX_AudioDecoder

Split from https://github.com/pygame-community/pygame-ce/pull/3843 since this is much simpler on its own. This is needed to create a pygame-ce 2 compatible mixer module on top of SDL3.

```python dev.py build --sdl3```

```py
import os

import pygame._audio as audio
import pygame._sdl3_mixer as mixer

mixer.init()

data_dir = r"C:\Users\charl\Desktop\pygame-ce\examples\data"

audio_dec = mixer.AudioDecoder(os.path.join(data_dir, "surfonasinewave.xm"))
print(audio_dec.spec)

chunk1 = audio_dec.decode(audio_dec.spec, size=100)
chunk2 = audio_dec.decode(spec=audio.AudioSpec(audio.S32, 2, 48000))

print(len(chunk1))
print(len(chunk2))
```


src_c/_sdl3_mixer_c.c
```diff
@@ -1372,6 +1377,176 @@ static PyType_Spec track_spec = {
...
+    SDL_AudioSpec spec;
+    spec.format = PyLong_AsInt(PyTuple_GetItem(spec_obj, 0));
+    spec.channels = PyLong_AsInt(PyTuple_GetItem(spec_obj, 1));
+    spec.freq = PyLong_AsInt(PyTuple_GetItem(spec_obj, 2));
+    // Check that they all succeeded
+    if (spec.format == -1 || spec.channels == -1 || spec.freq == -1) {
+        if (PyErr_Occurred()) {
+            return NULL;
+        }
+    }
```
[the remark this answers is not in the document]
Starbuck5: This way is consistent with 2 other spec parsing locations in the file, and callers who intentionally bypass Python level checks can deal with the consequences of their actions.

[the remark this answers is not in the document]
Starbuck5: It does not terminate the interpreter. If PyLong_AsInt receives a NULL, it raises `SystemError: D:\a\1\s\Objects\longobject.c:466: bad argument to internal function`.

src_c/_sdl3_mixer_c.c
```diff
@@ -1372,6 +1377,176 @@ static PyType_Spec track_spec = {
...
+    if (io == NULL) {
+        return -1;
+    }
+
+    self->audio_decoder = MIX_CreateAudioDecoder_IO(io, true, 0);
+    if (self->audio_decoder == NULL) {
+        PyErr_SetString(pgExc_SDLError, SDL_GetError());
+        return -1;
+    }
+    return 0;
```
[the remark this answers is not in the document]
Starbuck5: Wouldn't this finding apply to all objects throughout this library? I'm not sure that's the best way to deal with them, what about raising an exception if init is called repeatedly?

src_c/_sdl3_mixer_c.c
```diff
@@ -1372,6 +1377,181 @@ static PyType_Spec track_spec = {
...
+    }
+
+    SDL_IOStream *io = pgRWops_FromObject(file, NULL);
+    if (io == NULL) {
+        return -1;
+    }
+
+    self->audio_decoder = MIX_CreateAudioDecoder_IO(io, true, 0);
+    if (self->audio_decoder == NULL) {
+        PyErr_SetString(pgExc_SDLError, SDL_GetError());
```
oddbookworm, line 1402: The docs for `MIX_CreateAudioDecoder_IO` don't really say anything about SDL setting an error message internally, can we trust that they always will?

src_c/_sdl3_mixer_c.c
```diff
@@ -1372,6 +1377,181 @@ static PyType_Spec track_spec = {
...
+    // out.
+    SDL_AudioSpec spec;
+    spec.format = PyLong_AsInt(PyTuple_GetItem(spec_obj, 0));
+    spec.channels = PyLong_AsInt(PyTuple_GetItem(spec_obj, 1));
+    spec.freq = PyLong_AsInt(PyTuple_GetItem(spec_obj, 2));
+    // Check that they all succeeded
+    if (spec.format == -1 || spec.channels == -1 || spec.freq == -1) {
+        if (PyErr_Occurred()) {
+            return NULL;
+        }
```
oddbookworm, line 1442: It's fine if any of these is -1 but there's no internal python error set?

src_c/_sdl3_mixer_c.c
```diff
@@ -1372,6 +1377,181 @@ static PyType_Spec track_spec = {
...
+            // Get the new buffer write location by advancing the current
+            // location by the number of bytes read, and allocate the space
+            // for the next chunk.
+            buf = PyBytesWriter_GrowAndUpdatePointer(writer, chunk,
+                                                     buf + bytes_decoded);
+            if (buf == NULL) {
+                PyBytesWriter_Discard(writer);
+                return NULL;
+            }
+        }
```
oddbookworm, line 1480: The `while(true)` here with a break in the middle gives me the creeps... can we do something like
```c
char *buf = (char *)PyBytesWriter_GetData(writer);
int bytes_decoded;

while ((bytes_decoded = MIX_DecodeAudio(self->audio_decoder, buf, chunk,
                                         &spec)) > 0) {
    // Get the new buffer write location by advancing the current
    // location by the number of bytes read, and allocate the space
    // for the next chunk.
    buf = PyBytesWriter_GrowAndUpdatePointer(writer, chunk,
                                              buf + bytes_decoded);
    if (buf == NULL) {
        PyBytesWriter_Discard(writer);
        return NULL;
    }
}

// -1 on error
if (bytes_decoded < 0) {
    PyBytesWriter_Discard(writer);
    return RAISE(pgExc_SDLError, SDL_GetError());
}
```

src_c/_sdl3_mixer_c.c
```diff
@@ -1372,6 +1377,181 @@ static PyType_Spec track_spec = {
...
+                                     &size_obj)) {
+        return NULL;
+    }
+
+    // Python layer constructs spec tuple, we can assume it is correctly laid
+    // out.
+    SDL_AudioSpec spec;
+    spec.format = PyLong_AsInt(PyTuple_GetItem(spec_obj, 0));
+    spec.channels = PyLong_AsInt(PyTuple_GetItem(spec_obj, 1));
+    spec.freq = PyLong_AsInt(PyTuple_GetItem(spec_obj, 2));
```
oddbookworm, line 1437: I think this logic only holds up if we can assume nobody will **ever** call this manually and pass in their own arguments. Here's a quick script that causes an internal python failure that would be basically impossible to debug on the user-end
```py
import os

import pygame._sdl3_mixer as mixer
import pygame._sdl3_mixer_c as mixer_c

mixer.init()

audio_dec = mixer_c.AudioDecoder(os.path.join(os.path.abspath(os.path.dirname(__file__)), "examples/data/boom.wav"))

mixer_c.AudioDecoder.decode(audio_dec, spec="not a tuple")
```

src_c/_sdl3_mixer_c.c
```diff
@@ -1372,6 +1377,181 @@ static PyType_Spec track_spec = {
...
+        const int chunk = 4096;
+        PyBytesWriter *writer = PyBytesWriter_Create(chunk);
+        if (writer == NULL) {
+            return NULL;  // exception already set
+        }
+
+        char *buf = (char *)PyBytesWriter_GetData(writer);
+        while (true) {
+            int bytes_decoded =
+                MIX_DecodeAudio(self->audio_decoder, buf, chunk, &spec);
```
oddbookworm, line 1458: What do you think about releasing the GIL around MIX_DecodeAudio calls?

oddbookworm (commented):
A few things that come to mind

src_c/_sdl3_mixer_c.c
Starbuck5: Yes. Documentation for PyLong_AsInt states it will return -1 with an error set on error. -1 on its own is not an error. https://docs.python.org/3/c-api/long.html#c.PyLong_AsInt

Unless you're talking about input sanitization, us rejecting invalid values. In most places like this SDL does that. I tested here and it actually silently fails with invalid AudioSpec values (like channels=60, format=random number). So I've raised a bug for this specific spot, after checking the code it seems like it's just missing a check on a call internally. https://github.com/libsdl-org/SDL_mixer/issues/890

src_c/_sdl3_mixer_c.c
Starbuck5: This is the established pattern in our codebase, I'm not sure what I could do here that's reasonably simple and better. Worst case scenario it is pygame.error : empty string.

src_c/_sdl3_mixer_c.c
Starbuck5: Right now the public API is in the _sdl3_mixer module, and the internals are in _sdl3_mixer_c, which looks very similar. However, when released, the public API will be part of the mixer module, (and the c backend might be renamed to), but my point is that they will look more different than they do now.

If someone is importing internal modules and using non public APIs, I don't think we should be that concerned about the quality of the error messages they receive.

src_c/_sdl3_mixer_c.c
Starbuck5: Alright I've added a note about it. It's not safe on its own to just release them, I think, because multiple threads could be dealing with the same object and I don't think the decoder has its own internal locking.

src_c/_sdl3_mixer_c.c
Starbuck5: I restructured it into a do-while and bumped the chunk size. It's a little bit less efficient because it allocates an extra chunk at the end, but it fits better with future gil release/hold statements to have the decode audio not in the body of the comparison.

Also rebased this branch.

src_c/_sdl3_mixer_c.c
oddbookworm: That's exactly my point with my comment... `pygame.error: _______` is a pretty impossible thing to figure out what went wrong

src_c/_sdl3_mixer_c.c
oddbookworm: Fair enough

src_c/_sdl3_mixer_c.c
oddbookworm: I'm more talking about when any of these is -1 but no python error occurred. Like, what's a negative frequency?

Starbuck5: In other places, SDL handles that.
```py
print("Creating stream")
out = pygame.audio.AudioSpec(audio.S32, 2, -1)
stream = pygame.audio.AudioStream(audio_spec, out)
```
```
  File "C:\Users\charl\Desktop\projects\test_files\test_3657.py", line 31, in <module>
    stream = pygame.audio.AudioStream(audio_spec, out)
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "C:\Users\charl\Desktop\pygame-ce\src_py\_audio.py", line 361, in __init__
    self._state = _base_audio.create_audio_stream(
                  ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
pygame.error: Parameter 'dst_spec->freq' is invalid
```

oddbookworm (approved):
Sure, LGTM!

PR #3877, opened 2026-08-01
Move Rect docs to stubs

SuperRedingBros:
https://github.com/pygame-community/pygame-ce/issues/2757


oddbookworm:
@coderabbitai review as if this is a direct port PR. If there's something that needs to be changed from how it *currently* states it, then open a new issue for it.

buildconfig/stubs/pygame/rect.pyi
```diff
@@ -144,62 +243,143 @@ class _GenericRect(Collection[_N]):
...
     @overload
-    def inflate(self, inflate_by: Point, /) -> Self: ...
+    def inflate(self, inflate_by: Point, /) -> Self:
+        """Grow or shrink the rectangle size.
+        
+        Returns a new rectangle with the size changed by the given offset. The
+        rectangle remains centered around its current center. Negative values
+        will shrink the rectangle. Note, uses integers, if the offset given is
+        too small(< 2 > -2), center will be off.
+        """
```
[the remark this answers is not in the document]
oddbookworm: I thought you said that you only commented on formatting differences. Fully re-review it and only focus on direct translation from rst to stub

buildconfig/stubs/pygame/rect.pyi
```diff
@@ -28,6 +28,105 @@ _RectTypeCompatible_co = TypeVar(
...
+    All of these attributes can be assigned to:
+
+    ::
+
+        rect1.right = 10
+        rect2.center = (20,30)
+
+    Assigning to size, width, height, or relcenter changes the dimensions of the rectangle;
+    all other assignments move the rectangle without resizing it. Notice that
+    some attributes are integers and others are pairs of integers.
```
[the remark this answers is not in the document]
oddbookworm: Ignore this, it's out of scope of this PR

oddbookworm:
@coderabbitai fresh review

buildconfig/stubs/pygame/rect.pyi
```diff
@@ -144,62 +243,139 @@ class _GenericRect(Collection[_N]):
...
+
+        Returns a new rectangle having the same position and size as the original.
+
+        New in pygame 1.9
+        """
     @overload
     def move(self, x: float, y: float, /) -> Self: ...
     @overload
-    def move(self, move_by: Point, /) -> Self: ...
+    def move(self, move_by: Point, /) -> Self:
```
oddbookworm, line 260: Hmmmmm, this is a bit problematic... the signature of the function gets taken from the stub now and moved into the docs, so this now looks like this

Image

instead of this

Image

`Self` in static typing does imply an instance of the same type. But in the docs here, I'm going to go out on a limb and say that most people would read this in docs and think that it means it returns itself. @ankith26 thoughts as the typing guru?

aatle: Yes, it would be confusing for people who aren't very familiar with typing.

The `Self` type should stay in stubs because of subclass correctness but in the docs, perhaps some sort of replace to class name can be used. If not possible though, then I think it may have to stay `Self` in docs.

oddbookworm: I did come up with a rough fix for it, but I don't know offhand that this solution is bulletproof. It's a patch for `docs/reST/ext/documenters.py`

```patch
diff --git a/docs/reST/ext/documenters.py b/docs/reST/ext/documenters.py
index 3c9438eda..3214afbc7 100644
--- a/docs/reST/ext/documenters.py
+++ b/docs/reST/ext/documenters.py
@@ -3,6 +3,11 @@ import autoapi.documenters
 from autoapi._objects import PythonClass
 import re
 
+def _enclosing_class_name(object):
+    parts = object.qual_name.split(".")[:-1]
+    return parts[-1] if parts else None
 
 def build_signatures(object):
     name = object.short_name
@@ -24,10 +29,14 @@ def build_signatures(object):
     if object is None or object.obj.get("args", None) is None:
         return
 
+    enclosing_class = _enclosing_class_name(object)
+
     sigs = [(object.obj["args"], object.obj["return_annotation"])]
     sigs.extend(object.obj["overloads"])
 
     for args, ret in sigs:
+        if enclosing_class is not None:
+            ret = re.sub(r"\bSelf\b", enclosing_class, ret)
         arg_string = ""
         for modifier, arg_name, _, default in args:
             modifier = modifier or ""
```
Edit: updated the diff to be hopefully more robust and acceptable

aatle: This diff seems fine.
If it works on a type like `list[Self | None] | Self` e.g., then it's probably robust enough.

Though maybe you want to add a case for `typing.Self`.

As long as these PRs are well reviewed I think it is fine to implement without worrying too much about edge cases.

oddbookworm: @SuperRedingBros can you implement the diff above? I won't ask you to implement the `typing.Self` thing, we import it directly from `typing` or `typing_extensions` everywhere we use it. If it becomes a problem, then we can fix it at that point.

oddbookworm: I would put it into a separate PR, but that would block this one IMO because then we'd have a period of time where the docs are in an intermediate state

docs/reST/ext/documenters.py
```diff
@@ -24,10 +27,14 @@ def build_signatures(object):
...
         return
 
+    enclosing_class = _enclosing_class_name(object)
+
     sigs = [(object.obj["args"], object.obj["return_annotation"])]
     sigs.extend(object.obj["overloads"])
 
     for args, ret in sigs:
+        if enclosing_class is not None:
+            ret = re.sub(r"\bSelf\b", enclosing_class, ret)
```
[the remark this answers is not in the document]
oddbookworm: Open a new issue for this, referencing this pull request. Explicitly state that this pull request should merge before creating a pull request for fixing the new issue. I agree that there's an issue here, but as this is just a small fix to get it working, then I don't want to burden the author of this pull with it.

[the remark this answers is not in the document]
oddbookworm: Resolving this because it's out of scope

buildconfig/stubs/pygame/rect.pyi
```diff
@@ -211,67 +387,321 @@ class _GenericRect(Collection[_N]):
...
+
+        Same as the ``Rect.union()`` method, but operates in place.
+        """
+    def unionall(self, rect: SequenceLike[_RectTypeCompatible_co], /) -> Self:
+        """The union of many rectangles.
+
+        Returns the union of one rectangle with a sequence of many rectangles.
+        """
     def unionall_ip(
         self, rect_SequenceLike: SequenceLike[_RectTypeCompatible_co], /
```
oddbookworm: The only even minor issues I have with the new generated docs is that the signatures change to exactly what the stubs have, which can make it not-as-pleasant to read in docs, but that's not a nitpick of your review, more of "we should maybe look into making the docs more readable"

@aatle thoughts on this?

Here's an example from the generated docs

Image

oddbookworm (approved):
I have some minor nitpicks with how the generated docs came out, but that's not a "your pull request" issue, so I'm approving it. Thanks!

buildconfig/stubs/pygame/rect.pyi
zoldalma999: This specific one is a copy-paste error from #3002: https://github.com/pygame-community/pygame-ce/pull/3002/changes#diff-456cf12a8a2594fc2015129b4400c616cda456ba82cdfdfe637dda89e8b3bddfL225

zoldalma999:
I saw that you had issues with the inherited members not showing up automatically, this can be fixed with the `:inherited-members:` option, like so:
```rst
.. autopgclass:: Rect
   :members:
   :inherited-members:

```

This way you wouldn't have the duplication about both the docstring of the rect class itself and what method it needs to document. This leaves only the FRect signatures missing (do we even need this?). This can be fixed by adding it manually to the docs:

```python
class _GenericRect(Collection[_N]):
    """Pygame object for storing rectangular coordinates.
    | :sg:`FRect(left, top, width, height, /) -> FRect`
    | :sg:`FRect((left, top), (width, height), /) -> FRect`
    | :sg:`FRect(object, /) -> FRect`
    | :sg:`FRect() -> FRect`

    .. versionadded:: 2.2
    [...]
    """
```

However this is not the prettiest solution (especially in editors). @Starbuck5 thoughts?

buildconfig/stubs/pygame/rect.pyi
aatle: It should be named `rect`, yeah.
Can this be fixed right now?

```suggestion
        self, rect: SequenceLike[_RectTypeCompatible_co], /
```

aatle: The code needs to be formatted.

oddbookworm:
@coderabbitai resume and fresh review

oddbookworm (approved):
The only remaining gripes I have are about the signatures for generics making it harder to interpret in the docs, but I don't have a good solution for that at this point in time. Thanks! 🎉

buildconfig/stubs/pygame/rect.pyi
```diff
@@ -144,62 +247,139 @@ class _GenericRect(Collection[_N]):
...
     def inflate_ip(self, x: float, y: float, /) -> None: ...
     @overload
-    def inflate_ip(self, inflate_by: Point, /) -> None: ...
+    def inflate_ip(self, inflate_by: Point, /) -> None:
+        """Grow or shrink the rectangle size, in place.
+
+        Same as the ``Rect.inflate()`` method, but operates in place.
+        """
     @overload
     def scale_by(self, x: float, y: float = ...) -> Self: ...
```
aatle, line 307: The generated docs look a bit confusing because of how this is typed.

Can the overloads be changed to:
```
(self, x: float, y: float) -> Self
(self, scale_by: Point | float) -> Self
```
for both `scale_by` and `scale_by_ip`?

buildconfig/stubs/pygame/rect.pyi
```diff
@@ -144,62 +247,139 @@ class _GenericRect(Collection[_N]):
...
+        """
     @overload
     def update(
         self, left: float, top: float, width: float, height: float, /
     ) -> None: ...
     @overload
     def update(self, left_top: Point, width_height: Point, /) -> None: ...
     @overload
-    def update(self, single_arg: RectLike, /) -> None: ...
+    def update(self, single_arg: RectLike, /) -> None:
```
aatle, line 341: Better parameter name now that it matters more.
```suggestion
    def update(self, rect: RectLike, /) -> None:
```

buildconfig/stubs/pygame/rect.pyi
```diff
@@ -211,67 +391,319 @@ class _GenericRect(Collection[_N]):
     @overload
     def clipline(
         self, rect_arg: RectLike, /
```
aatle, line 393:
```suggestion
        self, rect: RectLike, /
```

buildconfig/stubs/pygame/rect.pyi
```diff
@@ -211,67 +391,319 @@ class _GenericRect(Collection[_N]):
...
-    ) -> bool: ...
+    def contains(self, left: float, top: float, width: float, height: float, /) -> bool:
+        """Test if one rectangle is inside another.
+
+        Returns true when the argument is completely inside the Rect."""
     @overload
     def collidepoint(self, x: float, y: float, /) -> bool: ...
     @overload
-    def collidepoint(self, x_y: Point, /) -> bool: ...
+    def collidepoint(self, x_y: Point, /) -> bool:
```
aatle, line 509: Might be easier to read:
```suggestion
    def collidepoint(self, point: Point, /) -> bool:
```

buildconfig/stubs/pygame/rect.pyi
```diff
@@ -211,67 +391,319 @@ class _GenericRect(Collection[_N]):
...
-    def unionall(self, rect: SequenceLike[_RectTypeCompatible_co], /) -> Self: ...
-    def unionall_ip(
-        self, rect_SequenceLike: SequenceLike[_RectTypeCompatible_co], /
-    ) -> None: ...
+    def union_ip(self, left: float, top: float, width: float, height: float, /) -> None:
+        """Joins two rectangles into one, in place.
+
+        Same as the ``Rect.union()`` method, but operates in place.
+        """
+    def unionall(self, rect: SequenceLike[_RectTypeCompatible_co], /) -> Self:
```
aatle, line 464: Plural parameters
```suggestion
    def unionall(self, rects: SequenceLike[_RectTypeCompatible_co], /) -> Self:
```

buildconfig/stubs/pygame/rect.pyi
```diff
@@ -211,67 +391,319 @@ class _GenericRect(Collection[_N]):
...
+        """The union of many rectangles.
+
+        Returns the union of one rectangle with a sequence of many rectangles.
+        """
+    def unionall_ip(self, rect: SequenceLike[_RectTypeCompatible_co], /) -> None:
```
aatle, line 469:
```suggestion
    def unionall_ip(self, rects: SequenceLike[_RectTypeCompatible_co], /) -> None:
```

buildconfig/stubs/pygame/rect.pyi
```diff
@@ -28,6 +28,109 @@ _RectTypeCompatible_co = TypeVar(
 )
```
aatle, line 28: This type variable unfortunately looks incredibly verbose in the docs.

I suggest renaming it to `_R_co`.
If that is too undescriptive, then the name `_RectLike_co` should still be much more concise.

buildconfig/stubs/pygame/rect.pyi
```diff
@@ -291,7 +740,24 @@ class _GenericRect(Collection[_N]):
...
+            hashable), so they must be converted to a tuple.
+            e.g. ``rect.collidedictall({tuple(key_rect) : value})``
+
+        .. versionchanged:: 2.4.0
+            ``values`` is now accepted as a keyword argument. Type Stub updated
+            to use boolean ``True`` or ``False``, but any truthy or falsy value
+            will be valid.
+        """
 
 # Rect confirms to the Collection ABC, since it also confirms to
```
aatle, line 762: Fix a few typos while we're here
```suggestion
# Rect conforms to the Collection ABC, since it also conforms to
```

buildconfig/stubs/pygame/rect.pyi
```diff
@@ -144,62 +247,139 @@ class _GenericRect(Collection[_N]):
...
+    def clip(self, left: float, top: float, width: float, height: float, /) -> Self:
+        """Crops a rectangle inside another.
+
+        Returns a new rectangle that is cropped to be completely inside the
+        argument Rect. If the two rectangles do not overlap to begin with, a Rect
+        with 0 size is returned.
+        """
     @overload
     def clipline(
         self, x1: float, x2: float, x3: float, x4: float, /
```
aatle, line 385: Small change to make it 2x clearer.
```suggestion
        self, x1: float, y1: float, x2: float, y2: float, /
```

buildconfig/stubs/pygame/rect.pyi
aatle, line 341: Very minor nitpick but can the single argument and quadruple argument overloads be swapped in order, to be consistent with the signatures of all of the other rect-taking methods?

aatle (commented):
Looks pretty good, just maybe some symbol renaming will help with the clarity issues on the generated docs.

Also, a relevant issue: #3890.

buildconfig/stubs/pygame/rect.pyi
SuperRedingBros: collidelist and collidelistall call their argument equivalent argument rect_list.
Should we change these to rect_list for consistency?

aatle: I personally prefer to instead rename `rect_list` to `rects`.
Reasons being it's more readable IMO (plurality), it's shorter, and the argument doesn't necessarily have to be a `list`; it can be any sequence-like.

I think that all of those plural parameter names should be consistent though, yes.
Stub parameter naming generally has never been particularly careful in the past.

buildconfig/stubs/pygame/rect.pyi
aatle, line 232: Checked generated docs and saw that the constructor also needs to be changed.
```py
    def __init__(self, rect_like: RectLike, /) -> None: ...
```
Not sure if `rect` (consistency) or `rect_like` (more accurate) is best here.
I would think saying `rect_like` (to suggest that lists of 4 numbers work) would be better.

(And the order of overloads should be kept the same of course.)

buildconfig/stubs/pygame/rect.pyi
```diff
@@ -23,11 +23,112 @@ _K = TypeVar("_K")
...
-    "_RectTypeCompatible_co", bound=RectLike, covariant=True
-)
+_RectLike_co = TypeVar("_RectLike_co", bound=RectLike, covariant=True)
 
 class _GenericRect(Collection[_N]):
+    """Pygame object for storing rectangular coordinates.
+    | :sg:`FRect(left, top, width, height, /) -> FRect`
+    | :sg:`FRect((left, top), (width, height), /) -> FRect`
+    | :sg:`FRect(object, /) -> FRect`
+    | :sg:`FRect() -> FRect`
```
aatle, line 33: These must be updated to match the style and parameter names of the generated `Rect` constructor docs.

aatle (commented):
Added a few more findings.

Additionally, the docs generation modification to replace `Self` with the class name has correctly affected some `Color` docs, which is causing tests to fail.
See https://github.com/pygame-community/pygame-ce/actions/runs/34722685588/job/103631306626?pr=3877#step:4:655.
To fix:
```
[dev.py] You need to run `python3 dev.py docs` and commit the changes
```

PR #3879, opened 2026-08-02
Fix rwobject test for 3.15

### LINKED ISSUE #3872: Refcount test failure on 3.15
`py -3.15 -m pygame.tests.rwobject_test -k test_refcount`

```
F
FAIL: test_refcount (__main__.RWopsEncodeStringTest.test_refcount)
  File "C:\Users\charl\AppData\Local\Programs\Python\Python315\Lib\site-packages\pygame\tests\rwobject_test.py", line 79, in test_refcount
    self.assertEqual(sys.getrefcount(bpath), before)
    ~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
AssertionError: 2 != 1

Ran 1 test in 0.001s

FAILED (failures=1)
```

It's unclear to me what has changed here to make this, is it a bug in 3.15? (probably not).
Do we just need to adjust the test? If so, why.

Starbuck5:
Previously, "before" was 2, but in 3.15 it became 1. Keep in mind sys.getrefcount is supposed to give an answer 1 over the true count, so it's reporting a true refcount of zero. I think because of 3.15 optimizations it's being put into a code constant and borrowed, instead of being a true local variable. Anyways, what we want to see here is that bpath has only 1 reference, e.g. 2 reported references.

Fixes #3872

### DIFF
--- a/test/rwobject_test.py
+++ b/test/rwobject_test.py
@@ -76,7 +76,7 @@ class RWopsEncodeStringTest(unittest.TestCase):
         upath = bpath.decode("ascii")
         before = sys.getrefcount(bpath)
         bpath = encode_string(bpath)
-        self.assertEqual(sys.getrefcount(bpath), before)
+        self.assertEqual(sys.getrefcount(bpath), 2)
         bpath = encode_string(upath)
         self.assertIn(sys.getrefcount(bpath), (before, before - 1))
 


oddbookworm:
Do you have anything you can point us to in CPython source or documentation that might explain this one? My biggest concern here is that we may have other refcounting issues if CPython is doing stuff under the hood, and our tests happen to not catch it.

Starbuck5:
I think it may be related to https://github.com/python/cpython/issues/132732

IMO the code that I'm proposing to merge makes way more sense the code before. I have a reference to something, I transform it into a new variable. Why do I care that the new variable has the same refcount the old variable had. That makes no sense.

Instead I've made it test that the new variable has a refcount of 1 (getrefcount reports a 1 higher refcount than real). Which it should, because it's a new object returned to the test, and pygame-ce isn't messing up and double increfing it or anything.

Starbuck5:
This was incorporated into https://github.com/pygame-community/pygame-ce/pull/3900

Issue #3707, opened 2026-02-09
Add color operations in OKLab space


kadir014:
**Description**

`pygame.Color` has very useful and wide range of functionality. One more addition which I think would make sense in game development space is adding color operations & conversions in OKLab space.

It's already being adopted in many similar engines and libraries because of it's strengths. You can read the [blog post by OKLab's creator](https://bottosson.github.io/posts/oklab/).

It most likely would have an API similar to this, I'd like to hear your opinions.
```py
# OKLab -> Linear RGB
# L: lightness
# a: green <-> red
# b: blue <-> yellow
pygame.Color.from_oklab(L: float, a: float, b: float)

# Polar OKLab -> Linear RGB
# L: lightness
# c: chroma
# h: hue
pygame.Color.from_oklch(L: float, c: float, h: float)

# Properties similar to pygame.Color.hsv
pygame.Color.oklab
pygame.Color.oklch
```
It makes it possible to have perceptually much better gradients and calculating similarity between colors more accurately.

Matiiss:
I love this idea ❤️
I suppose it'd have essentially the same interface as other color spaces, but with its own space name (i.e., OKLab)

ankith26:
+1 from my side as well.

PRs welcome :)

PR #3882, opened 2026-08-02
Add oklab and oklch color spaces

SuperRedingBros:
Implementation of the OkLab and OkLch color spaces for pygame.Color as per https://github.com/pygame-community/pygame-ce/issues/3707


src_c/color.c
```diff
@@ -1484,6 +1508,218 @@ _color_set_cmy(pgColorObject *color, PyObject *value, void *closure)
...
+static double
+_correct_srgb_channel_to_linear_inv(double x)
+{
+    if (x >= 0.0031308) {
+        return (1.055) * pow(x, 1.0 / 2.4) - 0.055;
+    }
+    else {
+        return 12.92 * x;
+    }
+}
```
oddbookworm, line 1535: You should also leave a link here explaining why these are the way they are

src_c/color.c
```diff
@@ -1484,6 +1508,218 @@ _color_set_cmy(pgColorObject *color, PyObject *value, void *closure)
...
+        They can be found here: https://bottosson.github.io/posts/oklab/
+    */
+
+    double l = 0.4122214708 * red + 0.5363325363 * green + 0.0514459929 * blue;
+    double m = 0.2119034982 * red + 0.6806995451 * green + 0.1073969566 * blue;
+    double s = 0.0883024619 * red + 0.2817188376 * green + 0.6299787005 * blue;
+
+    double l_ = cbrt(l);
+    double m_ = cbrt(m);
+    double s_ = cbrt(s);
```
oddbookworm: Why use unnecessary memory here? Just reassign them to themselves like you do for red, green, and blue. I know 3 doubles isn't exactly expensive, but I don't see a reason to use more memory than necessary (along with construction time). The straight copy should be faster I think

oddbookworm: Also, I suggest making as much const in these functions as possible to give the compilers opportunities to make optimizations

oddbookworm (changes requested):
This looks great, but I do have a few comments. And another thing I want to bring up is alpha. As far as I know, oklab and oklch don't have alpha baked in, but they do have an optional alpha channel that can be used, but I don't see any indicators that this implementation supports that, nor do you use any rgba colors to test with.

SuperRedingBros:
Good points, I based my implementation's interface off of cmy, which looking at it, is the only one to not support alpha, so I'll add it. I'll make sure to add some explanations for the linear RGB functions and handle the other little things too.
As an aside, should I open an issue about cmy not supporting alpha?

oddbookworm:
> As an aside, should I open an issue about cmy not supporting alpha?

You can, sure. I don't know about adopting it though. CMY(K) is a colorspace mostly designed around printers, where transparency is not a thing. So that issue will end up being a discussion issue to decide if we should even support CMYK(A)

test/color_test.py
```diff
@@ -832,6 +837,9 @@ def test_from_oklab(self):
         }
         for rgb, lab in colors.items():
             c = pygame.Color(rgb)
+            print(c.oklab)
+            c.oklab = c.oklab
+            print(c.oklab)
```
oddbookworm: No debug prints in test cases

test/color_test.py
```diff
@@ -804,19 +804,24 @@ def test_from_cmy(self):
 
     def test_from_oklab(self):
         # Test each method of construction
-        oklab = pygame.Color.from_oklab(0.62796, 0.22486, 0.12585)  # FF0000
+        oklab = pygame.Color.from_oklab(0.62796, 0.22486, 0.12585, 0.5)  # FF0000
         oklab_tuple = pygame.Color.from_oklab((0.62796, 0.22486, 0.12585))
         oklab_instance = pygame.Color(0, 0, 0, 0).from_oklab(0.62796, 0.22486, 0.12585)
 
-        self.assertEqual(oklab, (255, 0, 0))
+        self.assertEqual(oklab, (255, 0, 0, 128))
```
oddbookworm, line 811: I would like to see a few tests that take the results of `from_oklab/from_oklch`, convert back to `oklab/oklch`, and verify that the original and output are the same

oddbookworm (changes requested):

buildconfig/stubs/pygame/color.pyi
```diff
@@ -146,6 +146,37 @@ class Color(Collection[int]):
...
+    1]. ``L`` is perceived lightness, ``C`` is the chroma, ``h`` is the angle of hue in degrees, and ``A`` is alpha.
+    If ``A`` is not provided, 1.0 will be used.
+    Note that while ``h`` will always be returned in the [-180,180] range, any angle is acceptable as input.
+    Note that this will not return the absolutely exact ``Oklch`` values for the set ``RGB``
+    values in all cases. Due to the ``RGB`` mapping from 0-255 and the
+    ``Oklch`` mapping from 0-1, 0-.4, and -180-180 rounding errors may cause the ``Oklch`` values to
+    differ slightly from what you might expect. This is especially the case as chroma approaches zero.
+
+    .. versionadded:: 2.5.8
+    """
```
oddbookworm: I just have one more nitpick, and it's kinda my fault lol. These stubs for the properties rely on 4 components, so a 3-component assignment is going to make linters throw a hissy fit. Here's a quick patch to get it playing nicely:
```patch
diff --git a/buildconfig/stubs/pygame/color.pyi b/buildconfig/stubs/pygame/color.pyi
index 2f1971184..83ab1125b 100644
--- a/buildconfig/stubs/pygame/color.pyi
+++ b/buildconfig/stubs/pygame/color.pyi
@@ -146,7 +146,11 @@ class Color(Collection[int]):
     differ slightly from what you might expect.
     """
 
-    oklab: tuple[float, float, float, float]
+    @property
+    def oklab(self) -> tuple[float, float, float, float]: ...
+
+    @oklab.setter
+    def oklab(self, arg: tuple[float, float, float] | tuple[float, float, float, float]) -> None: ...
     """Gets or sets the Oklab representation of the Color.
 
     The ``Oklab`` representation of the Color. The ``Oklab`` components are in
@@ -161,7 +165,11 @@ class Color(Collection[int]):
     .. versionadded:: 2.5.8
     """
 
-    oklch: tuple[float, float, float, float]
+    @property
+    def oklch(self) -> tuple[float, float, float, float]: ...
+
+    @oklch.setter
+    def oklch(self, arg: tuple[float, float, float] | tuple[float, float, float, float]) -> None: ...
     """Gets or sets the Oklch representation of the Color.
 
     The ``Oklch`` representation of the Color. The ``Oklch`` components are in
```

buildconfig/stubs/pygame/color.pyi
```diff
@@ -255,6 +286,48 @@ class Color(Collection[int]):
...
+        cls, object: tuple[float, float, float] | tuple[float, float, float, float], /
+    ) -> Color: ...
+    @overload
+    @classmethod
+    def from_oklab(cls, l: float, a: float, b: float, /) -> Color: ...
+    @overload
+    @classmethod
+    def from_oklab(cls, l: float, a: float, b: float, alpha: float, /) -> Color: ...
+    @classmethod  # type: ignore
+    def from_oklab(cls, *args) -> Color:  # type: ignore
```
oddbookworm: What's this supposed to be for? This particular stub is giving me weird vibes (same with the oklch version below). Can you give me a valid use-case that isn't covered by the above (or couldn't be covered by the above, like replacing the tuples with a Sequence)?

oddbookworm (changes requested):
I just have a quick couple more comments on the actual content, but also a further question for consistency. Have you verified that say, setting rgba on a color, then setting oklab (without specifying alpha) has the same behavior as, say, `hsv`?

buildconfig/stubs/pygame/color.pyi
SuperRedingBros: I think that's the canonical implementation of the overloaded function, all the other colors also have them.

SuperRedingBros: Never mind, after further investigation that is not applicable to stub files, they are just redundant.
I'll remove them from oklab and oklch, should we do anything about them for the other from_color function?

oddbookworm:
@coderabbitai resume

buildconfig/stubs/pygame/color.pyi
oddbookworm: @coderabbitai open an issue about the other `from_color` functions that have this same weird stub

src_c/color.c
```diff
@@ -1484,6 +1508,258 @@ _color_set_cmy(pgColorObject *color, PyObject *value, void *closure)
...
+
+    _color_get_oklab_to_rgb(lab, color->data);
+    color->data[3] = (Uint8)round(alpha * 255);
+
+    return 0;
+}
+
+#ifndef M_PI
+#define M_PI 3.14159265358979323846
+#endif
```
oddbookworm, line 1676: Why is this here? It should be defined in the standard library includes that this file traces up to

SuperRedingBros: According to the things that I read, M_PI isn't necessarily part of the standard library, so I just defined it myself to be sure it exists.

oddbookworm: Sure, that's fine. However, we also do the same thing in `math.c`, so if we're going to have it defined in multiple places, we might as well move the define somewhere common, like at the end of `src_c/include/_pygame.h`

SuperRedingBros: Should I move it in this PR or should that be something seperate?

oddbookworm: I'll do it, we also do it in a few other places

oddbookworm: @coderabbitai create an issue to have me go fix the `M_PI` define to be in a single location across the codebase. It's currently defined in draw.c, geometry.h, mask.c, math.c, rotozoom.c, and SDL_gfxPrimitives.h. And this pull request is putting it in color.c

oddbookworm:
@coderabbitai fresh review

src_c/color.c
```diff
@@ -1484,6 +1508,258 @@ _color_set_cmy(pgColorObject *color, PyObject *value, void *closure)
...
+    item = PySequence_GetItem(value, 2);
+    if (!item || !_get_double(item, &(lab[2])) || lab[2] < -.4 ||
+        lab[2] > .4) {
+        Py_XDECREF(item);
+        PyErr_SetString(PyExc_ValueError, "invalid oklab value");
+        return -1;
+    }
+    Py_DECREF(item);
+
+    /* a */
```
oddbookworm, line 1655: Duplicated `/* a */`, do you mean to put something like `/* alpha */` to deconflict?

src_c/color.c
```diff
@@ -1484,6 +1508,258 @@ _color_set_cmy(pgColorObject *color, PyObject *value, void *closure)
...
+_color_set_oklab(pgColorObject *color, PyObject *value, void *closure)
+{
+    PyObject *item;
+    double lab[3] = {0, 0, 0};
+
+    DEL_ATTR_NOT_SUPPORTED_CHECK("oklab", value);
+
+    if (!PySequence_Check(value) || PySequence_Size(value) < 3 ||
+        PySequence_Size(value) > 4) {
+        PyErr_SetString(PyExc_ValueError, "invalid oklab value");
```
oddbookworm, line 1622: I would like to see this improved so that it's clearer to the caller *what* is invalid about it. Also, you're overwriting any error messages that python might set internally if `PySequence_Size` fails (like if `value` is not actually a sequence)

src_c/color.c
```diff
@@ -1484,6 +1508,258 @@ _color_set_cmy(pgColorObject *color, PyObject *value, void *closure)
...
+        if (!item || !_get_double(item, &(alpha)) || alpha < 0.0 ||
+            alpha > 1.0) {
+            Py_XDECREF(item);
+            PyErr_SetString(PyExc_ValueError, "invalid oklab value");
+            return -1;
+        }
+        Py_DECREF(item);
+    }
+
+    _color_get_oklab_to_rgb(lab, color->data);
```
oddbookworm, line 1668: This function validates that things aren't NaN except for the alpha, which seems like it would pass silently through?

oddbookworm (changes requested):
A few more, deeper comments

PR #3889, opened 2026-08-06
Add the `__setitem__` method to `PixelArray` stubs.

### LINKED ISSUE #3880: Setting pixels in PixelArray() gives PyCharm warning
pygame-ce 2.5.7 (SDL 2.32.10, Python 3.14.4)
PyCharm 2025.1.6.1 (Community Edition)

`this_pixel_array[0, 0] = transparent_pixel`

gives a warning in PyCharm:

**Class 'PixelArray' does not define '\_\_setitem\_\_', so the '[]' operator cannot be used on its instances**

Example code:

```
"""
Minimal repro to check whether PyCharm flags PixelArray's __setitem__
as unresolved, even though it's valid, documented pygame-ce usage.

It should print a small 4x4 grid of pixel colors after
setting a couple of them, with no runtime errors.
"""

import pygame

pygame.init()

# small throwaway surface just to wrap in a PixelArray
width, height = 4, 4
surface = pygame.Surface((width, height))
surface.fill((0, 0, 0))  # black background

transparent_pixel = surface.map_rgb((255, 0, 255))  # arbitrary "marker" color

this_pixel_array_image = surface
this_pixel_array = pygame.PixelArray(this_pixel_array_image)

# The lines PyCharm warns about:
this_pixel_array[0, 0] = transparent_pixel
this_pixel_array[1, 1] = transparent_pixel

for j in range(width):
    for k in range(height):
        this_pixel_array[j, k] = transparent_pixel if (j, k) in [(2, 2), (3, 3)] else this_pixel_array[j, k]

del this_pixel_array  # release the surface lock PixelArray holds

# confirm it actually worked at runtime
for k in range(height):
    row = [surface.get_at((j, k))[:3] for j in range(width)]
    print(row)

pygame.quit()
```

Assuming this is a genuine stub gap and not my misunderstanding, it is obviously still minor - so if you'd rather not have issues like this reported please let me know!

shaurya-sharma-dev:
Issue #3880 found that the `PixelArray` class does not have stubs for the `__setitem__` method. This PR adds those stubs to therefore fix #3880.

**Note:** All I did was copy the `__getitem__` stubs and removed the return value typehints to the `value` parameter of `__setitem__` (with some adjustments made). Please let me know if I made a mistake!


shaurya-sharma-dev:
@coderabbit review

Starbuck5:
Thanks for taking this on.

The solution is not entirely correct, due to the code being more flexible about input values than output values. It is not symmetric (nor should it be).

In the index_range: slice setter, colors can be broadcast, not just copied from another pixelarray/sequence.

In the ellipsis setter, the value can similarly be a pixelarray, a sequence, or a color. You just have it as a pixelarray.

Additions to celerity's code that works but still have a mypy type error:
```py
this_pixel_array[...] = [1,2,3,4] ##
this_pixel_array[:] = surface.map_rgb((123, 0, 123)) ##
```

shaurya-sharma-dev:
> It is not symmetric (nor should it be).

@Starbuck5 Thanks for pointing that out! I did make some changes, but it's good to know if i ever interact with stubs like this in the future.

What I'm seeing is that this:
```python
@overload
def __setitem__(self, ell: EllipsisType, value: PixelArray) -> None: ...
```

should be this:
```python
@overload
def __setitem__(self, ell: EllipsisType, value: PixelArray | SequenceLike[_PixelColor] | _PixelColor) -> None: ...
```

And that this code:
```python
@overload
def __setitem__(
    self, index_range: slice, value: PixelArray | SequenceLike[_PixelColor]
) -> None: ...
```

should be this:
```python
@overload
def __setitem__(
    self, index_range: slice, value: PixelArray | SequenceLike[_PixelColor] | _PixelColor
) -> None: ...
```

Just so you know, I'm not entirely if the code I just put in this comment to type hint the broadcast is correct, so your help would be much appreciated!

Once you check this, if it looks good to you, I'll add it to the PR.

Starbuck5:
> Just so you know, I'm not entirely if the code I just put in this comment to type hint the broadcast is correct, so your help would be much appreciated!

Why not? Make a test file, run it, run mypy on it.

shaurya-sharma-dev:
> Why not? Make a test file, run it, run mypy on it.

I'm running a bit low on storage right now, and I didn't think I have mypy, but after checking again, I just realized I already have mypy installed, so I'll make that test file right now and check! Thanks again for the guidance!

shaurya-sharma-dev:
I've just tested using mypy with the newest commit, and it runs successfully. The code I tested, for reference:
```python
# pxarray is 4x4
# pxarray2 is 1x4
pxarray[0, 0] = 0xFF00FF
pxarray[0, 0] = pygame.Color(255, 0, 255)
pxarray[0, 0] = (255, 0, 255)
pxarray[...] = [1,2,3,4]
pxarray[:] = surface.map_rgb((123, 0, 123))
pxarray[:1] = pxarray2
```

shaurya-sharma-dev:
@Starbuck5 When I used the below test program to test if everything was correct, it ran successfully, but mypy outputted this:
```
stubtest.py:22: error: Incompatible types in assignment (expression has type "PixelArray | None", target has type "PixelArray | SequenceLike[int | Color | tuple[int, int, int] | tuple[int, int, int, int]] | int | Color | tuple[int, int, int] | tuple[int, int, int, int]")  [assignment]
Found 1 error in 1 file (checked 1 source file)
```

I was able to trace this back to the fact that in line 42 of the pixelarray.pyi file, it has a return typehint of `PixelArray | None` and the typehints I added to `__setitem__` only accept `PixelArray`.

(Lines 39-42 for reference)
```python
# complicated, but I'm pretty sure this is guaranteed to return a PixelArray or None
# will only return None if the slice start and end are the same
@overload
def __getitem__(self, index_range: slice) -> PixelArray | None: ...
```

It doesn't make sense to me that `__getitem__` returns `None` if start and end are same. I can see no situation where the user couldn't check if the start and end are same manually rather than relying on this `None` return behavior. I'm not going to allow `None` in the `__setitem__` stubs because that's just wrong, but changing the existing behavior would break backward compatibility.

The test program I used, for reference:
```python
import pygame as pg

pg.init()

surface1 = pg.Surface((10, 10))
surface1.fill((0, 0, 0))

pxarray1 = pg.PixelArray(surface1)

surface2 = pg.Surface((10, 10))
surface2.fill((255, 0, 0))

pxarray2 = pg.PixelArray(surface2)

pxarray1[0, 0] = 0xFF00FF
pxarray1[0, 0] = pg.Color(255, 0, 255)
pxarray1[0, 0] = (255, 0, 255)
pxarray1[...] = range(10)
pxarray1[:] = pg.Color(255, 0, 255)
pxarray1[:1] = pxarray2[2:3]
pxarray1[:, :] = pxarray2
pxarray1[::2, :] = (255, 0, 0)
pxarray1[::2] = (0, 255, 0)
pxarray1[:, 1] = (0, 0, 255)
pxarray1[1, :] = (0, 0, 255)

# Incorrect usages (added to ensure mypy raises errors for them)
# pxarray1[1, 1] = pxarray2

pg.image.save(surface1, "test.png")
```

aatle:
I agree with the observations.

After inspecting the code and stubs, I see multiple issues with the original code.
See `_pxarray_subscript` and `_pxarray_ass_subscript` in [`pixelarray.c`](https://github.com/pygame-community/pygame-ce/blob/main/src_c/pixelarray.c), and see [`pixelarray.pyi`](https://github.com/pygame-community/pygame-ce/blob/main/buildconfig/stubs/pygame/pixelarray.pyi).

1. (typing only: bug) Missing `__getitem__` overloads for two-element tuples other than `tuple[int, int]`, as well as one-element tuples
2. (mentioned issue: API) Should not return `None` from `__getitem__` with empty slice (not how slicing should work)
3. (API/bug) `__getitem__` with single ellipsis argument returns self reference instead of new copy
4. (API/bug?) `None` is intended to be same as ellipsis, but `None` alone cannot be used as index for `__getitem__` or `__setitem__`
5. (typing only: improvement) Both 1D and 2D `PixelArray`s currently share the same signatures for type checking, which is a mess
6. (bug) Both `__getitem__` and `__setitem__` for both 1D and 2D mostly fail to gracefully handle an index that is a one-element tuple (e.g. `pxarray[("hello",)]`)

aatle:
In my opinion, this PR should be closed or wait until the other typing or API issues are resolved.
Specifically, `__getitem__` and `__setitem__` should be somewhat unified/similar, and I believe one can separate out 1D and 2D pixelarray overloads using generics for (5) above.

shaurya-sharma-dev:
> `None` is intended to be same as ellipsis

No? Ellipsis is meant for when any amount of items can be in an iterable in type hints or for indexing. If I'm missing your point and that's how it's used in the `PixelArray` API, than we should probably rethink that section entirely.

> the code being more flexible about input values than output values

As this comment by @Starbuck5 states, the `__getitem__` and `__setitem__` methods in `PixelArray` are not symmetrical, while I believe they should be; as far as I'm aware, there is no reason for the above statement to be true. For all of the below examples, you should be able to get and set using the same indexes:

```python
pxarray1[0, 0] = 0xFF00FF
pxarray1[0, 0] = pg.Color(255, 0, 255)
pxarray1[0, 0] = (255, 0, 255)
pxarray1[...] = range(10)
pxarray1[:] = pg.Color(255, 0, 255)
pxarray1[:1] = pxarray2[2:3]
pxarray1[:, :] = pxarray2
pxarray1[::2, :] = (255, 0, 0)
pxarray1[::2] = (0, 255, 0)
pxarray1[:, 1] = (0, 0, 255)
pxarray1[1, :] = (0, 0, 255)
```

I agree with @aatle that we should think about reworking `PixelArray` before worrying about this PR. Obviously, I want to avoid backward incompatibility, but there are certain cases where the current implementation doesn't make sense.

aatle:
I am not endorsing the use of `None` in indexing; that is just how I saw the implementation is. Because of backwards compatibility, it has to be part of the API, but it should at least be consistent within it.

It appears that for tuple indices, `None` is treated the same as ellipsis, but not in the case where it is just a single index value.

For example:
```py
# Succeeds
pxarray[None, None] = pxarray[..., ...]
pxarray[..., ...] = pxarray[None, None]
pxarray[...] = pxarray[...]
```
```py
# Fails
pxarray[None] = pxarray[None]
```

aatle:
As for Starbuck5's comment, I think that maybe it is meant that the *values* (colors) are asymmetrical in getting and setting, which is common in Python, rather than meaning that the *indices* are asymmetrical (which would not make sense).

Starbuck5:
I appreciate you taking a look aatle.

I would like to move this PR forward if it improves things, even if it does not fix everything or there are other API discussions to be had. 

Right now we’re developing 3.0, so there is potential for minimally breaking changes, like edge cases in PixelArray. 

We also may cut a 2.5.9 release at some point, which would be great to have improvements like this for OP of this issue, celeritydesign. If we have something that gets 95% of the way there it can be merged and then could be backported to the 2.0 branch.

@aatle would you like to open an issue with those findings / recommendations?

shaurya-sharma-dev:
> I would like to move this PR forward if it improves things, even if it does not fix everything or there are other API discussions to be had.

@Starbuck5 If you want, I can change the target branch of this PR to main-2.x. I don't think we should update the stubs in main for now until the API changes are made, seeing as it is for 3.0.0 development, and we are going to change it later anyway.

> As for Starbuck5's comment, I think that maybe it is meant that the values (colors) are asymmetrical in getting and setting, which is common in Python, rather than meaning that the indices are asymmetrical (which would not make sense).

Yes, I believe that was what it meant too. I misunderstood earlier.

Starbuck5:
Again I do not want the perfect to be the enemy of the good. This improves the stubs, we should do it. If we decide to change the API in the future that effort will be responsible for updating the stubs to match the new API.

aatle (approved):
This PR works fine for what it is trying to do. This can be merged, with refactoring done afterwards.

Starbuck5 (approved):
Thanks!

PR #3898, opened 2026-08-08
Deprecate `filename` argument of `print_debug_info` and add `get_debug_info` to get the raw string

### LINKED ISSUE #3886: Support Passing in `io.TextIOWrapper` objects into `pygame.print_debug_info`
**Description**

Currently, `pygame.print_debug_info` only supports a `filename` parameter that does not support passing in `io.TextIOWrapper` objects. While this is satisfactory for most use cases, in some cases, developers may already have a log file open, and reopening it or passing the log file path in two different places isn't a clean solution. 

While it is possible to use `contextlib.redirect_stdout` to work around this, this still isn't a clean enough solution, and given as it wouldn't be a difficult feature to implement (especially considering the function itself is written in Python), I believe that it should be added to pygame-ce.

**Note:** Because its always better to not break existing behavior, and keeping the parameter called `filename` would be confusing if it accepted a `io.TextIOWrapper` object as well, it could be easier to add this as a separate function and then use that function in `pygame.print_debug_info`. Example:

```python
import sys
def write_debug_info_to_file(file):
    # move the debug string logic from pygame.print_debug_info here and write it to the given file object.

def print_debug_info(filename: str=None):
    if filename is None:
        write_debug_info_to_file(sys.stdout)
    else:
        with open(filename, "w") as file:
            write_debug_info_to_file(file)
```

Edit: Fix class name.

oddbookworm:
#3886 pointed out that `print_debug_info` should handle file objects (say, you already have a file open for logging and you want to write into that existing file without capturing stdout yourself). Instead, I've made updates that deprecate `filename` from `print_debug_info` and add a new method to get the raw string to do with as you please (write to file, as an example)

Closes #3886


oddbookworm:
@coderabbitai resume

oddbookworm:
@coderabbitai fresh review

src_py/_debug.py
```diff
@@ -136,51 +176,39 @@ def default_return(linked=True):
...
-        if driver.upper() != "X11":
-            debug_str += f"Display Driver:\t\t{driver}\n"
-        else:
-            is_xwayland = (environ.get("XDG_SESSION_TYPE") == "wayland") or (
-                "WAYLAND_DISPLAY" in environ
-            )
-            debug_str += f"Display Driver:\t\t{driver} ( xwayland == {is_xwayland} )\n"
-    else:
-        debug_str += "Display Driver:\t\tDisplay Not Initialized\n"
+def print_debug_info(filename: str | None = None) -> None:
```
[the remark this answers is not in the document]
oddbookworm: The scope of this PR changed

Starbuck5 (commented):
Thanks for implementing the suggestion about get_debug_info, I think this is a good path forward.

I do have some notes, some of which are wider than the scope of this PR needs to be but are related.

Why does _debug.pyi exist? I was going to say why are we adding internal helpers into the public stubs, but more importantly why have public stubs for an internal module anyways?

Also, I would appreciate it if you updated the PR title/description (not super in depth or anything, but more in line with implementation so it's easy to parse later).

src_py/_debug.py
```diff
@@ -1,10 +1,12 @@
 """Debug functionality that allows for more useful issue reporting"""
 
 import importlib
+import io
 import platform
 import sys
 import traceback
 from os import environ
+from typing import Protocol
```
Starbuck5: These 2 import changes are not needed anymore

test/debug_test.py
```diff
@@ -12,13 +13,7 @@ def assert_stdout(self, expected_output, mock_stdout):
         self.assertEqual(mock_stdout.getvalue(), expected_output)
 
     def test_print_debug(self):
```
Starbuck5: I get that it's trivial now, but it's strange to have a test called test_print_debug that does not call the print_debug function.

src_py/_debug.py
```diff
@@ -136,51 +176,39 @@ def default_return(linked=True):
...
+            DeprecationWarning,
+            stacklevel=2,
+        )
 
-    else:
-        with open(filename, "w", encoding="utf8") as debugfile:
-            debugfile.write(debug_str)
+    debug_str = get_debug_info()
+
+    print(debug_str, end="")
```
Starbuck5: It seems to strange to have to specify this end= param? I see you're adding a newline at the end of get_debug_info, it seems like it would be simpler to not do that and then not need to do this.

Plus then print(pygame.get_debug_info) would actually equal pygame.print_debug_info().

oddbookworm:
> Why does _debug.pyi exist? I was going to say why are we adding internal helpers into the public stubs, but more importantly why have public stubs for an internal module anyways?

Because it does have two exposed functions, and mypy threw a hissy fit if I only exposed those two functions in stubs

Starbuck5:
> Because it does have two exposed functions, and mypy threw a hissy fit if I only exposed those two functions in stubs

I would expect it to be specified relative to the place it's exposed, "pygame", e.g. `__init__.pyi`.

src_py/_debug.py
oddbookworm: Weird, yes. It's purely because of the `\n` that `print` tacks on by default that this paradigm was what I went with. But I can just modify the test to handle it when I capture sys.stdout with the extra newline

Starbuck5: Why explicitly add a newline to the info string though? https://github.com/oddbookworm/pygame-ce/blob/debug_print_fileobject/src_py/_debug.py#L189

oddbookworm: Again... because of `print` implicitly adding one and the test would fail to match `stdout` to the bare string emitted. I wanted to make sure we were always emitting the exact same thing, even in the case of stdout capture. That's also why I put the `end=""` in the `print` call, to prevent it from adding its own newline

oddbookworm: Could I just do the end and not bother with a newline? Sure, but if you have stdout coming after that, then it would end up being super wonky

Starbuck5: I don't see why it would be an issue to do a normal print and not add an extra newline at the end in get_debug_info?

How does the string get to stdout? It gets printed, so I don't see a scenario where it gets super wonky.

A test testing stdout vs string could be adjusted use splitlines() and remove 1 empty line at the end of the stdout stream, or something like that.

ankith26 (approved):
LGTM, thanks! 🎉

src_py/_debug.py
```diff
@@ -136,51 +174,37 @@ def default_return(linked=True):
...
-    if filename is None:
-        print(debug_str)
+    debug_str = get_debug_info()
 
-    else:
-        with open(filename, "w", encoding="utf8") as debugfile:
-            debugfile.write(debug_str)
+    print(debug_str)
+
+    del warnings
```
Starbuck5: Warnings is local to this function, no need to delete it

test/debug_test.py
```diff
@@ -7,18 +7,15 @@
 
 class DebugTest(unittest.TestCase):
     @unittest.mock.patch("sys.stdout", new_callable=io.StringIO)
-    def assert_stdout(self, expected_output, mock_stdout):
+    def assert_stdout_equal(self, expected_output, mock_stdout):
         pygame.print_debug_info()
-        self.assertEqual(mock_stdout.getvalue(), expected_output)
+        # Stripping stdout to ensure that we don't end up with a newline at the end messing
+        # the compare up
```
Starbuck5: Stdout is not stripped?

docs/reST/ref/pygame.rst
```diff
@@ -205,23 +205,50 @@ object instead of the module, which can be used to test for availability.
...
 
+   .. note::
+      If the display has never been initialized using :func:`pygame.init`, :func:`pygame.display.init`, or by
+      constructing a Window object, then the display driver will be "Display Not Initialized" because that
+      information is not yet available.
+
+   .. note::
+      If ``pygame.mixer`` has not been initialized with :func:`pygame.init` or :func:`pygame.mixer.init`,
+      then the mixer driver will be displayed as "Mixer Not Initialized" because that information is not
+      yet available.
```
Starbuck5, line 230: All these notes apply to get_debug_info more than print_debug_info.

docs/reST/ref/pygame.rst
```diff
@@ -205,23 +205,50 @@ object instead of the module, which can be used to test for availability.
...
+
    .. versionadded:: 2.1.4
 
+   .. versionchanged:: 2.5.0 Added display and mixer driver outputs.
+
+   .. versionchanged:: 2.5.4 Added GIL status to the output.
+
+   .. versionchanged:: 3.0.0
+      Deprecated ``filename`` argument, use the new :func:`pygame.get_debug_info` function if you need to write to file.
```
Starbuck5: I think it would be good to be clear that the functionality is removed, not just deprecated.

And I think there's an argument to removing the argument in totality, because then it's a type error statically and a runtime error to force people to recover from, rather than the program changing behavior with a warning someone could miss in certain situations.

Starbuck5 (commented):
OK I feel like this is merge-able as is,  and I was tempted to approve as is but, I still do have some notes.

Starbuck5 (approved):
OK, LGTM. Thanks!

I would prefer this to be squashed-and-merged because the commit history is on the longer side and contains merge commits as well.

oddbookworm:
I'm good with a squash and merge

PR #3923, opened 2026-08-16
Have dependabot group updates into single PRs

Starbuck5:
Most updates we rubber stamp anyways, this just minimizes the amount of PRs and CI time that go towards this.

https://docs.github.com/en/code-security/reference/supply-chain-security/dependabot-options-reference#groups--


ankith26 (approved):
Good idea, we should be doing this. Thanks! 🎉

PR #3925, opened 2026-08-19
pre-commit: Add zizmor to security lint GitHub Actions

cclauss:
https://docs.zizmor.sh
* https://docs.zizmor.sh/integrations/#pre-commit

https://docs.zizmor.sh/audits/#excessive-permissions encourages setting permissions at the job level rather than at the workflow level.

https://docs.zizmor.sh/audits/#superfluous-actions encourages the use of `gh` instead of [`softprops/action-gh-release`](https://github.com/softprops/action-gh-release).

Also, `pre-commit autoupdate` updated `ruff format` which added a blank line to three Python files.


.github/workflows/release-gh-draft.yml
```diff
@@ -8,9 +8,7 @@ concurrency:
   group: ${{ github.workflow }}-${{ github.ref }}-gh-draft
   cancel-in-progress: true
 
-permissions:
-  contents: read
-  packages: write
+permissions: {}
```
ankith26: We need `contents: read` and `packages: write` permissions here because the sub workflows under this file require these permissions. https://github.com/pygame-community/pygame-ce/actions/runs/31305487273 this is an example of a fail we faced back when `packages: write` was missing from here

ankith26: The update you did is made on another file, accidentally I presume. It should be on this file not `build-debian-multiarch`

cclauss: Sorry... Fixed.

ankith26 (approved):
LGTM, thanks! 🎉

Issue #3539, opened 2025-07-24
Using keyword arguments to create Rects does not work.


Lirhio:
I was hesitant to come here, as this is my first time opening an issue on a GitHub project, but I think reporting this (even though you're already aware of it) could improve the experience for a PyGame noob when the bug is fixed (if it actually is one).

I'm talking about using keyword arguments to create Rects using `pygame.Rect`.

```
button: pygame.Rect = pygame.Rect((100, 100, 200, 50))
```
^ The above code works perfectly

```
button: pygame.Rect = pygame.Rect(single_arg=(100, 100, 200, 50))
```
^ This doesn't work, when I try to draw the Rect using `pygame.draw.rect`, nothing is drawn on the screen. It's weird, I don't think it's supposed to work like that. The same behavior is observed when using `(left=*, top=*, width=* height=*)` or `(left_top=*, width_height=*)`.

Update (as of this writing): The Rect is clearly created when I use print to check, but all of its information is reset. `print(button)` returns `Rect(0, 0, 0, 0)`.

I have no idea if other pygame functions have the same behavior. I apologize if I'm accidentally duplicating an existing issue. I can provide more information about the executed code if needed. I hope this helps!

`pygame.print_debug_info()` below:

```
Platform:               Windows-10-10.0.19045-SP0
System:                 Windows
System Version:         10.0.19045
Processor:              Intel64 Family 6 Model 158 Stepping 10, GenuineIntel        SSE2: Yes       AVX2: Yes   NEON: No
Architecture:           Bits: 64bit     Linkage: WindowsPE

Python:                 CPython 3.13.4 (tags/v3.13.4:8a526ec, Jun  3 2025, 17:46:04) [MSC v.1943 64 bit (AMD64)]
GIL Enabled:            True
pygame version:         2.5.5
SDL versions:           Linked: 2.32.6  Compiled: 2.32.6
SDL Mixer versions:     Linked: 2.8.1   Compiled: 2.8.1
SDL Font versions:      Linked: 2.24.0  Compiled: 2.24.0
SDL Image versions:     Linked: 2.8.8   Compiled: 2.8.8
Freetype versions:      Linked: 2.11.1  Compiled: 2.11.1

Display Driver:         windows
Mixer Driver:           wasapi
```

oddbookworm:
It's positional-only, that's why.

The stub for it could be improved to indicate this fact. Kwarg parsing is very slow, and rect creation needs to be as fast as possible.

[...]

PR #3926, opened 2026-08-20
Make Rect and FRect constructors reject keyword arguments 🤖🤖🤖

vhsgreed:
Fixes the silent zero-rect behaviour reported in https://github.com/pygame-community/pygame-ce/issues/3539.

## What changed

`Rect(left=1, top=2, width=3, height=4)` (or any keyword form, e.g. `Rect(single_arg=(1, 2, 3, 4))`) silently created `Rect(0, 0, 0, 0)` because `RectExport_init` returned early when there were no positional args, ignoring `kwds` entirely.

Per maintainer discussion in #3539 the constructor is **positional-only by design** (keyword parsing is too slow for the hot rect-creation path). So instead of adding kwarg support, `RectExport_init` now raises `TypeError` when keyword arguments are passed — for `Rect` and `FRect` alike (shared `rect_impl.h`), with the type name in the message. The `.pyi` stubs already mark `__init__` as positional-only (`/`), so runtime now matches the type hints.

```python
>>> pygame.Rect(left=1, top=2, width=3, height=4)
TypeError: pygame.rect.Rect() does not accept keyword arguments
```

## Verification

- Built pygame-ce from source (3.0.0.dev1, CPython 3.14, Windows) and confirmed:
  - all keyword forms raise `TypeError` for both `Rect` and `FRect`, including mixed positional+keyword;
  - positional forms are unaffected: `Rect(1, 2, 3, 4)`, `Rect((1, 2), (3, 4))`, `Rect((1, 2, 3, 4))`, `Rect()` → `Rect(0, 0, 0, 0)`.
- Added `test_rect_constructor__no_keywords` (inherited by `FRectTypeTest`).
- Full `test/rect_test.py` suite: **367 tests, all passing** (`python -m unittest test.rect_test`).

> [!INFO] *AI disclosure*: This contribution was prepared with AI assistance on behalf of karlsune. The change is ~10 lines of C; the AI drafted the patch and test, and karlsune verified the build, the runtime behaviour listed above, and the full rect test suite locally.

### DIFF
--- a/src_c/rect_impl.h
+++ b/src_c/rect_impl.h
@@ -885,6 +885,12 @@ RectExport_dealloc(RectObject *self)
 static int
 RectExport_init(RectObject *self, PyObject *args, PyObject *kwds)
 {
+    if (kwds && PyDict_GET_SIZE(kwds)) {
+        PyErr_SetString(PyExc_TypeError,
+                        ObjectName "() does not accept keyword arguments");
+        return -1;
+    }
+
     if (PyTuple_GET_SIZE(args) == 0) {
         return 0;
     }
--- a/test/rect_test.py
+++ b/test/rect_test.py
@@ -16,6 +16,22 @@ class RectTypeTest(unittest.TestCase):
     def _assertCountEqual(self, *args, **kwargs):
         self.assertCountEqual(*args, **kwargs)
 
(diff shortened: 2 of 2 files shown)


oddbookworm:
Nope, good first issue cannot be resolved by automated agent.

PR #3932, opened 2026-08-24
Simplify pkgdata implementation

Starbuck5:
The try/except ImportError is not necessary because this API was added in 3.9, all Python we support has it.

Without the fallback, the NotImplementedError try/catch is not needed (I tested on pyinstaller too to be safe), and I was able to rewrite it to use importlib.resources natively instead of through the compat layer functions, which is more straightforward and theoretically is more efficient (does not have to joinpath() twice).

### DIFF
--- a/src_py/pkgdata.py
+++ b/src_py/pkgdata.py
@@ -20,36 +20,7 @@ object (such as StringIO).
 __all__ = ["getResource"]
 import os
 import sys
-
-try:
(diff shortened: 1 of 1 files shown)


damusss (approved):
Fair enough, LGTM!

ankith26 (approved):
LGTM, thanks 🎉

PR #3934, opened 2026-08-24
Add `exclude` parameter to `spritecollide` and `spritecollideany`

### LINKED ISSUE #3193: Allow Sprite "sprite-vs-group" collide methods to ignore the calling sprite (it shouldn't hit itself!)

shaurya-sharma-dev:
Closes #3193 

# Changes
- Adds `exclude` parameter to `spritecollide` and `spritecollideany` and adds it to documentation and stubs accordingly.
- Adds proper tests for `exclude` for both `spritecollide` and `spritecollideany`.
- Refactors `spritecollide` and `spritecollideany` to use a lambda for `collided` that removes some code duplication.

# Testing
While I was adding tests for the new `exclude` parameter, I noticed a lack of the use of the `msg` parameter provided by `unittest.TestCase`. I'm not aware if this is convention, but I've decided to use `msg` because it allows for easier debugging if something breaks later. For reference, `msg` is a string argument that changes the error message raised if a call like `self.assertEqual` fails.


test/sprite_test.py
```diff
@@ -351,6 +351,80 @@ def collided_callback(
...
+        self.assertIs(
+            collided_sprite,
+            self.s3,
+            "self.s2 may not have been excluded or ordering may not be preserved.",
+        )
+
+        # Since self.ag only contains self.s1, this test ensures that excluding the
+        # only sprite in a group makes it return None.
+        collided_sprite = sprite.spritecollideany(self.s2, self.ag, None, {self.s1})
+        self.assertIs(collided_sprite, None, "self.s1 may not have been excluded.")
```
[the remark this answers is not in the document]
shaurya-sharma-dev: That was a copy paste error that I've resolved in the latest commit.

aatle:
I think that the performance improvement with `kill_if_do_kill` actually may decrease performance.

The original extra for loop is equivalent to a single extra operation on each collided sprite, which `dokill` (killing sprite) necessitates, so there is no way to further decrease the total number of operations.

The proposed code seems to have more operations over the original:
- If `dokill` is true, then there is an extra function call, tuple creation, and tuple access for each sprite
- If `dokill` is false, then there is an extra function call for each sprite

So, this optimization should be performance benchmarked (isolated from any other changes in this PR).

aatle:
I agree that using a lambda for `collided` for the default does improve code reuse.
However, in #3209 which most recently refactored the sprite collision functions, I kept separate pathways for performance reasons.

The new implementation necessitates an extra function call for each sprite with the lambda, which may introduce overhead.
I do not know if this has any noticeable impact, so a performance benchmark on this isolated change would be insightful.

buildconfig/stubs/pygame/sprite.pyi
```diff
@@ -196,6 +196,7 @@ def spritecollide(
     group: AbstractGroup[_SpriteT],
     dokill: bool,
     collided: Callable[[_HasRectT, _SpriteT], bool] | None = None,
+    exclude: set | None = None,
```
aatle: Specify the `set` item type by writing `set[_SpriteT]` instead of plain `set`.

src_py/sprite.py
```diff
@@ -1705,21 +1705,29 @@ def spritecollide(sprite, group, dokill, collided=None):
...
-        sprite_rect_collide = sprite.rect.colliderect
-        collided_sprites = [
-            group_sprite
-            for group_sprite in group
-            if sprite_rect_collide(group_sprite.rect)
-        ]
-    if dokill:
-        for group_sprite in collided_sprites:
-            group_sprite.kill()
+    exclude = exclude if exclude is not None else set()
```
aatle: I personally think that:
```py
if exclude is None:
    exclude = set()
```
Is more readable and concise because it avoids executing code (`exclude = exclude`) in the case where `exclude is not None`.

But a ternary operator is also great.

buildconfig/stubs/pygame/sprite.pyi
shaurya-sharma-dev: I've just fixed this on my local copy! Will be pushing soon after benchmarking the lambda.

src_py/sprite.py
shaurya-sharma-dev: I personally prefer ternary operators, primarily because they are inline, and the cost of doing `exclude = exclude` is practically nothing. However, I just checked how this is done in other lines of the sprite module, and I found this code in `LayeredDirty._find_dirty_area`:

```python
if spr.source_rect:
    _union_rect = _rect(spr.rect.topleft, spr.source_rect.size)
else:
    _union_rect = _rect(spr.rect)
```

So not using ternaries might be how the module itself is written. I won't change anything right now, but I may change it to a full `if` statement later if another maintainer wants me to preserve the non-ternary method instead.

shaurya-sharma-dev:
I benchmarked the ``kill_if_do_kill`` change with the following program to determine that the lambda change does decrease performance. 

However, it is work noting that in certain runs, the performance does actually increase from the lambda change.

In the below program, a negative percentage change means the ``kill_if_do_kill`` change did better than the subsequent iteration, and positive iteration means vice versa.

<details>
<summary>Test Program</summary>

    import pygame as pg
    import time
    import statistics

    def gen_ag():
        ag = pg.sprite.AbstractGroup()

        for _ in range(100):
            s = pg.sprite.Sprite()
            s.image = pg.Surface((10, 10))
            s.rect = s.image.get_rect(center=(0, 0))
            ag.add(s)

        return ag

    def spritecollide_dokill_lambda(sprite, group, dokill, collided=None, exclude=None):
        exclude = exclude if exclude is not None else set()

        collided = (
            collided
            if collided is not None
            else lambda sprite, group_sprite: sprite.rect.colliderect(group_sprite.rect)
        )

        kill_if_do_kill = (
            (lambda group_sprite: (group_sprite.kill(), group_sprite)[1])
            if dokill
            else (lambda group_sprite: group_sprite)
        )

        collided_sprites = [
            kill_if_do_kill(group_sprite)
            for group_sprite in group
            if group_sprite not in exclude and collided(sprite, group_sprite)
        ]
        return collided_sprites

    def spritecollide_dokill_subsequent_iter(sprite, group, dokill, collided=None, exclude=None):
        exclude = exclude if exclude is not None else set()

        collided = (
            collided
            if collided is not None
            else lambda sprite, group_sprite: sprite.rect.colliderect(group_sprite.rect)
        )

        collided_sprites = [
            group_sprite
            for group_sprite in group
            if group_sprite not in exclude and collided(sprite, group_sprite)
        ]

        if dokill:
            for sprite in collided_sprites:
                sprite.kill()

        return collided_sprites

    sprite = pg.sprite.Sprite()
    sprite.image = pg.Surface((10, 10))
    sprite.rect = sprite.image.get_rect(center=(0, 0))

    perc_changes = []

    for _ in range(100):
        ag = gen_ag()
        before_lambda_run = time.perf_counter()
        spritecollide_dokill_lambda(sprite, ag, True)
        lambda_time = time.perf_counter() - before_lambda_run

        ag = gen_ag()
        before_subsequent_iter_run = time.perf_counter()
        spritecollide_dokill_subsequent_iter(sprite, ag, True)
        sub_iter_time = time.perf_counter() - before_subsequent_iter_run

        perc_change = ((lambda_time - sub_iter_time) / abs(sub_iter_time)) * 100
        perc_changes.append(perc_change)
        print(f"Percentage change:\t{perc_change}%")

    print(f"Average percentage change:\t{statistics.mean(perc_changes)}%")
</details>

My Run (this was done on commit a7b1c824b4df7307d06391e86bbba5a15d3c2743 of the branch in this PR, which as of writing this is the latest commit):
```
pygame-ce 3.0.0.dev1 (SDL 2.32.10, Python 3.12.10)
Percentage change:      42.10992707658603%
Percentage change:      56.48996862159829%
Percentage change:      33.333328829491016%
Percentage change:      16.29422887858663%
Percentage change:      -34.895840122963676%
Percentage change:      39.75363103269952%
Percentage change:      -18.570661638600722%
Percentage change:      27.404717303284375%
Percentage change:      22.19073506130561%
Percentage change:      21.555764795220316%
Percentage change:      21.61904610116076%
Percentage change:      16.460500688342975%
Percentage change:      -18.588876907088196%
Percentage change:      23.06976634633503%
Percentage change:      -1.9980447104069459%
Percentage change:      20.99344986503678%
Percentage change:      27.95802050070531%
Percentage change:      16.794629030179024%
Percentage change:      -17.151698992493824%
Percentage change:      13.262731715220996%
Percentage change:      24.78953492677487%
Percentage change:      24.557126625134295%
Percentage change:      19.58296030031566%
Percentage change:      16.848370913763674%
Percentage change:      34.32282765063899%
Percentage change:      1.6535328954620256%
Percentage change:      17.338341447128677%
Percentage change:      15.555551663068211%
Percentage change:      6.661060211439598%
Percentage change:      2.6761828315053764%
Percentage change:      89.0015558744163%
Percentage change:      -26.310274903071125%
Percentage change:      16.8076082212168%
Percentage change:      56.450455087014326%
Percentage change:      71.22186839336915%
Percentage change:      -29.565669115490273%
Percentage change:      -5.104299060784636%
Percentage change:      16.176479579779805%
Percentage change:      28.250699145343823%
Percentage change:      22.592260581704615%
Percentage change:      22.66910315910896%
Percentage change:      25.592809093150343%
Percentage change:      16.272468562288733%
Percentage change:      -30.419977186032938%
Percentage change:      192.65060772217552%
Percentage change:      18.3673517368875%
Percentage change:      23.351152726875622%
Percentage change:      27.440142517655588%
Percentage change:      15.441871319634137%
Percentage change:      17.14015976240689%
Percentage change:      -18.92255656677487%
Percentage change:      66.488891843826%
Percentage change:      25.022837870917726%
Percentage change:      13.65678243436145%
Percentage change:      14.130437361516247%
Percentage change:      16.28569757576211%
Percentage change:      17.607648653021936%
Percentage change:      24.553819303300877%
Percentage change:      -7.382552156298636%
Percentage change:      13.741871465384548%
Percentage change:      -2.560820392655773%
Percentage change:      44.06777790845509%
Percentage change:      10.482693649745388%
Percentage change:      28.04935827229057%
Percentage change:      22.89156691501128%
Percentage change:      19.3160788697601%
Percentage change:      63.00340187003521%
Percentage change:      17.647061922153966%
Percentage change:      -2.597400287414238%
Percentage change:      10.862321964351898%
Percentage change:      25.149981432373213%
Percentage change:      16.651510075778543%
Percentage change:      19.903842628881687%
Percentage change:      40.831769942177445%
Percentage change:      14.11214821084882%
Percentage change:      19.79405209585883%
Percentage change:      27.756651197319854%
Percentage change:      15.820110916392624%
Percentage change:      116.00352951484739%
Percentage change:      20.353167983349145%
Percentage change:      -25.116673080334163%
Percentage change:      15.208159754054105%
Percentage change:      22.912799205179294%
Percentage change:      6.168830971605608%
Percentage change:      77.59412610403804%
Percentage change:      -41.50217234791533%
Percentage change:      10.385962762413167%
Percentage change:      23.020242971398925%
Percentage change:      -3.052804542502849%
Percentage change:      40.66231594434817%
Percentage change:      75.86208848697635%
Percentage change:      21.106257648404824%
Percentage change:      29.86948056307441%
Percentage change:      47.22906533603858%
Percentage change:      118.51525745893596%
Percentage change:      20.107715519212054%
Percentage change:      79.93873710773899%
Percentage change:      22.51539005820062%
Percentage change:      22.21197403884945%
Percentage change:      15.725042268377582%
Average percentage change:      23.12180664859752%
```

shaurya-sharma-dev:
CodeRabbit found this:

> Update both runtime callable synopses.
> 
> The spritecollide synopsis omits exclude. The spritecollideany synopsis also omits its optional arguments. help() therefore shows call signatures that do not match the public API.

I have fixed this on my local branch, but I won't be pushing right now so that it doesn't rerun CI.

buildconfig/stubs/pygame/sprite.pyi
shaurya-sharma-dev: Fix has been pushed.

aatle:
Thanks for the kill benchmark.

Some tips for potentially better benchmarking:
Use a larger group size, and use sleep and `gc.collect()` before running the function, and use f-string percentage formatting e.g. `f"{percentage:.2%}" for nice output.
The percentages seem to vary a lot so perhaps a larger group size would make things more consistent.

If possible, I would also like to see a benchmark testing if this new implementation is any faster or slower than the original implementation; specifically, the set contains checks and the `collided` lambda refactoring may change speed.

shaurya-sharma-dev:
@aatle Thanks for the feedback! I've gone ahead and added `time.sleep` for 100 milliseconds and `gc.collect` calls at the end of each run. I've also gone ahead and updated the group size to 200 (originally 100), and I've added a call to `print_debug_info` at the start.

Some overall statistics from this run:
- Highest percentage change: 100.82%
- Average: 36.82%
- Lowest: -27.08% (Interestingly, this was the only negative value. Definitely an outlier.)

<details> 
<summary>Updated Kill Benchmark Program</summary>


    import pygame as pg
    import time
    import statistics
    import gc

    pg.print_debug_info()

    def gen_ag():
        ag = pg.sprite.AbstractGroup()

        for _ in range(100):
            s = pg.sprite.Sprite()
            s.image = pg.Surface((10, 10))
            s.rect = s.image.get_rect(center=(0, 0))
            ag.add(s)

        return ag

    def spritecollide_dokill_lambda(sprite, group, dokill, collided=None, exclude=None):
        exclude = exclude if exclude is not None else set()

        collided = (
            collided
            if collided is not None
            else lambda sprite, group_sprite: sprite.rect.colliderect(group_sprite.rect)
        )

        kill_if_do_kill = (
            (lambda group_sprite: (group_sprite.kill(), group_sprite)[1])
            if dokill
            else (lambda group_sprite: group_sprite)
        )

        collided_sprites = [
            kill_if_do_kill(group_sprite)
            for group_sprite in group
            if group_sprite not in exclude and collided(sprite, group_sprite)
        ]
        return collided_sprites

    def spritecollide_dokill_subsequent_iter(sprite, group, dokill, collided=None, exclude=None):
        exclude = exclude if exclude is not None else set()

        collided = (
            collided
            if collided is not None
            else lambda sprite, group_sprite: sprite.rect.colliderect(group_sprite.rect)
        )

        collided_sprites = [
            group_sprite
            for group_sprite in group
            if group_sprite not in exclude and collided(sprite, group_sprite)
        ]

        if dokill:
            for sprite in collided_sprites:
                sprite.kill()

        return collided_sprites

    sprite = pg.sprite.Sprite()
    sprite.image = pg.Surface((10, 10))
    sprite.rect = sprite.image.get_rect(center=(0, 0))

    perc_changes = []

    for _ in range(200):
        ag = gen_ag()
        before_lambda_run = time.perf_counter()
        spritecollide_dokill_lambda(sprite, ag, True)
        lambda_time = time.perf_counter() - before_lambda_run

        ag = gen_ag()
        before_subsequent_iter_run = time.perf_counter()
        spritecollide_dokill_subsequent_iter(sprite, ag, True)
        sub_iter_time = time.perf_counter() - before_subsequent_iter_run

        perc_change = ((lambda_time - sub_iter_time) / abs(sub_iter_time)) * 100
        perc_changes.append(perc_change)
        print(f"Percentage change:\t{perc_change:.2f}%")
        time.sleep(.1)
        gc.collect()

    print(f"Average percentage change:\t{statistics.mean(perc_changes):.2f}%")
</details>

<details> 
<summary>New Results (unfortunately, the percentages are not being spaced from the labels properly)</summary>

pygame-ce 3.0.0.dev1 (SDL 2.32.10, Python 3.12.10)
Platform:               Windows-11-10.0.26200-SP0
System Version:         10.0.26200
Processor:              Intel64 Family 6 Model 140 Stepping 1, GenuineIntel     SSE2: Yes       AVX2: Yes       NEON: No

Python:                 CPython 3.12.10 (tags/v3.12.10:0cc8128, Apr  8 2025, 12:21:36) [MSC v.1943 64 bit (AMD64)]
GIL Enabled:            True
pygame version:         3.0.0.dev1
SDL versions:           Linked: 2.32.10 Compiled: 2.32.10
SDL Mixer versions:     Linked: 2.8.2   Compiled: 2.8.2
SDL Font versions:      Linked: 2.24.0  Compiled: 2.24.0
SDL Image versions:     Linked: 2.8.12  Compiled: 2.8.12

Percentage change:      38.17%
Percentage change:      50.92%
Percentage change:      28.52%
Percentage change:      31.61%
Percentage change:      37.36%
Percentage change:      84.01%
Percentage change:      29.57%
Percentage change:      29.00%
Percentage change:      44.24%
Percentage change:      32.44%
Percentage change:      65.76%
Percentage change:      25.00%
Percentage change:      20.80%
Percentage change:      30.17%
Percentage change:      28.69%
Percentage change:      26.27%
Percentage change:      55.03%
Percentage change:      34.15%
Percentage change:      29.25%
Percentage change:      29.95%
Percentage change:      34.75%
Percentage change:      32.69%
Percentage change:      27.92%
Percentage change:      26.06%
Percentage change:      78.40%
Percentage change:      42.85%
Percentage change:      39.58%
Percentage change:      15.85%
Percentage change:      81.32%
Percentage change:      16.02%
Percentage change:      63.83%
Percentage change:      45.73%
Percentage change:      100.82%
Percentage change:      48.91%
Percentage change:      34.62%
Percentage change:      68.73%
Percentage change:      68.69%
Percentage change:      72.92%
Percentage change:      27.92%
Percentage change:      79.58%
Percentage change:      30.32%
Percentage change:      32.82%
Percentage change:      55.63%
Percentage change:      18.02%
Percentage change:      69.05%
Percentage change:      62.40%
Percentage change:      19.53%
Percentage change:      29.84%
Percentage change:      33.57%
Percentage change:      42.81%
Percentage change:      29.21%
Percentage change:      31.76%
Percentage change:      34.90%
Percentage change:      38.38%
Percentage change:      32.74%
Percentage change:      63.63%
Percentage change:      32.93%
Percentage change:      36.91%
Percentage change:      94.01%
Percentage change:      84.08%
Percentage change:      30.71%
Percentage change:      31.83%
Percentage change:      30.52%
Percentage change:      15.69%
Percentage change:      20.51%
Percentage change:      29.98%
Percentage change:      31.02%
Percentage change:      45.15%
Percentage change:      31.10%
Percentage change:      38.47%
Percentage change:      -27.08%
Percentage change:      77.36%
Percentage change:      56.74%
Percentage change:      32.38%
Percentage change:      53.70%
Percentage change:      80.28%
Percentage change:      22.48%
Percentage change:      14.64%
Percentage change:      26.64%
Percentage change:      18.35%
Percentage change:      31.42%
Percentage change:      26.42%
Percentage change:      46.15%
Percentage change:      32.63%
Percentage change:      36.72%
Percentage change:      19.03%
Percentage change:      20.89%
Percentage change:      36.57%
Percentage change:      27.09%
Percentage change:      16.87%
Percentage change:      31.38%
Percentage change:      27.45%
Percentage change:      19.26%
Percentage change:      31.64%
Percentage change:      49.94%
Percentage change:      28.51%
Percentage change:      41.37%
Percentage change:      32.16%
Percentage change:      87.84%
Percentage change:      28.84%
Percentage change:      45.33%
Percentage change:      32.66%
Percentage change:      13.44%
Percentage change:      32.62%
Percentage change:      39.75%
Percentage change:      39.64%
Percentage change:      32.23%
Percentage change:      44.86%
Percentage change:      14.73%
Percentage change:      27.47%
Percentage change:      43.34%
Percentage change:      36.68%
Percentage change:      10.52%
Percentage change:      34.49%
Percentage change:      20.76%
Percentage change:      42.50%
Percentage change:      87.19%
Percentage change:      34.06%
Percentage change:      26.42%
Percentage change:      34.95%
Percentage change:      31.03%
Percentage change:      45.03%
Percentage change:      26.97%
Percentage change:      32.43%
Percentage change:      31.10%
Percentage change:      51.94%
Percentage change:      46.09%
Percentage change:      27.96%
Percentage change:      30.41%
Percentage change:      33.29%
Percentage change:      32.24%
Percentage change:      29.29%
Percentage change:      37.76%
Percentage change:      30.76%
Percentage change:      46.16%
Percentage change:      32.19%
Percentage change:      44.88%
Percentage change:      25.03%
Percentage change:      15.42%
Percentage change:      16.69%
Percentage change:      40.68%
Percentage change:      44.57%
Percentage change:      31.89%
Percentage change:      38.03%
Percentage change:      32.25%
Percentage change:      51.46%
Percentage change:      27.89%
Percentage change:      33.03%
Percentage change:      28.48%
Percentage change:      16.39%
Percentage change:      20.81%
Percentage change:      31.23%
Percentage change:      32.15%
Percentage change:      41.12%
Percentage change:      32.84%
Percentage change:      29.29%
Percentage change:      13.73%
Percentage change:      43.96%
Percentage change:      25.01%
Percentage change:      38.81%
Percentage change:      31.74%
Percentage change:      28.95%
Percentage change:      51.07%
Percentage change:      44.65%
Percentage change:      21.75%
Percentage change:      7.27%
Percentage change:      41.17%
Percentage change:      28.61%
Percentage change:      88.48%
Percentage change:      44.15%
Percentage change:      17.00%
Percentage change:      32.73%
Percentage change:      26.09%
Percentage change:      31.97%
Percentage change:      46.24%
Percentage change:      26.87%
Percentage change:      30.53%
Percentage change:      35.32%
Percentage change:      31.07%
Percentage change:      28.73%
Percentage change:      18.22%
Percentage change:      39.40%
Percentage change:      27.03%
Percentage change:      24.32%
Percentage change:      35.35%
Percentage change:      44.35%
Percentage change:      31.93%
Percentage change:      29.24%
Percentage change:      33.53%
Percentage change:      41.67%
Percentage change:      52.97%
Percentage change:      15.62%
Percentage change:      32.97%
Percentage change:      46.30%
Percentage change:      54.45%
Percentage change:      37.51%
Percentage change:      34.19%
Percentage change:      30.55%
Percentage change:      44.50%
Percentage change:      17.42%
Average percentage change:      36.82%

</details>

shaurya-sharma-dev:
@aatle I was hoping I could get your help with something. When I was benchmarking current vs new implementation, I noticed that even when I called the original function twice and benchmarked its time against itself, I got an average percentage change of 70.77% with 200 runs. I'm also suspecting that this affected the kill benchmark test too, so I'm going to try and add `time.sleep` and `gc.collect` in the middle of the two runs as well.

shaurya-sharma-dev:
I've decided that instead of continuing to paste benchmark code and results here, I'm going to put them in an orphan branch in my fork, so that's it's faster for me to upload benchmark data.

Below is the link to the first such benchmark, which is the benchmark for lambda vs. iteration at the end for dokill (lambda still decreases performance by 17.82% on average):

https://github.com/shaurya-sharma-dev/pygame-ce/tree/43b15ab7d73d76d5b83e38397813aa76c936cfd2/dokill-benchmarks

shaurya-sharma-dev:
Here is the next benchmark for this PR vs current implementation. This PR has a 15.59% slowdown from the current implementation, likely caused by checking if in `exclude` for each sprite in the group (even with empty `exclude set, this probably has some overhead).

https://github.com/shaurya-sharma-dev/pygame-ce/tree/9a6c196a5790f40f35606568c7bf6caf95564f0a/new-implementation-vs-current

aatle:
> @aatle I was hoping I could get your help with something. When I was benchmarking current vs new implementation, I noticed that even when I called the original function twice and benchmarked its time against itself, I got an average percentage change of 70.77% with 200 runs. I'm also suspecting that this affected the kill benchmark test too, so I'm going to try and add `time.sleep` and `gc.collect` in the middle of the two runs as well.

Variation may decrease with a sprite group size larger than 100.
Additionally, the `time.sleep()` should occur after both `gc.collect()` and `gen_ag()`, in order to ensure the actual function call starts in a more consistent CPU state. I am not sure of the best delay but likely even a small delay like `0.05` will work.

You can check the standard deviation of your results to verify whether the results are significant or not. If the abs of average is greater than around 1.5-3 standard deviations, then the result could be considered more reliable perhaps.

aatle:
I would also recommend that the set contains check be after the collision check because collision is rare, saving many set contains checks, while doing set contains first hardly saves many collision checks.

aatle:
Additionally, the proposed API is too restrictive with the type on `exclude`; groups, dictionaries, frozensets, even lists and tuples (for small sizes it is just as fast) should be allowed.

If I had to implement this, I would instead use `Container[_Sprite]` where `Container` is from `collections.abc`, and the default parameter value being the empty tuple `()` (which removes the default `exclude` `None` check).

aatle:
But whether this new API is a feature that should be implemented is up to the community and other members.
(It is preferred to discuss about new API on an issue rather than first making a PR for it, because of the common case that it does not get approved.)

Personally I am not sure yet if this feature is worth it, so I am waiting for other contributors to chime in.

shaurya-sharma-dev:
> It is preferred to discuss about new API on an issue rather than first making a PR for it, because of the common case that it does not get approved.

Issue #3193 was what I used as the basis of this PR. There was also another PR that fixed this issue, but that got stalled sometimes last year. I saw it and thought I would give it a go.

> groups, dictionaries, frozensets, even lists and tuples (for small sizes it is just as fast) should be allowed

The code right now only needs an object that has the `__contains__` magic method, its just the stubs that enforces the rule to only use `set`. I'm actually going to update this right now to allow `Container`, as this seems to be primarily a docs and stubs change.

shaurya-sharma-dev:
@aatle I ran a new benchmark with the improvements you suggested.

https://github.com/shaurya-sharma-dev/pygame-ce/tree/e1498e2c8a16eceaa4b95c6bb6c01b364d67cab6/new-implementation-vs-current

> You can check the standard deviation of your results to verify whether the results are significant or not.

This only applies if the data is normally distributed. I calculated population standard deviation just in case, but I highly doubt it will actually be useful for this reason.

aatle:
> Issue https://github.com/pygame-community/pygame-ce/issues/3193 was what I used as the basis of this PR. There was also another PR that fixed this issue, but that got stalled sometimes last year. I saw it and thought I would give it a go.

I assume you are referring to PR #3197.
I generally agree with your implementation of a fix for the issue rather than an `ignore_self` parameter.

> The code right now only needs an object that has the __contains__ magic method, its just the stubs that enforces the rule to only use set.

Yes, that is why I suggested `Container`. To be clear, I was only referring to the stubs for the change.

> I'm actually going to update this right now to allow Container, as this seems to be primarily a docs and stubs change.

I would suggest using a tuple for the default parameter value still.

> I have fixed this on my local branch, but I won't be pushing right now so that it doesn't rerun CI.

Don't worry about that, just push. Contributors need to see your new changes in order to review. CI runs plenty of times on inconsequential PRs like my #3935.

shaurya-sharma-dev:
> Don't worry about that, just push. Contributors need to see your new changes in order to review.

Got it, will note.

Just finished the stub and docs changes.

shaurya-sharma-dev:
> I would suggest using a tuple for the default parameter value still.

I did some research into this, and it does seem that sets to take more time to create, and while sets are generally faster for lookups, when both are empty, tuples are faster than sets because sets need to compute hashes.

Will be updating PR to use tuple as the default value.

shaurya-sharma-dev:
Benchmarked change to `tuple` default.

https://github.com/shaurya-sharma-dev/pygame-ce/tree/3441cfaddc9b942532f7ddabd538d3666b88642d/new-implementation-vs-current

src_py/sprite.py
```diff
@@ -1686,10 +1686,10 @@ def collide_mask(left, right):
     return leftmask.overlap(rightmask, (xoffset, yoffset))
 
 
-def spritecollide(sprite, group, dokill, collided=None):
+def spritecollide(sprite, group, dokill, collided=None, exclude=None):
```
aatle: Change to `exclude=()` and remove the `None` type from the stubs. (Use `()` to create an empty tuple as a literal, instead of using the `tuple` constructor.)
Then remove the `exclude` ternary conditional assignment.

Edit both `spritecollide` and `spritecollideany`.

Note: A tuple can safely be used as a default parameter value because it is immutable, while sets are mutable.

shaurya-sharma-dev: Didn't know this. Thanks for pointing this out! Will fix this.

shaurya-sharma-dev:
Not sure if this is outside of the scope of this PR, but the docstring for the `spritecollideany` function inconsistently documents  the `collided` argument when compared to the `spritecollide` function docstring, the `spritecollide` documentation, and even the `spritecollideany` function's own documentation. Every other location has the same documentation.

`spritecollide` function docstring:
```
....
The collided argument is a callback function used to calculate if two
    sprites are colliding. it should take two sprites as values, and return a
    bool value indicating if they are colliding. If collided is not passed, all
    sprites must have a "rect" value, which is a rectangle of the sprite area,
    which will be used to calculate the collision.
....
```

`spritecollide` function documentation:
```
....
The collided argument is a callback function used to calculate if two
   sprites are colliding. it should take two sprites as values, and return a
   bool value indicating if they are colliding. If collided is not passed, all
   sprites must have a "rect" value, which is a rectangle of the sprite area,
   which will be used to calculate the collision.
....
```

`spritecollideany` function documentation:
```
....
The collided argument is a callback function used to calculate if two sprites are
   colliding. It should take two sprites as values and return a bool value
   indicating if they are colliding. If collided is not passed, then all
   sprites must have a "rect" value, which is a rectangle of the sprite area,
   which will be used to calculate the collision.
....
```

`spritecollideany` function docstring (INCONSISTENT):
```
....
Collided is a callback function used to calculate if two sprites are
    colliding. It should take two sprites as values and return a bool value
    indicating if they are colliding. If collided is not passed, then all
    sprites must have a "rect" value, which is a rectangle of the sprite area,
    which will be used to calculate the collision.
....
```

EDIT: The newline differences in the `spritecollideany` function documentation don't change the actual text of the `collided` argument documentation compared to the others.

src_py/sprite.py
shaurya-sharma-dev: Fixed!

shaurya-sharma-dev:
Ran benchmark for default argument change.
https://github.com/shaurya-sharma-dev/pygame-ce/tree/fe22044532a43bcfa02ab34233f55b9d6be1be0a/new-implementation-vs-current

EDIT: Forgot to mention this earlier, but changing this does (very slightly) speed up performance on average.

PR #3935, opened 2026-08-24
Fix typos in pixelarray API docs

aatle:
Also rewrite explanation of `strides`.

### DIFF
--- a/docs/reST/ref/pixelarray.rst
+++ b/docs/reST/ref/pixelarray.rst
@@ -191,15 +191,13 @@
       To calculate the byte offset, the index for each dimension is multiplied
       by the corresponding stride, and then added up (for 2D pixel arrays).
 
-      For example, a standard PixelArray's strides
-      for a typical 10x20 Surface is ``(4, 40)``.
-
-      Each pixel in the Surface is 4 bytes: 1 byte for each of the standard 3 color
-      channels (RGB/BGR) plus the 1 potentially unused alpha channel (A).
-
-      The ``4`` x stride is because each pixel is 4 bytes.
-      The ``40`` y stride is 4 times width; each row of 10 pixels is 40 bytes, so
-      increasing the y index by 1 corresponds to moving 40 bytes forward to skip a row.
+      For example, a PixelArray's strides for a typical 32-bit 10x20 Surface
+      would be ``(4, 40)``. The ``4`` in the x stride is because each
+      pixel is 32-bits, 4 bytes. The ``40`` in the y stride represents the
+      number of bytes to skip forward a row. In this case, the surface pixels
+      are contiguous, so it is equivalent to the size of a pixel * the width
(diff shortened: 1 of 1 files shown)


docs/reST/ref/pixelarray.rst
```diff
@@ -184,10 +183,14 @@
...
-      A tuple or length :attr:`ndim` byte counts. When a stride is
-      multiplied by the corresponding index it gives the offset
-      of that index from the start of the array. A stride is negative
-      for an array that has is inverted (has a negative step).
+      A tuple of length :attr:`ndim` giving each stride in bytes. Multiplying
+      the index and its corresponding stride gives the offset in bytes for that
+      index, which are summed together from each index (for 2D pixel arrays)
+      to get the total byte offset from the start of the array.
+
+      For example, a standard PixelArray's strides for a 10x20 Surface is ``(4, 40)``.
```
Starbuck5: I don’t get this overall section (around highlighted line).

What the heck is an offset for an index, what does that actually mean

And why does a standard PixelArray have those strides. Is it off a 4 byte per pixel surface? If so, that should be clear to know where that 4 is coming from. And then thats 10x4 =40 presumably.

So a 2D PixelArray the first dimension is the pixel the second dimension is the width?

Starbuck5: Okay thanks aatle thats a big improvement. I would just like it to be added when you first introduce the surface that it is 4 bytes per pixel. I don’t want to be like “a standard surface ofc it’s 4 bpp”

docs/reST/ref/pixelarray.rst
```diff
@@ -184,10 +183,25 @@
...
+
+      For example, a standard PixelArray's strides
+      for a typical 10x20 Surface is ``(4, 40)``.
+
+      Each pixel in the Surface is 4 bytes: 1 byte for each of the standard 3 color
+      channels (RGB/BGR) plus the 1 potentially unused alpha channel (A).
+
+      The ``4`` x stride is because each pixel is 4 bytes.
+      The ``40`` y stride is 4 times width; each row of 10 pixels is 40 bytes, so
+      increasing the y index by 1 corresponds to moving 40 bytes forward to skip a row.
```
Starbuck5:
```suggestion
      For example, a PixelArray's strides for a typical 32-bit 10x20 Surface
      would be ``(4, 40)``. The ``4`` in the x stride is because each
      pixel is 32-bits, 4 bytes. The ``40`` in the y stride represents the
      number of bytes to skip forward a row. In this case, the surface pixels
      are contiguous, so it is equivalent to the size of a pixel * the width
      of the surface, 4 * 10. Some Surfaces can have strides greater than the
      width and pixel count.
```

How does this one read to you?

aatle: Seems good.

Starbuck5 (approved):
Thanks!

Issue #3929, opened 2026-08-21
Global argument parsing pass for pygame-ce 3.0


Starbuck5:
This is a mega issue for tracking a variety of audits we should do of our arguments. A major release is a great opportunity to go through things like this.

Numeric types:
- Floats
    - Do things accept them that shouldn't? Do things that should accept them not?
    - What happens to them, do they get truncated? Is that okay?
- Integer sizes
    - Python 3.14 introduces newer things like PyLong_AsInt32, PyLong_AsUInt64, lets use. (Available with our pythoncapi_compat.h header)
    - What happens on overflows?

General arguments (parameters)
- Can I pass things through that don't end up getting used at all? Should be an error
- Is it possible for strange arguments to get accepted? Like recursive parsers accepting deeply nested tuples.

Related / Examples:
- Using keyword arguments to create Rects does not work. https://github.com/pygame-community/pygame-ce/issues/3539
- FRect - incorrect positioning for negative values https://github.com/pygame-community/pygame-ce/issues/2409 -> I think blitting should use rounded floats, not truncated floats
- pygame.Rect values overflow for inputs greater than 2147483647 https://github.com/pygame-community/pygame-ce/issues/3815
- `pygame.Surface(size=(20,20), depth=32, masks=(0xFF, 65280.12345, 0xFF0000, 0xFF000000))` <- why does that accept a float. It shouldn't, imo. It's integer input.
- We have a whole `_pg_pylong_to_uint32` function in color.c, could that just be PyLong_AsUInt32?
- `pygame.Rect((10,20),(((((5,10))))))` <- this is valid code, just no
- Recursive parsing can crash the entire process: (stack overflow?)
```py
>>> import pygame
>>>
>>> class X: pass
...
>>> x = X()
>>> x.rect = pygame.Rect(1,2,3,4)
>>>
>>> x.rect = x
>>> pygame.Rect(x)
```
- https://github.com/pygame-community/pygame-ce/issues/2609

aatle:
I would suggest that floats be *floored* instead of truncated, and instead of rounded.

Truncation is clearly undesirable because it treats negative and positive values differently (rounds towards zero).

Flooring aligns with the convention that integer coordinates of a pixel are positioned exactly at the top left vertex of the pixel (e.g. `(0, 0)` is exactly the top left vertex of the surface), which is similar to OpenGL's convention.
(The center of a pixel is at `(x + 0.5, y + 0.5)`.)

[...]

PR #3938, opened 2026-08-25
Add `Window.show_system_menu`

damusss:
Is this game changing? no...
Will this be used by most of the users? no...
Are those good reasons not to implement `SDL_ShowWindowSystemMenu`? I don't believe so, but I'm open for criticism. For now the PR exists at least. It requires minimal amount of code anyways.

### DIFF
--- a/buildconfig/stubs/pygame/window.pyi
+++ b/buildconfig/stubs/pygame/window.pyi
@@ -1,5 +1,3 @@
-from typing import overload
-
 from pygame.locals import WINDOWPOS_UNDEFINED
 from pygame.rect import Rect
 from pygame.surface import Surface
@@ -513,18 +511,21 @@ class Window:
 
         .. versionadded:: 2.5.2
         """
+    def show_system_menu(self, pos: Point, /) -> bool:
+        """Show the system's default window menu at the given position in pixels.
 
-    @overload
-    def show_system_menu(self, x: int, y: int, /) -> None: ...
-    @overload
-    def show_system_menu(self, pos: Point, /) -> None: ...
-    def show_system_menu(self, *args) -> None:  # type: ignore
-        """Show the system's default window menu at the given position.
+        The system's window menu is a context menu provided and rendered by
+        the window manager. It lists available default operations that manage
+        the window like dragging, resizing, minimizing, maximizing, restoring,
+        and closing it.
 
-        The position is relative to the origin (top left) of the client area.
+        The position is relative to the origin (top left) of the client area,
+        which is the drawable area inside the system's window frame, title bar
+        and decoration. On borderless windows, it corresponds to the whole
+        window.
 
-        On platforms or desktops where this is unsupported, this method
-        does nothing.
+        Returns ``True`` if the operation succeeded and ``False`` if it's
+        not supported.
 
         .. versionadded:: 3.0.0
         """
--- a/buildconfig/stubs/pygame/window.pyi
+++ b/buildconfig/stubs/pygame/window.pyi
@@ -511,7 +511,8 @@ class Window:
 
         .. versionadded:: 2.5.2
         """
-    def show_system_menu(self, pos: Point, /) -> bool:
+
+    def show_system_menu(self, position: Point, /) -> bool:
         """Show the system's default window menu at the given position in pixels.
 
         The system's window menu is a context menu provided and rendered by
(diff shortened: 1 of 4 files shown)


buildconfig/stubs/pygame/window.pyi
```diff
@@ -511,4 +513,20 @@ class Window:
 
         .. versionadded:: 2.5.2
         """
+
+    @overload
+    def show_system_menu(self, x: int, y: int, /) -> None: ...
+    @overload
+    def show_system_menu(self, pos: Point, /) -> None: ...
+    def show_system_menu(self, *args) -> None:  # type: ignore
```
Starbuck5: For new code I would prefer to be more strict about input types. Lets just have it accept a position as one arg, instead of sometimes maybe 2 args.

src_c/window.c
```diff
@@ -1354,6 +1354,28 @@ window_flash(pgWindowObject *self, PyObject *arg)
     Py_RETURN_NONE;
 }
 
+static PyObject *
+window_show_system_menu(pgWindowObject *self, PyObject *args)
+{
+#if SDL_VERSION_ATLEAST(3, 0, 0)
+    int x, y;
+
+    if (!pg_TwoIntsFromObj(args, &x, &y)) {
```
Starbuck5: I have gripes with this function r.e. https://github.com/pygame-community/pygame-ce/issues/3929 but that global pass can occur after new features are added like this one. No action needed on this PR, I just wanted to share.

buildconfig/stubs/pygame/window.pyi
```diff
@@ -511,4 +513,20 @@ class Window:
...
+        """Show the system's default window menu at the given position.
+
+        The position is relative to the origin (top left) of the client area.
```
Starbuck5: (I know what it is because I ran test code, but as a user...). What is a "system default window menu" ??

What is the client area?? Is the position in pixels, or DPI independent "units", something else ?

src_c/window.c
```diff
@@ -1354,6 +1354,28 @@ window_flash(pgWindowObject *self, PyObject *arg)
...
+        return RAISE(PyExc_TypeError,
+                     "invalid position argument for Window.show_system_menu");
+    }
+
+    if (!SDL_ShowWindowSystemMenu(self->_win, x, y)) {
```
Starbuck5: According to SDL source if this is unsupported, this will return `false` with message "That operation is not supported". This is not the behavior claimed in the docs.

I like the behavior claimed in the docs, the other methods here will raise exceptions when unsupported and it means anyone on wayland will have a completely miserable time doing anything because plenty of stuff doesn't work there.

I guess the ideal would be to return true on success, false on failure? That way the app can know if the menu launched.

src_c/window.c
damusss: Thank you for sharing!

src_c/window.c
damusss: returning a boolean makes sense. I suppose most of the errors are raised because they involve properties, which can't really return a value (poor wayland users)... Should we change Window.flash to return a boolean as well? or is the try-except behavior suggested in the docs too ingrained in people's code? (It's technically not a breaking change)

Starbuck5: I would consider it a tiny break. But that's what this release is for. If that's changed it would definitely need a note in our nascent readme-migration. Feel free to open an issue about it, I'd want to get Ankith's opinion on it.

Starbuck5 (approved):
Alright, I think is nearly done, good enough to approve.

The things I see right now are that there needs to be a new line between the previous docstring and this one, and I think the client area explanation is overly complex. I think the whole thing could boil down to "the drawable area of the window" -- it's introducing and defining a new term "client area" instead of describing in more plain language.

damusss:
Thank you! I would have hoped the formatter would add that newline by itself.

you're right, I simplified it now. I still kept the borderless note because this function is intended for borderless windows (normal windows can already trigger the system menu on their titlebar) so the specification is still somewhat useful in a way.

I've also taken the liberty to rename the stub parameter from `pos` to `position`, @Starbuck5 (necessary ping since you already approved it) for the following reasons:
1) it's a positional-only argument, people can't even use a keyword, so why shorten it
2) In the whole Window API position is never shortened (for example Window.position), so why introduce it now.
3) The C code and the docs were already giving errors/information with the name "position"
(sorry for blabbing so long about a simple rename 😆)

PR #3940, opened 2026-08-25
Add `Window.progress_state` and `Window.progress_value`

damusss:
Implement `SDL_SetWindowProgressState`, `SDL_GetWindowProgressState`, `SDL_SetWindowProgressValue`, `SDL_GetWindowProgressValue` in the `Window.progress_state` and `Window.progress_value` properties. 5 constants were also added to reflect the possible progress states.

Very useful for any application in a loading state, for example when games/apps are loading assets, downloading files or performing some heavy async operation. It's the natural successor of `Window.flash` which was previously implemented.


src_c/window.c
```diff
@@ -966,6 +966,79 @@ window_get_utility(pgWindowObject *self, void *v)
...
+    if (SDL_SetWindowProgressValue(self->_win, progress_value) == SDL_FALSE) {
+        RAISERETURN(pgExc_SDLError, SDL_GetError(), -1);
+    }
+
+    return 0;
+#else
+    RAISERETURN(pgExc_SDLError, "Window.progress_value requires SDL 3.0.0+",
+                -1)
+#endif
+}
```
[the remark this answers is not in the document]
damusss: In /meson.build, the line
> sdl_ver = (sdl_api == 3) ? '3.4.0' : '2.32.10'

specifies the SDL3 version to be 3.4.0. if the checks fail, I'd be proven wrong, but otherwise I am confident the first SDL3 version supported by pygame 3 will be at least 3.4.0

src_c/window.c
```diff
@@ -966,6 +966,79 @@ window_get_utility(pgWindowObject *self, void *v)
...
+#else
+    return RAISE(pgExc_SDLError, "Window.progress_state requires SDL 3.4.0+");
+#endif
+}
+
+static int
+window_set_progress_state(pgWindowObject *self, PyObject *arg, void *v)
+{
+#if SDL_VERSION_ATLEAST(3, 4, 0)
+    int progress_state = (int)PyLong_AsLong(arg);
```
Starbuck5: I am not able to do a full review of this today, so forgive the bits-and-piece review, but this caught my eye.

`(int)PyLong_AsLong(arg);` ❌ 
`PyLong_AsInt(arg);` ✅ 

It's safer on platforms where long is not the same as int.

damusss: Still trying to figure out the history of how long came into existence without actually being longer most of the time.

Thank you for pointing it out :)

Starbuck5 (commented):
Alright, this is nifty!

I still have my note about PyLong_AsInt, and also my comments about SDL_FALSE and "This function should only be called on the main thread." apply here. Other than that looks good, I gave it a test and it looked cool.

damusss:
As I specified on discord, every single window operation requires the main thread. Specifying it on every method and property will get messy. The note on top of the window constructor should be enough, and I'll update it soon reflecting exact sdl3 requirements. I was unaware SDL_TRUE/SDL_FALSE were not a thing in sdl3, thanks.

Starbuck5 (approved):
Thanks!

ankith26:
We have `dev.py format` failing on this branch as of now

damusss:
@ankith26 thank you for reminding me, fixing the merge conflicts with the online editor didn't let me run the format command

PR #3947, opened 2026-08-30
Make repr subclass-aware for core geometry types 🤖🤖🤖

### LINKED ISSUE #3044: `repr()` of some classes doesn't use subclass name
**Environment:**

<details><summary> pygame.print_debug_info() </summary>

```
pygame-ce 2.5.0 (SDL 2.30.3, Python 3.12.4)
Processor:              Intel64 Family 6 Model 167 Stepping 1, GenuineIntel

Python:                 CPython 3.12.4 (tags/v3.12.4:8e8a4ba, Jun  6 2024, 19:30:16) [MSC v.1940 64 bit (AMD64)]
pygame version:         2.5.0
SDL versions:           Linked: 2.30.3  Compiled: 2.30.3

```
</details>

**Current behavior:**

In some classes such as `Surface`, `Rect`, `Vector2`, and others, the `repr()` behavior uses a hard-coded class name string instead of a dynamic access of the class name.
Subclass instances and normal instances appear the same with `repr()`.
Example with `Vector2` subclass:
```python
>>> repr(MyPosition(0.1, 0.2))
'Vector2(0.1, 0.2)'
```

**Expected behavior:**

The `repr()` should include the name of the actual instance class instead of the class that implemented the functionality.
```python
>>> repr(MyPosition(0.1, 0.2))
'MyPosition(0.1, 0.2)'
```

**Test code**

```python
import pygame as pg

class S(pg.Surface):
  __slots__ = ()

print(S([1,1]))

class R(pg.Rect):
  __slots__ = ()

print(R(0,0,5,5))

class V(pg.Vector2):
  __slots__ = ()

print(repr(V(1.2, 1.3)))
```

Side note:
I think the `Surface` repr should also say when it has per-pixel alpha.

bagel786:
## Summary

Use the unqualified runtime type name in repr strings for `Rect`, `FRect`, `Vector2`, `Vector3`, `Circle`, `Line`, and `Surface`. Subclasses now identify themselves correctly while base-class output remains unchanged.

Adds coverage for local and nested subclasses, including live and dead surfaces.

```python
class MyPosition(pygame.Vector2):
    pass

# Before: Vector2(0.1, 0.2)
# After:  MyPosition(0.1, 0.2)
```

This intentionally leaves `Surface` alpha reporting unchanged; that was discussed separately from the subclass-name fix in the issue.

Fixes #3044.

## Testing

- Native editable build completed successfully.
- `rect_test` — 367 passed.
- `math_test` — 185 passed.
- `geometry_test` — 147 passed.
- `surface_test` — 139 passed, 3 skipped.
- `python dev.py lint` — 10.00/10.
- `python dev.py format` — all pre-commit hooks passed, including clang-format.

## AI-generated code

AI generated 100% of the implementation, tests, and PR draft. Validation consisted of local diff inspection, a native build, and the affected test suites listed above.

### DIFF
--- a/src_c/circle.c
+++ b/src_c/circle.c
@@ -91,7 +91,8 @@ pg_circle_repr(pgCircleObject *self)
         return NULL;
     }
 
-    PyObject *result = PyUnicode_FromFormat("Circle((%R, %R), %R)", x, y, r);
+    PyObject *result = PyUnicode_FromFormat(
+        "%s((%R, %R), %R)", pgObject_TypeName((PyObject *)self), x, y, r);
 
     Py_DECREF(x);
     Py_DECREF(y);
--- a/src_c/line.c
+++ b/src_c/line.c
@@ -337,7 +337,9 @@ pg_line_repr(pgLineObject *self)
         return NULL;
     }
 
-    result = PyUnicode_FromFormat("Line((%R, %R), (%R, %R))", ax, ay, bx, by);
+    result = PyUnicode_FromFormat("%s((%R, %R), (%R, %R))",
+                                  pgObject_TypeName((PyObject *)self), ax, ay,
+                                  bx, by);
 
     Py_DECREF(ax);
     Py_DECREF(ay);
--- a/src_c/math.c
+++ b/src_c/math.c
@@ -1922,13 +1922,12 @@ vector_repr(pgVector *self)
     int tmp;
 
     if (self->dim == 2) {
-        tmp = PyOS_snprintf(buffer, STRING_BUF_SIZE_REPR, "Vector2(%g, %g)",
+        tmp = PyOS_snprintf(buffer, STRING_BUF_SIZE_REPR, "(%g, %g)",
                             self->coords[0], self->coords[1]);
     }
     else if (self->dim == 3) {
-        tmp =
-            PyOS_snprintf(buffer, STRING_BUF_SIZE_REPR, "Vector3(%g, %g, %g)",
-                          self->coords[0], self->coords[1], self->coords[2]);
+        tmp = PyOS_snprintf(buffer, STRING_BUF_SIZE_REPR, "(%g, %g, %g)",
+                            self->coords[0], self->coords[1], self->coords[2]);
     }
     else {
         return RAISE(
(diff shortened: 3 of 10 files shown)


Starbuck5:
It's a shame we can't use PyType_GetName, since we still support Python 3.10. And it's not in pythoncapi compat.h too. If that is equivalent to the function added here, there should be a comment explaining that to allow us to clean it up easily in the future and delete the helper on our end. 

I mean it's sort of unusual but maybe we could use PyType_GetName when it's available (3.11+) and the custom helper code when it is not, to make it feel super safe to transition to that in the future at the risk of maybe non matching behavior on 3.10 now. I'm just thinking this way because we will drop 3.10 support in 2 months when it is EOL.

For the ones that now have to go through PyOS_snprintf and PyUnicode_FromFormat, is there anything blocking it being all PyUnicode_FromFormat? That seems like it would be simpler.

In terms of docs, I'm okay with no docs. The behavior of subclasses repr is a pretty niche thing, and since we're going to 3.0 it will just be a 3.0 thing.

In terms of this fixing #3044, I think there may be other reprs besides these that should be edited, so I do not think this closes it. It is relevant, and could be merged on its own, but does not close it I believe.

oddbookworm:
> It's a shame we can't use PyType_GetName, since we still support Python 3.10. And it's not in pythoncapi compat.h too. If that is equivalent to the function added here, there should be a comment explaining that to allow us to clean it up easily in the future and delete the helper on our end.

It doesn't appear to be equivalent, but I think we could probably add it to the compat file by yoinking the implementation from cpython

Here's the raw definition: https://github.com/python/cpython/blob/26d9b25580159abeb69b80ae6fbf511385eb6e99/Objects/typeobject.c#L5796

And here's the definition of what it calls:
https://github.com/python/cpython/blob/26d9b25580159abeb69b80ae6fbf511385eb6e99/Objects/typeobject.c#L1487

Starbuck5:
I did look at the implementation in CPython, the implementation in _PyType_Name is similar to what OP has here.

Adding it to pythoncapicompat.h would need to be upstreamed, and that's too much of a burden to put on a contributor.

aatle:
Python 3.10 will be EOL in October 2026 so it shouldn't be too long of a wait.
Pygame-ce 3.0 will probably be 3.11+ anyways.

PR #3948, opened 2026-08-30
Fix wrong midi string type annotations

rmccampbell:
These functions actually take/return bytes, not str. write_sys_ex throws a TypeError on str.


oddbookworm:
I'm not sure this is the correct solution here. We've moved most of our bytes stuff to `str` since we definitely stopped supporting python2 paradigms. I think a better solution would probably be to actually change those values from bytes to str in the cython

rmccampbell:
In the case of write_sys_ex, it is accepting binary data not text data, so it absolutely should be bytes and not str; rather it is the documentation that seemingly hasn't been updated since Python 2 when "str" meant bytes.

For get_device_info: ideally these would be str as they are meant to be human-readable text, though I'm not certain if the underlying APIs in this case always return text in a deterministic encoding. There may be "wide" variants on windows that return UTF-16 though and I suppose on Unix UTF-8 is safe to assume.

rmccampbell:
Maybe it's good enough to just decode these with `locale.getpreferredencoding()` with errors="replace".

oddbookworm:
> For get_device_info: ideally these would be str as they are meant to be human-readable text, though I'm not certain if the underlying APIs in this case always return text in a deterministic encoding

It's only bytes currently because that's what cython default converts `char*` to, and we can choose to convert that to str instead

oddbookworm:
Here's the code for `_pypm.GetDeviceInfo`
```cython
def GetDeviceInfo(device_no):
    """Return device info tuple for MIDI device given by device_no.

    The returned tuple has the following five items:

    * underlying MIDI API (string)
    * device name (string)
    * whether device can be opened as input (1) or not (0)
    * whether device can be opened as output (1) or not (0)
    * whether device is currently opened (1) or not (0)

    """
    cdef PmDeviceInfo *info

    # disregarding the constness from Pm_GetDeviceInfo,
    # since pyrex doesn't do const.
    info = <PmDeviceInfo *>Pm_GetDeviceInfo(device_no)

    if info != NULL:
        return info.interf, info.name, info.input, info.output, info.opened
```

And here's `midi.get_device_info`
```py
def get_device_info(an_id):
    """returns information about a midi device
    pygame.midi.get_device_info(an_id): return (interf, name,
                                                input, output,
                                                opened)

    interf - a text string describing the device interface, eg 'ALSA'.
    name - a text string for the name of the device, eg 'Midi Through Port-0'
    input - 0, or 1 if the device is an input device.
    output - 0, or 1 if the device is an output device.
    opened - 0, or 1 if the device is opened.

    If the id is out of range, the function returns None.
    """
    _check_init()
    return _pypm.GetDeviceInfo(an_id)
```

rmccampbell:
But you need a known encoding to decode to str. I guess the PortMidi library doesn't really give any info on that, all I could find is https://portmidi.github.io/portmidi_docs/struct_pm_device_info.html; but most likely for Windows it would be the locale encoding.

Anyway changing to string would breaking existing code using it so I'm not sure if it's worth it.

oddbookworm:
> But you need a known encoding to decode to str.

This is true... when dealing with a `bytes` object in python. Not when dealing with a c-style string. A c-style string is literally just NULL-terminated array of characters. Each character is 8 bits (usually). So because a raw `char*` is what's actually returned, there's no hazard in having cython convert to a str instead of a bytes

rmccampbell:
I think you're mixing up C and Python "characters". A c-style `char` is a byte, and a sequence of chars must be interpreted in some encoding to convert it to logical "characters", which may be single-byte (e.g. Latin-1, Windows-1252) or multi-byte (e.g. UTF-8). A string in Python is not made of 8-bit bytes but of (up to) 32-bit Unicode code points. To convert from c strings to Unicode you need a known encoding. This is why ctypes `c_char_p` maps to Python bytes and not str.

In modern Unix it's generally pretty safe to just assume `char*` means UTF-8 but this is unfortunately more complex on Windows, where 8-bit `char` APIs tend to use legacy single-byte region-specific encodings like Windows-1252 (which should match `locale.getpreferredencoding()`), and you're instead recommended to use alternative non-Unix-compatible 16-bit "wide char" APIs. In modern times though the 8-bit char APIs may now use UTF-8 depending on the OS settings.

Also third-party libraries like PortMidi may or may not actually handle conversion from Windows APIs to UTF-8 behind the scenes, so that adds an extra layer of nuance here.

Starbuck5:
I agree with the approach in this PR of documenting the status quo properly. Switching to strings for those places it makes sense could be a separate issue. My main question for @rmccampbell is do you actually use the midi API? -- E.g. did you find this organically, or were you looking for type issues.

I believe this module is not widely used, so maybe not worth changing over to strings, or on the contrary maybe super worth changing to strings because we'll only disrupt a handful of people and we could put in the 3.0 release notes as a breaking change.

I do see evidence for UTF-8 in the portmidi source: https://github.com/PortMidi/portmidi/blob/master/pm_win/pmwinmm.c#L149-L159 . But this is something we could ask over there and maybe if it's UTF-8 they could add that to their header file to clarify.

One thing I noticed is that `get_device_info` is also supposed to be able to return `None`, it should be `tuple[bytes, bytes, int, int, int] | None` I believe.

rmccampbell:
I do have a personal project that used the pygame midi API, though I switched it to use rtmidi at some point as the pygame API has been a bit buggy and is somewhat harder to use, but I still have the old code. I just noticed this type error was highlighted by vs-code's type checker.

rmccampbell:
And yeah from that source it does seem like they're properly converting the names to UTF-8 so it probably would be safe to just assume that. I guess this wouldn't be that big of a deal for a breaking change in a major version given the presumed low usage.

Starbuck5:
For this PR lets keep the status quo, I just have my one note about get_device_info None still to resolve, and then future changes can be future changes.

I've opened an issue for it, https://github.com/pygame-community/pygame-ce/issues/3966

PR #3952, opened 2026-09-01
Improve and SDL3-fy the thread note in the Window constructor

damusss:
Modernized the thread note to reflect precise SDL3 requirements specified in the docs of every window-related operation:
"Thread Safety
This function should only be called on the main thread."
I'm open to corrections on my sentence, as this pull request is here more to push the improvement at all rather than enforce my precise wording.

### DIFF
--- a/buildconfig/stubs/pygame/window.pyi
+++ b/buildconfig/stubs/pygame/window.pyi
@@ -17,8 +17,8 @@ class Window:
     window class will continue to be developed, and we're excited to share
     the new functionality this class offers.
 
-    .. note:: Every window property, method and operation can only be
-        performed on the main thread the application started in.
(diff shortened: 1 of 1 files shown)


Starbuck5:
Thanks Damus.

I think the wording is awkward because properties and methods cannot be "performed." Of the 3 parallel words the verb only works with one of them; "operations" can be "performed"

`You can only interact with a Window (get/set properties, call methods, etc.) from your application's main thread.` ?

damusss:
Thank you, much better without that akwardness

Starbuck5 (approved):

ankith26 (approved):
LGTM, thanks! 🎉

Issue #3601, opened 2025-10-07
Calling `pygame.quit()` and then `Surface.get_bitsize()` overflowing and segfaulting


ImNotMeAnymore:
**Environment:**

```
pygame-ce 2.5.5 (SDL 2.32.6, Python 3.13.1)
Platform:		Linux-6.1.0-38-amd64-x86_64-with-glibc2.36
System:			Linux
System Version:		#1 SMP PREEMPT_DYNAMIC Debian 6.1.147-1 (2025-08-02)
Processor:			SSE2: Yes	AVX2: Yes	NEON: No
Architecture:		Bits: 64bit	Linkage: ELF

Python:			CPython 3.13.1 (main, Mar 19 2025, 22:50:29) [GCC 12.2.0]
GIL Enabled:		True
pygame version:		2.5.5
SDL versions:		Linked: 2.32.6	Compiled: 2.32.6
SDL Mixer versions:	Linked: 2.8.1	Compiled: 2.8.1
SDL Font versions:	Linked: 2.24.0	Compiled: 2.24.0
SDL Image versions:	Linked: 2.8.8	Compiled: 2.8.8
Freetype versions:	Linked: 2.13.3	Compiled: 2.13.3

Display Driver:		x11 ( xwayland == False )
Mixer Driver:		Mixer Not Initialized
```

**Current behavior:**

If you create a window with the `Window` class, save it's surface, then close pygame, and after print the surface you'll see overflow garbage instead of it's real size or, if you're lucky, a friendly `Fatal Python error: pygame_parachute: (pygame parachute) Segmentation Fault`

**Expected behavior:**

Surf should become a `<Surface(Dead Display)>` or similar after `pygame.close()` has been called. At the very least it should not segfault.

**Steps to reproduce:**

1. create a window, then get it's surface with `window.get_surface()`
2. call `pygame.quit()`
3. try to print the surface

**Test code**

```python
import pygame

window = pygame.Window()

surf = window.get_surface()
pygame.quit()

print(surf.size) # sometimes size overflows too, but this never seems to segfault
print(surf.get_bitsize()) # overflow or segfault which means print(surf) also segfaults
```

ankith26:
Hello, thanks for the bug report!

I cannot reproduce the segfault specifically but I do observe the undefined behaviour so there's definitely an issue here.

PR #3954, opened 2026-09-02
Prevent reinitialization for some object types

### LINKED ISSUE #3876: Protect objects from reinitialization
I think this applies to like every object that has external resources. Init can be explicitly called again on the object, and it leaks the external resources.

For example, running this program ends up with 3 windows on the screen.

```py
import pygame
pygame.init()

win = pygame.Window(size=(200,200))

win.__init__(size=(300,300))
win.__init__(size=(400,400))

print(win.get_surface())

while True: pygame.time.delay(50); pygame.event.get(); win.flip()
```

Internally, 2 of the SDL_Windows are leaked.

Maybe things should explicitly raise errors on reinitialization. Or they should run destructors first to clean up, then start fresh. I'd be interested to see what other projects have done with this.

jeevantenny:
**Changes:**
- Modify the init function of certain objects with external resources to check if the object is already initialized and raise a RuntimeError if it is.
- Add unit tests to make sure reinitializing these objects rause a Runtime error

Changed the init functions of `Window`, `Surface`, `Renderer` and `Texture` as suggested by @shaurya-sharma-dev. I didn't change the function for Image (I coundn't find the init function in render.c and it was commented out in the decleration of `pgImage_Type`) or Controller (it does not assign a new `SDL_GameController` upon reinitialization).

I also think reinitializing Sound will cause issues so maybe that should also be protected.

Closes #3876

### DIFF
--- a/src_c/render.c
+++ b/src_c/render.c
@@ -573,6 +573,11 @@ renderer_init(pgRendererObject *self, PyObject *args, PyObject *kwargs)
     int target_texture = 0;
     Uint32 flags = 0;
 
+    if (self->renderer) {
+        PyErr_SetString(PyExc_RuntimeError, "Object of type Renderer cannot be reinitialized");
+        return -1;
+    }
+
     char *keywords[] = {"window", "index",          "accelerated",
                         "vsync",  "target_texture", NULL};
     if (!PyArg_ParseTupleAndKeywords(args, kwargs, "O!|iipp", keywords,
@@ -1058,6 +1063,11 @@ texture_init(pgTextureObject *self, PyObject *args, PyObject *kwargs)
     int access = SDL_TEXTUREACCESS_STATIC;
     Uint32 Rmask, Gmask, Bmask, Amask;
     Uint32 format;
+
+    if (self->texture) {
+        PyErr_SetString(PyExc_RuntimeError, "Object of type Texture cannot be reinitialized");
+        return -1;
+    }
 
     char *keywords[] = {"renderer",  "size",   "depth",         "static",
                         "streaming", "target", "scale_quality", NULL};
--- a/src_c/render.c
+++ b/src_c/render.c
@@ -574,7 +574,8 @@ renderer_init(pgRendererObject *self, PyObject *args, PyObject *kwargs)
     Uint32 flags = 0;
 
     if (self->renderer) {
-        PyErr_SetString(PyExc_RuntimeError, "Object of type Renderer cannot be reinitialized");
+        PyErr_SetString(PyExc_RuntimeError,
+                        "Object of type Renderer cannot be reinitialized");
         return -1;
     }
 
@@ -1065,7 +1066,8 @@ texture_init(pgTextureObject *self, PyObject *args, PyObject *kwargs)
     Uint32 format;
 
     if (self->texture) {
-        PyErr_SetString(PyExc_RuntimeError, "Object of type Texture cannot be reinitialized");
+        PyErr_SetString(PyExc_RuntimeError,
+                        "Object of type Texture cannot be reinitialized");
         return -1;
     }
 
--- a/src_c/surface.c
+++ b/src_c/surface.c
@@ -485,7 +485,8 @@ surface_init(pgSurfaceObject *self, PyObject *args, PyObject *kwds)
 #endif
 
     if (self->surf) {
-        PyErr_SetString(PyExc_RuntimeError, "Object of type Surface cannot be reinitialized");
+        PyErr_SetString(PyExc_RuntimeError,
+                        "Object of type Surface cannot be reinitialized");
         return -1;
     }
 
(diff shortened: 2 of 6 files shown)


shaurya-sharma-dev:
@jeevantenny Not a maintainer, but I noticed that you have a lot of duplicated code in the unit tests to make sure `RuntimeError` is raised. 

On a side note, should it raise `RuntimeError` or `pygame.error`? Obviously, `pygame.error` is subclassed from `RuntimeError`, but it's still something that should be considered.

Just a disclaimer first before I show my "solution" to the duplicated code issue.

It could use much improvement, and we could just use another solution or just ignore it since the repetition isn't that bad.
I have also not tested this code, nor did I write it in an actual code editor (it was written in the GitHub UI).

Proposed Improvement (in some cases, this could actually decrease code readability, so a better way should probably be found):
```python
class TestCaseWithReinitError(unittest.TestCase):
    def __init_subclass__(cls, reinit_error_attr: str | tuple(str), reinit_kwargs: dict | tuple(dict)):
        cls.reinit_error_attr = reinit_error_attr
        cls.reinit_kwargs= reinit_kwargs

    def test_reinit(self):
        with self.assertRaises(RuntimeError):
            if type(self.reinit_error_attr) is str:
                kwargs = {k: eval(v) for k, v in self.reinit_kwargs.items()}
                getattr(self, self.reinit_error_attr ).__init__(**self.reinit_kwargs):
            else:
                for i, attr in self.reinit_error_attr:
                    kwargs = {k: eval(v) for k, v in self.reinit_kwargs[i].items()}
                    getattr(self, attr).__init__(**self.reinit_kwargs[i])
```

and then delete the individual `test_reinit` methods and add something like this:
```python
# This is an example for the RendererTest. This should change depending on the test class.
# Import TestCaseWithReinitError from a different file above.
class RendererTest(TestCaseWithReinitError, reinit_error_attr=("renderer", "texture"), reinit_kwargs=({"window": "self.window"}, {"renderer", "self.renderer", size: "(80, 60")):
# ...
```

EDITS: Rephrasing.

jeevantenny:
> but I noticed that you have a lot of duplicated code in the unit tests to make sure `RuntimeError` is raised.

The code repetition didn't seem that bad to me but I'm curious on what others think about it.

> On a side note, should it raise `RuntimeError` or `pygame.error`?

From what I understand, `pygame.error` represents an error thrown by SDL or other underlying C libraries. So `RuntimeError` seems to make more sense. I think `TypeError` would also work but I'm not too sure about it.

Starbuck5:
At least for surfaces, it is possible them to have their internal surface set to NULL.

```py
pygame.display.init()
print(screen)
pygame.display.quit()
print(screen)
screen.__init__((1000,1000))
print(screen)
```

```py
<Surface(500x500x32)>
<Surface(Dead Display)>
<Surface(1000x1000x32)>
```

This is the same behavior before and after your PR. And we will likely need to so more and to more objects in the future to prevent issues like https://github.com/pygame-community/pygame-ce/issues/3601.

jeevantenny:
Thanks for pointing that out @Starbuck5. To solve this issue I'm thinking we should add an `initialized` attribute to `pgSurfaceObject`.

```c
typedef struct {
    PyObject_HEAD struct SDL_Surface *surf;
    int owner;
    struct pgSubSurface_Data *subsurface;

    PyObject *weakreflist;
    PyObject *locklist;
    PyObject *dependency;
--> bool initialized; <--
} pgSurfaceObject;
```

After a surface has been fully initialized, the attribute can be set to true. I tested this solution locally and it raised the `RuntimeError` when running your code snippet. I can use this method for all the other objects in the list since it seems more robust then checking for the pointer to the SDL resource.

This is just my idea and there may be a better way to deal with this kind of issue.

Starbuck5:
I thought it about more, maybe it's okay to have reinitialization work in that case.

The point of the guard is to protect external resources, and if the external resource has been set to NULL (and hopefully freed correctly), then reinitializing is a just weird thing you could do.

It's just so elegant to have the resource pointer be the indicator and not need an external boolean flag.

Like the issue I linked we do have problems with objects sticking around after SDL quits. So maybe for some object types we have the module have a giant weakref list of them, or a linked list, and then when pygame.quit() happens we can explicitly destroy all the stuff before SDL says goodbye. @ankith26 may be interested. Having a module keep track of child object instances could also be useful for a freelist on some objects. But I don't think there any objects we'd want to have a freelist and also have external resources. A freelist of Surfaces/Windows/Audio would not make sense, but a freelist of Rect/Vector could.

jeevantenny:
So are you saying the current solution I have is good enough? If so I can implement this solution for other types like `pygame.Sound`. I'm not sure what other types need protection.

And just to clarify, implementing the weakref list for modules is not in the scope of this PR and @ankith26 may want to work on it instead.

PR #3959, opened 2026-09-03
Implementation of Vector2.from_polar and ClassObjectMethod descriptor

MichaelLesirge:
All credit to [joaquin30](https://github.com/joaquin30) in https://github.com/pygame/pygame/pull/3744

---

In pygame (not the community version), you can write

```python
some_function(pygame.Vector2.from_polar(length, angle))
```

as well as

```python
v = pygame.Vector2()
v.from_polar((length, angle))
some_function(v)
```

I was hoping to bring this feature to pygame-ce as well, as it quite convenient to be able to create Vector2()s from polar in-line. It reduces the need for two-step construction.

---

This makes from_polar to return a vector when used as a class method, and retain its original functionality when used as an object method. Although it is bit weird that it has different behavior when run as a class method or an instance method, it reduces the possibility of someone picking the wrong one if they were seporate. This is done though the "ClassObjectMethod descriptor"

This change has also be implemented for Vector3.from_spherical((r, theta, phi)).

Thanks for your time.


oddbookworm:
Heya @MichaelLesirge, thanks for wanting to contribute this!

Unfortunately, there's already an existing pull request for it, but it's unlikely to ever be a feature we include. Specifically, the implementation sucks and nobody wants to maintain the absolute mess that it is.

Existing pull request: https://github.com/pygame-community/pygame-ce/pull/2141

MichaelLesirge:
> Heya @MichaelLesirge, thanks for wanting to contribute this!
> 
> Unfortunately, there's already an existing pull request for it, but it's unlikely to ever be a feature we include. Specifically, the implementation sucks and nobody wants to maintain the absolute mess that it is.
> 
> Existing pull request: #2141

Shoot, should have checked first. Sorry about the duplication. Thanks for the quick response. 

Any chance I could implement another version of this (either as separate class method or even better just make the method always return itself)

oddbookworm:
I would love to see a better implementation, and I find that it's far more likely to merge than a port of the upstream implementation. I can't promise anything though, would need to see your implementation and it would need approval from multiple members

PR #3965, opened 2026-09-05
Feature options gate their optional dependencies

MrMusAddict:
The feature options (image, mixer, font, freetype, midi) weren't reliably gating their modules. When pkg-config couldn't see a library, `cc.find_library(..., required: get_option(...))` got wrapped in `declare_dependency()`, which always reports found. So `disabled` still built the module, and `auto` on a platform without the library configured fine and then failed at compile time. I hit it with midi on an android cross build, but the same pattern was on all 5, so this now covers all of them.

The 5 copies are replaced by one lookup loop. A disabled option skips both the pkg-config and find_library lookups, and only a library that was actually found gets wrapped. The emscripten branches still declare their own SDL deps and are skipped by the loop. On Windows, the legacy prebuilt block tied freetype.dll to the midi option, so `-Dmidi=disabled` failed configure on freetype; those are now independent.

Making the options real exposed a few things that had been unreachable: the ftfont-as-font.py fallback in src_py/meson.build never worked (`install_sources` has no `rename`, and it would circular-import through sysfont anyway), so it's removed; sysfont imported `Font` from `pygame.font` at module level, which made `pygame.ftfont` and `pygame.freetype` unusable without SDL_ttf; and `midi`, `freetype`, and `ftfont` now raise a clear ImportError when their compiled module wasn't built.

Tested on Linux with everything installed: default finds and builds all 5; each option disabled alone, all 5 disabled, and font+freetype disabled all build with the right modules absent and no stray linkage; auto with pkg-config hidden falls back to the bare libraries. `dev.py test` matches main (same 4 pre-existing WavPack mixer errors). Android cross build: disabled and auto skip pypm and produce a wheel, enabled still errors. Windows (MSVC, legacy prebuilts): before this, `-Dmidi=disabled` failed on freetype and `-Dfreetype=disabled` failed on portmidi; now all 4 combinations configure and the default gives the same summary as before.

| Platform | Configuration | Before | After |
| --- | --- | --- | --- |
| Linux | default (all libs installed) | all 5 modules built | same, `dev.py test` identical to main |
| Linux | `-Dmidi=disabled`, portmidi found via `.pc` | pypm built anyway | pypm skipped |
| Linux | `-Dmidi=disabled`, portmidi only findable as a bare lib | pypm built, `portmidi: true` in summary | pypm skipped, `portmidi: false` |
| Linux | each option disabled on its own | image/mixer/font/freetype: module still built (lib always found via `.pc`) | module absent, nothing else links the lib |
| Linux | all 5 disabled, `.pc` files hidden | summary reports all 5 as found, modules built | none found, none built, `pygame` imports with MissingModule placeholders |
| Linux | all 5 on `auto`, `.pc` files hidden | n/a (same wrap bug) | bare libs found via `find_library`, all 5 built |
| Linux | `-Dfont=disabled`, freetype on | configure error in src_py/meson.build (`rename` unsupported) | builds; `pygame.ftfont` and `pygame.freetype` usable |
| Linux | `-Dfreetype=disabled`, font on | `import pygame.freetype` -> `No module named 'pygame._freetype'` | clear ImportError naming the build option |
| Android cross | `-Dmidi=auto` (no portmidi) | configure OK, build fails compiling pypm | pypm skipped, wheel builds |
| Android cross | `-Dmidi=disabled` | pypm built, build fails | pypm skipped, wheel builds |
| Android cross | `-Dmidi=enabled` | configure error, portmidi not found | unchanged |
| Windows (MSVC, legacy prebuilts) | default | configures | same dependency summary |
| Windows | `-Dmidi=disabled` | configure error: freetype not found | configures, freetype found, portmidi skipped |
| Windows | `-Dfreetype=disabled` | configure error: portmidi not found | configures, portmidi found, freetype skipped |
| Windows | both disabled | configures but reports freetype2 as found | neither found |
| Windows | any option on `auto` | module silently skipped, or linked without its DLL when another block added the shared lib dir | prebuilt found and DLL bundled, same as `enabled` |

Not touched: the emscripten branches ignore the options entirely, and the newer CI prebuilt layout copies every DLL regardless of options (modules are still gated). Both seem like separate conversations.

AI disclosure: the initial meson.build diff was discovered by Fable 5.1, working from my android build experiments and my understanding of why `-Dmidi=disabled` wasn't taking effect. I reviewed it line by line and no changes were necessary. After review with ankith and starbuck5, additional edge cases were searched with Fable 5.1. Testing / sanity check was done by on my setup, and this description was reviewed for completeness.

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -449,16 +449,21 @@ endif
 
 endif # emscripten
 
-portmidi_dep = dependency('portmidi', required: false)
-if not portmidi_dep.found() and not plat.startswith('emscripten')
-    portmidi_dep = declare_dependency(
-        include_directories: pg_inc_dirs,
-        dependencies: cc.find_library(
-            'portmidi',
-            dirs: pg_lib_dirs,
-            required: get_option('midi'),
-        ),
-    )
+midi_opt = get_option('midi')
+# Start from a not-found dependency so pypm is only built when portmidi is
+# actually available and the midi option allows it.
+portmidi_dep = dependency('', required: false)
+if not midi_opt.disabled()
+    portmidi_dep = dependency('portmidi', required: false)
+    if not portmidi_dep.found() and not plat.startswith('emscripten')
+        portmidi_lib = cc.find_library('portmidi', dirs: pg_lib_dirs, required: midi_opt)
+        if portmidi_lib.found()
+            portmidi_dep = declare_dependency(
+                include_directories: pg_inc_dirs,
+                dependencies: portmidi_lib,
+            )
+        endif
+    endif
 endif
 
 portmidi_deps = [portmidi_dep]
--- a/meson.build
+++ b/meson.build
@@ -450,19 +450,20 @@ endif
 endif # emscripten
 
 midi_opt = get_option('midi')
-# Start from a not-found dependency so pypm is only built when portmidi is
-# actually available and the midi option allows it.
-portmidi_dep = dependency('', required: false)
-if not midi_opt.disabled()
-    portmidi_dep = dependency('portmidi', required: false)
-    if not portmidi_dep.found() and not plat.startswith('emscripten')
-        portmidi_lib = cc.find_library('portmidi', dirs: pg_lib_dirs, required: midi_opt)
-        if portmidi_lib.found()
-            portmidi_dep = declare_dependency(
-                include_directories: pg_inc_dirs,
-                dependencies: portmidi_lib,
-            )
-        endif
+# A disabled feature passed as `required` makes dependency() return not-found
+# without searching, so -Dmidi=disabled never picks up portmidi.
+portmidi_dep = dependency('portmidi', required: midi_opt.disabled() ? midi_opt : false)
+if not portmidi_dep.found() and not plat.startswith('emscripten')
+    portmidi_lib = cc.find_library(
+        'portmidi',
+        dirs: pg_lib_dirs,
+        required: midi_opt,
+    )
+    if portmidi_lib.found()
+        portmidi_dep = declare_dependency(
+            include_directories: pg_inc_dirs,
+            dependencies: portmidi_lib,
+        )
     endif
 endif
 
--- a/meson.build
+++ b/meson.build
@@ -260,7 +260,7 @@ if plat == 'win'
         dlls += sdl_lib_dir / '@0@.dll'.format(sdl)
 
         # SDL_image
-        if get_option('image').enabled()
+        if get_option('image').allowed()
             sdl_image_lib_dir = sdl_image_dir / 'lib' / arch_suffix
             pg_inc_dirs += fs.relative_to(sdl_image_dir / 'include', base_dir)
             pg_lib_dirs += sdl_image_lib_dir
@@ -281,7 +281,7 @@ if plat == 'win'
         endif
 
         # SDL_mixer
-        if get_option('mixer').enabled()
+        if get_option('mixer').allowed()
             sdl_mixer_lib_dir = sdl_mixer_dir / 'lib' / arch_suffix
             pg_inc_dirs += fs.relative_to(sdl_mixer_dir / 'include', base_dir)
             pg_lib_dirs += sdl_mixer_lib_dir
@@ -296,7 +296,7 @@ if plat == 'win'
         endif
 
         # SDL_ttf
-        if get_option('font').enabled()
+        if get_option('font').allowed()
             sdl_ttf_lib_dir = sdl_ttf_dir / 'lib' / arch_suffix
             pg_inc_dirs += fs.relative_to(sdl_ttf_dir / 'include', base_dir)
             pg_lib_dirs += sdl_ttf_lib_dir
@@ -304,12 +304,12 @@ if plat == 'win'
         endif
 
         # freetype, portmidi and porttime
-        if get_option('freetype').enabled()
+        if get_option('freetype').allowed()
             pg_inc_dirs += fs.relative_to(prebuilt_dir / 'include', base_dir)
             pg_lib_dirs += common_lib_dir
             dlls += common_lib_dir / 'freetype.dll'
         endif
-        if get_option('midi').enabled()
+        if get_option('midi').allowed()
             pg_inc_dirs += fs.relative_to(prebuilt_dir / 'include', base_dir)
             pg_lib_dirs += common_lib_dir
             dlls += common_lib_dir / 'portmidi.dll'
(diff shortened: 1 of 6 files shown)


meson.build
```diff
@@ -449,16 +449,21 @@ endif
...
-        ),
-    )
+midi_opt = get_option('midi')
+# Start from a not-found dependency so pypm is only built when portmidi is
+# actually available and the midi option allows it.
+portmidi_dep = dependency('', required: false)
+if not midi_opt.disabled()
+    portmidi_dep = dependency('portmidi', required: false)
+    if not portmidi_dep.found() and not plat.startswith('emscripten')
+        portmidi_lib = cc.find_library('portmidi', dirs: pg_lib_dirs, required: midi_opt)
```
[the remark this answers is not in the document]
MrMusAddict: This is pre-existing. `midi.py` was already in the unconditional install list, and before this PR, builds without portmidi either failed (pypm compile or configure) or shipped a pypm with unresolved portmidi symbols.

We could gate midi.py on `portmidi_dep.found()`, similar to font.py/ftfont. Happy to add it if you'd like it in this PR.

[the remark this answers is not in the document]
MrMusAddict: Pushed in 115db3e. midi.py is now installed only when pypm is built, matching the pont.py conditional. Default behavior is unchanged. An android cross-build still ships without it when midi is disabled.

src_py/meson.build
```diff
@@ -33,6 +32,11 @@ if sdl_api == 3
     py.install_sources('_audio.py', subdir: pg)
 endif
 
+# pygame.midi imports the pypm extension, so only install it when pypm is built
+if portmidi_dep.found()
+    py.install_sources('midi.py', subdir: pg)
+endif
```
ankith26: This change will cause `import pygame.midi` to raise an `ImportError`/`ModuleNotFound` error with a generic error message.

I suppose an alternative here is to have some error handling in `midi.py` that does a try-except while importing midi, and on failure raises a better error message?

MrMusAddict: I agree a try-except makes sense. I see something similar in `camera.py`. I'll push something here shortly. I'll revert 115db3e and edit `midi.py`

MrMusAddict: 99b9091 pushed

ankith26 (approved):
LGTM, thanks for contributing to pygame-ce! 🎉

Starbuck5:
Portmidi allegedly does work on Android by the way: https://github.com/PortMidi/portmidi/pull/102

Starbuck5:
I think the current proposal could be simplified 

Here's what I've come up with:
```meson
portmidi_dep = dependency('portmidi', required: false)
if not portmidi_dep.found() and not plat.startswith('emscripten')
    portmidi_lib = cc.find_library(
        'portmidi',
        dirs: pg_lib_dirs,
        required: get_option('midi')
    )
    if portmidi_lib.found()
        portmidi_dep = declare_dependency(
            include_directories: pg_inc_dirs,
            dependencies: portmidi_lib,
        )
    endif
endif
```

We just need to check if the find_library dependency was found before putting it into a declared dependency. I was pretty unhappy about needing to put it in a declared dependency at all, but the sticking point is if we want to grab the headers and the lib from random directories we need it. I found one other theoretically viable options with using project_options() to inject the extra headers into being included in all compilations, but that seemed more risky and weird than this.

This also applies to every other optional dependency. Freetype, SDL friends. If my understanding is correct, if they can't be found by pkg-config / whatever meson does they will be unconditionally wrapped in a declare_dependency that will report found() down the line no matter what cc.find_library's return says.

MrMusAddict:
@Starbuck5  I like where you're going with that, but it looks like that still builds pypm if portmidi.pc is visible to pkg-config but midi is disabled. 

Passing the option into dependency() when it's disabled skips the lookup the same way find_library already does, so disabled stays disabled without the extra branch. So here's what we could do:

```meson
midi_opt = get_option('midi')
portmidi_dep = dependency('portmidi', required: midi_opt.disabled() ? midi_opt : false)
if not portmidi_dep.found() and not plat.startswith('emscripten')
    portmidi_lib = cc.find_library(
        'portmidi',
        dirs: pg_lib_dirs,
        required: get_option('midi')
    )
    if portmidi_lib.found()
        portmidi_dep = declare_dependency(
            include_directories: pg_inc_dirs,
            dependencies: portmidi_lib,
        )
    endif
endif
```

ankith26 (commented):
I tested this on my linux which has portmidi installed. By default it finds midi and builds pypm as expected. I also tested passing `-Dmidi=disabled` and now after this PR it correctly skips portmidi detection and doesn't build pypm.

As starbuck mentioned earlier, the same kind of issue happens with other dependencies as well. So it would be good to adapt the same pattern/fix for those too. This is something you could take on in this same PR if you're interested.

meson.build
```diff
@@ -304,10 +304,15 @@ if plat == 'win'
...
+        if get_option('freetype').enabled()
             pg_inc_dirs += fs.relative_to(prebuilt_dir / 'include', base_dir)
             pg_lib_dirs += common_lib_dir
-            dlls += [common_lib_dir / 'freetype.dll', common_lib_dir / 'portmidi.dll']
+            dlls += common_lib_dir / 'freetype.dll'
+        endif
+        if get_option('midi').enabled()
+            pg_inc_dirs += fs.relative_to(prebuilt_dir / 'include', base_dir)
+            pg_lib_dirs += common_lib_dir
+            dlls += common_lib_dir / 'portmidi.dll'
```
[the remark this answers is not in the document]
MrMusAddict: Applied in 6829255 to all 5 prebuilt blocks rather than just these 2 since image, mixer and font had the same .enabled() check. Confirmed on Windows: before, -Dmidi=auto linked portmidi through freetype's lib dir but didn't bundle portmidi.dll, and all 5 on auto silently built none of them. Now auto behaves like enabled there, and default and disabled are unchanged.

src_py/meson.build
```diff
@@ -23,11 +23,6 @@ python_sources = files(
 )
 py.install_sources(python_sources, subdir: pg)
 
-# if not building font, install use ftfont for font.py
-if not sdl_ttf_dep.found() and freetype_dep.found()
-    py.install_sources('ftfont.py', subdir: pg, rename: 'font.py')
-endif
-
```
ankith26, line 30: We will probably end up removing this bit at some point in 3.0, but I don't think this change should be a part of this PR. For now we should still continue to do this I think

EDIT: oops looks like rename is indeed an invalid argument here, this piece of code was broken anyways so yeah might as well remove it now.

ankith26 (approved):
LGTM, I like how the dependency logic has been compactified! Tested it and all works as expected. Thanks again for contributing! 🥳

PR #3968, opened 2026-09-05
Restructure deps, update docs/licenses

ankith26:
- `manylinux-build` renamed to `deps`, `docker_base` renamed to just `base`: these changes are made to make it clearer that these scripts are not linux/docker specific but used on all platforms.
- `windependencies` and `macdependencies` moves to `deps/win` and `deps/mac` respectively to reduce clutter at the `buildconfig/` level
- Some legacy and unused files used for manylinux builds have been removed.
- All documentation consolidated and updated with currently relevant information in `deps/README.md`
- Updated licenses to cover the newer libraries we have been shipping since the last time this folder was updated.


Starbuck5:
Hey this now has a merge conflict, just in case you hadn't seen.

Starbuck5 (approved):
Seems reasonable. Thanks!

PR #3969, opened 2026-09-05
Remove legacy buildconfig

ankith26:
3.5k removed lines! I've been itching to make this PR for a while, and pygame-ce 3.0 seems to be the best time to do this.

The only thing that's not quite sorted is android support but I'm optimistic that we can land support for that as well in 3.0, but either way, that shouldn't block the removal of this legacy code which isn't maintained to work on modern python versions.


oddbookworm (approved):
The only thing I'm even slightly concerned about is removing the stuff from `.gitignore`, but that's just because someone out there might still have that stuff in their environment and might accidentally commit it. But that's not a blocker for me. LGTM! Let's get this legacy stuff out

damusss (approved):
The new build system has been working flawlessly for long while. No reason to keep outdated things, LGTM :D

Starbuck5 (approved):
This is a long time coming, and it allows us to be more flexible with our current system without needing to worry about causing bit rot in the old system. And 3.5k lines removed!

After merge there is a trailing location in the wiki that should be updated: https://github.com/pygame-community/pygame-ce/wiki/Advanced-compilation-guide-%28for-developers%29#build-system-notes

Thanks Ankith.

PR #3970, opened 2026-09-05
Document setup.py removal in README-migration, minor readme updates

Starbuck5:
I realized it would make sense to document Ankith's https://github.com/pygame-community/pygame-ce/pull/3969 as a readme-migration entry, even though it won't impact many people, it still is a significant change that redistributors may need to adjust to.

I also did a wording update in the README, and switched it to be commits since 2.5.8 instead of 2.5.7.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -102,8 +102,8 @@ or you want to contribute to pygame-ce, you will need to build pygame-ce
 locally from its source code, rather than pip installing it.
 
 Installing from source is fairly automated. The most work will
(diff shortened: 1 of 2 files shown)


damusss (approved):
It makes sense, thank you for documenting it!

Starbuck5:
Just force pushed a small wording change because the bot complained.

ankith26 (approved):
Thanks 👍

PR #3977, opened 2026-09-08
Fix finger event dict bug

JoKing-1999:
Pretty self explanatory code


aatle (approved):
LGTM!

ankith26 (approved):
LGTM, nice catch!

This probably needs to be backported to the 2.x branch as well

Issue #3953, opened 2026-09-02
PixelArray issues group


Starbuck5:
@aatle on PixelArray in https://github.com/pygame-community/pygame-ce/pull/3889#issuecomment-5399408651

> I agree with the observations.
> 
> After inspecting the code and stubs, I see multiple issues with the original code.
> See `_pxarray_subscript` and `_pxarray_ass_subscript` in [`pixelarray.c`](https://github.com/pygame-community/pygame-ce/blob/main/src_c/pixelarray.c), and see [`pixelarray.pyi`](https://github.com/pygame-community/pygame-ce/blob/main/buildconfig/stubs/pygame/pixelarray.pyi).
> 
> 1. (typing only: bug) Missing `__getitem__` overloads for two-element tuples other than `tuple[int, int]`, as well as one-element tuples
> 2. (mentioned issue: API) Should not return `None` from `__getitem__` with empty slice (not how slicing should work)
> 3. (API/bug) `__getitem__` with single ellipsis argument returns self reference instead of new copy
> 4. (API/bug?) `None` is intended to be same as ellipsis, but `None` alone cannot be used as index for `__getitem__` or `__setitem__`
> 5. (typing only: improvement) Both 1D and 2D `PixelArray`s currently share the same signatures for type checking, which is a mess
> 6. (bug) Both `__getitem__` and `__setitem__` for both 1D and 2D mostly fail to gracefully handle an index that is a one-element tuple (e.g. `pxarray[("hello",)]`)

shaurya-sharma-dev:
Should #3889 be backported to the main-2.x branch for a potential 2.5.9 release?

PR #3990, opened 2026-09-12
Optimize `PixelArray.__repr__` time complexity

aatle:
Use `str.join` instead of repeated string concatenation for `PixelArray.__repr__`, improving time complexity from quadratic to roughly linear.

No change in repr's output.

Related: #3953
This is a precursor to future changes to `PixelArray.__repr__`'s output.

<details><summary>Test program</summary>

```py
from hashlib import sha256
from time import perf_counter

import pygame as pg

size = 512, 512
color = 255, 0, 0, 255

surface = pg.Surface(size)
surface.fill(color)
pixel_array = pg.PixelArray(surface)

start = perf_counter()
string = repr(pixel_array)
end = perf_counter()
print(f"PixelArray.__repr__() ({size=}): {end - start:.6f} seconds")

hash_value = sha256(string.encode()).hexdigest()
print(f"(SHA-256 hash: {hash_value})")
```

</details>

Optimized version output:
```
PixelArray.__repr__() (size=(128, 128)): 0.003902 seconds
(SHA-256 hash: e63e87e9c1e8d2cbcd5942ef2e54b369b128d1c3db46476aa47ba94819f32795)
```
```
PixelArray.__repr__() (size=(512, 512)): 0.052563 seconds
(SHA-256 hash: e082901127482c57889f8c14bd0c011de76d38354692c2e48600cf8644b6976a)
```
Original version output:
```
PixelArray.__repr__() (size=(128, 128)): 0.032360 seconds
(SHA-256 hash: e63e87e9c1e8d2cbcd5942ef2e54b369b128d1c3db46476aa47ba94819f32795)
```
```
PixelArray.__repr__() (size=(512, 512)): 65.320605 seconds
(SHA-256 hash: e082901127482c57889f8c14bd0c011de76d38354692c2e48600cf8644b6976a)
```

Note: I am not experienced in working on the C extension modules, feedback is greatly welcomed.


Starbuck5:
So, this is a nice find of something to fix, and the solution seems very reasonable. It could use py Unicode_writer maybe to be fancy.

But I do think it’s a problem that if you print out a PixelArray for a large surface it will fill your terminal. Maybe it should be truncated in some way like a numpy array, or maybe it should just say “PixelArray”.

so I’m hesitant to accept this solution as is. But definitely a great find.

aatle:
I missed that `PyUnicodeWriter` is available to use in pygame-ce.

I agree that the array values should be truncated with ellipses to match the behavior of numpy arrays. This PR can be merged without those features I think; avoiding quadratic time complexity is applicable in any case.

Do you want me to rewrite with `PyUnicodeWriter` though?

PR #3991, opened 2026-09-12
Move Surface and Transform to their own directories

JoKing-1999:
Update of #3462, part of #3421. Functionally this does nothing, but visually, imo, it makes src_c/ a much nicer place.

It cuts the number of files in there from 92 to 69, which makes it way easier to look through and find the right file, and it also groups related files together, so when you're working on one thing it's easier to change all the files you need to. For now the tests still live in their own directory, we can figure out their future later.                                                                                                                                                                                                                                                                                  

I went with these two modules because they're among the biggest ones, and also because this is part of my SIMD reorganization and they're the modules that actually implement SIMD.


aatle:
@coderabbitai make an issue only for the `Unlock the surface before the PG_GetSurfaceDetails failure return` review comment, with no assignee.

Issue #3981, opened 2026-09-10
Return `False` instead of raising `TypeError` when using `__contains__` on arbitrary object


aatle:
Currently (from checking stubs), `Color`, `Rect`, `Vector_` and `sprite.Group` have an explicit `__contains__` method.

Performing a containment check like `"a" in pygame.Vector2()` (with an object not of the expected type) leads to a `TypeError` on these, except `Group`.
However, the correct or expected behavior is for this to return `False`.

`__contains__` is intended to be functionally equivalent to `any(obj == elem for elem in self)`, and equality will simply return `False` for different types.

Interestingly, there is actually a built-in fallback for a missing `__contains__` that iterates over the object.
For example, removing `__contains__` from `Color` entirely makes `"a" in pygame.Color("red")` evaluate to `False`, unlike the current implementation.

This minor issue affects `Color`, `Rect`, and `Vector_` as far as I'm aware.
Changing it is technically a breaking change.

This issue is also why there has needed to be `# type: ignore[override]` suppressions on those `__contains__` stubs.

PR #3998, opened 2026-09-14
Remove `Rect.contains()` functionality from `Rect.__contains__()`

### LINKED ISSUE #3980: Remove undocumented `rect1 in rect2` functionality in `Rect.__contains__`
There is a currently undocumented functionality of `Rect.__contains__` where one can check if another rect is completely contained inside the rect, the same functionality provided by the original `Rect.contains()` method.
Added by https://github.com/pygame/pygame/pull/2897.

I saw this from https://github.com/pygame-community/pygame-ce/issues/3531#issuecomment-3067323931.

The `__contains__` magic method is mainly intended for checking if an item is an element of a collection. Overloading it with other functionality for syntactic sugar is not necessary and a bit confusing.

So, I think this functionality should be removed from `__contains__`.

This affects #3877, which adds a small docstring to `__contains__` referencing this behavior (though it does not appear in the generated API reference).

aatle:
Closes #3980.

Blocks #3981.

Remove the functionality that `Rect.__contains__()` would also check `Rect.contains()`.
Then, simplified `Rect` code by integrating the redundant `RectExport_contains_internal` function into the existing `RectExport_contains` function.


pydsigner:
I'm not sure this change offers enough improvement to be worth it. It's backwards incompatible, shouldn't really have an impact on maintainability or performance, and makes the library less Pythonic.

